PR #180, opened 2014-05-19
Allow converting of r/n to /n when editing files in October backend

Jmeyering:
The commit message got messed up but this PR adds a new setting in cms.config called "convertLineEndings". When set to true the October backend will convert all /r/n's to /n in the onSave method via a new private method "convertLineEndings".

### DIFF
--- a/app/config/cms.php
+++ b/app/config/cms.php
@@ -169,4 +169,17 @@ return array(
 
     'defaultMask' => ['file' => null, 'folder' => null],
 
+    /*
+    |--------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


leancode:
Its no longer working. When editing something using the pages plugin with convertLineEndings=true users get a popup like this:
<code>
"Class RainLab\Pages\Controllers\Index does not have a method definition for convertLineEndings" on line 302 of /srv/www/oktick/html/vendor/october/rain/src/Extension/ExtendableTrait.php
</code>

PR #636, opened 2014-09-18
Ability to add and override redactor settings.

that0n3guy:
I needed the ability to easily add redactor settings (ie, from http://imperavi.com/redactor/docs/settings) so I added a simple way to do that.

The basic workflow:
* create a js file in a public location with all your redactor settings in it
* set the path to the file within your cms.php file.
* flush cache and the new settings will be added to your redactor settings.

Any setting you add that is already existing in the richeditor.js will overwrite the richeditor.js setting.

### DIFF
--- a/app/config/cms.php
+++ b/app/config/cms.php
@@ -216,4 +216,20 @@ return array(
 
     'convertLineEndings' => false,
 
+    /*
+    |--------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


daftspunk:
Added this to my TODO

daftspunk:
This appears to be underway as #1743 / #1744 for reference.

PR #748, opened 2014-11-04
Add Backend Settings to override file based config

karnold:
*Implement configurable timezones*

This adds a timezone setting in the interface to allow the override of app.timezone.


daftspunk:
After some investigationing, file based config has it's place as a fixed form of configuration, database settings also have their place. Having DB settings override File config has created a bit of confusion for people about where they should be setting these values. I think instead of using a blanket approach to override the file, we should do this sparingly as it has a performance cost.

daftspunk:
Also added a note to include Timezone as part of this list, same as locale.

PR #775, opened 2014-11-17
Fix backend search

alvaro-canepa:
Backend search doesn't work properly if exists a previous where sentence of `deleted_at`.
The problem is on the use of `orWhere`, without validate for previous wheres, making a final SQL like this

    `table_name`.`deleted_at` is null or (((lower(table_name.column_one) LIKE '%term%')

I don't know if my solution is the correct one. It work fine on my tests.

Thanks

### DIFF
--- a/modules/backend/widgets/Lists.php
+++ b/modules/backend/widgets/Lists.php
@@ -407,9 +407,19 @@ class Lists extends WidgetBase
          * Apply a supplied search term for primary columns
          */
         if (count($primarySearchable) > 0) {
-            $query->orWhere(function ($innerQuery) use ($primarySearchable) {
+            $whereBool = 'or';
+            $wheres = $query->getQuery()->wheres;
+            if(is_array($wheres) && count($wheres)) {
+                $lastWhere = array_pop($wheres);
+                if(is_string($lastWhere['column'])){
+                    if(strstr($lastWhere['column'], 'deleted_at')){
+                        $whereBool = 'and';
+                    }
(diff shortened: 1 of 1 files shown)


tresbach:
Hi, any update on this? List search really doesn't work along with SoftDeletingTrait/SoftDeletes..

daftspunk:
This is a dirty fix for a problem I cannot see demonstrated. I have created a list using a model with soft deletes, the list can be searched successfully.

daftspunk:
I was later able to replicate this, it has now been fixed in 8d93c9ffabca25950ca63dab40dbd014c1671952

alvaro-canepa:
I try this change before, and  fail on common queries (without softDelete). I know my solution was dirty, but this simple solution not work on all possible environments. 
Sorry my English  and thanks for this amazing CMS!

daftspunk:
Can you provide an example of a common query where it fails?

alvaro-canepa:
Let me make some tests on the Beta version. I still using the Laravel 4.2 version.

daftspunk:
Ok, that's fine, please submit as a new Issue if you discover any problems.

daftspunk:
I noticed there was a problem here when searching related columns, I managed to resolve it in 93dd61efc71e64092a53fe21fd3ef4b630d8e9be

alvaro-canepa:
Great, I will testing soon, actually have no time :(
Thanks daftspunk!!!

PR #848, opened 2014-12-23
Extend query of the filter options lists

nicolas-lescop:
For example if you want to exclude an element from filter options

```php
public function listFIlterExtendQuery($query, $scope)
{
    if ($scope->scopeName == 'status') {
        $query->where('status', '<>', 'all');
    }
}
```


daftspunk:
Looks good, just need time to test it. Thanks!

PR #850, opened 2014-12-27
Valid strtotime() format when time is left empty

scottbedard:
Date-type laravel validation ("date", "before", and "after") may not be used with datetime widgets containing empty time fields because they are not in a valid strtotime() format (example: 2014-12-27 :00).

This extra check ensures a valid strtotime() format by only adding the ":00" seconds value when a time value is present.


daftspunk:
Thanks

PR #875, opened 2015-01-08
Adds isHidden check to backend components widget

scottbedard:
Allows components to be hidden from the backend components widget by checking their isHidden value
https://github.com/octobercms/october/blob/master/modules/cms/classes/ComponentBase.php#L41


scottbedard:
*Edit*
I came to this PR while trying to help ndcisiv figure out how to hide his component in the cms backend. He just informed me that it may have already been behaving properly, and the component's puzzle piece was being hidden from the cms page. Sorry if we misunderstood how isHidden is supposed to work.
- Scott

daftspunk:
Looks fine to me, thanks

PR #876, opened 2015-01-09
Adding MarkdownServiceProvider to the list of service providers.

mossadal:
This complements the pull request I submitted on october/rain

### DIFF
--- a/modules/system/providers.php
+++ b/modules/system/providers.php
@@ -39,6 +39,7 @@ return [
     'October\Rain\Support\Scaffold\ScaffoldServiceProvider',
     'October\Rain\Support\FlashServiceProvider',
     'October\Rain\Mail\MailServiceProvider',
+    'October\Rain\Support\MarkdownServiceProvider',
 
     /*
      * Vendor providers


daftspunk:
I understand the need for this, but couldn't you simply hook the Model beforeSave() event and apply your own Markdown filter before the standard one does?

daftspunk:
Refs https://github.com/octobercms/library/pull/61

mossadal:
There are two problems with hooking into beforeSave. My primary use case at the moment is extending the Post model in Rainlab.Blog.

If I hook into `Post::beforeSave()`, using `Post::saving(...)`, my changes will be applied *before* the original `Post::beforeSave()` method is called, and any changes I make to `$post->content_html` will be overwritten. An alternative could be to add an extra field to the post model, put my parsed code there and override all partials referencing `$post->content_html`.

Right now, I instead do it by hooking into `Post::saved()` and recreate `$post->content_html` from `$post->content`and saving it directly via `DB::update`so the post `beforeSave()` logic isn't called again. This works reasonably well, but is a little clumsy.

The big problem with this approach is that it will be very difficult to maintain a consistent handling of Markdown across the entire site. For example (still in the Blog example), the Markdown parser is also called by the preview rendering code. I haven't yet figured out how to hook into that, but I guess that it's somehow possible to override the backend controller templates in another plugin?

By turning the `Markdown` helper class into a Facade, it would be enough to change *one* thing to get conistent behavior of the Markdown parsing across the entire site. This seems to be a more "Laravelesque" solution.

What would be the downside? I suppose a call to a proxied Parsedown via a Facade is ever so slightly slower, but I imagine the difference is negligible. The way I wrote the pull-request, the changes are completely transparent. 

I put the Facade `Markdown` class where the older helper class (even though it would have made more sense to put it in `october/rain/Support/Facade`and a call to `October\Rain\Support\Markdown::parse` behaves exactly as before.

daftspunk:
The concern with IoC + Plugins is this:

* Plugin A uses "Markdown Classic" and replaces the IoC instance
* Plugin B uses "Parsedown Github Flavor" and replaces IoC instance

Plugin A is now using the wrong library, because it has been replaced by Plugin B.

The solution is to operate by extension, two Markdown parsers can be applied twice, since Markdown can parse HTML just fine.

mossadal:
I suppose that could be an issue. Of course you could say the same about all the Facades you expose in `October\Rain\Support\Facades`.

Would you prefer that I propose a change to the Rainlab.Blog plugin to make it easier to override `Blog::formatHtml' instead? For example, firing events before and after the `formatHtml`code runs?

daftspunk:
Yes, and this is why you find we avoid IoC where possible.

I'd be more inclined to say let's focus on the Markdown class (October\Rain\Support\Markdown). Make it a singleton, keep the static parse() method, implement the event emitter trait, add events before and after the parsing takes place.

Something like this, I just wrote it off the top of my head, not sure if this will work but it should be enough to get you started:

```
<?php namespace October\Rain\Support;

use ParsedownExtra;

/**
 * Markdown content parser
 *
 * @package october\support
 * @author Alexey Bobkov, Samuel Georges
 */
class Markdown
{

    use \October\Rain\Support\Traits\Emitter;
    use \October\Rain\Support\Traits\Singleton;

    public static function parse($text)
    {
        return self::instance()->parseText($text);
    }

    /**
     * Parse text using Markdown and Markdown-Extra
     * @param  string $text Mardown text to parse
     * @return string       Resulting HTML
     */
    public function parseText($text)
    {
        if (
         ($event = $this->fireEvent('beforeParse', [$text], true)) ||
         ($event = Event::fire('markdown.beforeParse', [$text], true))
        ) {
            $text = $event;
        }

        $instance = new ParsedownExtra;
        $result = $instance->text($text);

        if (
         ($event = $this->fireEvent('parse', [$text, $result], true)) ||
         ($event = Event::fire('markdown.parse', [$text, $result], true))
        ) {
            $result = $event;
        }

        return $result;
    }
}

```

mossadal:
That sounds like a better approach. I'll try to come up with something later today or tomorrow.

mossadal:
How's that then? I thought it was cleaner to pass the data to the event handler with a microscopic helper class instead of passing a string by reference.

daftspunk:
Very clever, works for me!

PR #882, opened 2015-01-10
Adds editor content value setter

mod3st:


daftspunk:
This is no longer necessary, see latest code

PR #886, opened 2015-01-12
Update lang.php

mahony0:
Why these lines removed? On backend/backend/usergroups, strings not showing up..


daftspunk:
Not sure, thanks though

PR #897, opened 2015-01-19
Bug fix. There is no Aliasing to Model class in FormField class.

NemoD503:
This small typo is very dangerous, because without aliasing in method getValueFromData

    $result instanceof Model === false

always.
And the value of field-relation is NULL. (So we see unchecked checkboxes and inactive select in relation widget)


daftspunk:
Thanks for this

PR #900, opened 2015-01-20
Polish language

versedi:
#Polish localization - pretty everything (some single missing stuff as it was better to leave it english language cause of backend relationship).
    
Fonts should be replaced with external link (there are missing characters as: ą ś ć ł ź
    
 ```<link href='http://fonts.googleapis.com/css?family=Open+Sans:400,300,300italic,400italic,600,600italic,700&subset=latin,latin-ext' rel='stylesheet' type='text/css'>```
    
seeing from count of localization files it should be set to some IF statement somewhere if right character set for specific lang (i.e. Russian requires Cyrylic instead of Latin) so it should go with other font files - why load whole character set library if we could separate that (english don't requires any extended character set - so pure latin's ok).
    
        Signed-off-by: versedi <versedi@gmail.com>


Mr118:
damn you went too fast... i just noticed your message on OctoberCMS, if you had checked the forum you would probably have a lot less work:
https://octobercms.com/forum/post/backend-translation-pl?page=1#post-3985
(complete and already on several production portals)

We decided to not make pull request until some transifex-like option is enabled for OctoberCMS (I can't imagine making pull request for single plugins), but still, good work.

versedi:
It will be less work later on :) 
I'll keep this branch until some serious translation platform will be launched.

daftspunk:
This is going to the wrong branch and has garbage in it (app/config/* and themes/demo* files being changed).

versedi:
There was indeed trash in demo.phtml caused of testing polish characters (which are not working because you're not including extended latin character of OpenSans).
The rest of changed files based on branch history: (nothing touched /app/config/*) 
modules/backend/lang/pl/lang.php
modules/backend/model/BackendPreferences.php
modules/cms/lang/pl/lang.php
modules/cms/lang/en/lang.php
modules/cms/lang/ja/lang.php
modules/cms/lang/pl/lang.php
modules/cms/lang/pl/validation.php
modules/cms/lang/ru/lang.php
modules/cms/lang/sv/lang.php
modules/cms/lang/tr/lang.php
modules/cms/lang/ja/lang.php

daftspunk:
I was unable to merge this, I have added the language manually

PR #903, opened 2015-01-21
Corrections and updates for Build 182

jtherczeg:
Corrections and updates for Build 182 in the Hungarian language files


daftspunk:
Thanks!

PR #911, opened 2015-01-26
Richeditor - html preview tab

mod3st:
Textarea field is small by default. I removed auto height CSS property.
I don't understand why it was set earlier.

### DIFF
--- a/modules/backend/formwidgets/richeditor/assets/css/richeditor.css
+++ b/modules/backend/formwidgets/richeditor/assets/css/richeditor.css
@@ -1434,7 +1434,6 @@ body .redactor-box-fullscreen .redactor-editor {
   top: 0;
   bottom: 0;
   position: absolute;
-  height: auto !important;
 }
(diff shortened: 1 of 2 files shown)


daftspunk:
This is due to `stretch` class being applied, you should remove it if the redactor needs a size, and use a suitable size class

```
    &.size-tiny .redactor-editor     { height: (@size-tiny - @richeditor-toolbar-size) !important; }
    &.size-small .redactor-editor    { height: (@size-small - @richeditor-toolbar-size) !important; }
    &.size-large .redactor-editor    { height: (@size-large - @richeditor-toolbar-size) !important; }
    &.size-huge .redactor-editor     { height: (@size-huge - @richeditor-toolbar-size) !important; }
    &.size-giant .redactor-editor    { height: (@size-giant - @richeditor-toolbar-size) !important; }
```

mod3st:
But why can't we auto stretch WYSIWYG?

PR #913, opened 2015-01-26
Replaced short open tags

mariusvw:
Replaced short open tags due compatibility

### DIFF
--- a/modules/backend/views/access_denied.php
+++ b/modules/backend/views/access_denied.php
@@ -2,15 +2,15 @@
 <html lang="en">
     <head>
         <meta charset="utf-8">
-        <title><?= Lang::get('backend::lang.page.access_denied.label') ?></title>
-        <link href="<?= URL::to('/modules/system/assets/vendor/font-autumn/css/font-autumn.css') ?>" rel="stylesheet">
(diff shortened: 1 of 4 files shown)


daftspunk:
Can you explain what this is not compatible with? <?= $var ?> is fully supported in PHP 5.4 and above.

mariusvw:
On some servers short open tags are disabled.

This results that the installation script (and CMS) crash on short open tags.

After replacing all those it works fine.

NemoD503:
I think short open tags is BETTER in that case.

mariusvw:
I believe it can be better, but not every person has the ability to enable them on the server they use for hosting.

daftspunk:
AFAIK <?= $var ?> is not considered a short tag

<? echo $var ?> is an example of a short tag.

What version of PHP are you using @mariusvw ?

mariusvw:
I'm running version 5.5.x

daftspunk:
> Starting with PHP 5.4, short echo tag <?= is always recognized and valid, regardless of the short_open_tag setting.

Source: http://php.net/manual/en/language.basic-syntax.phpmode.php

PR #915, opened 2015-01-27
Bug-fix. Correct z-index for richeditor`s full-screen mode.

NemoD503:
When we have many richeditors on single page and switch one of them to full-screen mode we will see many toolbars.
P.S I use gulp-less for compiling, so it changet some another lines. If it`s critical, I can rebuild my pull request.
image


daftspunk:
Added this by hand

NemoD503:
Thanks!

PR #919, opened 2015-01-30
Add Mandrill support patch

carlos-arroyo:
Added support for Mandrill option in backend settings.

Additional translations will have to be added in backend translation files.


daftspunk:
Thanks for this

PR #923, opened 2015-02-02
Removed select from columns.yaml for User.php

SaifurRahmanMohsin:
SQLite doesn't support concat() and using it goes against Laravel's database abstraction. The getFullNameAttribute() in User.php already does this task well.

Also, removing this line fixed the issue where a SQLite based install of OctoberCMS threw gave this [error page](http://i.imgur.com/H8wGTmU.png) everything and prevented access to Administrator tab in backend and also overtime a new avatar was uploaded to user page.

### DIFF
--- a/modules/backend/models/user/columns.yaml
+++ b/modules/backend/models/user/columns.yaml
@@ -14,7 +14,6 @@ columns:
         searchable: true
     full_name:
         label: backend::lang.user.full_name
-        select: concat(first_name, ' ', last_name)
         searchable: true
     email:
         label: backend::lang.user.email


daftspunk:
Probably just need to parse the select using DbDongle so it is SQLite friendly

SaifurRahmanMohsin:
I haven't tried that but removing it fixed the problem instantly so I'm glad it works now. Please do check if the removal affects installs of october running on other DBs and if not apply the necessary fixes :)

daftspunk:
Fixed in https://github.com/octobercms/library/pull/70

PR #925, opened 2015-02-03
Added lang instead of plain text

SaifurRahmanMohsin:
Probably forgot to link the lang file instead because I can see these texts are already present in the lang.php file


daftspunk:
Thanks

PR #927, opened 2015-02-04
Display active plugin count on updates page

Flynsarmy:
If you have disabled plugins, the plugin counter will now show `activeCount (totalCount)` like so:


daftspunk:
I think this PR is missing a file? Scope should also be called `scopeIsEnabled`

Flynsarmy:
Fixed. Sorry about that!

daftspunk:
This is a weird feature, but ok.

Flynsarmy:
Maybe it should say 54 (23 Active) ?

Shahiem:
Or maybe you  can give it a green color

PR #930, opened 2015-02-06
Support multiple form contexts

Flynsarmy:
The RelationController creates a form with context 'relation' whether it's a create OR update form so it's impossible to target fields for one or the other. This PR allows you to do the following:

```php
$form->context = ['relation', 'create'];
```

It's fully backwards compatible.


Flynsarmy:
Sent to wrong branch, my bad.

PR #931, opened 2015-02-06
Better form context for RelationController forms

Flynsarmy:
Requires accepting #932.

The RelationController creates a form with context 'relation' whether it's a create OR update form so it's impossible to target fields for one or the other. This PR combined with #932 allows for context-aware form fields in RelationController forms.


daftspunk:
I see how this fixes your problem, however, we should add an option in the relation config to define a specific context for the manageWidget

```
manage
    context: somethingElse

#or

manage:
    context:
        create: somethingCreate
        update: somethingUpdate
```

PR #932, opened 2015-02-06
Support multiple form contexts

Flynsarmy:
The RelationController creates a form with context 'relation' whether it's a create OR update form so it's impossible to target fields for one or the other. This PR allows you to do the following:

```php
$form->context = ['relation', 'create'];
```

It's fully backwards compatible.

### DIFF
--- a/modules/backend/widgets/Form.php
+++ b/modules/backend/widgets/Form.php
@@ -474,7 +474,7 @@ class Form extends WidgetBase
              */
             if ($fieldObj->context !== null) {
                 $context = (is_array($fieldObj->context)) ? $fieldObj->context : [$fieldObj->context];
-                if (!in_array($this->getContext(), $context)) {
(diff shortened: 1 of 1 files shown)


daftspunk:
Forms should only really ever have one context. Fields can have multiple, why not just give you field context relation + create?

Flynsarmy:
Because when a forms context is set to something unhelpful (as it is in RelationController and perhaps could be in future plugins) it would be convenient to be able to add a more useful context to the form so it's better targetable by fields.

In this example setting my fields to 'relation create' would just result in them showing on both the update and create forms - because the form has its context set to 'relation' and the Forms widget just looks for any match as apposed to matching all provided field contexts.

I like your solution here https://github.com/octobercms/october/pull/931#issuecomment-73653220 but it really only solves this one particular issue and not any future ones that may crop up.

PR #933, opened 2015-02-07
Add support for getting all config

Flynsarmy:
This PR lets you do `$this->getConfig()` in a Controller to get all config. 

## Use case

Currently `setConfig()` overwrites the config rather than updating it - so the only way to update is to get all config, modify as desired then `setConfig`.


Flynsarmy:
Resubmitted to correct branch #941

PR #934, opened 2015-02-07
Add support for removing form fields

Flynsarmy:
Example Usage:

```php
    /**
     * Called after the form fields are defined.
     * @param \Backend\Widgets\Form $host The hosting form widget
     * @return void
     */
    public function formExtendFields($host, $fields)
    {
        $host->removeField('person');
    }
```


daftspunk:
LGTM

PR #961, opened 2015-02-23
Allow setting 'options' for a scope in filter config

tresbach:
Suppose you have a filter config like:
```
scopes:
    category:
        label: Category
        conditions: category in (:filtered)
        options:
            cat1: First Category
            cat2: Second Category
```

Currently, trying to do so raises a fatal error (call to undefined function).


daftspunk:
Thanks!

PR #966, opened 2015-02-25
Server for L5

NemoD503:
Now we do not have /bootstrap/paths.php, so we need to update server.php for starting local artisan server.


daftspunk:
Thanks

PR #967, opened 2015-02-25
Made config values configurable by .env file.

kuzyk:


daftspunk:
Thanks for this.

You're welcome to use config files for your config files in your own application, we don't feel it is a good idea to include this as part of the base platform because it creates a steeper learning curve for newcomers.

PR #977, opened 2015-03-01
Fix issue getting extending form field's options

Sylvert0:
Issue reproduction:

Plugin.php:
```
...
public function boot()
{
	UserController::extendFormFields(function($form, $model, $context)
	{
	    if (!$model instanceof UserModel)
	        return;

	    if (!$model->exists)
	    	return;

	    // Add tabs fields
	    $form->addTabFields([
	        'profile[country]' =>
	        [
	            'label'   => 'Country',
	            'tab'     => 'Profile',
	            'span'    => 'left',
	            'type'    => 'dropdown'
	        ]
	    ]);
	});
}

```

This return the error:
```
The model class Sylvert\Profile\Models\Profile must define a method getProfile[country]Options() returning options for the 'profile[country]' form field.
```

Instead of getProfile[country]Options() must be getProfileCountryOptions().

### DIFF
--- a/modules/backend/widgets/Form.php
+++ b/modules/backend/widgets/Form.php
@@ -887,7 +887,9 @@ class Form extends WidgetBase
          * Refer to the model method or any of its behaviors
          */
         if (!is_array($fieldOptions) && !$fieldOptions) {
-            $methodName = 'get'.studly_case($field->fieldName).'Options';
+            $fieldName = str_replace('[', '', $field->fieldName);
+            $fieldName = str_replace(']', '', $fieldName);
+            $methodName = 'get'.studly_case($fieldName).'Options';
             if (
                 !$this->methodExists($this->model, $methodName) &&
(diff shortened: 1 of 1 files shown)


daftspunk:
Hmm this should really be looking at the Profile model for options, then calling back to this concatenated version. Leave it with me.

daftspunk:
Ok, so this should be fixed in an upcoming build. The solution differs slightly to the proposed one here, instead you should define `getCountryOptions` inside your `Profile` model. If you really want to use `getProfileCountryOptions` in the `User` model then this method can be explicitly defined like so:


```
// Add tabs fields
$form->addTabFields([
    'profile[country]' =>
    [
        'label'   => 'Country',
        'tab'     => 'Profile',
        'span'    => 'left',
        'type'    => 'dropdown',
        'options' => 'getProfileCountryOptions',
    ]
]);
```

Note the `'options' => 'getProfileCountryOptions',` part.

Thanks for bringing this up!

PR #983, opened 2015-03-05
Added trigger on empty value

ChrisKanger:
With this code change the triggerAPI can also work with empty field values, for example:

fieldName:
type: text
trigger:
action: show
field: otherFieldName
condition: value[]


daftspunk:
LGTM

daftspunk:
LGTM

PR #986, opened 2015-03-06
Allow to pass 0 as column value in List widgets

ariponce:
**Problem**: When overriden the columns value, if we pass 0 or '0', the condition is not met and therefore the column has a null value.

**Solution**: Check for not null in the condtion


daftspunk:
Thanks for this!

PR #988, opened 2015-03-06
Polish translation fixes

keiosweb:
I've translated what remained untranslated and also changed a form in several places to neutral, previous translator used a masculine form in few places, while we sincerely hope that woman will use this CMS ;)


daftspunk:
Is nice!

PR #991, opened 2015-03-07
Respect radio-button value in form triggers

ebuster:
When I tried to use form triggers for radio button I found that this is impossible:

    field1:
        type: radio
        options:
            val1: ...
            val2: ...
    field2:
        type: ...
        trigger:
            action: hide
            field: field1
            condition: value[val1]

Proposed fix allows conditions like that.


daftspunk:
Thanks!

PR #992, opened 2015-03-08
Problem with using ColorPicker in modal window

kuzyk:


daftspunk:
This should be solved using CSS

daftspunk:
Added to #1021

PR #996, opened 2015-03-10
Fix relation controller JS issue on Chrome

ariponce:
When loading a relation, an "Uncaught token :" javascript error is thrown when using Chrome, which makes the select all checkbox not to work and most of the JS on the page to fail.


daftspunk:
Thanks

PR #997, opened 2015-03-11
Use External combiner paths

SaifurRahmanMohsin:
More efficient due to re-use of backend assets and adds transparency to location of backend assets.


daftspunk:
I'm not sure this is such a great idea, the frontend relying on the backend is bad practise.

SaifurRahmanMohsin:
Hmm. I didn't know that. I figured it's good re-use of code and also hides the require statements appearing in the front end so that way it even hides the location of the assets folder. Feel free to close the PR if you feel it's incorrect to use it this way.

daftspunk:
Doing away with the =require statements is a good idea though, will leave it open for that reason.

PR #1023, opened 2015-03-25
Force trailing slash

ariponce:
Do a 301 redirect if no trailing slash is present, to improve SEO.


daftspunk:
Laravel used to handle this internally, I think we should do it in PHP too since .htaccess is Apache-only

PR #1032, opened 2015-03-27
fixes HHVM relation formfield 

keiosweb:
Aliased Model class fails type check on hhvm, therefore relation data is not loaded. Affects belongsToMany in particular - checkbox list is not populated, ever.

Line 535 of FormField class is main the culprit:
if ($result instanceof Model && $result->hasRelation($key)) {

HHVM cannot resolve type if aliased dynamically, therefore use statement has to mention FQCN.


daftspunk:
Good good, thanks

PR #1039, opened 2015-03-29
Improved and added missing Swedish translation 

freestream:
Improved and added missing Swedish translation for core OctoberCMS modules.


daftspunk:
Thanks!

PR #1041, opened 2015-03-31
Updating .htaccess to correct protected uploads directory

chrisnharvey:
Since uploads are now stored in ```app/storage/uploads```. Should we not protect the ```app/storage/uploads/protected``` directory in .htaccess?


chrisnharvey:
Whoops! The directory is actually ```storage/app/uploads/protected```. I have corrected this in the PR.

daftspunk:
Fixed already! Thanks

PR #1043, opened 2015-04-01
Add missing ApplicationException dependency

dshoreman:
The relation backend formwidget throws an `ApplicationException` when a relation isn't defined. This exception is missing its `use` statement, causing a fatal error. See error below:

Screenshot showing error thrown


daftspunk:
Thanks!

PR #1044, opened 2015-04-02
Fixed custom error pages outputting headers

MatissJanis:
The bug:
screen shot 2015-04-02 at 09 55 05

If we don't have this method called, then the controller returns an object. From then on, [Symfony converts this object in to a string](https://github.com/symfony/HttpFoundation/blob/master/Response.php#L406). This string is **not just the content** of the view, but headers as well.

After attaching this method, `handleCustomError()` returns a string and everything works as expected.


daftspunk:
Merged this in to `develop`

PR #1048, opened 2015-04-06
Translations of error messages changed for pt-br

drFabio:
Changed some translations to be more organic for brazillian portuguese.
Some texts were wrong regarding the "gender" of the field so it was tryied to make some messages gender neutral by removing the "O attribute".
Some texts were also changed the email error message for example that was wrong regarding word order


tresbach:
I had fixed those on a bigger PR #1025 and now it can't be merged
Any reason it wasn't accepted, @daftspunk?

PR #1050, opened 2015-04-07
fixes getComponentProperties foreach iteration overriding method argument name

keiosweb:
Caused bug where properties of another page component where returned if cache was empty.

Problem was: argument name for method is $componentName and also foreach iteration key name is $componentName on line 365. Therefore, if cache was empty and method went into iteration part to retrieve full objectComponentPropertyMap, it was overriding method argument with iterator key and then tried to return properties for last component in iteration, not for requested component. Also, whole method could be separated into finder and obtainer methods to separate their responsibilities and lower method complexity (which obviously suffered from scoping issue due to size).


daftspunk:
Good find, thanks

PR #1054, opened 2015-04-08
Update Chinese (China) language

meteormatt:
Update Chinese (China) language.
Miss the `BackendPreferences.php` file.


daftspunk:
Thanks

PR #1055, opened 2015-04-08
Remove width for accountmenu

Rtransat:
In some languages the width is too small for see the full text


mrmlnc:
Excuse me, but:

> **Pull Requests**
>
> Do not edit compiled asset files such as `october.css` Instead, try to edit the LESS files inside the `less/` directory and then use a compiler.

daftspunk:
Duplicate of #1055

PR #1079, opened 2015-04-22
Add support for sorting sideMenu items

dshoreman:
Up until now, only main menu items have supported the 'order' parameter to set the sort order. This PR adds the same support to side menus as well.

This is especially useful when one plugin extends the side menu of another. With the existing functionality, if you were to use `$manager->addSideMenuItems()`, and menu items that you add will automatically be appended to the end of the existing ones provided by the plugin you're extending.

By adding the `order` parameter to side menus, plugin authors will be able to inject their extra menu item(s) wherever they want. However, this functionality comes at a small cost.

All existing plugins that have multiple items in their side menu are likely to find they now appear out of order. I've updated the `cms` module as part of this PR to maintain existing order. Plugin authors will need to manually update their `Plugin.php@registerNavigation()` methods if they don't want PHP to decide how their items should be sorted.

### DIFF
--- a/modules/backend/classes/NavigationManager.php
+++ b/modules/backend/classes/NavigationManager.php
@@ -47,6 +47,7 @@ class NavigationManager
         'url'         => null,
         'counter'     => null,
         'counterLabel'=> null,
+        'order'       => 100,
         'attributes'  => [],
         'permissions' => []
     ];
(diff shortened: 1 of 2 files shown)


dshoreman:
@daftspunk I'll be AFK as from about 06:00 my time (it's 22:50 right now) and won't get back until Tuesday night. On the offchance that you see this before I disappear, ping me on IRC if you need any extra info.

Once I'm back, I don't mind spending some time updating a few of the Rainlab plugins. Would be nice if users had new versions of plugins at the same time this gets released, although there's only so much we can do about that.

daftspunk:
I don't like that we have to update existing plugins, I think some method to assign defaults is needed, i.e take the existing order and incrementally add 100,200,300 to each item as a default (if not already defined). Will add this.

daftspunk:
Added best of both worlds in cceae628121b24decc99328f6f4d52fe075610e1
It's a bit magical because you'd need to calculate the order in increments of 100, but it should do the trick.

Thanks!

PR #1083, opened 2015-04-25
Artisan commands related to themes

krisawzm:
- Adds artisan command `october:fresh` for removing the demo included with October.

- Adds new artisan commands related to themes:
  `theme:list`     List all available themes (in marketplace.)
  `theme:use`      Switch the active theme.
  `theme:install`  Install a theme from the marketplace.
  `theme:delete`   Delete an existing theme.

This could make the development workflow for setting up a new project alot quicker.
For instance, take a look at https://github.com/krisawzm/blank-theme
This is just a blank scaffolding theme that you can build on. You could install it by simply running
`php artisan theme:install krisawzm.blank` - after running `october:fresh`, of course.

### DIFF
--- a/modules/system/ServiceProvider.php
+++ b/modules/system/ServiceProvider.php
@@ -310,7 +310,7 @@ class ServiceProvider extends ModuleServiceProvider
         $this->registerConsoleCommand('october.update', 'System\Console\OctoberUpdate');
         $this->registerConsoleCommand('october.util', 'System\Console\OctoberUtil');
         $this->registerConsoleCommand('october.mirror', 'System\Console\OctoberMirror');
-        $this->registerConsoleCommand('october.mirror', 'System\Console\OctoberFresh');
+        $this->registerConsoleCommand('october.fresh', 'System\Console\OctoberFresh');
 
         $this->registerConsoleCommand('plugin.install', 'System\Console\PluginInstall');
         $this->registerConsoleCommand('plugin.remove', 'System\Console\PluginRemove');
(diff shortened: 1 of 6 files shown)


modules/system/ServiceProvider.php
```diff
@@ -310,10 +310,17 @@ public function register()
         $this->registerConsoleCommand('october.update', 'System\Console\OctoberUpdate');
         $this->registerConsoleCommand('october.util', 'System\Console\OctoberUtil');
         $this->registerConsoleCommand('october.mirror', 'System\Console\OctoberMirror');
+        $this->registerConsoleCommand('october.mirror', 'System\Console\OctoberFresh');
+
```
daftspunk: There is a typo here, this needs to state `october.fresh`

daftspunk:
This needs to be documented in the help-console.md docs file

LucasZdv:
Hello,
I have to make a change of template through an SSH connection.
http://laravelcollective.com/docs/5.0/ssh
It does not work because the confirmation request may escape from it?

    public function switchTheme(){

       $commands = [
            "cd /srv/octomain/clients/one-project",
            "php artisan theme:use thistheme", # how i can confirm this command?
            "php artisan cache:clear"
        ];

        SSH::run($commands, function($line)
        {
            $this->response.= $line.PHP_EOL;
        });

        return $this->response;
    }

# response: 
********************************************************** * Do you really want to change the active theme? * ********************************************************** 
Command Cancelled! 
Application cache cleared!

PR #1087, opened 2015-04-26
Updates language nb-no

krisawzm:
Includes a critical commit which fixes a syntax error in `modules/system/lang/nb-no/validation.php` and should be pulled asap.


daftspunk:
Thanks!

PR #1088, opened 2015-04-27
Route parameters isn't filling when enableRoutesCache is enabled

shina:
This PR fixes the empty parameters when __enableRoutesCache__ is enabled.

I was testing my site with production configs and found this problem.


daftspunk:
I think the purpose of the cache is to not run getRouterObject(), I have fixed this in 10ec177c544d27b551d5d446321197fb49cdadfb

PR #1097, opened 2015-05-01
Remove list columns

arall:
[Extending list columns](https://octobercms.com/docs/backend/lists#extend-list-columns) allows to add new columns to a list, but not to remove them.

This PR allows it, so in controllers now its possible (for example) to do something like I needed:

```php
public function listExtendColumns($list)
{
    // Permisions
    if (!$this->user->hasAccess('acme.plugin.acces_clients')) {
        $list->removeColumn('client');
    }
}
```

### DIFF
--- a/modules/backend/widgets/Lists.php
+++ b/modules/backend/widgets/Lists.php
@@ -651,6 +651,18 @@ class Lists extends WidgetBase
     }
 
     /**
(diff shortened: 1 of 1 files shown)


daftspunk:
LGTM! Thanks

PR #1101, opened 2015-05-03
Create _field_string.htm File

astucieuxzephyr:
To have the ability of creating a string field (as it is shown in the video : Extending the User plugin - guide on adding extra fields to the RainLab.User plugin by Samuel Georges).

### DIFF
--- a/modules/backend/widgets/form/partials/_field_string.htm
+++ b/modules/backend/widgets/form/partials/_field_string.htm
@@ -1,0 +1,16 @@
+<!-- Partial String -->
+<?php if ($this->previewMode): ?>
+    <span class="form-control"><?= $field->value ? e($field->value) : '&nbsp;' ?></span>
+<?php else: ?>
+    <input
+        type="text"
+        name="<?= $field->getName() ?>"
+        id="<?= $field->getId() ?>"
+        value="<?= e($field->value) ?>"
+        placeholder="<?= e(trans($field->placeholder)) ?>"
+        class="form-control"
+        autocomplete="off"
+        maxlength="255"
+        <?= $field->getAttributes() ?>
+    />
+<?php endif ?>


daftspunk:
Can you help me understand how this is different to `text` form field type?

astucieuxzephyr:
Very good question. The string field is not in the list of available field types indicated in the [October's Documentation](https://octobercms.com/docs/backend/forms#field-types), and I don't know if there's any difference between `string` form field type and `text` form field type. But an error **_field_string.htm file is missing** occured for me when adding a `string` field in one of my migration files (in this case : `updates/create_profile_table.php` migration file).
To solve this error, instead of changing my field type (and put a `text` type instead of a `string` type in my `updates/create_profile_table.php` migration file), I added a new file `_field_string.htm` with the content of `_field_text.htm`, and it's perfectly working now for me.

thiagovarela:
@astucieuxzephyr, aren't you confusing october form widget types with laravel's schema builder types?
When you're defining your backend form, october uses yaml to define form fields and the text type to represent any type of char, varchar or text.  
What I think is that you tried to use **string** as a type when you extended the user controller.

astucieuxzephyr:
@thiagovarela, you are right, thank you for your answer. When extending the User Controller, I did this :
```php
            $form->addTabFields([
                'profile[about]' => [
                    'label' => 'About me',
                    'tab' => 'Profile',
                    'type' => 'textarea',
                ],
                'profile[role]' => [
                    'label' => 'Role',
                    'tab' => 'Profile',
                    'type' => 'string', <======= the mistake !!
                ],
            ]);
```
What I did at the beginning is just to follow the [video](https://vimeo.com/108040919) : at 2'13 you can see that Samuel is using a `string` type, (and not a `text` type), and just after, (at 2'17 in the video) he is using a `text` type. This misleaded me, and that's why I thought that `string` was a possible type for forms...

So, finally, what I purpose is : instead of creating a new *_field_string.htm* file, it could be useful to change the error caption to be more precise and tell the user something like : *string is not an allowed type in forms, use `text` type* instead of telling that the field file is missing.

PR #1103, opened 2015-05-03
Fix two minor media manager bugs

niclasleonbock:
* Fixed wrong link to video poster (`widgets` instead of `Widgets`).
* Fixed wrong link to public URL (add slash if it is external, e.g. `https://google.de/`).

### DIFF
--- a/modules/cms/classes/MediaLibrary.php
+++ b/modules/cms/classes/MediaLibrary.php
@@ -61,8 +61,11 @@ class MediaLibrary
     {
         $this->storageFolder = self::validatePath(Config::get('cms.storage.media.folder', 'media'), true);
         $this->storagePath = rtrim(Config::get('cms.storage.media.path', '/storage/app/media'), '/');
+
         if (!preg_match("/(\/\/|http|https)/", $this->storagePath)) {
             $this->storagePath = Request::getBasePath() . $this->storagePath;
+        } else {
+            $this->storagePath .= '/';
         }
 
         $this->ignoreNames = Config::get('cms.storage.media.ignore', $this->defaultIgnoreNames);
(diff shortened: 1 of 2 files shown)


daftspunk:
Not sure why the trailing slash is needed?

niclasleonbock:
Let's say we have a different domain we want our media data to be available from (e.g. `media.octobercms.com`).

The storage path is right-trimmed, which causes our URL to be faulty (e.g. `http://media.octobercms.commyimage.png` instead of `http://media.octobercms.com/myimage.png`). To avoid that I had to add a slash in the `MediaLibrary` class if the storage path is external (starts with http oder https).

alekseybobkov:
The logic doesn't seem to be right. The URL should not depend on whether the protocol specified on not. 

Media Manager was tested with external libraries (AWS and Rackspace). When URL of a file is requested from the library with [getPathUrl](https://github.com/octobercms/october/blob/master/modules/cms/classes/MediaLibrary.php#L399), the input path is processed with [validatePath](https://github.com/octobercms/october/blob/master/modules/cms/classes/MediaLibrary.php#L377), which adds a slash to the beginning of the input path, so that myimage.png is first converted to /myimage.png and then concatenated with the library URL. If it doesn't work in your case, there could be a bug in another place, or in the way how you use the classes. Please provide steps for reproducing the problem.

niclasleonbock:
In my specific use case I dont use an external library but want my files to be accessible at `cdn.domain.tld`. I use Syncthing to mirror my files on several different servers (which listen on `cdn.domain.tld`). Whether this is reasonable or not doesn't matter, it is in particularly just a test.

What I noticed is that if I configure `https://cdn.domain.tld/` as storage path in the media manager config, I get a faulty public URL (e.g. `https://cdn.domain.tldimage.png` instead of `https://cdn.domain.tld/image.png`, the slash is missing). Using my "fixed" code it works like a charm.

If the URL isn't supposed to depend on the protocol, why is there this regex thats checks for the protocol (and prepends the base path if it is not external)?

Maybe I just don't understand the purpose of the storage path?

PR #1132, opened 2015-05-13
Responsive scoreboard items

gergo85:


daftspunk:
Can you please explain what you're doing here? Screenshots or sample code would be helpful.

gergo85:
Before:
oc-before

After:
oc-after

daftspunk:
These are already responsive, you can click and drag with the mouse or with your finger on a touch device.

gergo85:
You are right. I didn't know about it.

Issue #1124, opened 2015-05-11
[MediaManager] Permissions


alxy:
Permissions need to be registered for the MediaManager to maintain access to the manager. Possibly one should be able to also allow/deny for specific content types (images, videos, ...).

Needs to be implemented here: https://github.com/octobercms/october/blob/master/modules/cms/ServiceProvider.php#L153-L183

Note: Maybe it is possible to also hide/show the icons in the richeditor widget according to the permissions of the current user, so he doesnt see these buttons if he isnt allowed to manage media files.

daftspunk:
Great, can you submit a pull request?

PR #1134, opened 2015-05-13
Add permissions for the MediaManager

alxy:
Refs https://github.com/octobercms/october/issues/1124

Not sure if that is how to do it, but it should work.


daftspunk:
Need to look at a more complete solution that hides the button from the rich editor too.

PR #1135, opened 2015-05-14
Adds option to set chart value by data-value

scottbedard:
This allows for more expressive charts by separating the value and the display value. For example, values like these could now be used without breaking the chart.

```
<div class="scoreboard-item control-chart" data-control="chart-pie">
    <ul>
        <li data-value="11024">Foo <span>~11k</span></li>
        <li data-value="1234">Bar <span>1,234</span></li>
    </ul>
</div>
```


daftspunk:
Looks good, thanks!

PR #1139, opened 2015-05-16
Fixed grid widget breaking backend menu scrolling.

justin-lau:
The issue is described in rainlab/translate-plugin#5.

It was caused by a `mousewheel` jquery plugin living in the `jquery.handsontable.js` file, which broke the `jquery-mousewheel` plugin already included in the `vendor-min.js` file.


daftspunk:
Well spotted. 

Just FYI - the handsontable (DataGrid) widget will be removed entirely soon, in favor of DataTable. It just needs some more features to become ready.

daftspunk:
@alxy You might be pleased to know this is fixed...

justin-lau:
Caused me quite a few hours digging..! But it was the last bone my client
can pick =)

alxy:
Yeah. Thanks @justin-lau for the fix!

Issue #1108, opened 2015-05-06
Bug rendering bar chart at various screen widths


scottbedard:
I've got a bar chart in a report widget that is loading irregularly at certain resolutions. My normal resolution of 1920x1080 loads just fine, but the chart gets progressively more distorted the lower the resolution goes.
![1920](https://cloud.githubusercontent.com/assets/7980426/7500725/7d0efcae-f3fd-11e4-8c7f-dd6b5b993882.jpg)
![1680](https://cloud.githubusercontent.com/assets/7980426/7500727/7f0b5296-f3fd-11e4-9857-4f3f7e4c1c12.jpg)
![1260](https://cloud.githubusercontent.com/assets/7980426/7500730/8179281e-f3fd-11e4-8826-23829e031e85.jpg)
![850](https://cloud.githubusercontent.com/assets/7980426/7500734/8436838a-f3fd-11e4-906f-27fbda30c510.jpg)

The distortion happens if I am resizing my window, as well as when I refresh a window at that size. I will poke around and see if I can find what's going wrong here.

scottbedard:
I wasn't able to find a solution to this bug, but I have determined that the bug starts occurring when `chartWidth` is less than `chartHeight`.

scottbedard:
Both tall and wide bar charts now work.

![screenshot 2015-05-29 20 01 22](https://cloud.githubusercontent.com/assets/7980426/7894732/ba978332-063d-11e5-8cdf-dc905d6010e9.png)

PR #1152, opened 2015-05-21
Solves issue #1108

scottbedard:
Finally found what was causing this issue. The base of the bars was being calculated using the chart width rather than the height. When the chart is wider than the height, this isn't a problem because the bottom of the bars will just extend beyond the bottom of the chart. But when the height is the larger value, the bottom of the bars was being shifted upwards.

Now we can have bar charts taller than they are wide =)

skinny-bar-chart


daftspunk:
Well done

PR #1153, opened 2015-05-21
Update of FRENCH translation for October CMS

Galixte:
Update for lang.php file.


Galixte:
Bad file (english instead of French :/).

PR #1165, opened 2015-05-26
allow context to be set on construct

ChadStrat:
currently the report container can no longer be utilized on other backend pages. (this used to work).  If context is not set on instantiation, the context reverts to 'dashboard' and cannot be set after initialization.  This code passively allows other controllers to utilize the widget and properly contextualize the use.

http://laravel.io/bin/E338m

### DIFF
--- a/modules/backend/widgets/ReportContainer.php
+++ b/modules/backend/widgets/ReportContainer.php
@@ -73,8 +73,11 @@ class ReportContainer extends WidgetBase
     /**
      * Constructor.
      */
-    public function __construct($controller)
+    public function __construct($controller, $context=null)
     {
+        if($context){
+            $this->context = $context;
+        }
         $configFile = 'config_' . snake_case($this->alias) . '.yaml';
         $path = $controller->getConfigPath($configFile);
         if (File::isFile($path)) {


daftspunk:
I think I may need to see this demonstrated, so it was working previous and suddenly stopped?

ChadStrat:
You and Aleksey are both contributors on the helpdesk app.  I just updated to the latest rework. I've only started reworking the backend so far.  (/helpdesk/controllers/Reports.php) But you can see and play with doing as you say - setting the property publicly from the controller.  It will:

- only save the 'dashboard' context to the "backend_user_preferences" table even if context is set.
- show the correct context when logging the $widget after setting the context even though the view render is actually the 'dashboard' context. (this to me means the context is pulled to early.)

My fix allows it to work by making sure the context is what I want.  But it might actually be a method ordering issue.  Please review.  https://github.com/ChadStrat/helpdesk-plugin

All I am after is a custom reportContainer (that used to work just fine.)

daftspunk:
Fixed by 8e3d3d5e305df99bf9e6615f2b1fadf93dec07c2

PR #1168, opened 2015-05-26
Add PHP .env (DOTENV)

jovib:
Loads environment variables from .env to env('LOAD_VARIABLES'),

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,8 @@
+APP_ENV=local
+APP_DEBUG=true
+APP_KEY=SomeRandomString
+
+DB_HOST=localhost
+DB_DATABASE=homestead
+DB_USERNAME=homestead
+DB_PASSWORD=secret


daftspunk:

jayalfredprufrock:
I hope you'll reconsider this at some point. I'd prefer to keep just my theme code version controlled so a deploy would simply be a `composer create-project` followed by a `git pull` regardless of whether I was deploying to a development, staging, or a production server. This is especially useful in the context of docker containers which makes working with environment variables particularly easy. Of course there are workarounds but there is a reason the Laravel community adopted the .env approach. Regardless, october does so many other things right I'll just have to deal with it :)

daftspunk:
Here's how you solve that:

1. Create a repo called *october-environment*
1. Just like the theme, git pull the repo to ~/config/thiscanbeanything
1. Set `APP_ENV="thiscanbeanything"`

Inside the repo include config files that use `env()`, eg **database.php**

```
<?php

return [
    'default' => 'mysql',
    'connections' => [
        'mysql' => [
            'host'      => env('DB_HOST', 'localhost'),
            'database'  => env('DB_DATABASE', 'forge'),
            'username'  => env('DB_USERNAME', 'forge'),
            'password'  => env('DB_PASSWORD', ''),
        ],
    ]
];
```

jayalfredprufrock:
Thanks @daftspunk. This is actually exactly what I am doing and it seems to work pretty well. Again, it's just an extra step that would be eliminated with the env() calls built in. I don't quite get the argument that env() adds complexity, since it's part of Laravel 5 now and with defaults specified and no .env file present it would work exactly the same as it does now. Anyways, I do appreciate the help. Thanks again

daftspunk:
A minor inconvenience for advanced deployments is the sacrifice we make for entry-level developer accessibility. Thank-*you* for working with us on this :-)

PR #1169, opened 2015-05-26
Loads environment variables from .env

jovib:
'key' => 'KEY_VARIABLE', 'debug' => 'DEBUG_VARIABLE',


daftspunk:

PR #1170, opened 2015-05-26
Loads environment variables to database config from .env

jovib:


daftspunk:

PR #1178, opened 2015-05-29
Fixed error when using {% flash type %}

dshoreman:
While `{% flash %}` works fine for displaying all flash messages, if a user wants to show only a certain type with `{% flash success %}` or `{% flash error %}` for example, it errors with **Undefined variable: message**.

This fixes that error (reported in IRC by user *lowerends*)


daftspunk:
Pending investigationing

PR #1181, opened 2015-06-01
Number of users by administrator groups

gergo85:


daftspunk:
This value should be eager loaded, will add this solution.

PR #1195, opened 2015-06-07
Fixes #1194

### LINKED ISSUE #1194: Can not use Repeater widget in theme data.
When adding a Repeater widget to the `form` section in the `theme.yaml` file, an exception is thrown when trying to save the form:

    "Unexpected type of array, should attribute "social_buttons" be jsonable?" on line 1064 of /path/to/october/vendor/october/rain/src/Database/Model.php

### Steps to reproduce:

- Set up a `theme.yaml` file with a repeater form field defined:

```
name: foo
author: bar

form:
    fields:
        something:
            type: repeater
            form:
                fields:
                    somefield:
                        label: I do not work
```

- In the backend, go to CMS > Front-end theme > Customize
- Save the form

krisawzm:
Repeater form fields store arrays and must be jsonable.

### DIFF
--- a/modules/cms/models/ThemeData.php
+++ b/modules/cms/models/ThemeData.php
@@ -75,6 +75,15 @@ class ThemeData extends Model
     public function afterFetch()
     {
         /*
+         * Repeater form fields store arrays and must be jsonable.
+         */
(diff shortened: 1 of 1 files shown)


daftspunk:
Thanks!

PR #1196, opened 2015-06-08
Fix showing some value instead of NULL

vojtasvoboda:
**Bug summary:** When relation has NULL value (or relation doesn't exist), dropdown shows first value, which can be confusing in update mode, because it looks like valid value of relation field.

**Reproduce steps:**
Case 1) Create record with relation to another entity. E.g. create Floor 1 within Building A. Delete related entity (Building A).
Case 2) Create record with relation to another entity. E.g. create Floor 1 within Building A. Set NULL to foreign key (object_id = NULL).


daftspunk:
Can't this be solved by defining an `emptyOption` for the form field?

vojtasvoboda:
Yes you are right, when I define emptyOption it works. But when developer forget to define emptyOption I just add condition to show empty option. But maybe it's too complicated.

daftspunk:
Yeah, I think it is too complicated. emptyOption should be used if this behavior is not desired.

PR #1197, opened 2015-06-08
Added more file extensions

gergo85:


daftspunk:
I dont think `ico` is supported by the IMG reference, is it?

gergo85:
You are right.

Issue #1133, opened 2015-05-13
Record Finder field is not bind by scope


mplodowski:
Record Finder field is not constrained by scope or conditions I specified in my model relation definition.

I have in Order model:

    public $belongsTo = [
        'client' => [
            'Total4Net\Clients\Models\Client',
            'scope'      => 'currentCompany'
        ]
    ];

and Record Finder is not appling my scope to query.

devillom:
Oh i need fix it (

rosuji-mms:
Bump to also needing a fix for this. Both scope & conditions are being ignored. To elaborate looking at the logs 3 queries get generated the first applies conditions & scope. But the later queries do not. I am not sure what is going on.

roscoecairney:
Bump - I'm having the same issue

daftspunk:
This is fixed in https://github.com/octobercms/october/commit/3bad70f0416b99aefb2c0ffbf7de2ad224361a77

PR #1203, opened 2015-06-11
applying relation scope model in RelationController

cybear-d:
this will apply relation scope model in relation behavior

### DIFF
--- a/modules/backend/behaviors/RelationController.php
+++ b/modules/backend/behaviors/RelationController.php
@@ -80,6 +80,11 @@ class RelationController extends ControllerBehavior
     protected $initialized = false;
 
     /**
+     * @var array Relationship definition
+     */
+    public $relationDefinition;
+
+    /**
      * @var string Relationship type
      */
     public $relationType;
@@ -300,6 +305,7 @@ class RelationController extends ControllerBehavior
          * Relationship details
          */
         $this->relationName = $field;
+        $this->relationDefinition = $this->model->getRelationDefinition($field);
         $this->relationType = $this->model->getRelationType($field);
         $this->relationObject = $this->model->{$field}();
         $this->relationModel = $this->relationObject->getRelated();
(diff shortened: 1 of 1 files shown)


daftspunk:
Can you provide a complete example (documentation) on how this is used?

Thanks

cybear-d:
if relation have scope in definition, relation controller will not applying it because it will use new query from list event.  So for consistency scope in relation, applying scope after it querying the new query for list widget will be better.

Example:
```
class Post extends Model
{
    public $belongsToMany = [
        'categories' => [
            'Acme\Blog\Models\Category',
            'scope' => 'isActive'
        ]
    ];
}

class Category extends Model
{
    public function scopeIsActive($query)
    {
        return $query->where('is_active', 1)->orderBy('name', 'desc');
    }
}
```

it will display the list for relation list widget with the definition scope isActive

daftspunk:
I had assumed this behavior is already applied as part of the Relation filter in the Model (`applyRelationFilters`). Can you prepare a sample plugin that demonstrates the issue?

cybear-d:
sorry for late response.

Relation filter in the Model ( applyRelationFilters ) only applied when the relation object is called.
RelationController widget list only using Relation Model from the relation object, so it will not call the applyRelationFilters (the widget only using query from the Model itself, not from the Relation).

```
$config->model = $this->relationModel;  // only the model, not the relation
....
$widget = $this->makeWidget('Backend\Widgets\Lists', $config);
```

Example:
https://github.com/cybear-d/oc-test-plugin

i think this issue also affect #1133 

if you want i'll make pull request for the record finder too.

daftspunk:
Installed the test plugin, thanks!

It would appear the List creates a new query and the basic constraints are being added, not those defined in the relationship. There should be a way to apply these "extra constraints" (`addDefinedConstraints`) in the model/relation API. This will result in some reorg of the Database package, so will have a look when I find some time.

cybear-d:
thank you, will wait for your update

PR #1206, opened 2015-06-14
Add a order field for file config form.

json-wong:
Add a order field for file config form, I think the file will be better able to sort.


daftspunk:
Not sure this is a good idea, it adds to confusion. The sorting feature has recently been improved.

PR #1209, opened 2015-06-15
Added more file extensions

gergo85:


daftspunk:
Couple of problems, flv is flash not HTML5 and webm is video not audio. I will fix this for you.

Thanks

PR #1210, opened 2015-06-15
getAttributes maxlength="255" Fix

LucasZdv:
This lets you change the maxlength from field config with "attributes"
Otherwise all input, use maxlength="255"


daftspunk:
Thanks - I fixed this using a slightly different approach

PR #1216, opened 2015-06-17
Fix csrf in framework js

keiosweb:
Related to: https://github.com/octobercms/library/issues/105, extracts XSRF token from cookie and injects it into AJAX request as a X-XSRF-TOKEN header (works exactly the same as in angular.js).


keiosweb:
Twin PR: https://github.com/octobercms/library/pull/116

keiosweb:
Posted updated PR with better, more global AJAX solution: https://github.com/octobercms/october/pull/1217

PR #1217, opened 2015-06-17
Fix csrf in framework js

keiosweb:
Related to: octobercms/library#105, extracts XSRF token from cookie and injects it into AJAX request as a X-XSRF-TOKEN header (works exactly the same as in angular.js).

Upgraded previous PR by implementing $.ajaxPrefilter. This will allow requests not using $.request syntax ($.ajax, $.get) to get X-XSRF-TOKEN header too.

### DIFF
--- a/modules/system/assets/js/framework.js
+++ b/modules/system/assets/js/framework.js
@@ -11,6 +11,21 @@ if (window.jQuery === undefined)
 
 +function ($) { "use strict";
 
+    /*
+     * Get XSRF-TOKEN from cookie
+     */
+    function getXsrfToken() {
+        var cookies = document.cookie.split(';')
+        var xsrfToken
+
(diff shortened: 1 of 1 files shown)


keiosweb:
Twin PR: https://github.com/octobercms/library/pull/116

daftspunk:
Still thinking about the best way to tackle this -- see https://github.com/octobercms/library/issues/105#issuecomment-113749344 for notes.

LukeTowers:
@daftspunk Please copy content from linked issue as the library repository no longer supports issues.

daftspunk:
Let's leave this for now. We can to keep the front-end open and free. We need a non intrusive way to approach it, perhaps CSRF can be implemented as a plugin.

PR #1222, opened 2015-06-19
Update MediaLibrary.php

b0gok:
Using $this->storagePath.$path in return output wil be, for example, '/october/storage/app/media/october/storage/app/uploads/public/553/655/04c/55365504c9cf5288658383.jpg'. It's wrong path to file.


daftspunk:
Will it? Where? Can you provide more detail?

daftspunk:
Unable to replicate.

PR #1224, opened 2015-06-20
Translate component group

scottbedard:
Just a minor convenience, I think it makes sense to have this property auto-translated as well.


daftspunk:
:+1:

PR #1229, opened 2015-06-22
Change a string

felixinx:
Change `Identifiant` to `S'identifier`. `Identifiant` mean `username` and `S'identifier` mean `Login`.


felixinx:
Also, the string is longer then the button, so we have problem when we display.

felixinx:
Sorry for the wrong branch!

PR #1258, opened 2015-07-06
Set more CSV parameters

vojtasvoboda:
Set more params for CSV builder.

**Reason:** I needed to set `;` as a delimiter.


daftspunk:
Nice one!

PR #1259, opened 2015-07-07
update zh-cn lang files

awaysoft:


awaysoft:
Update zh-cn lang file with en lang file

PR #1264, opened 2015-07-07
Added Greek language support

tsioukas:
This change add el value on locales array to enable Greek language support. Βackend, CMS and System Greek language files already pulled.


tsioukas:
Thanks, sorry for the wrong way I am new on GitHub. Later I will make some changes to translate some new phrases that I found in dev branch.

PR #1265, opened 2015-07-07
Login page - more space for other locales

vojtasvoboda:
More space for Deutsch and Czech login button. See before/after screenshot:

patch-login

Yes, it is not systematic solution, but better then before.


daftspunk:
Agreed, thanks!

Shahiem:
+1

PR #1279, opened 2015-07-11
Added new phrases on Greek Language

tsioukas:
Change/Insert new phrases from develop branch


tsioukas:
I see that in Master (latest build) you forgot to add Greek option for select list in Backend Plug plugin. Also there is some changes in previously commits.

daftspunk:
Thanks!

tsioukas:
There no problem in Backend Plug, it just take for the language selection the values of array "locale"
from modules/system/lang/en/lang.php. Now it's ok.
 
I will try to keep up to date the Greek language.

Thanks

PR #1285, opened 2015-07-13
Update Polish translation

bilgorajskim:


daftspunk:
Thanks! Please try to use less commits next time

Issue #47, opened 2014-05-13
Support for JSON and XML templates


ronnyandre:
Because of it's simple and extensible GUI, I'd like to use October as a backend for publishing XML/JSON data for an iPhone app. Is that possible?

PR #1286, opened 2015-07-13
Turn off replaceDivs

felixinx:
Hi,

One of the most anoying problem, can be saved easily. When we make code, we should be able to save it as **original** and not as the editor one.

See issue [#47 in `rainlab/pages-plugin`](https://github.com/rainlab/pages-plugin/issues/47)

### DIFF
--- a/modules/backend/formwidgets/richeditor/assets/vendor/redactor/redactor.js
+++ b/modules/backend/formwidgets/richeditor/assets/vendor/redactor/redactor.js
@@ -119,7 +119,7 @@
 		maxHeight: false,
 
 		linebreaks: false,
-		replaceDivs: true,
+		replaceDivs: false,
 		paragraphize: true,
 		cleanStyleOnEnter: false,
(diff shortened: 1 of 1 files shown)


jayalfredprufrock:
I really hope this gets pulled in. If replaceDivs is on by default then there should at least be some kind of backend option to turn it off. I'd build a plugin for this, but from what I understand there is no clean way of extending redactor right now short of destroying and recreating the instance on every page.

daftspunk:
You're modifying the vendor code. This is not allowed.

felixinx:
Ok, I understand. But how can we fix this?

Le samedi 25 juillet 2015, Samuel Georges <notifications@github.com> a
écrit :

> Closed #1286 <https://github.com/octobercms/october/pull/1286>.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/octobercms/october/pull/1286#event-365114429>.
>


-- 

Félix Desjardins

PR #1297, opened 2015-07-18
Translatable prompt on recordfinder

bombozama:
Prompt could only be hardcoded into yaml files. Now able to work with language files.


daftspunk:
Thanks!

PR #1304, opened 2015-07-22
Ensuring version order of database histroy.

psoots:
Currently there's nothing that ensures the order that migrations are run when refreshing a plugin. I discovered this when working with a postgres database that returned the history records in a completely different order than the mysql database did.


daftspunk:
Interesting, thanks!

PR #1306, opened 2015-07-24
Allow overriding number field pattern attribute

Flynsarmy:
This PR allows you to set a custom pattern for number fields using:
```yaml
myField:
    label: My Field
    attributes:
        pattern: [0-9]+
```

This is helpful for example if you only want positive integers entered, or only a 2 digit currency value etc. It's fully backwards compatible and will default back to the current `-?\d+(\.\d+)` if not specified.


daftspunk:
Thanks!

PR #1308, opened 2015-07-25
Update Chinese language

awaysoft:


modules/system/lang/en/lang.php
```diff
@@ -26,7 +26,7 @@
         'se' => 'Swedish',
         'sk' => 'Slovak (Slovakia)',
         'tr' => 'Turkish',
-        'zh-cn' => 'Chinese (China)',
+        'zh-cn' => '简体中文',
```
daftspunk: This should be the word "Chinese" in English

awaysoft: I think use native language will more friendly, maybe user does not know little English?

awaysoft: And Chinese has two:Simplified Chinese and Traditional Chinese, I think 简体中文 and 繁体中文 is more clearly to users.
PS:Traditional Chinese is used in Taiwan,China.

daftspunk: English (China) is fine. English should show English words.

awaysoft: OK,I have pull a new request.

PR #1315, opened 2015-07-27
Fix the modified items counter in the backend sidenav

### LINKED ISSUE #1309: Incorrect ItemCode in SideNav counter
In the sidenav template, the code reference a `$sideItemCode` which is a 0-indexed value while the javascript code for the counter update expect the text code (like *page* or *content*).

The `$sideItemCode` should be replaced by `$item->code`.

https://github.com/octobercms/october/blob/master/modules/backend/layouts/_sidenav.htm#L24

gabsource:
When a user edit an item in the CMS module, a red counter should appear.
The `_sidenav.htm` partial used the 0-indexed key `$sideItemCode` but the javascript (and documentation) uses the *string* code (the key in the configuration array of the plugin).
When changing `$sideItemCode` to `$item->code`, the counter shows up. - Fixes #1309

### DIFF
--- a/modules/backend/layouts/_sidenav.htm
+++ b/modules/backend/layouts/_sidenav.htm
@@ -21,7 +21,7 @@
                                 </a>
                                 <span
                                     class="counter <?= $item->counter === null ? 'empty' : null ?>"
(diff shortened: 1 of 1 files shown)


daftspunk:
Thanks!

PR #1316, opened 2015-07-27
add new language zh-tw / Chinese Taiwan

kiang:
as title


daftspunk:
Patched manually, thanks!

PR #1318, opened 2015-07-28
Fix the switching tab bug when calling ocTab.updateTab()

gabsource:
__How to reproduce:__

1. In CMS Page open 2 page tabs *(Error 404 and Error 500 for example)*
2. Select the second tab *(Error 500)*
2. In javascript console run command below to force the first tab to update
```js
$.oc.cmsPage.reloadForm($('form[data-change-monitor]:first'))
```
3. Now click on the first tab : 

 - The tab title is highlighted (*Error 404* is selected)
 - The tab pane (main content) is not updated (you still see the second tab content *Error 500*)

__Cause:__ 
*Bootstrap* tab.js uses the `.data('target')` to switch tab, but `ocTab.updateTab()` only changes the `attr('data-target')` ; *jQuery* does not re-read attributes after the first time (it's cached).

__Resolution:__
Call `$link.data('target', $link.attr('target'))` after the ìnitTab()` to update *jQuery* cache.

Proposed patch also update the `$link.data('title')` in case another library needs it.

After patching `modules/system/assets/ui/js/tab.js` we need to re-compile the assests
```sh
 php artisan october:util compile assets
```


daftspunk:
Yes, I see.. thanks for this!

PR #1321, opened 2015-07-29
Filter out empty CSV rows before import

Flynsarmy:
CSVs with an empty line at the end result in an import error. As this is a relatively common occurrence in the CSV world, we should filter out empty rows before hitting our import models `importData` method.


daftspunk:
Thanks

PR #1327, opened 2015-07-30
Fix height issues introduced in Chrome 45

dshoreman:
The `height:100%` rule applied to .layout is inherited by both the backend navbar, and the search box of the settings sidebar.

This fixes the issue by specifically applying `height:auto` to the affected elements.

2015-07-30-084236_960x512_scrot

### DIFF
--- a/modules/backend/assets/css/october.css
+++ b/modules/backend/assets/css/october.css
@@ -385,7 +385,7 @@
 .control-treeview ol > li.drop-target.has-subitems > div:before{background-position:0px -102px}
 }
 .sidenav-tree{width:280px;background:#34495e}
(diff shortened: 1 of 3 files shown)


daftspunk:
Yes, please update the source files too :+1:

flip4bytes:
Thanks for this fix! It has been broken in Canary for over a month now. Was just hoping eventually it'd fix itself, but doesn't seem like it was heading that way. I use Canary for development so it has been somewhat of a pain the past few weeks. Appreciate it!

dshoreman:
@daftspunk LESS files have now been updated. Travis is yet to re-run its checks but this should be good to go.

daftspunk:
Thanks!

Issue #1329, opened 2015-07-30
Relation controller interface in backend forms doesn't work with a belongsTo relationship if the primary / master record hasn't been created first


lthurston:
I have a fork of the oc-test-plugin that serves as a proof of concept. I had to make some modifications to it because this issue was obscured by virtue of which views the relation manager was set up in. Note that I removed some validation on name fields to get to the heart of the issue. With the validation enabled, the bug doesn't occur, but that's only because the primary / master record cannot be validated nor saved.

My fork is here: https://github.com/lthurston/oc-test-plugin

The readme provides a bug summary and instructions to experience the issue yourself.

Expected result: ideally, using the relation controller in unsaved records would work, and the unsaved master record would be created only once the record was saved by clicking "save" and not before.

Actual result: saving the slave record results in the creation of a blank master record with only the relationship field set. When the editor saves the record with their details, a second record gets created, without the relationship.

lthurston:
I wanted to follow up with some thoughts on the solution that I've provided here via this pair of pull requests. The core problem isn't so much that belongsTo relation and deferred bindings don't work, but more that the relationship controller will create the master record when the slave is saved. If the master record form itself created the master record along with the belongsTo relation to the slave when the user clicked "save", then this wouldn't be a problem.

It's a little counter-intuitive that deferred bindings would be needed on belongsTo relationships. The slave record is the first to be saved, and since the relation to that slave is on the master, why would you need to defer the binding at all? In most cases you probably don't. But when there's a degree of separation between the creation of the master and slave -- like in this exact case of the relationship manager interface -- deferred binding actually feels like an elegant solution. This situation is a case in point of the reason why disallowing deferred binding on belongsTo doesn't make a ton of sense. It's a edge case, but it's still a real case.

That said, this isn't the only solution to the problem. The other obvious solution would be to not save the master record when the slave is saved but instead save from the form only. This would likely require some conditional logic in the relationship manager and the master form. I haven't dug into this.

Although the changes that I provided in the pull requests may not be the perfect solution, I'm not sure the other solution I just mentioned is in any way better.

As always, I'm open to suggestions. But one way or another saving belongsTo relations via the relation manager should be possible in cases when the master record has already been saved and when it hasn't. The solution that allows the unusual case of deferred bindings for belongsTo feels like the right one to me. I'm happy to modify my changes in any way to get them into the core.

lthurston:
[...]

PR #1330, opened 2015-07-31
Creates deferred bindings for belongsTo relationships

lthurston:
This is related to https://github.com/octobercms/october/issues/1329

There's going to be an associated pull request in octobercms/library momentarily. With both of these change sets merged in, and the form configuration for the belongs to relationship controller set to deferredBinding: true in the configuration, the relationship can be established between an uncreated master and uncreated slave record. Not much additional testing has been done. This is primarily to prove that this issue can be resolved, although the code changes, particularly in octobercms/library are a bit on the smelly side.


daftspunk:
Fixed by https://github.com/octobercms/october/issues/1329

PR #1332, opened 2015-07-31
Reload relations in RelationsController::relationRefresh()

lthurston:
This pull request contains some changes from a previous pull request, on which I believe this change is dependent for this fix to work.

I'm not opening a new issue for this because it's so closely related to: https://github.com/octobercms/october/issues/1329

What this change does is reload the relations that wouldn't otherwise get updated before the slave preview form partial is regenerated. My specific example:

Exhibition relation:
  belongsTo => DigitalAsset

DigitalAsset relation:
  attachOne => SystemFile

The attached file information was not being updated when the re-rendering of the relation controller preview of the slave, so no thumbnail would appear based on the system file. By reloading the relations, I have access to the system file and can generate a thumbnail, even when the Exhibition itself hasn't been saved.

### DIFF
--- a/modules/backend/behaviors/RelationController.php
+++ b/modules/backend/behaviors/RelationController.php
@@ -393,6 +393,10 @@ class RelationController extends ControllerBehavior
     {
         $field = $this->validateField($field);
 
+        if(is_object($this->relationModel)) {
+            $this->relationModel->reloadRelations();
+        }
+
         $result = ['#'.$this->relationGetId('view') => $this->relationRenderView($field)];
         if ($toolbar = $this->relationRenderToolbar($field)) {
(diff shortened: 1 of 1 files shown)


daftspunk:
Should be fixed by #1329

lthurston:
It looks like this one didn't get merged (although perhaps you managed it in another way). I can open another issue or offer further explanation on #1329 for why this is necessary.

daftspunk:
I probably just need to see this demonstrated in the the test plugin, if it's a separate issue.

lthurston:
No problem. I'll put something together and will open a new issue. Easy to demo by putting an attachOne to a system file on the person, and previewing a thumbnail (or even just the file path) in the relation controller on phone.

PR #1342, opened 2015-08-03
Improve french translation (more grammatically accurate)

gabsource:
Prevents wrong messages like "Le auteur" (should be "L'auteur") by using "Le champ auteur".


daftspunk:
Squashed to 70a335e1fd7b671e6a433384762433bc7da90f9a

Thanks

PR #1343, opened 2015-08-03
Fix error when extending plugin by adding a new side menu item

gabsource:
When booting a plugin _which extends another plugin side menu_ like in the code below, the framework complains about `StdClass:$owner` not existing. This because `addSideMenuItems` method was not setting the `code` and `owner` properties (provided as the array key or call parameter). This fix adds both missing properties by merging the array with default values (like in `addMenuItems`).

```php


    public function boot()
    {

        Event::listen('backend.menu.extendItems', function ($manager) {

            $manager->addSideMenuItems('RainLab.User', 'user', [
                'foo' => [
                    'label'       => 'author.plugin::lang.menu.label',
                    'icon'        => 'icon-leaf',
                    'url'         => Backend::url('author/plugin/foo'),
                    'permissions' => ['author.plugin.access_foo']
                ]
            ]);

        });

    }
```

__Without this fix, the error below is thrown:__

    Undefined property: stdClass::$owner
    modules/backend/Classes/NavigationManager.php line 424


daftspunk:
Thanks again!

PR #1345, opened 2015-08-04
Make backend auto closing option label translatable

powelski:
`Auto close tags and special characters` option in Code Editor preferences is currently non-translatable. This pull request fixed this and provides English translation as well, which serves as a fallback for all other languages.


daftspunk:
LGTM!

PR #1347, opened 2015-08-04
Escape HTML for .txt contents

### LINKED ISSUE #1272: .txt content have HTML interpreted
Content files with .txt extensions are interpreted like HTML. I think they should be always escaped by default, because this is supposed to be plain text, not HTML.

powelski:
Close #1272


daftspunk:
Thanks!

PR #1351, opened 2015-08-05
Update bootstrap plugins

gpasztor87:


daftspunk:
Why do we need to update these?

PR #1353, opened 2015-08-06
Remove 'compact-container' body class from import and export view.

gpasztor87:


daftspunk:
It would appear your controller is polluting these actions with a body class, you should narrow the scope of this logic, or use the proposed solution externally:

    public function export()
    {
        $this->bodyClass = '';
        return $this->asExtension('ImportExportController')->export();
    }

PR #1355, opened 2015-08-06
Fix back-end auth layout, storm.css and storm.js are missing

gabsource:
The back-end `auth.htm` layout needs to load `storm.css` and `storm-min.js` because they are not bundled in other files.

Without this import the back-end login form is not styled and javascript errors are thrown because `$.oc.foundation` does not exist.


daftspunk:
Oops, fixed in 891c034496b99e7e4986bdbafd38825b567b0190 before I saw this

PR #1360, opened 2015-08-08
Show mediamanager sidebar preview when "Go up" is selected.

jofrysutanto:
Currently when "Go up" is selected (by using keyboard arrow keys or other means), the sidebar will display empty file information, with `null` public url.

select-go-up-before

This PR adds additional template to sidebar which hide file information and show "Go up" label with `icon-level-up` 

select-go-up-after


daftspunk:
Very cool! Thank-you!

daftspunk:
We liked this so much we thought we could improve upon it some more by allowing the "Go Up" to be selected with the mouse too.

jofrysutanto:
Cool stuff! Keep up the good work.

PR #1362, opened 2015-08-08
Disable SweetAlert forced focus when a Bootstrap modal is already open in back-end

gabsource:
Bootstrap is enforcing focus on it's modal box, sweet alert too. Focus on sweet alert buttons is not really important, keyboard actions are still available. 

This patch prevents SweetAlert from requesting focus if Bootstrap modal is present in DOM (checking presence of an element with `.control-popup.modal` selector).

capture d ecran 2015-08-08 20 30 11

Assets must be recompiled

     php artisan october:util compile assets


daftspunk:
Thanks!

Issue #1356, opened 2015-08-06
Plugin migration files not being refreshed in correct order.


justin-lau:
The change in #1304 makes plugin migration files not refreshing in the correct (reversed) order with the `plugin:refresh` command.

**Reproducing:**
1. Create a plugin with 2 migration files that have models with a one to many relationship.
2. Make sure they are listed in correct order in `version.yaml`.
3. Run `php artisan plugin:refresh` for the first time and it should migrate properly.
4. Run `php artisan plugin:refresh` again. It should throw a foreign key violation SQL exception.

Commenting out the PR's code does resolve the issue.

lazycms:
兄弟 看英文老费劲了

gabsource:
Can not reproduce your issue with the step suggested. __The rollback succeed even with #1304__.

Test files : 

```yaml
1.0.1:
 - First
 - create_first_table.php
1.0.2:
 - Second
 - create_second_table.php
```

```PHP
<?php namespace RainLab\User\Updates;

use Schema;
use October\Rain\Database\Updates\Migration;

class CreateFirstTable extends Migration
{

    public function up()
    {
        Schema::create('fix_demo_table_1', function($table)
        {
            $table->engine = 'InnoDB';
            $table->increments('id');
            $table->timestamps();
        });
        echo '1';
    }

    public function down()
    {
        Schema::dropIfExists('fix_demo_table_1');
    }

}

```

```PHP
<?php namespace RainLab\User\Updates;

use Schema;
use October\Rain\Database\Updates\Migration;

class CreateSecondTable extends Migration
{

    public function up()
    {
        Schema::create('fix_demo_table_2', function ($table) {
            $table->engine = 'InnoDB';
            $table->integer('fix_id')->unsigned();
            $table->foreign('fix_id')->references('id')->on('fix_demo_table_1')->onDelete('restrict');
        });
        echo '2';
    }


    public function down()
    {
        Schema::dropIfExists('fix_demo_table_2');
    }

}

```

```
Rolled back: Fix.Version
Reinstalling plugin...
Fix.Version
 - v1.0.1:  First
 - v1.0.2:  Second
```

justin-lau:
@gabsource try using both migration files in a same version, and you should see the exception:

```yaml
1.0.1:
 - First
 - create_first_table.php
 - create_second_table.php
```

gabsource:
Confirmed, the sorting from PR #1304 was inconsistent. Patch proposed in PR #1366.

[...]

PR #1366, opened 2015-08-09
Ensure versions history are retrieved by id ASC (bug introduced by PR #1304)

gabsource:
Fix issue #1356 introduced by previous PR #1304.

We can't just use `version_compare` because there are multiple entries for the same version number (comments and scripts). The items should have been inserted in the correct, so forcing sort order by `id ASC` should prevent strange default sort order from RDBMS


psoots:
This looks good to me.

daftspunk:
Good stuff! Thanks!

PR #1373, opened 2015-08-14
Fix component partials not being rendered with multiple components and nested partials

gabsource:
There was a bug when, in a partial, the user includes multiple component.

__theme/partials/header.htm__
```htm
[Menu menu]
code = "main-menu"

[account]
paramCode = "code"
==
{% component 'menu' %}
{% component 'account' %}
```

__theme/partials/menu/default.htm__
```htm
<ul id="nav-main">
    {% partial __SELF__ ~ "::items" items=__SELF__.menuItems %}
</ul>
```

__theme/partials/account/default.htm__
```htm
{% if not user %}
{% partial __SELF__ ~ '::signin' %}
{% else %}
{% partial __SELF__ ~ '::activation_check' %}
{% partial __SELF__ ~ '::logout' %}
{% endif %}
```

`renderPartial()` in `Cmd\Classes\Controller` is called for `theme/partials/header.htm`, the `partialComponentStack` is built like this :
- Component `menu` is push
- Component `account` is push

Then `menu` component is rendered, and recursively, `renderPartial` is called with the same controller instance on nested partials. The last `renderPartials` of nested elements calls `array_pop` (but we are in a nested partial rendering not the root) this will pop `account` (not `menu` so menu is not rendered because it is not fount in the stack anymore), the the call is going up the stack and the another call to `array_pop` is made at the end of `menu` rendering (the stack is empty).

xdebug

There is two point here : 
- the components are added in the settings order, but the markup can reference them in any order so we can't just `array_pop` or `array_shift`
- the nested partials must not remove elements from the stack as they did not push element to the stack.

The proposed patch keeps track of the root partial of a component (the first time it encounters the alias), then let nested partials render and removed itself from the stack at the end.

### DIFF
--- a/modules/cms/classes/Controller.php
+++ b/modules/cms/classes/Controller.php
@@ -898,12 +898,6 @@ class Controller
         $result = $template->render(array_merge($this->vars, $parameters));
         CmsException::unmask();
 
-        if ($partial instanceof Partial) {
-            if ($this->partialComponentStack) {
-                array_pop($this->partialComponentStack);
-            }
-        }
-
         $this->vars = $vars;
         $this->componentContext = null;
         return $result;
--- a/modules/cms/classes/Controller.php
+++ b/modules/cms/classes/Controller.php
@@ -781,9 +781,9 @@ class Controller
                         return false;
                     }
                 }
-            /*
-             * Component alias is supplied
-             */
+                /*
+                 * Component alias is supplied
+                 */
             }
             else {
                 if (($componentObj = $this->findComponentByName($componentAlias)) === null) {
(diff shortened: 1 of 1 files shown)


daftspunk:
This is a complicated issue. I see another problem in that array_push() can be potentially called multiple times (if multiple components are used) but array_pop() is only called once per partial.

Can you please prepare a test theme that demonstrates the issue?

gabsource:
I've cloned Vanilla and put my test on the homepage
https://github.com/gabsource/vanilla-theme.git

demo-theme

This illustrate my first message.

gabsource:
With a level-2 nested partial and PR patch it's stills display the component but each component must have a unique alias through all the page (all the components, partials including on the page). If the same name is used they are not rendered.

daftspunk:
I'm getting a missing partial error:

```
An exception has been thrown during the rendering of a template ("The partial 'account::logout' is not found.").
```

gabsource:
I don't have have those partials in the overloaded theme, they are loaded throm RainLab plugins on my installl. I've just added 2 partials in the theme just in case.

daftspunk:
The problem is we are stacking components and unstacking partials. We need to be stacking partials, storing their components and unstacking partials.

Solution incoming...

daftspunk:
Fixed by 85933facbc41d6a78ec52076222854ac71dfc04a

Thanks for bringing this to our attention

PR #1374, opened 2015-08-16
Fix Repeater widget behavior.

voidshah:
Bug's demonstration:
fields.yaml
```yaml
fields:
    testcol:
        type: repeater
        form:
            fields:
                title:
                    label: ITEM
                item:
                    type: repeater
                    prompt: Add new subitem
                    form:
                        fields:
                            subtitle:
                                span: left
                                label: subitem
```
I create item with two subitems:
repeater1
When if I add a new item, I get the following picture:
repeater2
But I expected get the following:
repeater3
But the data saved in database properly. I should just  delete excess subitems manually.  Bug occurs in backend only when I create a new ITEM.

### DIFF
--- a/modules/backend/formwidgets/Repeater.php
+++ b/modules/backend/formwidgets/Repeater.php
@@ -99,6 +99,8 @@ class Repeater extends FormWidgetBase
             $loadValue = array_keys($loadValue);
         }
 
+        if (!$loadValue) return;
+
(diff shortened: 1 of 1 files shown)


voidshah:
Sorry, I spent a deeper testing and discovered that my solution is wrongю (((

PR #1381, opened 2015-08-18
Fix Repeater formwidget behaivor when using nested repeaters.

voidshah:
fields.yaml
```yaml
fields:
    testcol:
        type: repeater
        form:
            fields:
                title:
                    label: ITEM
                item:
                    type: repeater
                    prompt: Add new subitem
                    form:
                        fields:
                            subtitle:
                                span: left
                                label: subitem
```
I create item with two subitems:
repeater1
When if I add a new item, I get the following picture:
repeater2
But I expected get the following:
repeater3

This patch solves the problem. I made tests with two  and three nested repeaters and have not found errors.

### DIFF
--- a/modules/backend/classes/Controller.php
+++ b/modules/backend/classes/Controller.php
@@ -120,10 +120,19 @@ class Controller extends Extendable
     protected $statusCode = 200;
 
     /**
+     * @var bool Event flag. If Repeater->onAddItem has occurred, it is set to true.
+     */
+    protected $onAddNewRepeater = false;
+
+    /**
      * Constructor.
      */
     public function __construct()
     {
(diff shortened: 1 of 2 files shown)


daftspunk:
Maybe we should include an option to set "do not populate with a blank item by default"

voidshah:
I thought about it, but i think then we can not save the empty items. Сan we create two consecutive empty items after? In my version, you can save a tree repeaters with empty items.

daftspunk:
Nested repeaters are not officially supported because it opens a big can of worms, so you should use them like this at your own risk (ideally for simple functions only).

I'm not sure we need to be polluting the backend controller with these concerns, so I have modified the code to do the same using a static property on the Repeater widget itself.

Thanks!

Issue #1384, opened 2015-08-20
Lang SE folders are wrong.


bjorngrunde:
The Swedish lang folders in cms, backend and system are named "SV", but the system expects "SE". The translation wont work unless you change the name of the folders.

daftspunk:
Can someone please submit a pull request to change this?

gabsource:
Nothing wrong with the code.
sv	Swedish
se	Sami (Northern)

SV as in `https://sv.wikipedia.org` for (Swedish Wikipedia)

bjorngrunde:
No. Just no. Im looking at the folder structure right now here on github. There is no SE folders. Wich should be swedish. SV is the swedish shortname for Sweden(**SV**erige) in swedish. The global is SE and the system expects SE folders that does not exist, so i cannot translate backend by option in profile>settings unless i add the folders myself. wich i did..

gabsource:
`lang` folder names reflect language code  
- `en` like in `en_US` 
- `sv` like in `sv_SE`.
The ISO 639 language for Swedish is `sv` as stated here : http://www.loc.gov/standards/iso639-2/php/code_list.php, the region code is `SE` or `FI`. But October does not use region code just the langague code.

To use the `sv` locale (Swedish) you must put ` 'locale' => 'sv', ` in your `config/app.php` file. Then the lang files in `sv`folder will be used. And if you have created `se` folders for new translations you should rename them to `sv` (or use existing `sv` folder). If the system is looking for `se`, it's because you've written `se` in the `config/ app.php`, no ?

bjorngrunde:
Setting the locale to se was the first thing i did and choose the language in the backend still did not work. Not untill i created new folders. Try it yourself, new install and try change backend Lang to swedish from settings, they expect se folders, not sv.

gabsource:
The code is wrong in the `en` locales list which is used everywhere to list locales. PR submitted to fix it.

bjorngrunde:
Thank you, much appreciated! :)

daftspunk:
Fixed by #1395

PR #1395, opened 2015-08-26
Fix Swedish language code reference 'se' to 'sv' to reflect ISO code

gabsource:
An mentioned in issue #1384 the system was looking for `se` folder instead of `sv` folder for Swedish language. `sv` is the ISO code for Swedish like in locales `sv_SE` or `sv_FI`.

PR includes the compiled JS (renamed from `se` to `sv`) and also the Hungarian which were not compiled (still english).

The flag-icon is using another ISO standard than for locales, that's why the class name has not changed.


gabsource:
Other solution is to rename all `sv`folder to `se` but `se` is not an the ISO code for Swedish.

daftspunk:
Thanks for this!

Issue #1378, opened 2015-08-18
The Emmet is not working properly


gergo85:
When I click on the TAB key I get the following html code:

br -> `<br></br>`
img -> `<img></img>`
input -> `<input></input>`

The original Emmet plugin of Ace Editor is working properly.

luizbills:
+1

daftspunk:
Fixed by #1396

PR #1396, opened 2015-08-26
Fix Emmet expand by forcing syntax to html when editor is set to Twig

gabsource:
As mentioned in issue #1378 Emmet does not recognized the `twig` syntax passed by the editor when interpreting commands. 

`codeeditor.js` is patched to force `html` syntax for Ace Emmet extension when syntax is set to `twig`.
`emmet.js` did have encoding errors in strings and comments.

JS assets must be recompiled : 

    php artisan october:util compile js

After the patch is applied : 

    br + [tab]    -->     <br>
    input + [tab]    -->     <input type="text">

### DIFF
--- a/modules/backend/formwidgets/codeeditor/assets/js/codeeditor.js
+++ b/modules/backend/formwidgets/codeeditor/assets/js/codeeditor.js
@@ -338,6 +338,24 @@
         this.editor.focus()
     }
 
+    // FIX EMMET HTML WHEN SYNTAX IS TWIG
+    // ==================================
+
+    +function (exports) {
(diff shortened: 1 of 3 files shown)


daftspunk:
Good find, although this will likely get wiped out next time Ace is updated. Is there a way to implement this behavior externally from the vendor files?

gabsource:
I think we can override the AceEmmetEditor prototype in the codeeditor.js so we could inject a patch there. I'll give it a try.

gabsource:
Moved the patch to `codeeditor.js` by overriding the `AceEmmetEditor` prototype `getSyntax`

daftspunk:
Rockin'!

Thanks

Issue #1389, opened 2015-08-23
Plugin permissions should be automatically be tabbed by PluginName


jBOKA:
I think Plugin permissions should be either tabbed/grouped automatically by PluginName or the docs should be updated to include a hint to use the `tab` array key under `registerPermissions()` to group them. Otherwise one comes up with an ampty tab name heading in the permissions tab.

daftspunk:
Fixed the empty tab and added tab to the docs, thanks!

PR #1397, opened 2015-08-27
Set plugins permissions default tab title to plugin name

gabsource:
As mentioned in Issue #1389 the default tab title is blank if not provided in the definition array.

This patch uses the plugin details to extract the plugin name and use it if the permission definition does not provided a tab item.


daftspunk:
I think we might keep it simple and go with the standard undefined tab string.

daftspunk:
Thanks!

PR #1398, opened 2015-08-27
Made $viewPath public

MatissJanis:
You may be asking: **why?**

Because if `$viewPath` is a public variable, we can simply extend the backend forms with custom view files. For example, I don't like the markup of tabs. With this change in place I can simply change this markup in a new file.

```php
<?php
Event::listen('backend.form.extendFields', function($widget) {
    $widget->viewPath = [getcwd() . '/themes/acme/widgets', $widget->viewPath];
});
```

Now I can create a new file `_form_tabs.htm` in the `themes/acme/widgets/` folder and I don't need to worry about the OctoberCMS system changing.

### DIFF
--- a/modules/system/traits/ViewMaker.php
+++ b/modules/system/traits/ViewMaker.php
@@ -23,7 +23,7 @@ trait ViewMaker
     /**
      * @var string Specifies a path to the views directory.
      */
-    protected $viewPath;
+    public $viewPath;
 
     /**
      * @var string Specifies a path to the layout directory.


psoots:
I'd rather see some comprehensive feature that makes it easier to add to or override widget views and partials.

daftspunk:
This is quite a fragile solution since the view files are tightly coupled to their owning classes, a minor change could see the implementation break because there is no mediating logic.

You might be better off building a generic wrapper form widget that handles this. Here is an example: https://github.com/rainlab/translate-plugin/blob/master/formwidgets/MLRichEditor.php#L72-L84

MatissJanis:
A generic wrapper does work for simple form widgets, but this solution would not work for the tabs widget. Only viable solution that I could think of was this implementation.

Do You have a better idea how this could be done?

daftspunk:
If I understand correctly you're needing this [to use back-end forms in the front-end](https://octobercms.com/forum/post/using-backend-forms-in-frontend). This might be considered a dangerous idea because of the reason I mention above, the solution is fragile and can break easily by changes in the back-end system. I should point out the instructions in the "How to" article are already broken because the file `/modules/backend/assets/css/controls.css` no longer exists. October's front and back-ends are decoupled by a strong API so system updates should never break the front end.

If you really want this, to make it safe, you would need to advise developers to disable core updates (cms.disableCoreUpdates) and never update their systems. This could easily be coupled with an instruction to make the `$viewPath` property public. This creates other problems and is probably not such a great idea either.

My advice would be to not cut corners and build a complete and dedicated front-end system that can generate forms and lists from YAML files, based on the back-end system that you want (copy and paste), rather than trying to extend it.

MatissJanis:
I understand Your concerns with this and have to agree that this is a dirty approach.

Is there any developement or plans on doing something similiar to the forums post (frontend forms via yaml)? Many plugins currently are using their own hard-coded frontend forms. I believe it can be done much better and in a cetralized way.

Thanks!

Issue #1399, opened 2015-08-27
The "SSL required" backend smtp mail setting not used for sending mail


jajacobs:
Sending mail always uses the default encryption setting (TLS) in mail.php  even when "SSL required" is checked in backend mail setting.

daftspunk:
Fixed by #1400

PR #1400, opened 2015-08-27
Fix SMTP encryption settings to reflect Laravel encryption config.

gabsource:
In reference to issue #1399, the mail settings interface was displaying a `SSL` checkbox that wasn't used anywhere in the code. This patch replaces the SSL checkbox with a dropdown list which allow the user to select an encryption option (none, tls, ssl).

The configuration is applied in the Laravel `mail.encryption` setting.


daftspunk:
Thanks!!

Issue #1390, opened 2015-08-24
After backend form fields are hidden via a trigger, tabs without any visible fields should be hidden too


lthurston:
Currently one can set triggers on fields so that they show / hide based on the values of other fields. This show / hide should -- in a perfect world -- bubble up to the tab level, and tabs with no fields should hide as well. This certainly would add some complexity to the form JS, but would be so magical.

daftspunk:
[We have logic for this already](https://github.com/octobercms/october/blob/master/modules/backend/widgets/form/assets/js/october.form.js#L134-L147) that is applied when `dependsOn` is used, but not for `trigger`, should be pretty simple to implement.

lthurston:
Could dependsOn replace trigger in this particular case? I'm not seeing an example in the documentation that is doing value-based dependent show / hide.

daftspunk:
[This is probably what you're after](http://octobercms.com/docs/backend/forms#filter-form-fields). However, good news: it looks like there is a PR incoming that will resolve the issue for the trigger API!

lthurston:
Woot!

daftspunk:
Fixed by #1401

This will become available in Build 293+

PR #1401, opened 2015-08-27
Hide FormWidget tabs when they contain no fields or only hidden field

gabsource:
In reference to issue #1390 this patch triggers  `toggleEmptyTabs` on the parent `FormWidget` when a form field with a `toggle` action is updated.

In `input.trigger.js` a new event `oc.triggerOn.afterUpdate` is triggered after an update.

In `october.form.js` the event is catched to trigger the tab refresh. The CSS selector for "hidden tabs" has been updated because the toggle action doesn't remove the form groups but hides them with a `hide` class.

### DIFF
--- a/modules/backend/widgets/form/assets/js/october.form.js
+++ b/modules/backend/widgets/form/assets/js/october.form.js
@@ -34,6 +34,7 @@
         this.bindCheckboxlist()
         this.toggleEmptyTabs()
         this.bindCollapsibleSections()
+        this.bindEvents()
 
         this.$el.one('dispose-control', this.proxy(this.dispose))
     }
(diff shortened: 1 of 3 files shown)


daftspunk:
Thanks for this @gabsource, I think it would be better if there were a clear separation of concerns and this entire set of logic be contained inside october.form.js withjout polluting input.trigger.js. Perhaps the `oc.triggerOn.update` event could be used?

gabsource:
Yes, you're right, it needs separation. The `oc.triggerOn.update` event is more an action event ("do this") than a reaction event ("this has been done"), no ?
> - oc.triggerOn.update - triggers the update. Trigger this event on the element the plugin is bound to to force it to check the condition and update itself. This is useful when the page content is updated with AJAX.

Maybe the `triggerOn.updateTarget` should fire a new event `oc.triggerOn.afterUpdate` that would bubbles up and get catched by the formwidget ?

daftspunk:
Ah yes, of course. Sounds good!

gabsource:
I'll look into this.

gabsource:
I've squashed the commits and updated the PR text to reflect the changes.

daftspunk:
Thanks!

PR #1408, opened 2015-08-31
Fix #1407

### LINKED ISSUE #1407: Relation Manager with pivot data and validation in pivot model.
I used pivot model and added validation trait into it. I added some validation rules. 

I opened relation manager popup with pivot data. Filled form data and submit it. I received validation error, but popup was closed and there is no way to correct form errors.

That bug occurs only when using pivot model with relation manager.

Popup should reopen if there were any validation errors.

mplodowski:
Now when validation error occurs the popup will reopen and user will be able to correct form.

### DIFF
--- a/modules/backend/behaviors/relationcontroller/partials/_manage_pivot.htm
+++ b/modules/backend/behaviors/relationcontroller/partials/_manage_pivot.htm
@@ -38,3 +38,9 @@
         </div>
     <?= Form::close() ?>
 </div>
+<script>
+    setTimeout(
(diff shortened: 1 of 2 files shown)


daftspunk:
Looks good! Thanks!

daftspunk:
I noticed when the pivot model fails validation, the object(s) are still added to the relationship. This is fixed using a database transaction (e8b4bf9d72cbbfc3458977d4fb6dccea88b6d812).

PR #1413, opened 2015-09-01
Added permission for /backend/system/settings

joris-vdlp:
## Permission for Settings menu

I have added permission for /backend/system/settings overal page and backend preferences setting page.

With this option you can exclude the backend user(s) for the settings part in the backend. Useful for user that only are allowed to manage content for example.

If you are developing a plugin thats has to be accessed thru backend/settings controller you should add this __'permissions' => ['system.general']__ to you __registerSettings__ in your plugin in order to be in sync with the permissions.

In the permissions page for groups or users i have added 2 extra options, __Backend preferences__ and __Manage general settings__ .

## Why??
We needed this for a client project, we dont want the customer to have access to settings part of the backend, and dont want to explain why the settings menu item is there and why he may not use it! :)

### DIFF
--- a/modules/backend/ServiceProvider.php
+++ b/modules/backend/ServiceProvider.php
@@ -113,6 +113,10 @@ class ServiceProvider extends ModuleServiceProvider
                 'backend.manage_branding' => [
                     'label' => 'system::lang.permissions.manage_branding',
                     'tab'   => 'system::lang.permissions.name'
+                ],
+                'backend.preferences' => [
+                    'label' => 'system::lang.permissions.manage_backend_preferences',
+                    'tab'   => 'system::lang.permissions.name'
                 ]
             ]);
         });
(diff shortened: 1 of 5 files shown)


daftspunk:
We are always nervous when [a PR has to explain why](https://github.com/octobercms/october/pull/1398), since a lack of necessary explanation is a hallmark of simplicity. This proposed solution certainly would solve the problem for your customer, however it may add a level of complexity for other developers.

Take a look at the recent addition to the `NavigationManager` class (1c8d021a358fc6ee17da422ccd99fe570ce7912f) which allows you to dynamically remove main items using events. Here is some example code:

    {
        [...]

        public function boot()
        {
            \Event::listen('backend.menu.extendItems', function($manager) {

                $manager->removeMainMenuItem('October.System', 'system');

            });
        }
    }

joris-vdlp:
Thanx for your feedback @daftspunk. Your example sounds good! Next time i will contact you before i do a pull request. I think you can decline my pull request for now :) and i have to build a plugin todo this kind of actions.

daftspunk:
Thanks!

PR #1416, opened 2015-09-02
Fix evaluate zero value as null

vojtasvoboda:
**Bug summary:** Zero value in backend list is represented as null value. But zero is normal value as any other.

**Expected result:** Zero is printed as "0", not as "".

**Use case:** When I have product in eshop a want to save stock level. _Zero_ represents empty stock (all items sold) and _null_ represents stock waiting for fill its value.

**Solution:** Fix condition.

**Screenshot** before/after with testing dump() which show real value in database:

null


daftspunk:
Thanks!

Issue #922, opened 2015-02-01
contain composer.lock in repository


wppd:
see this article https://blog.engineyard.com/2014/composer-its-all-about-the-lock-file

flip4bytes:
Yeah, for hosting on Heroku, the repository must contain the composer.lock file. I always remove the composer.lock file from the .gitignore on my installs.

daftspunk:
I always thought this was a great idea, I just wasn't sure if we should do it because I don't see other projects doing it. There must be a reason why they don't do it, to make a decision we need to know why.

wppd:
@daftspunk  it's so sad... I thought the link explains the "why" very well.

shina:
The latest build of october beta is broken because of laravel 4.2.17




My lucky is that I had another project with 4.2.16 on composer.lock then I am using it.

wppd:
@shina @daftspunk lol, what a coincident! that's the proof.

that's the reason why you need to keep the composer.lock.

users do the "composer update" but the third packages got upgraded and break BC after the "release" of octobercms.

anyway, just my 2 cents. your project your call.

dgleba:
I have read the article and agree that composer.lock should be included. I have tried to install and run many projects and only ended in errors due to version issues. How can composer.lock hurt?

saleksin:
Why, after this discussion, in [this](https://github.com/octobercms/october/commit/8c689d63ea5d580d9e0981da273cf5cae061220b#diff-a084b794bc0759e7a6b77810e01874f2) commit someone added composer.lock file to gitignore again?

daftspunk:
This is to add composer.lock to the git project, we decided against it because it added complications. Composer will merge plugin dependencies in to the lock file.

PR #1438, opened 2015-09-17
Remove composer.lock from .gitignore

MatissJanis:
This file should be included in the codebase so that all developers get exact copies of libraries and their corresponding versions.

Explained in more detail [here](https://blog.engineyard.com/2014/composer-its-all-about-the-lock-file).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 /bootstrap/compiled.php
 /vendor
-composer.lock
 composer.phar
 .DS_Store
 .idea


daftspunk:
[This has been discussed before](https://github.com/octobercms/october/issues/922), still unsure why other projects don't include composer.lock. There hasn't been any issues with packages not respecting semantic versions in the past. This fixes a problem that doesn't really exist at this point in time. Although it is good to know we have this solution if we need it some day.

MatissJanis:
October is based on Laravel. Since Laravel does not ignore the `composer.lock` file, neither should we ignore it. 

https://github.com/laravel/laravel/commit/3c9d67ad657290c889969c85867677f35e6ae8f1

daftspunk:
I don't see it in the repo: https://github.com/laravel/laravel

MatissJanis:
Yes, they do not ignore it, but at the same time they do not commit it either.

They do not commit it because this would introduce unneccessary clutter in the codebase, **but** at the same time it is not ignored because other developers (those who do not contribute to the Laravel repo) may have multiple servers/machines on which this code is running on. The lock file helps all of these machines have the same exact version of Laravel.

So esentially - contributors must ignore this file locally, but everyone else should commit the lock file up to their private repositories.

I'd suggest that October takes the same approach.

mplodowski:
I also think October should not ignore composer.lock file.

I had an issue with memory limit error on my production machine when I didn't push composer.lock file and run "composer update" command. Composer use that way a lot of memory, sometimes over 1GB.

Solution for this problem is to have composer.lock file on server and run "composer install" command.

I deploy my October apps from git repositiories, so for default composer.lock file is not present there.

I always end up with copying composer.lock file by hand.

daftspunk:
If this results in PRs trying to commit a composer.lock file then we will have to put it back.

PR #1442, opened 2015-09-20
Updated AssetList widget to use correct folder permissions

jBOKA:
when using mkdir directly, the umask config is skipped


daftspunk:
Nice! :+1:

PR #1445, opened 2015-09-21
Translate the "CMS" string and minor cleanup

gergo85:


daftspunk:
Translations should occur as close to the view as possible
Fixed in 44bfda46ccd668a6a68b25cc83e024ec0f19000b

daftspunk:
Thanks!

PR #1446, opened 2015-09-22
Allow \ character in Page URLs. Fixes #1441

### LINKED ISSUE #1441: \ not allowed in CMS page URL regex
In the [URL Syntax docs](http://octobercms.com/docs/cms/pages#url-syntax) you have `/blog/:post_name?|^[a-z0-9\-]+$` however if you actually try to save a page with this URL you'll get the error

> Invalid URL format. The URL should start with the forward slash symbol and can contain digits, Latin letters and the following symbols: ._-[]:?|\/+*^$

Remove the \ from the regex section of the URL above and the page saves successfully. This bug is possibly related to https://github.com/octobercms/october/issues/614

Flynsarmy:


daftspunk:
*No description provided.*? What URL are you trying to use? So we can write tests.

daftspunk:
Thanks!

Ref #1441

PR #1448, opened 2015-09-22
Fix relation manager searching + pagination issue

MatissJanis:
With this fix users will now be able to search for records in the relation modal window AND continue on to the next page (paginate properly).

Lets say that there are 100 relation records and we paginate them at 10 per page. If we search for something using the search bar we could reduce the amount of results to les say 30. These results still get paginated in 3 pages. At this stage we can observe the bug. If we try to click on the next page button, the total record count will be reset back to 100.

So basically..

In relation manager pagination + search did not work.

**This fix makes sure that both the manager and list view are using the same search widget.**

### DIFF
--- a/modules/backend/behaviors/RelationController.php
+++ b/modules/backend/behaviors/RelationController.php
@@ -484,15 +484,25 @@ class RelationController extends ControllerBehavior
     // Widgets
     //
 
-    protected function makeSearchWidget()
+    protected function makeSearchWidget($widget)
     {
         $config = $this->makeConfig();
         $config->alias = $this->alias . 'ManageSearch';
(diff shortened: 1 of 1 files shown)


daftspunk:
Thanks for pointing out this bug and even more thanks for your efforts in attempting to fix it

See dd6e0a383631ec88d14804f8cfe334b21d807377 for the fix

MatissJanis:
Thanks!

immeyti:
hello.
i want add search to relation manager. 
how can do it?

LukeTowers:
@immeyti commenting on old issues is not the way to solve your problem.

PR #1453, opened 2015-09-23
Fixed relation render onclick in list view to obey $model->primaryKey value

tschallacka:
When people have a differently named primary key than id the relation render cannot do the popup when clicking on the list view, because it can't find the field id.

This will make sure that it will obey the primaryKey in the model to define which field to open which will resolve such issues.


tschallacka:
ffs, wrong branch again

PR #1455, opened 2015-09-23
Small fixes for localization

Mr118:
One missing translation in backend/lang/pl/lang.php and one missing translation string in Lists.php


Mr118:
damn wrong branch

PR #1456, opened 2015-09-23
fixes in localization

Mr118:
one missing polish translation added and one string replaced with translation key that exists in all lang files and should be there from the start.


daftspunk:
Thanks

PR #1458, opened 2015-09-23
Update the CSS Browser Selector php

gergo85:


daftspunk:
Please provide a description, why are you updating this? Is there a browser that is not covered by the current version?

gergo85:
The new version supports more browsers (e.g. EDGE) and platforms (e.g. BlackBerry).

daftspunk:
This doesn't fix anything

PR #1460, opened 2015-09-24
Improve the date format

gergo85:
Reference: https://en.wikipedia.org/wiki/Date_format_by_country


daftspunk:
Pretty sure we can [use Carbon](https://github.com/briannesbitt/Carbon) for this.

dshoreman:
I'll post a new PR, PHP can do all this for us (although Carbon uses `setlocale()` which is unreliable).

daftspunk:
Fixed by #1462

PR #1461, opened 2015-09-24
Security tweaks

dshoreman:
Adds 3 new permissions to allow removing the corresponding links from the backend account dropdown. This is particularly useful in demo scenarios where you don't want people to be able to change the backend language or demo password for example.

Also contains a fix that prevents admins from creating their own superuser account or promoting other users.

### DIFF
--- a/modules/backend/ServiceProvider.php
+++ b/modules/backend/ServiceProvider.php
@@ -110,6 +110,14 @@ class ServiceProvider extends ModuleServiceProvider
                     'label' => 'system::lang.permissions.manage_other_administrators',
                     'tab'   => 'system::lang.permissions.name'
                 ],
+                'backend.manage_preferences' => [
+                    'label' => 'system::lang.permissions.manage_preferences',
+                    'tab'   => 'system::lang.permissions.name'
+                ],
+                'backend.manage_editor' => [
+                    'label' => 'system::lang.permissions.manage_editor',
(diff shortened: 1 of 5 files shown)


daftspunk:
This covers a variety of issues, the main one being the postback data from a backend form is not getting filtered. While I see you have addressed the issue for the superadmin value alone, I think we need a more robust solution that covers other areas.  I have something in mind already, the FormController should check to see if the postback value exists in the form field definition and only commit those attributes to the model. In other words, no field definition, no assignment.

Leave it with me.

SaifurRahmanMohsin:
This is great! It would also be nice to see an option to “Allow switching maintenance mode ON/OFF” because currently only super admins can do that.

dshoreman:
@SaifurRahmanMohsin You may have to write a plugin for it. As much as I'd love to add better permissions in October, I had to strip most of this PR because they don't want "too many" permissions.

SaifurRahmanMohsin:
Then the entire maintenance mode feature should be plugin-only. I suggested this only because it’s core and not a separate plugin.

dshoreman:
@SaifurRahmanMohsin The permissions I removed from this PR were for menu items under the backend user dropdown.

In most cases I think maintenance mode would probably only need to be disabled by superadmins, but there could always be some use for a permission. Feel free to post a new issue though, it's worth a shot...

PR #1462, opened 2015-09-24
Localise last modified date for mediaManager items

dshoreman:
Updates the string version of the date to use PHP's IntlDateFormatter.
That gives us translated dates in the appropriate format, based on the user's locale set in backend preferences.

If ever there's a case where locale wasn't set or didn't match anything, it would default to the system locale.

Note that this PR supercedes #1460.

### DIFF
--- a/modules/cms/classes/MediaLibraryItem.php
+++ b/modules/cms/classes/MediaLibraryItem.php
@@ -2,6 +2,8 @@
 
 use File;
 use Config;
+use IntlDateFormatter;
+use Backend\Models\UserPreferences;
 
 /**
  * Represents a file or folder in the Media Library.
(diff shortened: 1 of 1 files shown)


daftspunk:
Thanks for this

daftspunk:
FYI this code executes on every backed controller:

```
        /*
         * Set the admin preference locale
         */
        if (Session::has('locale')) {
            App::setLocale(Session::get('locale'));
        }
        elseif ($this->user && ($locale = BackendPreferences::get('locale'))) {
            Session::put('locale', $locale);
            App::setLocale($locale);
        }
```

mplodowski:
Probably some users will experience the following error in media library after this commit:

    'Class 'IntlDateFormatter' not found'

That is because like me they do not have installed INTL php module extension.

For Ubuntu users:

    sudo apt-get install php5-intl

Check if you have uncommented:

    extension=php_intl.dll

in your php.ini file.

PR #1464, opened 2015-09-26
Use correct folder permissions in ThemeImport/Export

Flynsarmy:
Similar to #1442 we should be using the `File` facade rather than `mkdir` and `unlink` so we don't ignore umask config.


daftspunk:
Thanks bud!

PR #1468, opened 2015-09-26
Minimum password length should be 4 characters

mul14:
Because the minimum password length on User model is 4
https://github.com/octobercms/october/blob/master/modules/backend/models/User.php#L26-L27


dshoreman:
Is there a reason you're limiting the password to 64 characters here?

daftspunk:
Thanks!

PR #1474, opened 2015-09-29
Fix undefined variable $filePath

adrenth:
Also added proper phpDoc to the PluginBase class

And updated Dutch language for CMS


daftspunk:
Thanks!

PR #1475, opened 2015-09-29
Allow minDate/maxDate to be entered in date format for DatePicker. Fixes #909.

### LINKED ISSUE #909: 'minDate/maxDate' datepicker options not set correctly on form
If I have a form field like so:
```yaml
published_at:
  label: Published on
  type: datepicker
  mode: datetime
  minDate: 1900-01-01
  maxDate: 2020-01-01
```
The datepicker element on the form will show the year as 1970 despite it being 2015 in the database. This is caused by the YAML file converting 1900-01-01 to -2208988800 and 2020-01-01 to 1577836800 rather than simply leaving them as strings:
```html
<!-- Date -->
<div
    id="DatePicker-formPublishedAt-published_at"
    class="field-datepicker"
    data-control="datepicker"
    data-min-date="-2208988800"
    data-max-date="1577836800">
    <input
        type="text"
        id="DatePicker-formPublishedAt-input-published_at"
        name="Post[published_at]"
        value="2015-01-24"
        class="form-control align-right"
        autocomplete="off">
</div>
```

I've whipped up a demo repo complete with steps to replicate [here](https://github.com/Flynsarmy/oc-baddate-plugin.git).

patrickward:
The Symfony Yaml parser converts dates into timestamps via `strtotime` whenever a value matches a regex pattern that looks like a date. So, when adding a date for the `minDate` and `maxDate` options without quoting them as strings, the Yaml parser converts them into an integer timestamp. This fix looks for minDate and maxDate properties that are integers and turns them back into a date string with the form `Y-m-d`. 

Tested with multiple date formats. For example: 

* 2015-01-01 
* 01/01/2015 
* January 1, 2015 
* 2015 01 01 
* 2015-01-01 11:11:11
* 2015 January 01 11:11:11
* January 01 2015 11:11

Should work with any valid PHP date format.

### DIFF
--- a/modules/backend/formwidgets/DatePicker.php
+++ b/modules/backend/formwidgets/DatePicker.php
@@ -46,13 +46,33 @@ class DatePicker extends FormWidgetBase
      */
     public function init()
     {
-        $this->fillFromConfig([
-            'mode',
-            'minDate',
-            'maxDate',
-        ]);
-
+        $this->fillDatesFromConfig();
+        $this->fillFromConfig(['mode']);
         $this->mode = strtolower($this->mode);
+    }
+
+    /**
+     * Transfers minDate and maxDate config values stored inside the
+     * $config property directly on to the root object properties.
+     *
+     * This method checks for Yaml parsed dates that have been converted
+     * into integer timestamps via the Symfony Yaml parser and converts them
+     * back into strings for use with the datepicker.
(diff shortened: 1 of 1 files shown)


daftspunk:
Ref #909

daftspunk:
Thanks!

PR #1487, opened 2015-10-04
Fix broken link in README

svenluijten:
http://octobercms.com/docs/help/install -> http://octobercms.com/docs/setup/installation


svenluijten:
Noticed this was duplicate, closed :smile:

PR #1496, opened 2015-10-07
Added support for arrays in the dropdown form option when Many To Many Relations are used

tschallacka:
This way if someone specifies a belongsToMany or other many to many relation field types it can still select the correct one to show as selected.
This is solely for the niche cases where this is used, But it's useful to be able to use the dropdown when needed for relationship displaying.

### DIFF
--- a/modules/backend/widgets/form/partials/_field_dropdown.htm
+++ b/modules/backend/widgets/form/partials/_field_dropdown.htm
@@ -18,7 +18,7 @@
                 if (!is_array($option)) $option = [$option];
             ?>
             <option
-                <?= $value == $field->value ? 'selected="selected"' : '' ?>
+                <?= ($value == $field->value || (is_array($field->value) && in_array($value,$field->value))) ? 'selected="selected"' : '' ?>
                 <?php if (isset($option[1])): ?>data-<?=strpos($option[1],'.')?'image':'icon'?>="<?= $option[1] ?>"<?php endif ?>
                 value="<?= $value ?>">
                     <?= e(trans($option[0])) ?>


tschallacka:
Some test code where I needed this :-)

The model where it has two belongs to many relationships with a linking table to refer to itself
One side is basically a belongsTo(the brand_hotel) but since belongsTo doesn't support an external table for the join I had to use this construct.

    class HotelManagerModel extends Model
    {
        public function getBrandHotelOptions($keyValue) {
            $result = HotelManagerModel::where('is_chain_definition','=',1)->get();
            $arr = [''=>'Select a brand if applicable'];
            foreach($result as $res) {
                if($res->getKey() != $this->getKey()) {
                    $arr[$res->getKey()] = $res->hotel_name;
                }
            }
            return $arr;
        }
        public function scopeis_chain_definition($query) {
            return $query->where('is_chain_definition',true);
        }
        public $belongsToMany = [
                'child_hotels' => [
                        'ExitControl\HotelManager\Models\HotelManagerModel',
                        'key' => 'brand_id',
                        'otherKey' => 'hotel_id',
                        'table' => 'hotel_to_brand'
                ],
                'brand_hotel' => [
                        'ExitControl\HotelManager\Models\HotelManagerModel',
                        'key' => 'hotel_id',
                        'otherKey' => 'brand_id',
                        'scope' => 'is_chain_definition',
                        'table' => 'hotel_to_brand'
                ]
        ];
    }

The fields.yaml for brand_hotel

    brand_hotel:
        label: exitcontrol.hotelmanager::lang.hotelmanagermodel.brand_hotel_name
        type: dropdown
        span: left

And for the child_hotels I used a custom relation manager thats set up for belongsToMany relationships. No worries there.

But for the dropdown I just need this extra line added, so it can recognise which value is selected.

Now, this line wouldn't be neccesary if there'd be support for external linking table in belongsTo, but I can imagine that would give several headaches too :-) Especially since this is a niche case that should't pop up often.

daftspunk:
The problem with this: forcing a "has many" relationship to behave like a "has one" relationship. In a dropdown there can be only one, so this logic error is normal and expected, since in "has many" the value returned is an array, even if only one is selected. Your options:

1. Use `belongsTo` instead
1. Or....

```
primary_brand_hotel:
    label: exitcontrol.hotelmanager::lang.hotelmanagermodel.brand_hotel_name
    type: dropdown
    span: left
```

```
public function getPrimaryBrandHotelOptions($keyValue) {
    //...
}

public function getPrimaryBrandHotelAttribute() {
    return $this->brand_hotel()->first();
}
```

PR #1500, opened 2015-10-08
Added a web.config file for IIS servers. 

tschallacka:
Added a web.config file for IIS servers. Does basically everything the default .htaccess does


daftspunk:
[Added this to the documentation](https://github.com/octobercms/docs/commit/ab00be2f77a4fcd43ed452c3d018a6c1d2b589d7) instead.

PR #1502, opened 2015-10-08
Fire model events on update context

alxy:
Currently, the ``FormController::update()`` method saves the model rather than updating it. That leads to the unexpected behavior, that ``beforeUpdate()`` and ``atferUpdate()`` do not work at all using the behavior. This commit fixes the described issue by actually calling ``update`` on the saved models.


alxy:
Ok this PR is entire nonsense. I totally fucked it up an editted the onCreate method...

Still I think this line should be reviewed: https://github.com/octobercms/october/blob/master/modules/backend/behaviors/FormController.php#L261
I don't even know if - in update context - it makes sense to supply the $sessionKey. Can any deffered binding magic happen in update context? In my eyes, it should simply be ``$modelToSave->update([]);``, but I dont know if this has other implications.

As for now, I'll close the PR.

PR #1504, opened 2015-10-09
Add support for default values in form widgets

Flynsarmy:
Form Widgets like colorpicker ignore the default value you specify in your YAML. This PR makes them work.

Example YAML:
```yaml
my_colorpicker:
    label: My Colorpicker
    type: colorpicker
    default: #27ae60
```

My PR is just a modified version of what you already have for standard form fields:
https://github.com/octobercms/october/blob/master/modules/backend/widgets/Form.php#L778-L788


daftspunk:
Thanks!

PR #1506, opened 2015-10-09
Fixed form field parsing to accept no config

tschallacka:
If a programmer would define a yaml like

    fields:

        shortcut:

        country:

It would fail with message

    "array_key_exists() expects parameter 2 to be array, null given" on line 275 of ..\modules\backend\classes\FormField.php

The yaml parsing returns a null. This check intercepts that null value being passed to the function and turns it into a proper array.


daftspunk:
LGTM

PR #1509, opened 2015-10-11
fixed FormModelSaver ignoring fillable/guarded field rules fixes #1042

### LINKED ISSUE #1042: Relation manager obeys they $fillable/$guarded, but regular page does not.
This bug was driving me crazy, but here is how to duplicate it:

* Add a relationship manager to an "update" form with something (like blogpost model) like `<?= $this->relationRender('comments') ?>`
* Then set a field in fields.yaml in comments like `from_email`.  A related column should not exist in the DB.
* in your comments model, make sure you have `from` field to be jsonable
* make sure you have a `from` column in your database.
* add the accessor and mutators for the `from_email` attribute in your comments model, like so:

```
    public function getFromEmailAttribute() {
        if( isset ($this->from['from_email']) ){
            return $this->from['from_email'];
        }
        return '';
    }

    public function setFromEmailAttribute($value) {
        
        $this->from = array_replace_recursive((array)$this->from, ['from_email' => $value]);
    }
```

* Try to edit a comment from the relation manager on blogpost update form.  It won't complain (no error) and wont save the `from_email` field.

*I've verified that setFromEmailAttribute() is NOT fired via a breakpoint with xdebug.  `getFromEmailAttribute` is working fine though.*

Also, this DOES work if you are creating/editing a comment normally (like on a page like: backend/oca/mypluginname/comments/update/2).  It works perfectly from a normal page... **just not from the relation manager.**

jwilson8767:
As noted in #1042, saves performed by FormControllers ignore fillable/guarded rules. This is because FormModelSaver is responsible for the saving, but it directly sets the fields without checking if that is allowed (or using $model->fill($attributes)).

### DIFF
--- a/modules/backend/traits/FormModelSaver.php
+++ b/modules/backend/traits/FormModelSaver.php
@@ -60,7 +60,7 @@ trait FormModelSaver
             if ($isNested && is_array($value)) {
                 $this->setModelAttributes($model->{$attribute}, $value);
             }
-            elseif ($value !== FormField::NO_SAVE_DATA) {
+            elseif ($value !== FormField::NO_SAVE_DATA && $model->isFillable($attribute)) {
                 $model->{$attribute} = $value;
             }
         }


daftspunk:
We need to do the opposite here, `RelationController` should ignore fillable instead. Form fields are already constrained by their `fields.yaml` definition (values not defined in the yaml will not be saved) so we don't need to double dip by enforcing fillable too.

jwilson8767:
Oh, I see what you mean. Thanks for looking into this!

Issue #1349, opened 2015-08-05
No error thrown when upload fails


vojtasvoboda:
When I have PHP config `post_max_size = 8M` and try to upload 12MB file, media manager shows green status line with "Upload done" msg, but no file is visible in list:

![snimek obrazovky 2015-08-05 v 11 26 26](https://cloud.githubusercontent.com/assets/374917/9082481/9fe5fd96-3b64-11e5-83c4-5cb1e3cf94b1.png)

POST request returns HTTP 200 code.

daftspunk:
Fixed by #1518

PR #1518, opened 2015-10-14
Display media manager upload errors more clearly

firoxer:
This is a fix for #1349

Failed uploads only caused a "Error uploading file" alert to pop up, while the upload bar implied otherwise (as per the given issue). I turned the bar's color back to default (gray in the demo theme) and changed the text to "Upload failed".

Some problems:
* There are no translations (other than English) for the "Upload failed" message.
* The demo stylesheet had no styles for a "bad" upload bar. I would have added a class to turn it red, but then I'd break all the other themes.


daftspunk:
Great job, thanks!

PR #1519, opened 2015-10-15
Allow extend query on reorder controller, on 'simple' sortMode

jofrysutanto:
Only tested on 'simple' sortMode, hence placed within `$this->sortMode == 'simple'` condition.


daftspunk:
Please share why you need this with code we can replicate

daftspunk:
Can't this be achieved with the following?

```
public function reorderGetModel()
{
    return MyModel::withTrashed();
}
```

daftspunk:
Looks good! Merged in f0b469116878bcb05a5e7986d040157e5ffccd63
Thanks

PR #1520, opened 2015-10-15
Translate missing items

alxy:
Some local names may need some more attention. Im unsure how to translate them (best).


daftspunk:
Goood

PR #1521, opened 2015-10-15
Add YAML-specific settings

lesaff:
Force text editor to indent using 2 spaces on `editorconfig` enabled text editor.


daftspunk:
We use `indent_size = 4` for YAML...

PR #1522, opened 2015-10-15
Add libwww-perl blocker

lesaff:
Block libwww-perl access user agent

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -13,6 +13,12 @@
     # RewriteBase /
 
     ##
+    ## Block libwww-perl access user agent
+    ## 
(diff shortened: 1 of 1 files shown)


daftspunk:
Why?

mrmlnc:
[It seems because of botnets](http://serverfault.com/questions/346540/how-to-create-htaccess-and-block-useragent-libwww-perl)

daftspunk:
This will cause confusion for people with a genuine need to use this library.

lesaff:
To prevent spammers and spam-bots. Also for security measure to prevent backdoor perl access.

PR #1524, opened 2015-10-15
Add ``conditions`` config to RecordFinder

alxy:
Allows custom SQL conditions on the record finder. Relations already support this, but the finder doesnt seem to respect the conditions due to some limitations. I think it is a good idea to bring conditions back with this code.


daftspunk:
Duplicate of #1203

alxy:
Its not exactly a duplicate, however, it is related. My PR is targeted against the RecordFinder, wheras the other issue is looking for a ore complete solution for the RelationController, where this feature is indeed also missing.

As the record finder has really limited functionality, I'd supose to merge this as easy solution (limited to SQL conditions only). Maybe this can be fixed to use the relation definition later, but I did not find any good way to do so.

alxy:
Any chance to merge this until a more robust solution is found there? I have this requirement to limit the items of the recordfinder and did not find any good way without modifying the sources...

PR #1525, opened 2015-10-16
Allow RelationController to list attachOne/attachMany relations

Flynsarmy:
I have an attachMany field I'm trying to display in RelationController however I currently get the error:

> Undefined variable: widget in /path/to/modules/backend/Behaviors/RelationController.php line 666

This PR fixes the above error and displays my attachMany relation.


daftspunk:
To confirm: You are wanting to display a list of file attachments specifically?

Can you provide an example plugin.

daftspunk:
This needs to be resubmitted with a proper test scenario.

PR #1527, opened 2015-10-16
Unbinding javascript event handlers from html elements

tschallacka:
When event handlers are bound to elements within the refreshed partial object they don't
get unbound when the element is refreshed and the elements are not valid anymore.
However, new event handlers will be bound to the new elements  by coders.
Both eventhandlers will usually fire, remain scoped, etc.. and get triggered.
By calling empty first on the selected target, which is safe to do because all contents
are overwritten anyways, we make jQuery "clean up" all lingering event handlers.

### DIFF
--- a/modules/system/assets/js/framework.js
+++ b/modules/system/assets/js/framework.js
@@ -199,7 +199,7 @@ if (window.jQuery === undefined)
                             $(selector.substring(1)).prepend(data[partial]).trigger('ajaxUpdate', [context, data, textStatus, jqXHR])
                         } else {
                             $(selector).trigger('ajaxBeforeReplace')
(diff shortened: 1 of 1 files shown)


tschallacka:
According to the jQuery documentation this is not needed as .html() does it's own unbinding call etc...
But in my field test I found that some of the event handlers did live on with just .html, and didn't live on with the call to .empty.
So I don't really know what's going on there. I'll leave it up to you guy's discretion to see if you wish to implement it or not.

daftspunk:
If jQuery docs said it isn't necessary then this a bug that needs to be raised with jQuery (mismatch of functionality and spec)

PR #1529, opened 2015-10-16
fixed some PHPDocs

daniel-mueller:
Hi,

I was reading the source of your project and thought it might help to correct some PHPDocs while I'm at it. I hope this is in your interest. If you think like there could be some more I could do feel free to contact me. 

Thank you for considering,
Daniel


daftspunk:
The comments look good, however please refrain from modifying the PHP code.

daniel-mueller:
I removed every change to the code and it's behaviour. Only the PHPDocs are left. You tagged the PR as 'wrong branch'. I guess develop is the right one?

daftspunk:
Thanks

PR #1534, opened 2015-10-18
Add generator meta tag and improve HTML5 support

gergo85:


daftspunk:
I think it's more of a preference, HTML5 doesn't specify one way or the other...

```
<!-- Both valid -->
<meta />
<meta>
```

PR #1541, opened 2015-10-21
Revert 'hasMany/belongToMany' onRelationManageCreate 

a2thek26:
...so that children models save with the parent's id. Deals with af657e9 where child models where not saving with parent id and throwing constraint errors.

### DIFF
--- a/modules/backend/behaviors/RelationController.php
+++ b/modules/backend/behaviors/RelationController.php
@@ -912,13 +912,21 @@ class RelationController extends ControllerBehavior
         $sessionKey = $this->deferredBinding ? $this->relationGetSessionKey(true) : null;
 
         if ($this->viewMode == 'multi') {
-            $newModel = $this->relationModel;
-            $modelsToSave = $this->prepareModelsToSave($newModel, $saveData);
-            foreach ($modelsToSave as $modelToSave) {
-                $modelToSave->save(null, $this->manageWidget->getSessionKey());
+            if($this->relationType == 'hasMany') {
+                $newModel = $this->relationObject->create($saveData, $this->manageWidget->getSessionKey());
+            } elseif ($this->relationType == 'belongToMany') {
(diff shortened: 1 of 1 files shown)


daftspunk:
I was worried this might happen but was unable to replicate the error. Can you provide replication instructions using the [Test plugin](https://github.com/daftspunk/oc-test-plugin).

daftspunk:
Unable to replicate

a2thek26:
I created https://github.com/daftspunk/oc-test-plugin/pull/2 to create the issue.

Assuming you've migrated the database to that pull request. - I created a `Notes` model to demonstrate a scenario where a `Person` would have many `Notes`...

Go to `Playground > People > click New Person`.

Add a name or whatever. 

Click the Notes tab and Create a new note. When you try to save that note it throws a constraint error. It doesn't look to be grabbing the parent id. Even if you store the `Person` model and come back to create a new note, it throws the error.

When I revert to `this` pull request, for the relationObject storage it works just fine.

daftspunk:
Thanks! I will check it soon

daftspunk:
The only difference between the new code and the old code is the following:

```
; Old code output
insert into `october_test_notes` (`person_id`, `title`, `body`, `updated_at`, `created_at`) values (, dfgssdgf, sdfgsdfg, 2015-10-31 05:28:44, 2015-10-31 05:28:44))

; New code output
insert into `october_test_notes` (`title`, `body`, `updated_at`, `created_at`) values (dfgssdgf, sdfgsdfg, 2015-10-31 05:28:50, 2015-10-31 05:28:50))
```

It would be worthwhile ensuring that your FK constraint is nullable so it can be compatible with deferred binding. i.e the `person_id` field will need to be empty until the person created.

Regardless, I have pushed c30b34e4d94560a7590104dbce20cabac7262285 that will retain the previous functionality and keep the new improvements.

a2thek26:
Sounds good and makes sense. Thanks.

PR #1552, opened 2015-10-24
Allow plugins to be symlinks

JoelESvensson:
This change makes sure that plugins can be loaded even though they are symlinks. The change makes it easier to have custom schemes to handle version control with zero-downtime deployment.


jwilson8767:
I researched this approach earlier this week and can confirm that this is both a useful change and this is an appropriate way of implementing it.

daftspunk:
LGTM!

PR #1560, opened 2015-10-27
Access management controller

MatissJanis:
Adds an easy way to manage the access to various backend sections.

For example, I may need to have a backend user that can see (preview) a table, but he cannot edit. This takes care of that.

If this gets merged, I can write a section in the documentation.

Cheers!

**Usage:**

In the model class definition:

```
  public $implement = ['Backend.Behaviors.AccessController'];
  public $accessConfig = 'config_access.yaml';
```

In config file (config_access.yaml):

```
access:
   index: acme.demo.access_clients
   update:
       permission: acme.demo.update_clients
       redirectUrl: acme/demo/clients/preview/:id
   onDelete: delete_clients
```

### DIFF
--- a/modules/backend/behaviors/AccessController.php
+++ b/modules/backend/behaviors/AccessController.php
@@ -1,0 +1,114 @@
+<?php namespace Backend\Behaviors;
+
+use Backend;
+use BackendAuth;
+use Backend\Classes\BackendController;
+use Backend\Classes\ControllerBehavior;
+use Event;
+use October\Rain\Router\Helper as RouterHelper;
+use Request;
+use Response;
+use View;
+
+/**
+ * Access Controller Behavior
+ * Helps manage accesses.
+ *
+ * Usage:
+ *
+ * In the model class definition:
+ *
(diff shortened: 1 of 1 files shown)


daftspunk:
This looks good, I think it might be a safe bet to wait for the Builder plugin before committing to this pattern

MatissJanis:
Sure thing. Just let me know if it needs any changes

gergo85:
+1

daftspunk:
It's nice but is it useful?

LukeTowers:
I think so, especially the redirect option when the user doesn't have that permission. I think it needs to be further reviewed for it's specific implementation, it's a bit out of date right now. I'd be willing to take a look at implementing it into core once I have a bit more time

daftspunk:
I would've thought access management would be handled internally by the behavior or by the controller itself. I mean, it's just an IF statement right?

```
if (!$this->user->hasPermission())
```

Not really sure this adds enough value to justify itself. There could also be more complex scenarios needed that are not covered. Still on the fence with this one; just can't see myself using it.

PR #1561, opened 2015-10-27
Added a check for the Failed Jobs table, in case you have already cre…

Lickwid:
…ated it from laravel migration


daftspunk:
In what scenario would the `failed_jobs` table already exist?

PR #1565, opened 2015-10-28
Remember previous page in list view

MatissJanis:
Steps to reproduce this bug:

1. Open up a list;
2. Go to page 20;
3. Open any record;
4. Go "back";
5. You get thrown back to the first page.

This commit fixes that issue.

### DIFF
--- a/modules/backend/widgets/lists/assets/js/october.list.js
+++ b/modules/backend/widgets/lists/assets/js/october.list.js
@@ -131,4 +131,17 @@
         $('[data-control="listwidget"]').listWidget();
     })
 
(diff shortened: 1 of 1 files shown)


tschallacka:
support!

daftspunk:
This creates a few more problems

1. Pressing the back button on pagination navigation does nothing: Navigate to page 2, page 3, then press back twice. I expect page 2 and page 1 to reappear.
1. Searching the list now stays on the selected page: Navigate to page 2, search for something, it will keep us on page 2. Previously the pagination would reset back to page 1 when searching.

PR #1575, opened 2015-11-01
Make recordfinder's popup list title configurable

SaifurRahmanMohsin:
When selecting record using the recordfinder, by default the popup list
shows **Find Record** as the title. Made this field a user configurable
field, and kept **Find Record** as default if not specified. Also,
added this default string to the lang file.


daftspunk:
Domo arigato!

PR #1576, opened 2015-11-02
Update ServiceProvider.php

FrancescoPassanti:
Update permissions of October.System


daftspunk:
Sorry this is lacking detail, what problem does this solve? The current proposal here will hide the Settings menu item for plugins that need it. Settings should always be visible for this reason. You may remove it using a hook, [as per the documentation](https://octobercms.com/docs/plugin/extending#extending-backend-menu).

PR #1579, opened 2015-11-03
Added support for 'nameFromSelect' in relation form fields

JoelESvensson:
For example, if you have a `first_name` and `last_name` column, you can now combine them to display a full name. This is done by adding the custom column selection in the new config field `nameFromSelect`.

The described example above would look like this:
```
user:
    label: User
    type: relation
    nameFromSelect: concat(first_name, ' ', last_name)
```

This behaviour could very well just be moved into `nameFrom` but i'm a bit unsure about backwards compatibility. Anyway, i'm open to renaming `nameFromSelect` to anything logical. I also volunteer to write the documentation.

### DIFF
--- a/modules/backend/formwidgets/Relation.php
+++ b/modules/backend/formwidgets/Relation.php
@@ -5,6 +5,7 @@ use Backend\Classes\FormWidgetBase;
 use ApplicationException;
 use SystemException;
 use Illuminate\Database\Eloquent\Relations\Relation as RelationBase;
+use DB;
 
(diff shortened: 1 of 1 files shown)


daftspunk:
Great! Let's go with the convention used in list columns

```
user:
    label: User
    type: relation
    select: concat(first_name, ' ', last_name)
```

I think we should also add support for `nameFrom` to select model attributes as well. We will implement this today. Thanks!

PR #1582, opened 2015-11-04
Add .stub extension to stub file

JoelESvensson:
As the stub namespaces.php file is not a valid .php file it should have a stub extension. The main point of the change is to tell tools (like linters and CS-fixers) that do code analyzis that the file should not be analyzed.


daftspunk:
Thanks

PR #1585, opened 2015-11-06
Add filetypes of FileUpload formwidgets in uppercase too

brunoti:
I was not able to make upload of images with uppercase extensions like ``.JPG `` so this is how i solve the problem. 

Maybe this is not the best way of doing this! I think this is a simple but important thing so if you like the idea but not the code, i can learn and do better :smiley:.


daftspunk:
Thanks!

PR #1593, opened 2015-11-10
Make record finder configurable to allow for only displaying related …

a2thek26:
…models in list config.

This essentially makes the record finder only return the models that are not already associated with a relationship. Should be able to be controlled by field config for the relationship by adding `showAvailable` key. There was some commented code in there for how to extend the widget list. I simply implemented a way for that to be realized.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -1,0 +1,3957 @@
+{
+    "_readme": [
+        "This file locks the dependencies of your project to a known state",
+        "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
+        "This file is @generated automatically"
+    ],
+    "hash": "82178672a3502110901475209157f0ae",
+    "content-hash": "39a8cc6f64156c7970e195ed8fda0a4d",
+    "packages": [
+        {
+            "name": "classpreloader/classpreloader",
+            "version": "1.4.0",
+            "source": {
+                "type": "git",
+                "url": "https://github.com/ClassPreloader/ClassPreloader.git",
+                "reference": "b76f3f4f603ebbe7e64351a7ef973431ddaf7b27"
+            },
+            "dist": {
+                "type": "zip",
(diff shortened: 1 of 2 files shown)


daftspunk:
Argh, this PR contains a `composer.lock` file. We will merge this manually today. Thanks

daftspunk:
This is producing an error with this using the [test plugin](https://github.com/daftspunk/oc-test-plugin):

```
Column not found: 1054 Unknown column 'person_id' in 'field list' (SQL: select `person_id` from `october_test_people`)'
```

Can you elaborate on what you are hoping to achieve with this feature? Perhaps there is another solution we can use. What is your usage case?

a2thek26:
Well the goal was to have a configuration to allow the record finder to be able to filter out records that were already assigned to other models. So, I was trying to get the ids of the models that do not already have assignment, but as far as I can tell, I can only do that by knowing what the foreign key field is of the relationship. 

So, this would return a list of ids that are already assigned:

```
$relatedAlias = $this->formField->fieldName;
$relationDefinition = $this->model->getRelationDefinition($relatedAlias);

if(array_key_exists('key', $relationDefinition)) {
    $foreignKey = $relationDefinition['key'];
} else {
    $foreignKey = snake_case($relatedAlias).'_id';
}

return $this->model->lists($foreignKey);
```

Then, the list query would get extended by pulling only the records that are not in the list:

```
$widget->bindEvent('list.extendQueryBefore', function($query) {
    $existingIds = $this->findExistingRelationIds();
    if (count($existingIds)) {
        $query->whereNotIn('id', $existingIds);
    }
});

```

So, I have to imagine that it's getting caught up with the condition when the key is not defined in the relationDefinition and it's trying to assume the foreign_key `snake_case($relatedAlias).'_id'`.

daftspunk:
This can be implemented by using the `searchScope` option we recently added.

Reference: https://github.com/octobercms/october/commit/f2d8fcaf1488a250bc995b393ccc8a13b866e405

PR #1599, opened 2015-11-13
Fix small bug with ".noclick" column in ListController

darkdarin:
**Bug:**
On the page with a list of items (ListController) add a column and assign property cssClass in config_list.yaml value "nolink", to click on the column is not processed. (Example: Column button switch element visibility).
If this is the first column, then click stops processing on all columns line, ie you can not go to the editing of the element.
If this is not the first column, the error does not occur.
**Reason:**
Handler click on the column takes the value of href attribute from the first found in the string if this reference, as well as the parent <td> no class "nolink". The template "list_body_row" such a link is automatically inserted in the first column. Since in our example, the column has the class "nolink", the handler does not find a single link, and ignoring clicks on the line.
**Decision:**
Change the click handler so that he took the reference value of the data-attribute row (tr). Accordingly, the change template "list_body_row".

### DIFF
--- a/modules/backend/widgets/Lists.php
+++ b/modules/backend/widgets/Lists.php
@@ -544,7 +544,7 @@ class Lists extends WidgetBase
 
         $columns = array_keys($record->getAttributes());
         $recordOnClick = RouterHelper::parseValues($record, $columns, $this->recordOnClick);
-        return Html::attributes(['onclick' => $recordOnClick]);
+        return $recordOnClick;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


daftspunk:
Added a `clickable` option to list columns, see Backend > Lists docs

https://github.com/octobercms/docs/commit/d7e5ac5a316c695f2486b99c870b9e8eb74857e1

daftspunk:
This is a noble and worthy solution, thank-you for it. We decided to go with a slightly more elegant approach that doesn't require rewriting the docs and potentially breaking markup. See https://github.com/octobercms/october/commit/04e61adac302e9f203d5ed240290379f3e4d217d

PR #1601, opened 2015-11-16
Custom filename for thumb images

fuunnx:
Small changes going with [my PR on octobercms/library](https://github.com/octobercms/library/pull/148)


daftspunk:
Does the backend really need SEO enhanced images?

PR #1602, opened 2015-11-17
Fix "Unsupported operand types"

MatissJanis:
On a fresh installation of October with PHP 5.6.13 when AJAX calls a `dependsOn` AJAX handler, the following error is being thrown:

    "Unsupported operand types" on line 363 of /path/to/modules/backend/widgets/Form.php

Checking if `$eventResult` is array fixed this problem and everything was functioning as expected.

### DIFF
--- a/modules/backend/widgets/Form.php
+++ b/modules/backend/widgets/Form.php
@@ -360,7 +360,9 @@ class Form extends WidgetBase
         );
 
         foreach ($eventResults as $eventResult) {
-            $result = $eventResult + $result;
+            if (is_array($eventResult)) {
+                $result = $eventResult + $result;
+            }
         }
 
         return $result;


daftspunk:
Looks like it's working in [the test plugin](https://github.com/daftspunk/oc-test-plugin). Can you provide replication instructions?

Test plugin:
1. Open URI: backend/october/test/countries/update/1
2. Check/uncheck "Active" checkbox that uses dependsOn

MatissJanis:
Thanks for the test plugin!

I installed it on a fresh installation of october and everything seemed to be working. The same plugin wouldn't work on my site, so I started removing plugins to narrow down the problem.

It seems that the problem lies in the [Bedard.Debugbar](https://octobercms.com/plugin/bedard-debugbar) plugin. It causes this weird error in October.

Issue #1604, opened 2015-11-17
Error with save russian text


valiknet18:
Hello, i get error, when try save russian text. In result i get cropped text =C
Before: 
![image](https://cloud.githubusercontent.com/assets/5772019/11216079/1413c1fa-8d52-11e5-9dca-fc1d9d73ecd2.png)
Result: 
![image](https://cloud.githubusercontent.com/assets/5772019/11216096/2181b00e-8d52-11e5-8ce6-773952eb344e.png)
i don't know, it's plugin bug(rainlab.translation) or editor, help plz resolve him

luizbills:
Test without the translation plugin.

valiknet18:
it's bug(maybe) in rich editor, i fix him in https://github.com/octobercms/october/pull/1605

daftspunk:
Should be fixed as part of #2005 and should no longer occur. If the issue still exists, please log a new issue for it.

PR #1605, opened 2015-11-18
fix bug with crop text in richeditor

valiknet18:
fix this issue https://github.com/octobercms/october/issues/1604


daftspunk:
Superseded by #2005

PR #1607, opened 2015-11-18
Ignored composer.json. Fixed failing tests

mjacobus:
Tests were failing because the indexes were not the same.


daftspunk:
Merged in 8c689d63ea5d580d9e0981da273cf5cae061220b

Thanks!

PR #1609, opened 2015-11-19
Add events for sign in and out backend user.

mplodowski:
Add events for sign in and out backend user.

Remove unused import.

### DIFF
--- a/modules/backend/controllers/Auth.php
+++ b/modules/backend/controllers/Auth.php
@@ -7,13 +7,13 @@ use Redirect;
 use Validator;
 use BackendAuth;
 use BackendMenu;
-use Backend\Models\User;
 use Backend\Models\AccessLog;
 use Backend\Classes\Controller;
 use System\Classes\UpdateManager;
 use ApplicationException;
 use ValidationException;
 use Exception;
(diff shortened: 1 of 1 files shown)


mplodowski:
I am afraid this alone is not enough.

This events cannot be listened in Plugin boot method. Any idea how to use those events in Plugin development?

The idea is to have a way to add custom functionality after user sign in and sign out.

daftspunk:
Can you give more detail on you are trying to achieve?

mplodowski:
In my plugin I created some session variables and I want to forget them when user is sign out. In Laravel Auth class there is an event for this and for sign in as well.

There is core module which logs when user sign in to backend. I would like to extend this with my plugin, add action column and log some other backend events along with sign in.

I think developers should be able to modify how CMS works with event handlers and not modify code itself and create PRs.

Currently this PR is no good, because this events are not fired when I bind them in plugin boot method or init.php.

daftspunk:
You may want to set the `$elevated` property in your plugin base class to `true`, this will allow the event to work.

daftspunk:
Perhaps fixed by #1796

PR #1610, opened 2015-11-20
allow using get attribute mutators as nameFrom

martinemmert:
Adding those few lines enable using getAttribute mutators as nameFrom property on the RelationWidget.

Example:

*fields.yaml*
```
...
author:              
  type: relation     
  nameFrom: full_name
...
```

*Model.php*
```
...
public $belongsTo = [
    "author" => 'Backend\Models\User'
];
...
```
*Result*
```
print_r($fields->$options)

array(
    [0] => "first_name last_name"
)
```


daftspunk:
This feature was recently added in 030e1977e8f290d67645cf39f4a2a6e54aaf7a99

PR #1629, opened 2015-12-01
Simplify the account menu

gpasztor87:


daftspunk:
The previous solution showed a dark overlay when opening the user menu.

gpasztor87:
I fixed it.

Issue #1600, opened 2015-11-14
File Upload forms does not work with Relation Behavior and Pivot Model


myst6re:
When you add a file upload form in a pivot field in the controller's relation configuration, the attachment is not saved/attached with the pivot model: the image is uploaded, an entry seems to be added in system_files table, but after saving the page, the association between the pivot table and system_files is not correctly saved.

tschallacka:
+1 I'm encountering the same issue.
 
owning model RichIndexPage


    public $belongsToMany = ['richblock' => ['\ExitControl\RichIndexPageManager\Models\RichBlock',
    		                                  'key' =>'rich_page_id',
                                                'otherKey'=>'rich_block_id',
                                                'table' => 'rich_block_to_page',
                                                'order' => 'sortorder',
                                                'pivot'=>'sortorder',
                                                'timestamps'=>'true']];


Relation model RichBlock

    public $belongsToMany = ['richpage' => ['\ExitControl\RichIndexPageManager\Models\RichIndexPage',
    		                                  'key' =>'rich_page_id',
                                                'otherKey'=>'rich_block_id',
                                                'table' => 'rich_block_to_page',
                                                'pivot'=>'sortorder',
                                                'timestamps'=>'true']];
    public $attachOne = ['image' => ['System\Models\File',]];

tschallacka:
It has to do with deferred binding messing up, if I add this in 
FileUpload.php on line 263 ish

    if(strpos($this->alias,'Pivot')){
        $this->getRelationObject()->remove($file);
    }
    else {
        $this->getRelationObject()->remove($file, $this->sessionKey);
    }

then it works ish, although the ajax postback still fails miserably.

tschallacka:
I made a dirty workaround to allow me to do both things in the mean time...

in the relation render fields yaml add a partial where you will define your form fields for the pivot data

    <div class="form-group  text-field span-full   " data-field-name="subtitle" id="Form-relationRichblockManageForm-field-RichBlock-subtitle-group">
        
        <?php
        $pivot = '';
        foreach($this->vars['formModel']->richblock as $block) {
            if($model->getKey() == $block->getKey()) {
                $pivot = $block->pivot->sortorder;
            }
        }
        ?>
        
        <input type="hidden" name="RichBlock[pivot][owner]" value="<?= $this->vars['formModel']->getKey(); ?>">
        
[...]

PR #1641, opened 2015-12-05
Commit deferred data when creating/updating a pivot table

myst6re:
When you add an attachment to a pivot model, deferred binding is created but never commited.
This patch fix that (#1600).

**Note:** the pivot table must have an "id" auto increment and primary column, along with standard foreign and other keys (which cannot be primary). This column "id" must be explicitly listed in the parent model relation.

**Example** (from modified october/test plugin):

    // In updates/create_tables.php
    Schema::create('october_test_users_roles', function($table)
    {
        $table->engine = 'InnoDB';
        $table->increments('id');
        $table->integer('user_id')->unsigned();
        $table->integer('role_id')->unsigned();
        $table->unique(['user_id', 'role_id']);
        $table->string('clearance_level')->nullable();
        $table->boolean('is_executive')->default(false);
        $table->timestamps();
    });

    // In models/User.php
    public $belongsToMany = [
        // ...
        'roles_pivot_model' => [
            'October\Test\Models\Role',
            'table' => 'october_test_users_roles',
            'pivot' => ['id', 'clearance_level', 'is_executive'], // Adding id
            'timestamps' => true,
            'pivotModel' => 'October\Test\Models\UserRolePivot',
        ],
    ];

    // In models/UserRolePivot.php
    public $attachOne = [
        'image' => ['System\Models\File']
    ];


daftspunk:
Looks alright, merged in 55bceba32cf8455aae709a091434c076bbf8f3ca

mihailbishkek:
Image is successful add, but how I can see it? What I can write in fields.yaml???

PR #1651, opened 2015-12-11
Feature to allow filtering lists by date scopes (single date or range)

gabsource:
> Assets need to be recompiled with ` php artisan october:util compile assets`

~~# Pikaday translation~~

~~Using the built-in `$.oc.lang` module, the patch injects the i18n object into the datepicker init call.~~

> Moved to PR #1857 

# Lists filter by date scope

The proposed feature extends the built-in `Filter` class to allow two new types `datepicker` and `daterangepicker` which will trigger a datepicker UI to select a single date or two dates.

Dates value are passed to the conditions in the `Y-m-d` format. Dates are displayed on screen using the format set in the two language files (backend for PHP  / client for JS). If the time is needed (timestamp column type), the time can be manually entered in the conditions as in the examples below.

## Single

The same keyword `filtered` is used in the conditions with this new scope type.

```yaml
# ===================================
# Filter Scope Definitions
# ===================================
scopes:

    created_at:

        # Filter name
        label: Date

        # Filter type
        type: datepicker

        # Scope Conditions (match against a timestamp column for the whole day)
        conditions: created_at >= ':filtered 00:00:00' AND created_at <= ':filtered 23:59:59'
```

The refresh event is triggered on date selection.

single

## Range

Two new keywords are used `after` and `before` in the conditions with this new scope type.

```yaml
# ===================================
# Filter Scope Definitions
# ===================================
scopes:

    created_at:

        # Filter name
        label: Date

        # Filter type
        type: daterangepicker

        # Scope Conditions (match against a range of timestamp columns)
        conditions: created_at >= ':after 00:00:00' AND created_at <= ':before 23:59:59'
```

The refresh event is triggered when clicking the `Filter` button.

range-1
range-2


modules/backend/widgets/Filter.php
```diff
@@ -94,6 +97,14 @@ public function prepareVars()
      */
     public function renderScopeElement($scope)
     {
+        switch ($scope->type) {
+            case 'date':
+            case 'range':
+                // Load datepicker assets
+                new DatePicker($this->controller, new FormField('dummy','dummy'));
```
gabsource: This is a hack to load pikaday without duplication of assets files (maybe a cleaner solution can be proposed)

chesterx:
i am waiting for this

mosid:
+1
This is definitely a must have feature

null-byte-0:
+1

gergo85:
+1

brenodouglas:
+1

vojtasvoboda:
+1

mosid:
@daftspunk don't want to bug you, but  any input on this PR?
eagerly waiting for this feature.

daftspunk:
This is quite possibly the coolest PR feature to date. Bravo @gabsource

daftspunk:
Although the PR is accepted. This feature is partially implement. It still requires some more work.

Notes:
- We need to find a nice way to streamline the Reset/Apply buttons. It is inconsistent with the `group` filter type.
- Consider adding an `X` next to each input once it is populated. This could be a generic control type and can be useful in other places (Settings > Search area for example)
- The Reset button functionality appears broken: Select date, refresh page, reset date. Date still appears.
- The displayed date on the PHP side needs to be localized via `Backend::dateTime` format alias *dateMin*
- Explore if there is a better approach than adding the time to the `conditions` as `00:00:00` and `23:59:59`. There are benefits to flexibility, but the code could be nicer.
- Still need to do testing using model `scope` approach.

gabsource:
@daftspunk  Thanks for merging this PR

I've spotted a bug in the FilterWidget.getLang (typo in variable name). I will open a PR It cause the reset button not to update the label when clicked.

I will try to look into the points you've mentioned as soon as possible. The backend code as evolved quite a bit since I first posted this PR so it needs some refactoring to catch up with the new features like timezone.

I think adding an `X` button (on the left) when the input is populated is a good solution for the single date mode.
As far as I remember I did not manage to find a good way to get rid of the _Apply/Reset_ for the *range* mode. I had the feeling that auto-applying (when two dates are populated) and auto-reseting (when the two inputs are empty) was harsh.

daftspunk:
After some more thought, the X button is a "nice to have" we can add later. Here is a visual representation of the continuity requirement:

oc-filter-apply-reset

daftspunk:
To elaborate on the Timezone conversion point. Take a look at [the new docs for the datepicker control](https://github.com/octobercms/october/blob/d50c7591cfef6e243aabba28bef8a07784170440/modules/system/assets/ui/docs/datepicker.md#locale-and-timezone-handling). The value selected by the user should be converted from the "backend timezone" to the "app timezone" when it gets sent to the server. It should happen seamlessly behind the scenes.

gabsource:
Thank you very much for your replies. I will look into it.

For the conditions (query), the time is not mandatory, it depends on the database engine and the use case. It could also be something like this for MySQL

`DATE_FORMAT(created_at, GET_FORMAT(DATE,'ISO')) = DATE_FORMAT(:filtered, GET_FORMAT(DATE,'ISO'))`

or `DATE_TRUNC('DAY', ...)` for Postgres.

As I see it, it really depends on what the user wants to do. If he stores a simple date, no need to call extra SQL functions, but for timestamps we need to use something to get rid of the time part.

daftspunk:
In general, we always like KISS (keep it simple) to be the primary approach across all implementations. When complexity is needed, the syntax becomes more complex too. For example:

Simple every day use

        # Filter type
        type: daterange

        # Scope Conditions (match against a range of timestamp columns)

Complex requirement use

        # Filter type
        type: daterange

        # Scope Conditions (match against a range of timestamp columns)
        conditions: created_at >= ':afterDate 00:00:00' AND created_at <= ':beforeDate 23:59:59'

In the above examples the variables `:before` and `:after` should append the `00:00:00` and `23:59:59` time values respectively. It makes sense that the default behavior be inclusive of the selected dates. For those who need the more control (complexity), the alternates `:afterDate` and `:beforeDate` are available.

gabsource:
The _daterange_ type is simpler to handle in this case as there is already `after` and `before` placeholders.

For the _date_ there is only one for the moment `:filtered` which make it quite harsh to handle a timestamp column nicely. Maybe we should populate `after` and `before` keys for the _single date_ type also. It would be more consistent across both date type : 
- for _date_ and _daterange_ types : 
  * `after` holds the full datetime _Y-m-d 00:00:00_
  * `before` holds the full datetime _Y-m-d 23:59:59_
- for _date_ type only : 
  * `filtered` holds the short date _Y-m-d_  (keeping the same name as other filter widgets)
- for _daterange_ type only : 
  * `afterDate` holds the full datetime _Y-m-d_
  * `beforeDate` holds the full datetime _Y-m-d_

So users should write this for timestamp column using single date type (same as daterange) : 

```
    # Filter type
    type: date

    # Scope Conditions (match against a range of timestamp columns)
```

and this for date column : 

```
    # Filter type
    type: date

    # Scope Conditions (match against a range of timestamp columns)
    conditions: created_at = ':filtered'
```

PR #1663, opened 2015-12-20
Keep the migrations directory

pespantelis:
To avoid something like this:
http://stackoverflow.com/questions/34378313/failed-to-open-stream-error-when-executing-migratemake/34378428#34378428


daftspunk:
The standard interaction with October should see that you would never need this directory. If you are starting with a Laravel application base and including October via composer, you should get this directory as part of the boilerplate. What steps did you follow that required you to have this directory in existence?

PR #1665, opened 2015-12-21
Check if email subject set in closure

BOOMER74:
If subject setted in template, changing in closure like:

Mail::queue("template", [], function ($message) use ($subject, $email) {
	$message->subject($subject)->to($email);
});

has no effect.


daftspunk:
Merged in 922cf30cb6d47f8ec891e5ebe0e68b6cdd79c805

Thanks

PR #1668, opened 2015-12-23
Hotfix for mysql 5.7

### LINKED ISSUE #1619: october:up vagrant problem
I updated my virtual machine. I use vagrant - laravel/homestead box on windows. I want use php artisan october:up , because my  database disappeared and i want to rebuild. But i always get this error
bug1

  [Illuminate\Database\QueryException]
  SQLSTATE[42000]: Syntax error or access violation: 1067 Invalid default value for 'created_at' (SQL: create table `deferred_bindings` (`id` int unsi
  gned not null auto_increment primary key, `master_type` varchar(255) not null, `master_field` varchar(255) not null, `slave_type` varchar(255) not n
  ull, `slave_id` varchar(255) not null, `session_key` varchar(255) not null, `is_bind` tinyint(1) not null default '1', `created_at` timestamp defaul
  t 0 not null, `updated_at` timestamp default 0 not null) default character set utf8 collate utf8_unicode_ci engine = InnoDB)

I think it's because of the new homestead box, but i can't find any real solution.

### LINKED ISSUE #1667: Installation issue with Database
Installing on MySQL 5.7.9:
SQLSTATE[42000]: Syntax error or access violation: 1067 Invalid default value for 'created_at' (SQL: create table `deferred_bindings` (`id` int unsigned not null auto_increment primary key, `master_type` varchar(255) not null, `master_field` varchar(255) not null, `slave_type` varchar(255) not null, `slave_id` varchar(255) not null, `session_key` varchar(255) not null, `is_bind` tinyint(1) not null default '1', `created_at` timestamp default 0 not null, `updated_at` timestamp default 0 not null) default character set utf8 collate utf8_unicode_ci engine = InnoDB)

jwilson8767:
MySQL 5.7 defaults to NO_ZERO_DATE which makes Laravel's Schemabuilder timestamps() throw this error when running migrations: `Syntax error or access violation: 1067 Invalid default value for 'created_at'`
According to [laravel/framework issue #3602](https://github.com/laravel/framework/issues/3602) our options are to make timestamps nullable by default, default to something valid (like the current time), or to turn offthe NO_ZERO_DATE flag in mysql config. In the short term, setting strict=true allows October to be used with MySQL 5.7. However, this is a blanket fix and should be reviewed as the Laravel community addresses/solves this issue.

fixes #1667 
fixes #1619

### DIFF
--- a/config/database.php
+++ b/config/database.php
@@ -62,6 +62,7 @@ return [
             'charset'   => 'utf8',
             'collation' => 'utf8_unicode_ci',
             'prefix'    => '',
+            'strict'    => true
         ],
 
         'pgsql' => [


daftspunk:
Is this the same as setting `sql_mode=STRICT` in my.cnf/my.ini files?

jwilson8767:
Nope, this is telling Eloquent to produce output that is compliant with sql_mode=STRICT.

jwilson8767:
Sorry about targeting wrong branch. I know better.

adrenth:
I was wondering were that came from since I was using MySQL 5.7. Thanks for the fix.

jwilson8767:
As discussed in [the MySQL docs for 5.7](http://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sql-mode-changes), MySQL 5.7.8+ enables `NO_ZERO_DATE` by default, but strict mode does not include `NO_ZERO_DATE`. Thus the workaround to enable strict mode really does work. However, it is only a temporary fix and will in the future cause the very same issues to return.

Thus the final, long-lasting solution will be to create your **timestamp** columns with `$table->timestamp('created_at')->useCurrent();` followed by `$table->timestamp('updated_at')->useCurrent();` instead of using `$table->timestamps();`

PR #1671, opened 2015-12-26
Fix Template Caching

dqsully:
When developing an October site using an external editor, sometimes even October's own editor, the template cache does not always reflect the source files correctly (http://octobercms.com/forum/post/template-asset-caching-during-development). I stumbled across this and dug into October to find this:

(note: Atom decided to remove the excess whitespace, keep it if you like)


dqsully:
I will do better.

PR #1672, opened 2015-12-28
Fix twig caching

dqsully:
When October would load a file from its changed source, Twig would not see the message until it had gone. See `Cms\Classes\Loader->isFresh`. This meant a template would not update unless you deleted the Twig cache, or that template's TTL expired. Fix: add another variable (`freshness`) that would only change after being observed, and accurately reflected if a template's source had been modified. This includes adding a function to manage and return any object's freshness.

### DIFF
--- a/modules/cms/classes/CmsObject.php
+++ b/modules/cms/classes/CmsObject.php
@@ -12,6 +12,8 @@ use RecursiveDirectoryIterator;
 use RecursiveIteratorIterator;
 use ArrayAccess;
 use Exception;
+use DateTime;
+use DateInterval;
 
 /**
  * This is a base class for all CMS objects - content files, pages, partials and layouts.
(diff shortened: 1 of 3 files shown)


daftspunk:
This code looks alright, just not sure what problem it actually solves, can you provide some replication instructions for the bug?

daftspunk:
### Replication instructions:

1. Using the `demo` theme supplied with October.
1. Open the **home.htm** file and include a link to the plugins page anywhere in the markup:

    `{{ 'plugins'|page }}`
1. Save the **home.htm** file
1. Open the **plugins.htm** file and make any change, i.e change the heading to *Plugin components!!!!!* 

    *Important*: Do not open the [Plugin components] page in the browser

1. Open the [Home] page in the browser
1. Navigate to the [Plugin components] page and the changes are not seen

daftspunk:
Thanks!

daftspunk:
This PR was rolled back, alternative fix implemented: 5a70cec7a2f17494a5e519236095e831bac9daa3

PR #1676, opened 2015-12-31
Fix permissions required for theme settings

dshoreman:
If you create a non-superuser group, users of that group won't see the *Front-end theme* or *Mantenance mode* pages under Settings, despite having the *Manage themes* permission.

This fixes the issue by updating the setting definitions to require the correct permission codes, [defined in the service provider](https://github.com/octobercms/october/blob/96959e0683def248806f8b56a1baaea3be01d9a9/modules/cms/ServiceProvider.php#L186-L190) and referenced from the [CMS Themes controller](https://github.com/octobercms/october/blob/51446f6a27b5680ba062c5afba71f95c6fda7c46/modules/cms/controllers/Themes.php#L36).


daftspunk:
Good find, thanks!

PR #1688, opened 2016-01-10
Use Grammar from Query in List Widget

CSNWEB:
When using models with a different database software than the default, the old code produced a syntax error. 
Using the Grammar from the query ensures that the alias is always wrapped correctly. 

Note: The code should be scanned for similar errors, produced by using DB Facade.


CSNWEB:
Sry, i should have read the guidelines before. Created a merge request for the dev branch now.

Issue #1578, opened 2015-11-03
Repeater type does not validate field type


BOOMER74:
I have this fields in plugin:
```yaml
products:
    type: repeater
    prompt: Add new product
    form:
        fields:
            product:
                span: left
                label: Product
                type: dropdown
            quantity:
                span: right
                label: Quantity
                type: number
```
**quantity** has rendered as number (right aligned), but not validate on create/save. I know validator in model cannot catch this, because it json, but maybe need add JS validator for this? And if set **default: 1** this work only on create new item, when reopen item and chage **quantity** has empty value.

jBOKA:
+1
You may extended the Validator in the Plugin's `boot` method like this
```php
Validator::extend('validationname', function($attribute, $value, $parameters) {
    // $value will be an php array in the structure of the repeater's fields structure
    // all other things like customMessages can then be configured in the usual way in the Model
    if ($something) {
        return true;
    } else {
        return false;
    }
});
```
This works, since the json conversion is processed after the validation on saving into the database. When fetching the data from the database, the model will again hold an array instead of a json string, which is the magic of the jsonable trait.

I don't know if this is the best solution, but it integrates nicely into the backend. The only drawback is, that you cannot address the index of the repeated form.

nozreh:
Hi Guys, I have added a custom validation rule, with custom message, but how can I validate the value correctly?

I have a repeater Carton qty and I want to validate it as numeric only, same as BOOMER74 but how can I use the existing rule 'numeric' in my custom validation?

Validator::extend('numeric_loose_carton_qty', function($attribute, $value, $parameters) {
			foreach ($value as $v) {
				
				if( !is_numeric($value) ){
					return false;
				}

			}

			return true;
});

Do i have to loop the value? tried this it's not working.

jBOKA:
Yes, by defining the new rule `numeric_loose_carton_qty` you are actually validating the whole repeater form. Therefore you have to run your validation logic for each repetition but throw an error for the whole repeater array.
As for your question, you would do
```PHP
Validator::extend('validation_rule_name_for_carton_repeater', function($attribute, $value, $parameters) {
    foreach ($value as $v) {
        $validator = Validator::make(
            //first values
            $v,
            //then rules
            [
                'product' => '...whatever',
                'quantity' => 'numeric'
            ]
        );
        if ($validator->fails())
            return false;
    }
    return true;
}
```
As I said before, there's no way to address the index within the repeater.

nozreh:
[...]

PR #1691, opened 2016-01-12
Convert number field in repeater to float

BOOMER74:
This also fix [this](https://github.com/octobercms/october/issues/1578) issue.

### DIFF
--- a/modules/backend/formwidgets/Repeater.php
+++ b/modules/backend/formwidgets/Repeater.php
@@ -102,7 +102,22 @@ class Repeater extends FormWidgetBase
      */
     public function getSaveValue($value)
     {
-        return (array) $value;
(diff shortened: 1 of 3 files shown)


daftspunk:
Hmmm, I'm not sure if the `Form` class should really be concerned about `Repeater`, so this logic is in the wrong replace. It should live with the `Repeater` class and get pre-processed there.

BOOMER74:
@daftspunk I move code from Form to Repeater. Maybe now checking for type is not necessary. About ajax: need to change framework.js:268. Sorry, its in one commit.

daftspunk:

LukeTowers:
Closed due to lack of follow up from @BOOMER74

PR #1693, opened 2016-01-12
Flash twig node now displays all messages of all types.

### LINKED ISSUE #1225: Markup : flash session doesn't display multiple messages of the same type
I found a bug in markup that prevents multiple flash messages of the same type to show width the {% flash %} tag.

Exemple :
if I register 5 error flash messages and 5 warning messages, {% flash %} {% endflash %} will only show up the first registered error message and the first warning one.
However, we can pass over this bug by specifying a message type to show.
To make my exemple work, I have to write {% flash error %} {% endflash %}  {% flash warning %} {% endflash %}. Works but not fine as I have to repeat code in some cases.

Best regards

jwilson8767:
Fixes #1225

### DIFF
--- a/modules/cms/twig/FlashNode.php
+++ b/modules/cms/twig/FlashNode.php
@@ -33,13 +33,17 @@ class FlashNode extends Twig_Node
         ;
 
         if ($attrib == 'all') {
-            $compiler
+           $compiler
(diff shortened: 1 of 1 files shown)


daftspunk:
Thanks

PR #1700, opened 2016-01-14
Minor fix in timepicker

gpasztor87:
When I use Revisionable trait. I become the following record many times on system_revisions table:

field: start_time
old_value: 08:00:00
new_value: 08:00:00:00


daftspunk:
Thanks

PR #1704, opened 2016-01-15
feature/improvement

adrenth:
Code improvements and added proper PhpDoc


daftspunk:
`array_key_exists` is a much slower function than `isset`

adrenth:
You're right! Although the PHP inspection tells me to replace it with array_key_exists for better readability. I reversed the specific lines. Should now be ok.

daftspunk:
Merged in 386f958b6e427137a0a5cd25b92b9730827e7e38

Turns out `===` is faster than `==` too! Good job

Issue #84, opened 2014-05-15
Sort CMS pages by URI/Name


JohannesMP:
When you have nested pages it becomes really difficult to manage them if they ordered by Title.

I think it's necessary to add the option to have the list of pages ordered by URI, because doing so would logically group pages together that share the same parent URI.

Here is a mockup of how this could work:
![order_mockup](https://cloud.githubusercontent.com/assets/662874/2982904/400f3ecc-dc1b-11e3-97cd-3d50ecbc73ea.png) (similar in style to the ordering options on http://octobercms.com/plugins/)

As you can see, even with only a few pages this is more logical to follow. The Root/Home directory `/` is automatically the first page in the list, and pages nested under the same directory such as `/demo/` are grouped together.

alekseybobkov:
Implemented in 6e54433bc14e1eb8473e97e1148ce4c1301bc2dd. The original default sorting option is the file name, not the title. It's default now too, but you also can choose Title or URL in the drop-down.

PR #1705, opened 2016-01-16
Feature/#84

adrenth:
See issue #84 

Added a function which sorts the page items; by default it is sorting the items by URL (ascending).
But the `sortPages` method is also able to sort items by `Title` (asceding/descending).

This is just a first attempt. The UI part is something I left for later. What do you think about it?


daftspunk:
#84 is assigned to @alekseybobkov to look at when this cycle of work begins.

This can't really be merged because it is incomplete. Thanks for this, it will be useful for research purposes when the time comes to implement this feature.

PR #1708, opened 2016-01-18
Add extra file types

gpasztor87:


daftspunk:
Thanks, could you add them to [the database default file types](https://github.com/octobercms/library/blob/develop/src/Database/Attach/File.php#L661) too?

gpasztor87:
Yes, I added them.

daftspunk:
Thanks

Refs https://github.com/octobercms/library/pull/170

PR #1711, opened 2016-01-19
Example security: nginx.conf

emtudo:


vinicius73:
+1

nginx.conf.example
```diff
@@ -0,0 +1,34 @@
...
+  root          /var/www/october;
+  index         index.php;
+
+  # Deny access to dotfiles
+  location ~ /\. {
+    deny  all;
+  }
+
+  # Deny access to framework directories
+  location ~ /(bootstrap/|config/|modules|plugins/|storeage/|themes/|vendor/) {
```
jwilson8767: "storage" not storeage

emtudo:
fixed!

nginx.conf.example
emtudo: fixed! thanks!

daftspunk:
These improvements should be made to [the documentation](https://github.com/octobercms/docs/blob/master/setup-configuration.md#nginx-configuration).

This is not something we really want to include in the core files.

PR #1720, opened 2016-01-23
Phpdoc cleanup

jwilson8767:
Most of the changes here represent phpdoc-only changes. However, there are few places where return types did not match what was in the phpdoc or seemed to be expected by the usages of the relevant method. I have done my best to ensure that all return types in the files I touched are now consistent with the way the return values are used.

Phpstorm was used to validate phpdoc blocks. 

At your request I can also clean up other files as part of this pr. I only cleaned the files immediately associated with a form issue I was troubleshooting today.

### DIFF
--- a/modules/backend/widgets/Form.php
+++ b/modules/backend/widgets/Form.php
@@ -1,8 +1,6 @@
 <?php namespace Backend\Widgets;
 
-use App;
 use Lang;
-use Input;
 use Event;
 use Form as FormHelper;
 use Backend\Classes\FormTabs;
@@ -95,7 +93,7 @@ class Form extends WidgetBase
     ];
 
     /**
-     * @var array Collection of all form widgets used in this form.
+     * @var FormWidgetBase[] Collection of all form widgets used in this form.
      */
     protected $formWidgets = [];
 
@@ -110,7 +108,7 @@ class Form extends WidgetBase
     public $previewMode = false;
 
     /**
-     * @var Backend\Classes\WidgetManager
+     * @var \Backend\Classes\WidgetManager
      */
     protected $widgetManager;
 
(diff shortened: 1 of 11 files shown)


modules/backend/widgets/Form.php
```diff
@@ -338,7 +353,7 @@ public function onRefresh()
                 if (!isset($this->allFields[$field])) {
                     continue;
                 }
-
+                /** @var FormWidgetBase $fieldObject */
```
jwilson8767: Inline `@var` tags like this allow IDEs to resolve the method used on objects (and therefore also their return types)

modules/system/traits/ConfigMaker.php
```diff
@@ -149,11 +153,11 @@ public function getConfigPath($fileName, $configPath = null)
         foreach ($configPath as $path) {
             $_fileName = $path . '/' . $fileName;
             if (File::isFile($_fileName)) {
-                break;
+                return $_fileName;
             }
         }
 
-        return $_fileName;
+        return '';
```
jwilson8767: This change prevents returning `$_fileName` which may be undefined.

modules/backend/widgets/Form.php
```diff
@@ -526,17 +545,18 @@ public function addFields(array $fields, $addToArea = null)
...
     public function addTabFields(array $fields)
     {
-        return $this->addFields($fields, 'primary');
+        $this->addFields($fields, 'primary');
     }
 
     public function addSecondaryTabFields(array $fields)
     {
-        return $this->addFields($fields, 'secondary');
+        $this->addFields($fields, 'secondary');
```
jwilson8767: `$this->addFields` returns void, and should not be returned.

daftspunk:
Merged in 6f3d7a58f2d18e834a5d6cf51bc1702ac5bef117

Thanks for this, a quick note: we won't be using phpDoc for API docs, more than likely it will be a custom solution.

PR #1721, opened 2016-01-23
Fix view bag

### LINKED ISSUE #1571: ViewBag component required properties
Currently, if I have a partial that uses the ViewBag component, and the component property is not defined - this will result in an error.

**Example partial:**
````
<div class="container">
    {{ viewBag.someProperty }}
    {{ viewBag.someOtherProperty }}
</div>
````
**Example page:**
````
url = "/"

[viewBag]
someProperty = "TEST"
==
{% partial 'somename' %}
````

**Expected result:** 200
**Actual result:** 500 - An exception has been thrown during the rendering of a template ("The component 'Cms\Classes\ViewBag' does not contain a method 'someOtherProperty'.").

**Example working page:**
````
url = "/"

[viewBag]
someProperty = "Hello World"
someOtherProperty = "TEST"
==
{% partial 'somename' %}
````

**Workaround partial**
````
<div class="container">
    {{ viewBag.properties.someProperty }}
    {{ viewBag.properties.someOtherProperty }}
</div>
````

### PR DESCRIPTION
**Pull request by jwilson8767**
Fixes #1571 by 1) changing the behavior of ViewBag::__call()  and 2) by finishing `@todo Check required properties` in PropertyContainer.

### DIFF
--- a/modules/system/traits/PropertyContainer.php
+++ b/modules/system/traits/PropertyContainer.php
@@ -1,4 +1,5 @@
 <?php namespace System\Traits;
+use October\Rain\Exception\ValidationException;
 
 /**
  * Property container trait
@@ -36,7 +37,15 @@ trait PropertyContainer
         }
         $properties = array_merge($defaultProperties, $properties);
 
-        // @todo Check required properties
+        // Verify required properties are present
+        $missingRequiredProperties = [];
+        foreach ($definedProperties as $name => $information) {
+            if (isset($information['required']) && $information['required'] && empty($properties[$name])) {
+                $missingRequiredProperties[$name] = isset($information['validationMessage'])
+                    ? $information['validationMessage']
+                    : trans('config.required', ['location' => get_class($this), 'property' => $name]);
+            }
+        }
 
         return $properties;
     }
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by jwilson8767** [comment]
Please verify that the `ValidationException` thrown is correct in this context and that the message it is given is acceptably clear.

File: modules/system/traits/PropertyContainer.php [hunk 1]
```diff
@@ -1,4 +1,5 @@
 <?php namespace System\Traits;
+use October\Rain\Exception\ValidationException;
```
dshoreman: Doesn't look like this needs to be here

daftspunk:
Alt fix implemented.

Not quite sure what the addition to `PropertyContainer` was supposed to do?

jwilson8767:
the addition there validated that required properties are actually present, I believe one place this is used is [in components](https://octobercms.com/docs/plugin/components#component-properties)

daftspunk:
I mean this code literally does nothing except populate an unused `$missingRequiredProperties` variable?

        // Verify required properties are present
        $missingRequiredProperties = [];
        foreach ($definedProperties as $name => $information) {
            if (isset($information['required']) && $information['required'] && empty($properties[$name])) {
                $missingRequiredProperties[$name] = isset($information['validationMessage'])
                    ? $information['validationMessage']
                    : trans('config.required', ['location' => get_class($this), 'property' => $name]);
            }
        }

jwilson8767:
Huh, you're right. I think next was to throw a `ValidationException` or some other way to report that there were missing required properties.

PR #1722, opened 2016-01-24
Fix negative count

tresbach:


daftspunk:
Aha, good pick up! Thanks

PR #1723, opened 2016-01-24
Make default 404 page use 404 status code

Herlevsen:
On a fresh installation, when the default 404 page is visited at domain.com/404, the HTTP status code is 200. That would make search engines able to index the 404 page. IMO the url should act as if the 404 page itself did not exist.

### DIFF
--- a/modules/cms/classes/Controller.php
+++ b/modules/cms/classes/Controller.php
@@ -190,7 +190,7 @@ class Controller
         /*
          * If the page was not found, render the 404 page - either provided by the theme or the built-in one.
          */
-        if (!$page) {
+        if (!$page || $url === '404') {
             $this->setStatusCode(404);
 
             // Log the 404 request


jBOKA:
Shouldnt the 404 page be marked invisible? There's no need to make the content of the 404 available through normal URL.

Herlevsen:
By invisible, do you mean hidden? I'm all new to October.

jBOKA:
Hey, yes I mean the "hidden" checkbox.
I have to correct myself. The page indeed needs to have the URL /404.
But If you switch it to "hidden", every visitor, that is not logged in to the backend gets redirected to the proper 404 with the correct status code (due to "hidden") . Even if he visits /404 directly.
Just keep in mind to log out of the backend for testing or use an incognito window.

Herlevsen:
But since the 404 url is hardcoded into the system, i think it would make sense to always have it return a 404 status code. Alternatively, the 404 page should be always be hidden, and the checkbox should be disabled.

Personally i prefer the first one. What do you think?

jBOKA:
I guess you're right about the URL being hardcoded, so +1 for the standard
error code. Let the developers decide.

Jens Herlevsen <notifications@github.com> schrieb am Mo., 25. Jan. 2016 um
15:31 Uhr:

daftspunk:
The URL `/404` is a fixed system page, so opening /404 on any website showing a 404 page should be a hallmark of someone running an October website? Well, no, most websites would return 404 on this URI.

@Herlevsen is correct, it should return 404 status code, to emulate a missing page and remove *response code 200 on /404 URI* as a hallmark.

Now what to do about the `/error` URI? ;-)

PR #1727, opened 2016-01-26
Fix for static primary key naming relationcontroller

tschallacka:
RelationController forced belongs to relations to have a primary key named id.
If primary key was named differently it wouldn't register the onclick.
This fixes that by taking the primary key name from the model.


jfoliveira:
+1 please merge this and make OctoberCMS usable with non `id` PKs.

daftspunk:
Thanks

PR #1728, opened 2016-01-27
Make sure Project ID is always included in POST

dshoreman:
Attempting to install a paid plugin through Artisan would return an 'authority not found' error, when the same would work fine when clicking "Update" in the backend.

This moves the code that adds the project ID into the applyHttpAttributes method to ensure it gets added for all post requests, and not just those that are actually requesting a file.


daftspunk:
Need to see if this can be fixed at the API level. Fetching plugin details shouldn't need authorization.

PR #1736, opened 2016-01-29
Fix framework's extra loading indicator

jakerator:
There is a bug with loading strape. It initiates before Jquery's document.ready. So if you have some large external javscrtips after framework extras, it may show the JS error "Cannot read property 'show' of undefined"  if you click some button with "data-request" attribute quickly before page fully loaded.
Simple fix is to move indicator initialization into $(document).ready also.


devWebMntq:
+1

daftspunk:
Dupe #1737

PR #1738, opened 2016-01-29
Add vim and emacs keybindings support to CMS

ariponce:
Everytime i use the CMS i struggle due to my vim habits and seeing that Ace editor offers support for vim keybindings, i decided to add them to the CMS editor.

I've also added the Emacs because, why not?

They can be activated on the Editor preferences page.


modules/backend/formwidgets/codeeditor/assets/js/codeeditor.js
```diff
@@ -64,7 +64,7 @@
         showPrintMargin: true,
         highlightSelectedWord: false,
         hScrollBarAlwaysVisible: false,
-        readOnly: false
+        readOnly: false,
```
myst6re: This extra comma will cause an error on some old JS interpreter.

modules/backend/formwidgets/CodeEditor.php
```diff
@@ -99,7 +104,8 @@ public function init()
             'fontSize',
             'margin',
             'theme',
-            'readOnly'
+			'readOnly',
+			'keyboard_binding'
```
myst6re: Wrong indentation, you must be consistent.

daftspunk:
Please use spaces instead of tabs. `keyboard_binding` should be `keyboardBinding` too I think.

dshoreman:
+1. Would love to see support for vim bindings in the editor.

dshoreman:
@ariponce A lot of the files still have a few tabs lingering in them. It's probably easier if you tell your editor to convert tabs to spaces rather than manually editing them.

@daftspunk Looking at the `keyboardBinding` var, are you sure? All other preferences use `snake_case` format, now it's the odd one out

ariponce:
Ok, i think i converted all tabs to spaces.

Regarding the `keyboardBinding` var, i've used snake case on the lang file because i thought that was the standard (all other variables are snake case). I can change it to camel case if you want me to.

ariponce:
Any news on this?

daftspunk:
The editor page has since been rebuilt, sorry for the delay. Please resubmit at your earliest convenience.

Issue #1359, opened 2015-08-07
Morphed Backend Relations


ab192130:
Hi OctoberCMS Team,

I have a Eloquent Model with morphed relations. **Product** model morphs to the models like **Car** and **Estate**. The table for products looks like this:

```
| id | title         | object_id | object_type |
|----|---------------|-----------|-------------|
| 1  | Mercedes Benz | 1         | car         |
```

How can I manage product model in my backend with **Backend Relations** feature? I tried it but it throws an error like "unknown variable - $widget". 

What is your best practices with it? Thanks...

ARH-Digital:
Unfortunately they do not permit morphic relations in the backend.
https://octobercms.com/docs/backend/relations#relationship-types

This can be fixed by adding morphs into the RelationController.php
I will look at fixing this and submitting a pull request. Hopefully they accept it as I have used this a couple of times now.

ab192130:
Thanks you, @ARH-Digital. It made up my day. I will look forward to it.

daftspunk:
Fixed by #1741

Arekusei:
Many To Many Polymorphic relations are not supported yet.

In file https://github.com/octobercms/october/blob/develop/modules/backend/behaviors/RelationController.php#L1341 presents duplicated code for morph relations.

daftspunk:
Thanks @Arekusei, please submit a Pull Request to the [test plugin for investigation](https://github.com/daftspunk/oc-test-plugin) that demonstrates the issue. Don't forget to include step by step instructions how how to replicate it *using the test plugin*.

Arekusei:
@daftspunk Okay, I'll do later pull request for the test module.

Arekusei:
@daftspunk, done.

LukeTowers:
@Arekusei Should be fixed by #2275: see https://github.com/octobercms/october/issues/2561#issuecomment-277541963 for further information

PR #1741, opened 2016-02-01
Add morphic relations to backend

ARH-Digital:
This will fix issue #1359 

I have needed this feature one a couple of projects now and have either had to do work arounds or edit the core files.

This Pull Request contains a modified RelationController.php
The changes are adding 'morphOne' and 'morphMany' into the controller to permit polymorphic relations to display and be edited in the backend.
Attempting to do this previously would result in error of:
'Undefined variable: widget'

### DIFF
--- a/modules/backend/behaviors/RelationController.php
+++ b/modules/backend/behaviors/RelationController.php
@@ -1102,7 +1102,7 @@ class RelationController extends ControllerBehavior
                 if ($this->relationType == 'belongsToMany') {
                     $this->relationObject->detach($checkedIds);
                 }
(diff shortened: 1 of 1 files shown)


grantjames:
Are you sure this works? I applied your changes and tried setting up a polymorphic many-to-many relationship (the taggables example from the docs) but it still didn't work. Could you provide a link to a working codebase where you've got this working?

Thanks!

daftspunk:
This requires a test case in the [test plugin](https://github.com/daftspunk/oc-test-plugin).

ARH-Digital:
I will do up a test shortly when time permits. Work got busy sorry.

PR #1748, opened 2016-02-04
Allow nested form field widgets to send AJAX requests

Flynsarmy:
This PR allows nested widgets to send AJAX requests.

Usually you can only have an ajax handler like `config::onGetAutocompleteOptions` however nested form fields will send `config[rates]::onGetAutocompleteOptions` which isn't allowed currently.

### Example
Say I have a DataTable widget of name `config[rates]`. If I click a field in the table it tries to send an AJAX request with its handler name `config[rates]::onGetAutocompleteOptions` resulting in a JS error
> Invalid handler name. The correct handler name format is: "onEvent".

### DIFF
--- a/modules/backend/classes/Controller.php
+++ b/modules/backend/classes/Controller.php
@@ -401,7 +401,7 @@ class Controller extends Extendable
                 /*
                  * Validate the handler name
                  */
-                if (!preg_match('/^(?:\w+\:{2})?on[A-Z]{1}[\w+]*$/', $handler)) {
+                if (!preg_match('/^(?:\w+(?:\[\w+\])*\:{2})?on[A-Z]{1}[\w+]*$/', $handler)) {
                     throw new SystemException(Lang::get('cms::lang.ajax_handler.invalid_name', ['name'=>$handler]));
                 }
 
--- a/modules/system/assets/js/framework.js
+++ b/modules/system/assets/js/framework.js
@@ -21,7 +21,7 @@ if (window.jQuery === undefined)
         if (handler == undefined)
             throw new Error('The request handler name is not specified.')
 
-        if (!handler.match(/^(?:\w+\:{2})?on*/))
+        if (!handler.match(/^(?:\w+(?:\[\w+\])*\:{2})?on*/))
             throw new Error('Invalid handler name. The correct handler name format is: "onEvent".')
 
         /*


daftspunk:
I think the DataTable should probably be using a more appropriate alias instead.

```
$config->alias = studly_case($this->fieldName);
```

Flynsarmy:
I'm not sure such a thing is always possible. For example in this case the DataTable that I add programatically to my form spawns its own Table widget which uses the DataTable's fieldName as its alias. So I'd need to change the alias of the spawned widget(s), not the formfield itself.

If in my controller I do something like:
```php
public function create($context = null)
{
    $this->asExtension('FormController')->create($context);
    $tableWidget = $this->formGetWidget()->getFormWidget('config_data[rates]')->getTable();
    $tableWidget->alias = studly_case(\October\Rain\Html\Helper::nameToId($tableWidget->alias));
}
```

it merely results in the ajax events failing with error
> A widget with class name 'ConfigDataRates' has not been bound to the controller

The PR I've provided works perfectly and acts just like every other AJAX call in October. Requiring a bunch of custom code in some specific circumstances "just 'cause" sounds needlessly confusing and redundant. Should we not be trying to keep these things as seemless as possible?

daftspunk:
I simply mean updating [this line instead](https://github.com/octobercms/october/blob/develop/modules/backend/formwidgets/DataTable.php#L156)

Flynsarmy:
Just studly_casing keeps the [] however I could send a PR to change it to
```php
$config->alias = studly_case(\October\Rain\Html\Helper::nameToId($this->fieldName));
```
There are a couple of issues with this approach though.
* Downright changing its alias in this way could result in duplicate field conflicts. In the above example `config_data[rates]` is changed to `ConfigDataRates` which could potentially conflict with an existing model field - we can't know for certain.
* I have a feeling if I saved the page it'd also try to save into `$model->ConfigDataRates` instead of my `$model->config_data` array and I'd need to manually fix that up in a pre validation method.

The Controller classes `runAjaxHandler` method already supports nested events so I'm not sure why we're trying so hard to avoid allowing them to be used through validation rules. You can (and the `runAjaxHandler` method does) call $this->widget->{"config_data[rules]"} just fine. It'll work.

daftspunk:
Happy for the potential conflict to occur for the sake of a simple solution. In the unlikely event of a conflict, we can add the ability to manually define the alias in the form field definition. Eg:

```
config_data[rates]:
  type: datatable

config_data_rates:
  type: datatable
  alias: ConfigDataRates2
```

I sincerely doubt it will ever come to this though. Adding a suffix of `datatable` will further reduce the probability of a conflict, it would then require the same formwidget of the same generated identifier.

```
use \October\Rain\Html\Helper as HtmlHelper;

$config->alias = studly_case(HtmlHelper::nameToId($this->fieldName)) . 'datatable';
```

Flynsarmy:
Your code is confirmed to work. I'll send a PR.

My PR stands though - I don't see why we aren't allowing nested fields through validation when they work with your event handler. You can close this PR if you don't think it's a good idea for some reason.

PR #1749, opened 2016-02-04
Remove brackets from DataTable's table widget

Flynsarmy:
See #1748 for full discussion.

Say I have a DataTable widget of name config[rates]. If I click a field in the table it tries to send an AJAX request with its handler name config[rates]::onGetAutocompleteOptions resulting in a JS error
> Invalid handler name. The correct handler name format is: "onEvent".

This PR removes the brackets from `config[rates]` fixing the validation errors and letting the form submit correctly.


daftspunk:
Thanks

PR #1762, opened 2016-02-09
Feature component autowiring

mcustiel:
Hi. 
I would like to have the ability to inject dependencies in my component classes. I noticed that they are registered in a way that it is not allowed because they need some specific parameters on the constructor method. 
I added the option to register components factories in the registerComponents method of the plugin registration class, so this factory can create the component adding it's required arguments and all other dependencies required by it.

----

This change would require a similar change in SnippetManager class from rainlab.pages plugin.

### DIFF
--- a/modules/cms/classes/ComponentManager.php
+++ b/modules/cms/classes/ComponentManager.php
@@ -4,6 +4,7 @@ use Str;
 use Illuminate\Container\Container;
 use System\Classes\PluginManager;
 use SystemException;
+use Illuminate\Support\Facades\App;
 
 /**
  * Component manager
@@ -132,7 +133,7 @@ class ComponentManager
         return $this->codeMap;
     }
 
-    /** 
+    /**
      * Returns an array of all component detail definitions.
      * @return array Array keys are component codes, values are the details defined in the component.
      */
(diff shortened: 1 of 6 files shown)


daftspunk:
This is an interesting approach. Why not simply update Line 214 of ComponentManager from this

```
$component = new $className($cmsObject, $properties);
```

to this

```
$component = \App::make($className, [$cmsObject, $properties]);
```

Then we have the ability to use DI in the constructor, eg `Demo\ToDo` component:

```
<?php namespace October\Demo\Components;

use Illuminate\Contracts\Config\Repository as Config;
use Cms\Classes\CodeBase;
use Cms\Classes\ComponentBase;
use ApplicationException;

class Todo extends ComponentBase
{

    public function __construct(CodeBase $cmsObject = null, $properties = [], Config $config)
    {
        parent::__construct($cmsObject, $properties);
    }

    //...
}
```

mcustiel:
Nice tip. I don't have much of an idea about laravel so I did not know about passing parameters to the make method. I will change it.

daftspunk:
Can you squash these additions in to a single commit?

mcustiel:
Ok. Will close this one and reopen.

Issue #1761, opened 2016-02-08
Is it really possible to use OctoberCMS if your primary keys are not named `id`?


jfoliveira:
Hi, I'm working with OctoberCMS and I must use and existing database whose adopted naming convention for primary keys is to use the table name as the PK column name.

If I add an `id` column to the `Category` table everything works fine: a dropdown is displayed on the product form with the current category correctly selected.

Unfortunately I can't add the `id` column to the `Category` table as this is an existing production database.

Here are my models, notice the `belongsTo` relation definition:

```php
<?php namespace My\Plugin\Models;
use Model;

class Product extends Model
{
    public $table = 'Product';
    protected $primaryKey = 'Product';

    protected $guarded = ['*'];
    protected $fillable = [];
    public $hasOne = [];

    public $belongsTo = [
        'category' => [
            'My\Plugin\Models\Category',
            'key' => 'Category',
            'otherKey' => 'Category',
        ],
    ];

    public $belongsToMany = [];
    public $morphTo = [];
    public $morphOne = [];
    public $morphMany = [];
    public $attachOne = [];
    public $attachMany = [];

}
```

And the category model:

```php
<?php namespace My\Plugin\Models;

use Model;

class Category extends Model
{
    public $table = 'Category';
    protected $primaryKey = 'Category';

    protected $guarded = ['*'];
    protected $fillable = [];
    public $hasOne = [];
    public $hasMany = [];
    public $belongsTo = [];
    public $belongsToMany = [];
    public $morphTo = [];
    public $morphOne = [];
    public $morphMany = [];
    public $attachOne = [];
    public $attachMany = [];
}
```

My `Product` model `fields.yaml` file:

```yaml
    category:
        label: Category
        span: left
        type: relation
        nameFrom: Description
        emptyOption: Select a category
```

No exception is thrown/logged when the `id` field doesn't exist. As can be seen by the `belongsTo` relation above I'm clearly specifying the `Category` field to be used as the relation key.

Any ideas why by just adding the `id` field everything works? What can I do to use OctoberCMS with PK fields that are not named `id`?

tschallacka:
in the model define a property named

    public $primaryKey = 'primary_key_name';

and then it'll work fine. just also modify controllers etc.. everything that defaults to id.

https://octobercms.com/forum/post/howto-use-a-different-primary-key-than-id-in-a-model-or-controller

tschallacka:
for your use case, define a $primaryKey in your Category  where you define it's key.

[...]

PR #1768, opened 2016-02-10
Fix hardcoded column name `id` on recordfinder form widget

jfoliveira:
See: #1761
Thanks @tschallacka!

It is currently not possible to use the `recordfinder` field type if a database table does not contain an `id` column. This is due to the fact this `id` name was hardcoded on the plugin code.

This PR assumes the `keyFrom` property will be defined in the field options:
```yaml
category:
    label: Category
    type: relation
    nameFrom: Description
    keyFrom: Category      # <------- Important if your table doesn't have an id field
    emptyOption: Select a category
```

It would be great if this could be merged as it is a show stopper if one cannot change the database table to include an `id` column.


daftspunk:
LGTM!

PR #1770, opened 2016-02-11
Deepen nesting limit of CmsObjects in directories

### LINKED ISSUE #1707: Partials not working in deeper nested directories
* `/partials/header.htm` works.
* `/partials/layout/header.htm` works.
* `/partials/layouts/main/header.htm` fails.

It's the exact same file, but it doesn't work if it is nested two deep. It would be nice to be able to nest deeper.

jwilson8767:
I attempted to make the most minimally invasive change that would fix #1707. The lists in the backend now show groups like: `b`, `b/c`.


jwilson8767:
Close for wrong branch.

PR #1771, opened 2016-02-11
Deepen nesting limit of CmsObjects up to 4 directories (up from 2)

### LINKED ISSUE #1707: Partials not working in deeper nested directories

jwilson8767:
I attempted to make the most minimally invasive change that would fix #1707. The lists in the backend now show groups like: b, b/c.

### DIFF
--- a/modules/cms/classes/CmsObject.php
+++ b/modules/cms/classes/CmsObject.php
@@ -190,7 +190,7 @@ class CmsObject implements ArrayAccess
      */
     protected static function getMaxAllowedPathNesting()
     {
-        return 2;
+        return 4;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


danny12mb:
Thanks, I need this for my website. Could somebody accept this PR, please?

axomat:
I would also find this useful. I have hundreds of pages now and managing them would be so much easier with more levels.

pedzed:
> I would also find this useful. I have hundreds of pages now and managing them would be so much easier with more levels.

I hate to say "me too", but in the hope of getting a higher priority, "me too". It's ridiculous for medium to large websites to put everything in one subfolder...

daftspunk:
Cms Objects are now controlled by Halycon. We need to discuss if this is really necessary, when discussed in the initial design it was decided not, since it affects the back-end UI simplicity and possible app performance.

PR #1772, opened 2016-02-11
Workaround for Microsoft SYLK bug

vojtasvoboda:
There is some wierd Microsoft SYLK bug which appears when you have first two characters of CSV uppercased (for example "ID", which is a very common situation in OctoberCMS).

It is reported for:
- Windows - https://support.microsoft.com/en-us/kb/323626
- Mac - https://support.microsoft.com/en-us/kb/215591

**Solutions** are at least two:
a) Insert apostrophe as first character, which works only when inserting apostrophe directly in Excel, not while exporting.
b) Make this two first characters not uppercased (I made it in this PR). It means convert "ID" to "Id" what is still well readable.

**Screenshots** from Win and Mac:

snimek obrazovky 2016-02-11 v 15 11 45

excel-sylk-error

Tested with latest build 317, unit tests passed.

### DIFF
--- a/modules/backend/behaviors/ImportExportController.php
+++ b/modules/backend/behaviors/ImportExportController.php
@@ -566,9 +566,11 @@ class ImportExportController extends ControllerBehavior
          * Add headers
          */
         $headers = [];
+        $index = 0;
         $columns = $widget->getVisibleColumns();
         foreach ($columns as $column) {
-            $headers[] = Lang::get($column->label);
+            $label = Lang::get($column->label);
+            $headers[] = ($index++ == 0) ? ucfirst(mb_strtolower($label)) : $label;
         }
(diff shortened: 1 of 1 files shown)


daftspunk:
This appears to be an [issue with old Microsoft Office versions](https://support.microsoft.com/en-us/kb/323626) and probably should be addressed at the `league/csv` package level to see if it is an issue worth addressing natively.

This could be a slippery slope, I'm not sure if we want to be adding monkey patches for every qwirk of every application that imports a CSV. 

The user can overcome this for the moment by moving the ID when exporting so it does not appear as the first column. The developer can also make their plugin do this by default.

vojtasvoboda:
Yes you are totally right, it should by solved at `league/csv` package or I can handle it directly in my plugin.

PR #1777, opened 2016-02-13
Add a new event on login - backend.user.login

filipac:
I wanted to listen to an event when a users logs-in in the backend and I realised that October does not yet have that. It would be great to support this out of the box to avoid users make changes to core files or listen to hack-ish events to do something after login.


daftspunk:
Please submit to `develop` branch

Issue #1127, opened 2015-05-12
[Bug]? Cannot upload fonts in assets manager


bestxp:
When i try to upload fonts ( font awesome ) i saw 

Ошибка загрузки файла 'fontawesome-webfont.woff2': Разрешены только файлы следующих типов: jpg, jpeg, bmp, png, gif, css, js, woff, svg, ttf, eot, json, md, less, sass, scss

Why i cant upload any file? Why woff2 or any ext of file is restricted?

scottbedard:
I cannot replicate this, I have no problems uploading eof, ttf, svg, woff, or woff2
https://www.dropbox.com/s/8cd4zi60cf8gev0/Screenshot%202015-05-20%2022.50.04.png?dl=0

Perhaps there is something wrong with your permissions?

jBOKA:
I had the same problem. I think woff2 in opposite to woff files are not allowed... it also doesnt show up in the message of allowed types...

daftspunk:
Fixed by #1783

PR #1783, opened 2016-02-15
Update AssetList.php

iBourgeois:
Allow .otf and .woff2 font formats by default. These file types are included in popular packages such as Bootstrap and Font Awesome. 

Should fix issue #1127 where users are unable to upload these file types to the asset manager in the CMS module.


daftspunk:
Thanks

iBourgeois:
No problem.

PR #1784, opened 2016-02-15
Preserve redirect response attributes

shina:
This PR is for complex ajax responses.
Sometimes we want to send a custom response to the client, not only the "result" string.

This make possible the following:

1. Add attributes to the response body.
2. Include custom cookies.
3. Include custom headers.

It works for either Response or RedirectResponse

### DIFF
--- a/modules/cms/classes/Controller.php
+++ b/modules/cms/classes/Controller.php
@@ -650,7 +650,20 @@ class Controller
                     $responseContents['X_OCTOBER_REDIRECT'] = $result->getTargetUrl();
                 }
 
-                return Response::make($responseContents, $this->statusCode);
+                if ($result instanceof \Symfony\Component\HttpFoundation\Response) {
+                    $content = json_decode($result->getContent(), true);
+                    if ($content !== null) {
+                        $responseContents = array_merge($responseContents, $content);
+                    }
+
+                    $result->setStatusCode(200);
+                    $result->setContent(json_encode($responseContents));
+                }
+                else {
+                    $result = Response::make($responseContents, $this->statusCode);
+                }
+
+                return $result;
             }
             catch (ValidationException $ex) {
                 /*


daftspunk:
Can you provide an example of how you are using this? For documentation and greater understanding.

shina:
It's not possible to return a Response from a component ajax response (onXXX method)
It's possible to return string, array or RedirectResponse.
I can't , for example, include a cookie or a response header.

This PR try to solve it.
It checks if the result is a Response and use it instead of remake another Response.

My use case:
The FacebookUser plugin uses an ajax request called onLoginWithFacebook to either register or sign in an user.
On the javascript side, I want to know if the user was registered or signed in.
Today, if I return a Redirect from the onLoginWithFacebook with some informations, these informations will be lost.

daftspunk:
Code example?

shina:
```
public function onSigninOrSignup() {
        $data = post();
        $user = User::where('email', $data['email'])->first();

        if (!$user) {
            $isSignup = true;
            $user = Auth::register($data, true);
        }

        Auth::login($user);

        if ($isSignup) {
            $response = Redirect::to('/when/sign-up');
            $response->setContent(json_encode(['result' => 'sign-up']));
        }
        else {
            $response = Redirect::to('/when/sign-in');
            $response->setContent(json_encode(['result' => 'sign-in']));
        }

        $response->header('Content-Type', 'application/json');

        return $response;
}
```

shina:
```
public function onFoo() {
        $response = \Response::make('foo');
        $response->headers->set('MY_CUSTOM_HEADER', 'bar');
        $response->headers->setCookie(\Cookie::make('MY_CUSTOM_COOKIE', 'foo'));
        
        return $response;
}
```

daftspunk:
Here's the thing I'm not sure about, do we really need to be merging the `$responseContents` with the custom response? I think if you are returning a custom response, it should just return it as is.

It will disable some features, like partial updates, but that is to be expected. What do you think?

shina:
Maybe we can use the custom response on an index called 'result' just like regular Response.

daftspunk:
Thanks for this @shina, it has uncovered a general inconsistency with how responses are handled in October. For example, in a CMS page `onStart()` method, we may return a `Model` class directly, and it will be serialized and sent to thr browser. However, if the same logic is applied to an AJAX handler (eg: `onDoSomething`) the response will be ignored entirely -- this is not consistent.

The issues has been addressed in 7baea87068e496d83d6afd46fccee699779cb2ad. Check the notes for the deeper rationale.

> [Typo]: Importantly to note when a custom ~~redirect~~ response is sent, October's workflow is completely wiped out, so any partial updates, redirects, etc. will not occur. This is a normal and fair expectation.

PR #1788, opened 2016-02-17
Support access parameter in partial onStart and fix css in form relation with repeater

brenodouglas:
Variables passed in parameter partils not accessed in onStart method in the partial.

With merged vars before render partial, stays like this::

```php
function onStart() {
   $this['param'];
} 
```

And repeater widget in form relation rendered color is not correct

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,10 +13,10 @@ matrix:
 
 sudo: false
 
-install: travis_retry composer install --no-interaction --prefer-source
-
-before_script:
-  - git reset --hard HEAD
+install:
   - composer self-update
+  - travis_retry composer install --no-interaction --prefer-source
+
+before_script: git reset --hard HEAD
 
 script: vendor/bin/phpunit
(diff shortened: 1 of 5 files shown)


daftspunk:
We need to be careful here not to allow partials variables to pollute the global scope.

Also you have updated some compiled code here with **modules/system/assets/ui/storm.css**, please update the LESS files.

brenodouglas:
In relation to pollute the global scope with partials variables, in this case I keep the variables on a side and they only exist in the global scope while rendering the partial.

And I commit a proposed feature, which is installing plugins dependencies in the main vendor for better reuse of modules and packages. Using [wikimedia/composer-merge-plugin](https://github.com/wikimedia/composer-merge-plugin) for that.

brenodouglas:
And my build failed because is necessary run: 'composer self-update' in build.

daftspunk:
I see there is no pollution occurring, it looked like code was original intended but never finished (`$vars` was set but not used). This has been merged to the `develop` branch.

The composer feature is interesting, we will trial this and see if people like it.

Thank-you for your contributions.

PR #1789, opened 2016-02-17
Update FrameworkNode.php

sNuuFix:
Fixed framework extra stylesheets for successful validator.w3.org HTML5 validation.


daftspunk:
Can you squash these two commits?

Issue #1778, opened 2016-02-14
CombineAssets -> makeCacheId


christianWilling:
Could you add an event in the `makeCacheId` function in the `System\Classes\CombineAssets` class to modify the cache id. So you can add included files to the File like less or scss import.

christianWilling:
the request comes from christianWilling/Scss#1

daftspunk:
Can you submit a PR for this? Include the example code you want to use as well.

christianWilling:
yes #1794 

´´´
 protected function makeCacheId(array $assets)
     {
         //build cacheId for every file
         $cacheIds = [];

         foreach ($assets as $asset) {
             $cacheId = $asset;

             //calls event so plugins can modify the cacheId
             Event::fire('cms.combineassets.makecacheid', [$cacheId, $asset, $this]);

             $cacheIds[] = $cacheId;
         }

         //put all cacheIds together
         return md5($this->localPath . implode('|', $cacheIds));
    }
´´´

PR #1794, opened 2016-02-19
add makeCacheId Event

christianWilling:
adds event so plugins can modify the cacheId

### DIFF
--- a/modules/system/classes/CombineAssets.php
+++ b/modules/system/classes/CombineAssets.php
@@ -2,6 +2,7 @@
 
 use App;
 use URL;
+use Event;
 use File;
 use Lang;
 use Cache;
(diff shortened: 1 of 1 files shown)


christianWilling:
#1778

daftspunk:
Please include example code of how this is used, i.e the code that contains `Event::listen`

daftspunk:
Still not sure what this is for and how it is used. This feature requires documentation and example code. Please resubmit when you are ready to provide more details.

christianWilling:
Im sry I was on hollday.


And I'm sry again  this pull request is useless I made an mistake makeChaheId is the wrong function.


The problem I won't to fix is if I have a page with with an include less or scss and in the this file I have included some other file if I change something in this file and reload the page it won't change because the caching only effects the file wichbis included in the layout.

I will create an new pull request for this.

I think the problem is that the assets for the assetsCache class don't have filter because in filter you can add an filter cache hash wich includes the included files for example.

PR #1796, opened 2016-02-20
Add a new event on login - backend.user.login

filipac:
I wanted to listen to an event when a users logs-in in the backend and I realised that October does not yet have that. It would be great to support this out of the box to avoid users make changes to core files or listen to hack-ish events to do something after login.

### DIFF
--- a/modules/backend/models/User.php
+++ b/modules/backend/models/User.php
@@ -3,6 +3,7 @@
 use Mail;
 use Backend;
 use October\Rain\Auth\Models\User as UserBase;
+use Event;
 
 /**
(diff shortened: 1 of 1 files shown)


filipac:
OLD PR on wrong branch - https://github.com/octobercms/october/pull/1777

daftspunk:
The event should come after the logic as per the developer guidelines. If the event comes before the logic, it should be called `beforeLogin`.

mitjarobic:
where do i put the listener for this event? Plugins boot method doesn't work.

PR #1799, opened 2016-02-23
Added the ability to inject dependencies into components

mcustiel:
Hi.
I would like to have the ability to inject dependencies in my component classes.
Would be nice to have this also in rainlab.pages plugin for snippets registration from components.


daftspunk:
Great work, thanks for this.

PR #1804, opened 2016-02-24
fix output of function isDisabled in accordance to phpdoc and function naming

haegemon:
more purely accordance to code style


daftspunk:
Thanks

PR #1805, opened 2016-02-25
Add index_onDelete method to ListController behavior.

gpasztor87:


daftspunk:
For the builder plugin I'm guessing?

gpasztor87:
Yes.

daftspunk:
I see approval from @alekseybobkov 

Alrighty then

Bit of scope creep here I think, now lists are capable of deleting records even if the button isn't shown this could potentially be exploited. There is an assumption that forms can do this, the assumption should now be extended to lists. Something to think about and be aware of.

PR #1809, opened 2016-02-26
Add event to extend relation form widget query

MarvinDurot:


daftspunk:
This PR lacks detail. At minimum please provide the code you are using to hook this event.

PR #1811, opened 2016-02-26
Add event to extend relation form widget query

MarvinDurot:
### Description
ListController and FormController have both some extend events which are useful to filter elements.
This pull request allow you to extend the query used by the Relation widget form to populate dropdown or checkbox list.

### Example
You have a model with a belongsToMany relation, and you want his form controller to display it thanks to a field of type 'relation'. If you want to restrict results of the checkbox list generated by your relation field, you can write something like that :
````
// Extend Rainlab User controller form
        Event::listen('backend.relation.extendQuery', function ($widget, $query) {
            if (!$widget->getController() instanceof \RainLab\User\Controllers\Users) return;
            if (!$widget->model instanceof \RainLab\User\Models\User) return;
            
            // No restriction for super users
            if (BackendAuth::getUser()->isSuperUser()) return;
            
            // Filter available groups using backend user permissions
            $query->whereIn('id', BackendAuth::getUser()->frontend_groups);
        });
````
In this code, I extended Backend User model with a JSON attribute which contains allowed frontend groups.

*Sorry for duplicated pull request in the master branch...*

### DIFF
--- a/modules/backend/formwidgets/Relation.php
+++ b/modules/backend/formwidgets/Relation.php
@@ -129,7 +129,7 @@ class Relation extends FormWidgetBase
             /**
              * Extensibility
              */
-            Event::fire('backend.relation.extendQuery', [$this, $query]);
+            Event::fire('backend.relation.extendQuery', [$this, $attribute, $query]);
 
             // The "sqlSelect" config takes precedence over "nameFrom".
             // A virtual column called "selection" will contain the result.
--- a/modules/backend/formwidgets/Relation.php
+++ b/modules/backend/formwidgets/Relation.php
@@ -2,6 +2,7 @@
 
 use Db;
 use Lang;
+use Event;
 use Backend\Classes\FormWidgetBase;
 use ApplicationException;
 use SystemException;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by daftspunk** [comment]
Please try this instead to see if it meets your requirements

    /**
     * Provides an opportunity to manipulate the manage widget.
     * @param Backend\Classes\WidgetBase $widget
     * @param string $field
     */
    public function relationExtendManageWidget($widget, $field)
    {
        $widget->bindEvent('list.extendQuery', function($query) {
            // No restriction for super users
            if (BackendAuth::getUser()->isSuperUser()) return;

            // Filter available groups using backend user permissions
            $query->whereIn('id', BackendAuth::getUser()->frontend_groups);
        });
    }

This will extend the list shown when you click "Add", to extend the view list override the `relationExtendViewWidget` method instead.

**Comment by MarvinDurot** [comment]
Hi, thank you for your review !

Your solution is about the relation manager controller, not about the relation form widget (`modules/backend/formwidgets/Relation.php`).

**Comment by daftspunk** [comment]
Thanks for clarifying. This will be worked on as part of #1524

**Comment by daftspunk** [comment]
Actually, #1524 mentions it is possible to do already, use a scope in the relation definition to apply your filter. Eg

```
class Post extends Model
{
    public $belongsToMany = [
        'categories' => [
            'Acme\Blog\Models\Category',
            'scope' => 'isActive'
        ]
    ];
}

class Category extends Model
{
    public function scopeIsActive($query)
    {
        return $query->where('is_active', true)->orderBy('name', 'desc');
    }
}
```

Issue #1801, opened 2016-02-24
Create a file definition helper class


daftspunk:
We should introduce a class to the Filesystem package that simply provides file extensions used throughout the app. We have these in various places so far:

- Ignore file names ['.svn', '.git', '.DS_Store']
- 'image' => ['gif', 'png', 'jpg', 'jpeg', 'bmp'],
- 'video' => ['mp4', 'avi', 'mov', 'mpg', 'mpeg', 'mkv', 'webm'],
- 'audio' => ['mp3', 'wav', 'wma', 'm4a', 'ogg']
- safe to upload (white list)
- unsafe to upload (black list)
- CMS Asset list: 'jpg', 'jpeg', 'bmp', 'png', 'gif', 'css', 'js', 'woff', 'svg', 'ttf', 'eot', 'json', 'md', 'less', 'sass', 'scss'

The class should allow config to override definitions. Refit existing code to use this class in:

- FileUpload widget
- Media Manager classes
- Asset List widget

daftspunk:
Black list extensions pulled from #1812: 
- 'asp','avfp','aspx','cshtml','cfm','go','gsp','php','hs','jsp','ssjs','js','lasso','lp','op','lua','p','cgi','ipl','pl','php','php3','php4','phtml','py','rhtml','rb','rbw','smx','tcl','dna','tpl','r','w','wig'

PR #1812, opened 2016-02-26
Verify file extensions against server-side scripting file upload

kamesatou:
Additional file extension verification against server-side scripting upload.
A list of default disallowed file types is from https://en.wikipedia.org/wiki/Server-side_scripting, Line 51-92.
            Line 1009-1022:
            // Use the simple extension validation.
            $disAllowedFileTypes = Config::get('cms.disAllowedFileTypes');
            if (!$disAllowedFileTypes) {
                $disAllowedFileTypes = $this->disAllowedFileTypes;
            }
            /*
             * Convert uppcare case file extensions to lower case
             */
            $extension = strtolower($uploadedFile->getClientOriginalExtension());
			// Verify file extensions against server-side scripting file upload
            if (in_array($extension, $disAllowedFileTypes)) {
				throw new ApplicationException('Error uploading file');
            }

### DIFF
--- a/modules/cms/widgets/MediaManager.php
+++ b/modules/cms/widgets/MediaManager.php
@@ -48,6 +48,49 @@ class MediaManager extends WidgetBase
      */
     public $cropAndInsertButton = false;
 
+	/**
+     * @var array A list of default disallowed file types.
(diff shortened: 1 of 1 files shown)


daftspunk:
Duplicate of #1801

Thanks for this, will keep for reference.

PR #1825, opened 2016-03-03
Added missing backend translations (German)

manogi:


manogi:
As this is my first ever pull request - am I doing this right? Should I be doing something differently?

PR #1849, opened 2016-03-16
Expanded select-no-search to consider container class. Fixes #1817 

### LINKED ISSUE #1817: feature request: allow 'select-no-search' to be defined in yaml file
Consider adding css rule to allow the 'select-no-search' class to be defined in yaml field definition files, so that the 'select' tag can inherit the definition from the 'drop-down-field' class.

fields.yaml

```
type: dropdown
cssClass: select-no-search
```

css definition

```
.drop-down-field.select-no-search > select {
    // hide search ui
}
```

jwilson8767:
select-no-search may now be used inside yaml form configs like so:
`````````
form:
  fields:
    shirt_size:
      label: Shirt Size
      type: dropdown
      cssClass: select-no-search
      options: ...
```````

This should be a documented option [here](https://octobercms.com/docs/backend/forms#field-dropdown), but I will leave the verbiage to you, @daftspunk 

Cheers

### DIFF
--- a/modules/system/assets/ui/js/select.js
+++ b/modules/system/assets/ui/js/select.js
@@ -58,7 +58,7 @@
                 }
             })
 
-            if ($element.hasClass('select-no-search')) {
+            if ($element.hasClass('select-no-search') || $element.parent().hasClass('select-no-search')) {
                 extraOptions.minimumResultsForSearch = Infinity
             }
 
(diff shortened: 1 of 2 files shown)


daftspunk:
Done in https://github.com/octobercms/october/commit/0aef151d9a73555209d43ede12be4ad059b0deae, with docs in https://github.com/octobercms/docs/commit/f0116bc66e3a053e907b803af9d690ed3e3ada35.

PR #1853, opened 2016-03-19
Make available to extends Lists records

leocavalcante:
Makes it available to extends `Collection` enabling sorting, mapping, filtering etc.

### DIFF
--- a/modules/backend/behaviors/ListController.php
+++ b/modules/backend/behaviors/ListController.php
@@ -157,6 +157,10 @@ class ListController extends ControllerBehavior
             $this->controller->listExtendQuery($query, $definition);
         });
 
+        $widget->bindEvent('list.extendRecords', function ($records) use ($definition) {
+            $this->controller->listExtendRecords($records, $definition);
+        });
+
         $widget->bindEvent('list.injectRowClass', function ($record) use ($definition) {
             return $this->controller->listInjectRowClass($record, $definition);
         });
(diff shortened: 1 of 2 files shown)


daftspunk:
Missing usage example for documentation.

leocavalcante:
Sorry.
With this change you should be able to access `$records` before they are set to [`Lists`](../blob/master/modules/backend/widgets/Lists.php) `$vars` enabling mutations over the [`Collection`](https://laravel.com/docs/5.1/collections) like [`map`](https://laravel.com/docs/5.1/collections#method-map), [`sort`](https://laravel.com/docs/5.1/collections#method-sort), [`reject`](https://laravel.com/docs/5.1/collections#method-reject) etc.

Then you can set/override column properties by computed ones:
```php
public function listExtendRecords($records)
{
    return $records->map(function ($record) {
        $record->bar = $record->computedProp(2);
    });a
}
```
Also, the use case that leads me to this PR, was sorting by a computed value or something else them an `ORDER BY` over a table column.
```php
public function listExtendRecords($records)
{
    return $records->sort(function ($a, $b) {
        return $a->computedVal() > $b->computedVal();
    });
}
```

LukeTowers:
Done and documented. Modified slightly in https://github.com/octobercms/october/commit/9f3c2b8bf6340af6f843212db1de32eef25cc975 to use `$this->fireSystemEvent()` instead of calling both local and global events separately

daftspunk:
Documentation found here: https://github.com/octobercms/docs/commit/d1b6049cf8b0c6c8747bffeb16d2cd6d98251f11

marryj:
There is an issue on the sorting functionality.
The following example does not work:
`public function listExtendRecords($records)
{
    return $records->sort(function ($a, $b) {
        return $a->computedVal() > $b->computedVal();
    });
}`

A solution is returning the record in Backend\Behaviors\ListController.php
`
public function makeList($definition = null) {
...
        $widget->bindEvent('list.extendRecords', function ($records) use ($definition) {
            **return** $this->controller->listExtendRecords($records, $definition);
        });
...
}
`

daftspunk:
@marryj you should probably report this as a new issue so it isn't missed

LukeTowers:
@daftspunk I feel like it was

LukeTowers:
@daftspunk @marryj https://github.com/octobercms/october/issues/2965

PR #1855, opened 2016-03-20
French translations for import/export errors strings

gabsource:
Externalized english to localization file and added translation for french.


daftspunk:
Thanks!

PR #1856, opened 2016-03-20
Enable setting source file options on backend import page (as it exists for export)

gabsource:
Mainly copy/pasted the existing logic from export feature into the import one.

Added a new option to select the source file encoding and a patch to prevent a Symfony exception when the response (containing some CSV lines) can not be encoded to Json (when using non utf-8 chars in CSV source file).

image


modules/backend/behaviors/ImportExportController.php
```diff
@@ -174,15 +175,22 @@ public function onImport()
...
+                'sessionKey' => $this->importUploadFormWidget->getSessionKey()
+            ];
 
             if ($optionData = post('ImportOptions')) {
                 $model->fill($optionData);
             }
 
-            $model->import($matches, [
-                'sessionKey' => $this->importUploadFormWidget->getSessionKey(),
-                'firstRowTitles' => post('first_row_titles', false)
```
daftspunk: Oops, this was lost!

daftspunk:
Wow @gabsource , thanks for this!

2a60e3455806ccdd4abd812856b4bd03c9301d26 - It has been merged to the `stable` branch and will appear in the next major release.
a035acdc2c5dafc83c6bfb9f4d6395642f12cbae - There is some minor code restructuring performed, please test these changes too.

PR #1857, opened 2016-03-20
Allow Pikaday translation for backend datepickers

gabsource:
Extracted from PR #1651 

Using the built-in $.oc.lang module, the patch injects the i18n object into the datepicker init call.

Assets needs to be recompiled `php artisan october:util compile assets`


daftspunk:
Merged in f836b43bb3775710e47d114ece498d93e3b1fbf4

Thanks

Issue #1750, opened 2016-02-04
.htaccess not mirrored when using october:mirror


philipptempel:
**Bug summary:** ```october:mirror``` does not mirror apache2's ```.htaccess``` file (refs #122)
**Request location:** n/a
**Reproduce steps:**
```
$ composer create-project october/october myoctober dev-master
$ cd myoctober
$ php artisan october:mirror public/
```
Then access October using your favorite browser at http://localhost/myoctober/public
**Expected result:** October shows up
**Actual result:** 500 Internal server error issued by apache

**Proposed fix:** Make ```october:mirror``` also copy the .htaccess file (think this shouldn't be an issue in case lighttpd or nginx are used.

PR #1865, opened 2016-03-22
Fix #1851 no assets load for subfolder/public mirror install

### LINKED ISSUE #1851: Install in subfolder and public mirror folder, no assets loaded
This issue is similar to many of those that have been posted before (#100, #257, #59, #829), however in this case assets load for both frontend and backend when **_not_** using a public folder (#1331) but no assets are loaded when using a public folder (`october:mirror`).

Apache conf: `Alias /october /var/www/october/public`
```
    <Directory /var/www/october/public>
         Options Indexes FollowSymLinks
         AllowOverride All
         Order allow,deny
         allow from all
     </Directory>
```
.htaccess: `RewriteBase /october/`
```
# a2enmod rewrite
Module rewrite already enabled
```
It seems the only thing not getting loaded is the file returned by `getCombinedUrl` function in `CombineAssets.php` as indicated by the error log:
```
... [error] ... File does not exist: /var/www/october/public/combine
```
and it only occurs when using a public folder.

scosist:
Should this be on master or another branch like stable?

### DIFF
--- a/modules/system/console/OctoberMirror.php
+++ b/modules/system/console/OctoberMirror.php
@@ -30,6 +30,7 @@ class OctoberMirror extends Command
         'favicon.ico',
         'robots.txt',
         'sitemap.xml',
+        '.htaccess',
     ];
 
     protected $directories = [


daftspunk:
Usually `develop` branch, but this is already in `stable` https://github.com/octobercms/october/blob/stable/modules/system/console/OctoberMirror.php#L29

Duplicate #1750

Issue #201, opened 2014-05-22
[Issue] Cannot use object of type Cms\Classes\Page as array


DanielHitchen:
Why do I get this `Exception` when using:

```
function onBeforePageStart()
{
    $this->page['var'] = 'value';
}
```

in a layout?

Issue #1847, opened 2016-03-16
Compression for transferring combine


vebam:
Hello,
Why /combine/* stylesheet is gziped,
/combine/* script don't.

![clipboard01](https://cloud.githubusercontent.com/assets/6676468/13806086/8fbd627e-eb59-11e5-9da0-89019b44c4e5.jpg)

gabsource:
Apache `/etc/apache2/mods-available/deflate.conf` may look like this (should be the defaults on Ubuntu 14.04) : 
```
    AddOutputFilterByType DEFLATE text/css
    AddOutputFilterByType DEFLATE application/x-javascript application/javascript application/ecmascript
```

So it does not recognize `text/javascript` Content-Type header as a candidate to compression.

Adding `AddOutputFilterByType DEFLATE text/javascript` in `.htaccess` should trigger compression for combined javascript.

`\System\Classes\CombineAssets@getContents` sets `$mime = 'text/javascript';` at line 178.

vebam:
Thanks, is working.

I will add:
AddOutputFilterByType DEFLATE text/html
AddOutputFilterByType DEFLATE text/xml
AddOutputFilterByType DEFLATE text/css

daftspunk:
@gabsource Should this be changed to `application/javascript`?

gabsource:
`text/javascript` is deprecated per RFC 4329 in favor of `application/javascript`
https://www.rfc-editor.org/rfc/rfc4329.txt

```
 Use of the "text" top-level type for this kind of content is known to
   be problematic.  This document thus defines text/javascript and text/
   ecmascript but marks them as "obsolete".  Use of experimental and
   unregistered media types, as listed in part above, is discouraged.
 The media types,

      * application/javascript
      * application/ecmascript

   which are also defined in this document, are intended for common use
   and should be used instead.
```

PR #1885, opened 2016-03-30
Added system wide gzip compression

mahony0:
Also helper method proposed to “octobercms/library” as PR [#201](https://github.com/octobercms/library/pull/201)


daftspunk:
You are welcome to include this logic in your index file if you like (the index file is never updated).

I thought this issue was addressed in 2a3a63857c1f6dd403d17908bea6faa0bf0bbb4d

See #1847

PR #1890, opened 2016-03-31
Add newline conversion to code data section too

Abdillah:
When saving through Editor, OctoberCMS only convert newline of markup section once config 'cms.convertLineEndings' enabled.
In the code too, I've only seen [the conversion](https://github.com/octobercms/october/blob/master/modules/cms/controllers/Index.php#L158) of newline for markup.
Then, I propose the conversion of code section data too.

### DIFF
--- a/modules/cms/controllers/Index.php
+++ b/modules/cms/controllers/Index.php
@@ -158,6 +158,10 @@ class Index extends Controller
             $templateData['markup'] = $this->convertLineEndings($templateData['markup']);
         }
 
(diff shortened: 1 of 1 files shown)


daftspunk:
Merged in 6bb5926e197ac5a7f34a292b49bbf4b6668a4842

Thanks

Nyorai:
This is still a problem. The settings front matter and the open/close PHP tags in the code section still have windows line endings, which leads to files with mixed line endings. Surely all line endings should be converted in each file section?

LukeTowers:
Open a new issue for that please @Nyorai

PR #1892, opened 2016-03-31
add the Unzip feature to assets manager

matheusrrocha:
Sometimes we have a huge folder to import to assets manager. And it's a pain to import one by one of these files. This feature enables an Unzip option button, which allow us to submit a zip file and unzip then into the current asset folder.

### DIFF
--- a/modules/cms/lang/en/lang.php
+++ b/modules/cms/lang/en/lang.php
@@ -208,7 +208,7 @@ return [
         'path' => 'Path',
         'unzip' => 'Unzip',
         'is_a_directory' => 'Only files are allowed. You\'ve selected a directory.',
-        'error_open_zip' => 'Error open the zip file'
+        'error_open_zip' => 'Failed to open the zip file'
     ],
     'component' => [
         'menu_label' => 'Components',


gergo85:
+1

modules/cms/lang/en/lang.php
```diff
@@ -205,7 +205,10 @@
         'error_moving_file' => 'Error moving file :file',
         'error_moving_directory' => 'Error moving directory :dir',
         'error_deleting_directory' => 'Error deleting the original directory :dir',
-        'path' => 'Path'
+        'path' => 'Path',
+        'unzip' => 'Unzip',
+        'is_a_directory' => 'Only files are allowed. You\'ve selected a directory.',
+        'error_open_zip' => 'Error open the zip file'
```
jurchiks: Failed opening*

matheusrrocha: thanks @jurchiks. Fixed.

daftspunk:
There are a couple of major flaws with this:

* Security risk: zipped file may act as a malicious payload
* This does not appear to be compatible with [CDN features](https://octobercms.com/docs/cms/mediamanager)

For these reasons this feature will not be accepted. Please submit feature ideas as issues for comment before spending time on code for pull requests. See [the contribution guidelines](https://github.com/octobercms/october/blob/master/CONTRIBUTING.md#reporting-a-bug-with-octobercms) for more details.

PR #1896, opened 2016-04-02
Added deflate compression if server supports

mahony0:
This is an alternative for PR [#1885](https://github.com/octobercms/october/pull/1885)


daftspunk:
See #1885 and #1847
This should be configured by default on Apache and should not be necessary now that we switch to application/javascript

PR #1898, opened 2016-04-02
french lang correction

Alexandre-Gadiou:
french words to say "Upload a file" are "Déposer un fichier"


daftspunk:
Fixed by #1940

PR #1903, opened 2016-04-03
Using Assetic less filter

gpasztor87:
Using kriswallsmith/assetic less filter you don't need to use the oyejorge/less.php library.


daftspunk:
Not sure exactly what this is supposed to fix. We used this filter originally and it caused headaches.

gpasztor87:
Ok, I just find an another less compiler in the [assetic](https://github.com/octobercms/library/blob/master/composer.json#L21) package. And I wanted to add SCSS compiler with the same [PR](https://github.com/gpasztor87/library/commit/15bb257f93fa983e75c2db8b141abd6c2987baee)

PR #1907, opened 2016-04-05
Customize backend navbar style

alvaro-canepa:
### Inline
Inline

### Tiles
Tiles

### Collapse
Collapse


alvaro-canepa:
Is Done in build 325

PR #1914, opened 2016-04-07
Improve code of Backend Behavior FormController

adrenth:
- Added missing PhpDoc
- Improved readability of PhpDoc
- Small type checking changes
- Add missing type hinting in some method

Please let me know if this is OK, I'd like to improve more of these files of the backend modules since I use them a lot in my projects.


daftspunk:
I don't think this is a good idea sorry. This doesn't really add much value and causes headaches for outstanding PRs and spams the commit history. We can improve the code as we go, although I don't see anything wrong with the code in its current state. The API docs will need to be addressed when we build a docs generator. Just bear with us until then.

adrenth:
No problem, I understand it can conflict outstanding PR's. Thanks for your feedback. Code documentation is becoming extremely important since current IDE's such as PhpStorm benefits from it. Let's keep that in mind ;-)

PR #1919, opened 2016-04-09
Datepicker: allow specify display format

jfoliveira:
Adds the ability to specify a display format to be used with the [datepicker](https://octobercms.com/docs/backend/forms#widget-datepicker).
Reads the `format` property from field settings and apply it to the `datepicker` instance.

Example usage:
```yaml
created_at:
    label: Date
    type: datepicker
    format: DD/MM/YYYY
```
or
```yaml
created_at:
    label: Date
    type: datepicker
    format: L
```

See [Moment.js docs](http://momentjs.com/docs/#/displaying/format/) for a detailed description of supported display formats.

### DIFF
--- a/modules/backend/formwidgets/DatePicker.php
+++ b/modules/backend/formwidgets/DatePicker.php
@@ -20,6 +20,11 @@ class DatePicker extends FormWidgetBase
     //
 
     /**
+     * @var string Display format.
+     */
+    public $format = 'YYYY-MM-DD';
+
+    /**
      * @var bool Display mode: datetime, date, time.
      */
     public $mode = 'datetime';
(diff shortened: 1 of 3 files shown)


daftspunk:
Good job

bhargavy:
When will be merged into the Master?

daftspunk:
They are in `master` already.

bhargavy:
Think it got overwritten at 'Complete rebuild of datepicker form widget'

For example at Line 128 below code is missing Is missing
`$this->vars['format'] = $this->format;`

daftspunk:
I see. That's right, the format is determined by the locale selected in the user's Backend preferences. If this isn't suitable, please create another issue with your expectations and provide some prototype code on how you want it to function.

PR #1924, opened 2016-04-11
Add url input to fileupload

FrancescoPassanti:
Add an url input to fileupload formwidgets is a very useful tool for create interactive gallery, specially in the homepage. I have a banner then I can click on a Image and browse in other connected page.


daftspunk:
This doesn't make a lot of sense to be adding to the core. It could be solved in other ways using a plugin to meet the requirements.

PR #1926, opened 2016-04-11
Config option to specify combined assets path

ShaaD:
Add the combineAssetsPath option in cms.php config that allow to change domain name and path for combained assets.

This option is necessary to use combined assets with proxy CDN.

_Probably better to reuse themesPath variable, but with this approach we will not have backwards compatibility issues._


daftspunk:
I don't understand exactly what is the problem? Why does the combine URL need to be changed?

ShaaD:
It is common practice for highload websites to use separate domain for frontend assets, it is allow parallel assets loading and simplify CDN integration. Github itself use https://assets-cdn.github.com domain for example.

So for assets included in twig templates with **|theme** filter we can just set **themesPath** variable and all works just fine.

`{{  'assets/css/app.css' |theme }}`

But, not when combiner involved

`{{  ['assets/css/normalize.css', 'assets/css/app.css'] |theme }}`

Commit fix this problem.

daftspunk:
Are you saying the combiner will work with an origin pull CDN?

ShaaD:
Yes, it is the main purpose of the commit.

daftspunk:
Thanks @ShaaD, I am still in the dark about one thing. How are you using this? We need solid and real examples to confirm the code operates as expected.

Can you provide the URLs you are using? Feel free to change them if they are sensitive.

- What does your config look like?  `'combineAssetsPath' => 'XXX',` 
- What does the output URL look like? `href="XXX"`
- What October URL does the reverse CDN server use for the first capture? `http://mysite/...`

And finally, do all reverse origin pull CDNs operate in the same way? For example is it possible that the output URL might be:

```
http://mycdn.com/abcxyz1234
```

Yet the first capture URL might be

```
http://mycdn.com/combine/abcxyz1234
```

If this is true, we need to consider the output URL and capture URL as separate entities. This would mean give them their own config entries. Thanks for helping with this.

ShaaD:
Ok, so there is an example: https://bless.101xp.com.

As u can see, we are using combine for both css and js files and load resulted files from https://sc-bless.101xp.com.

In cms.php its looks like:

`'combineAssetsPath' => 'https://sc-bless.101xp.com/combine',`

And output url is:
`https://sc-bless.101xp.com/combine/690e11204859529bd282620b6837bbeb-1470066797`

>If this is true, we need to consider the output URL and capture URL as separate entities. This would mean give them their own config entries
In config option we define only one value, and if it is full URL with domain name we are just have it as is on templates and the file itself will be available on our server with same URI.

daftspunk:
Thanks for the reply.

Ok so the CDN URL: https://sc-bless.101xp.com/combine/...
Will pull origin content from URL: https://bless.101xp.com/combine/...

Is that correct? If we consider some CDNs may have different structures, it could be possible to use a configuration like this (in future)

```
'combineAssetsPath' => [
    'capture' => '/combine',
    'output' => 'https://sc-bless.101xp.com/combine',
]
```

Where the `capture` key defines the origin content route, a bit like the `cms.backendUri` config option.

LukeTowers:
@ShaaD Are you able to provide input on @daftspunk's latest question? If not, let us know and we'll move on.

LukeTowers:
Closing as it has been over a month since any activity on this occurred.

PR #1932, opened 2016-04-13
Add SES configuration for mailer

ariponce:
Laravel ships with a SES transport class, even on the 5.0 branch that's currently used by October, but currently there's no option for it when configuring the mail settings via the backend.


daftspunk:
Nice! :+1:

PR #1943, opened 2016-04-18
Bug: Fix for components that support autowiring

mcustiel:
Hi. 
Some weeks ago I created a pull request with changes to support autowiring in components: https://github.com/octobercms/october/pull/1799
It was merged to master recently, but still does not work because there is another place where components are being created: Cms\Widgets\ComponentsList
I created a fix for this, so the autowiring works for the components also in /backend/cms section.

### DIFF
--- a/modules/cms/widgets/ComponentList.php
+++ b/modules/cms/widgets/ComponentList.php
@@ -1,5 +1,6 @@
 <?php namespace Cms\Widgets;
 
+use App;
 use Str;
 use Lang;
 use Input;
@@ -110,7 +111,7 @@ class ComponentList extends WidgetBase
             foreach ($components as $componentInfo) {
                 $className = $componentInfo->className;
                 $alias = $componentInfo->alias;
-                $component = new $className();
+                $component = App::make($className);
 
                 if ($component->isHidden) {
                     continue;


daftspunk:
Why the need for NullCodeBase? Can't we simply use this:

```
$component = App::make($className);
```

mcustiel:
Hi!
The problem is that the change is to support autowiring in the components. So in the components where I use autowiring the default values can not be specified in the constructor like this:
```php
public function __construct(CodeBase $cmsObject = null, $properties = [], MyDependency $test = null)
```
 or `App::make($classname)` does not load them (in this case, MyDependency).
So my constructor should look more like this:
```php
public function __construct(CodeBase $cmsObject, $properties, MyDependency $test)
```
In this case, App::make autowires myDependency to the constructor.
but I need the NullCodeBase to pass it as argument instead of the default null.

Probably you could have a better idea, since you know how laravel works. I am not experienced with that framework.
Please let me know.

mcustiel:
I made some test with this and it actually works. So discard my last comment. This laravel's IoC container is awesome! Will do the suggested change.

daftspunk:
Oh yeah that's more like it!

mcustiel:
Do you think it would make sense to change every place where `new $className` is used to App::make($className)? I would like to autowire classes in form widgets too.
Maybe there is some design decision to use `new $className`?

PR #1958, opened 2016-04-25
Export with utf-8 characters

ITCorporate:
for cyrillic it will:
screen shot 04-25-16 at 06 42 pm

but it must be in utf-8:
screen shot 04-25-16 at 06 42 pm 001


daftspunk:
OK, let's try it

PR #1961, opened 2016-04-26
Added missing backend/  translations for Latvian language

canthis:
Added missing strings, translated whole Import/Export section, fixed found typos.


daftspunk:
Thanks

PR #1963, opened 2016-04-27
Added web.config for IIS, includes URL rewrite

asimr:
Added a web.config file for IIS Server, the /backend (And most of the other directories) won't work and will throw 404 on IIS, since it ignores .htaccess.


daftspunk:
This should be added to the documentation

https://github.com/octobercms/docs/blob/master/setup-configuration.md#iis-configuration

PR #1967, opened 2016-04-27
Fixed parse error which occurs after upgrading to edge version while …

daslicht:
…using php7

see: https://i.imgur.com/fbAYYcN.png


daftspunk:
Already fixed in `develop`

daslicht:
yeah!

PR #2001, opened 2016-05-11
Fix var name typo in FilterWidget javascript

gabsource:
`name` should be `key`


daftspunk:
Was thinking the other way around 75003fb87793fc06ed8d1f8bae857e08c5d879cf

PR #2003, opened 2016-05-12
Allow ajax framework to handle File download response 

jofrysutanto:
This allow the convenience of returning `response()->download('file.txt')` in Backend controller and Component during AJAX request, and have the client initiates file download.

Credit to this thread on Javascript part: http://stackoverflow.com/questions/16086162/handle-file-download-from-ajax-post

The checking can be pulled up a bit higher in the `execAjaxHandlers` since no partials will be pushed back to client as part of response.

Usage example:

```
// CMS page
<button type="button"
    class="btn btn-primary"
    data-request="demoTodo::onExportList">
    Export Todo List
</button>

// Todo Component
public function onExportList()
{
    return response()->download('path_to_todo_file.csv');
}
```

### DIFF
--- a/modules/backend/classes/Controller.php
+++ b/modules/backend/classes/Controller.php
@@ -24,6 +24,7 @@ use October\Rain\Exception\ApplicationException;
 use October\Rain\Extension\Extendable;
 use Illuminate\Database\Eloquent\MassAssignmentException;
 use Illuminate\Http\RedirectResponse;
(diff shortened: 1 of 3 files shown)


daftspunk:
Thinking about this. The solution seems hacky and support is limited... at the moment we simply return a redirect to a regular request cycle that downloads the file and the UX is the same.

We should get some peer review on the approach.

jofrysutanto:
Agree with everything you just said. 👍

PR #2008, opened 2016-05-14
Improved list date / daterange filter implementation

gabsource:
Follow up of #1651 

Let me know if it needs more improvements.

#### Continuity requirement

Popup buttons are streamlined : 
image

#### Timezone conversion

The displayed date on the PHP side are now displayed with Backend::dateTime format alias dateMin
The client side code converts the datepickers value to app timezone before submitting.

User selection is converted from 00:00:00 / 23:59:59 in the backend timezone (user timezone) to the application timezone (database timezone).

For example, if the app timezone is UTC and the backend timezone UTC+4 then selecting _after_ date `2016-05-16` will send to the server `2016-05-15 20:00:00` and _before_ date `2016-05-16` will send to the server `2016-05-16 19:59:59`.

#### Nicer syntax

New placeholders are populated in order to allow simpler conditions writing : 
- for _date_ and _daterange_ types : 
  * `after` holds the full datetime _Y-m-d HH:mm:ss_  (00:00:00 in the backend timezone converted to app timezone)
  * `before` holds the full datetime _Y-m-d HH:mm:ss_  (23:59:59 in the backend timezone converted to app timezone)
- for _date_ type only : 
  * `filtered` holds the short date _Y-m-d_  (keeping the same name as other filter widgets)
- for _daterange_ type only : 
  * `afterDate` holds the short date _Y-m-d_
  * `beforeDate` holds the short date _Y-m-d_

Conditions for timestamp type columns can be written like this : 

```
    # Filter type
    type: date

    # Scope Conditions (match against a timestamp column)
```

```
    # Filter type
    type: date

    # Scope Conditions (match against a timestamp column)
```

Conditions for date type columns can be written like this : 

```
    # Filter type
    type: date

    # Scope Conditions (match against a date column)
    conditions: birthday = ':filtered'
```

```
    # Filter type
    type: daterange

    # Scope Conditions (match against a date column)
    conditions: birthday >= ':afterDate' AND birthday <= ':beforeDate'
```

#### Scope queries

Date type scoped query is passed a `Carbon` instance that correspond to _Y-m-d 00:00:00_ in the backend timezone but converted to the app timezone. See above in the timezone conversion.

        scope: FilterDate

```

    public function scopeFilterDate($query, $date)
    {
        return $query
            ->where('published_at', '>=', $date)
            ->where('published_at', '<=', $date->copy()->addDay()->addMinutes(-1));
    }

```

Daterange type scoped query is passed two `Carbon` instances that correspond to _Y-m-d 00:00:00_ and _Y-m-d 23:59:59_ in the backend timezone but converted to the app timezone. See above in the timezone conversion.

    scope: FilterDateRange

```

    public function scopeFilterDateRange($query, $after, $before)
    {
        return $query
            ->where('published_at', '>=', $after)
            ->where('published_at', '<=', $before);
    }

```

#### Bugfixes

- Submitting null dates is now working (two name refactoring were missing from date(range)picker to date(range))

### DIFF
--- a/modules/backend/widgets/Filter.php
+++ b/modules/backend/widgets/Filter.php
@@ -499,21 +499,56 @@ class Filter extends WidgetBase
         }
 
         switch ($scope->type) {
-            case 'datepicker':
-                if ($scope->value instanceof Carbon && $scopeConditions = $scope->conditions) {
-                    $query->whereRaw(strtr($scopeConditions, [':filtered' => $scope->value->format('Y-m-d')]));
+            case 'date':
+                if ($scope->value instanceof Carbon) {
+                    $value = $scope->value->setTime(0, 0, 0);
+
+                    /*
+                     * Condition
+                     */
+                    if($scopeConditions = $scope->conditions) {
+                        $query->whereRaw(DbDongle::parse(strtr($scopeConditions, [
+                            ':filtered' => $value->format('Y-m-d'),
+                            ':after'    => $value->format('Y-m-d H:i:s'),
+                            ':before'   => $value->copy()->setTime(23, 59, 59)->format('Y-m-d H:i:s')
+                        ])));
+                    }
+
+                    /*
+                     * Scope
+                     */
+                    if ($scopeMethod = $scope->scope) {
+                        $query->$scopeMethod($value);
+                    }
                 }
 
                 break;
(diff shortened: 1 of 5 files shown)


daftspunk:
Solid. Thanks for all your hard work on this.

PR #2009, opened 2016-05-14
Fix List Date Filter error thrown when `$date` is not set

gabsource:
Sorry, a test was missing in the PR branch #2008 to prevent an error when no date was sent in the partial parameters.


daftspunk:
Thanks, I notice some other strangeness. With the range mode, when I want to supply only one date:

screenshot 2016-05-15 05 05 42

PR #2010, opened 2016-05-14
defaultTab option does not work

amel13:
defaultTab option does not work


daftspunk:
The `defaultTab` setting gives a tab for fields that do not specify a `tab`. It doesn't necessarily specify which tab should be opened by default.

PR #2011, opened 2016-05-14
Allow specifying half bound daterange filter values

gabsource:
Follow up of #2009 

Submitting a _daterange_ filter with only on value is now allowed to support _after_ and _before_ queries in addition to the previous _between_.
- when the after date is empty the "zero" value "0000-00-00 00:00:00" is used.
- when the before date is empty a far away value "2999-12-31 23:59:59" is used.

| after | before |
|----|----|
| image | image |

> Assets must be recompiled `php artisan october:util compile assets`

### DIFF
--- a/modules/backend/widgets/Filter.php
+++ b/modules/backend/widgets/Filter.php
@@ -116,10 +116,26 @@ class Filter extends WidgetBase
                     $scope->value[0] && $scope->value[0] instanceof Carbon &&
                     $scope->value[1] && $scope->value[1] instanceof Carbon
                 ) {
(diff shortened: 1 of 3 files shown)


daftspunk:
Looks we made our own Y3K bug. Good thing we won't be around to see it!

Good job :+1:

gabsource:
Thanks ! I think MySQL will face it's own Y2K38 bug, as for now TIMESTAMP has a range of '1970-01-01 00:00:01' UTC to '2038-01-19 03:14:07', I hope we will be around at that time though :-)

Issue #2026, opened 2016-05-18
Input Converter: Preset URL for German


codebeauty:
If you create a new page the input converter (title to url) is a bit useless for Germans. It needs to converts 'ä' to 'ae' instead of just 'a'.

Example for a simple German map:
```
    var GERMAN_MAP = {
        'Ä': 'AE', 'Ö': 'OE', 'Ü': 'UE',
        'ä': 'ae', 'ö': 'oe', 'ü': 'ue',
        'ß': 'ss', 'ö': 'oe', 'ü': 'ue',
    }
```
We need something like a German map inside **input.preset.js**. Or can we update the Latin map? Any ideas?

daftspunk:
Yes, perhaps submit a PR for this to be added. Add it as specifically a `GERMAN_MAP` so we can pick and choose based on the locale in future.

gabsource:
All the chars @codebeauty is reporting are already bundled in the `LATIN_MAP` in the `develop` branch. See at https://github.com/octobercms/october/blob/develop/modules/system/assets/ui/js/input.preset.js#L26

At least it works for me by copy-pasting the char from the JS to the input field. 

![image](https://cloud.githubusercontent.com/assets/1851314/15448665/74a732ec-1f68-11e6-8991-413ba8cbd733.png)

Those are missing though in the `LATIN_MAP` though : 
```
'œ': 'oe', 'ō': 'o'
```
I will open a PR for those to show as :

![image](https://cloud.githubusercontent.com/assets/1851314/15448739/35e0a8a2-1f6a-11e6-9709-7af89948225a.png)

daftspunk:
I think the problem is that it converts to "a" instead of "ae", something distinct to German by the look of it. @alxy can you comment?

codebeauty:
@daftspunk you are right.  Like my example map above, in German 'ä' needs to be 'ae', 'ü' will be 'ue' and so on.

| Converter Input  | Converter Output |
| ------------- | ------------- |
| ä | ae  |
| ü | ue |
| ö | oe |
| ß | ss |

daftspunk:
So it needs its own MAP, then we should use the backend-locale to determine which map to use.

```
<meta name="backend-locale" content="en">
```

codebeauty:
Yes, @daftspunk looks like a good way. The converter can check for the backend-locale information and builds the map under this condition.

gabsource:
[...]

PR #2031, opened 2016-05-21
Add missing LATIN_MAP preset chars and allow specific map based on configured locale

gabsource:
In ref to #2026 added two missing LATIN chars to the preset list.


gabsource:
Added a commit to illustrate a specific map solution to #2026

Issue #910, opened 2015-01-25
Better formatting requried for Logs


Flynsarmy:
I don't know how to write acceptable targeted CSS for this given your class names but it's very hard to read PHP tracebacks in the event logs due to the line breaks. Here's a before and after pic of how they should look IMO:

Before:
![](http://i.imgur.com/U6bvXg5.png)

After:
![](http://i.imgur.com/hPDV6Qw.png)

flip4bytes:
I agree, the logs formatting could be improved. Here is an example of how papertrail displays their logs for instance:

![Logs](https://s3.amazonaws.com/f.cl.ly/items/0q2J3Z3b280x3N180Q1c/Image%202015-01-25%20at%206.12.15%20AM.png)

They use both color and indentation to break up each line making it a lot easier to scan (with wordwrap on).

Also, @Flynsarmy if you don't mind me asking, what plugin(s) are you using for the payments on the backend? Or did you build that payment settings section yourself?

Thanks!

Flynsarmy:
[responsiv/pay](https://github.com/responsiv/pay-plugin).

flip4bytes:
@Flynsarmy Ah, it's a WIP, gotcha. I'll check it out, thanks!

daftspunk:
Fixed by #2035

PR #2035, opened 2016-05-22
Feature to beautify event log in backend

gabsource:
My attempt to pretty print the logs. Refs #910 

It produces something like this : 

#### Stack trace message

image

##### Open with dialog

When `env` is `local` or `dev` the links on filepaths open an URL scheme for common editors (with an option to save the value in the sessionStorage - until the tab or browser is closed). Tested with PhpStorm at least.

image

#### Message with no stack trace

Displayed as usual with newlines characters replaced by `<br>` and tab or 4 spaces replaced by `&nbsp;&nbsp;&nbsp;&nbsp;`

image

### DIFF
--- a/modules/system/assets/css/eventlogs/preview.css
+++ b/modules/system/assets/css/eventlogs/preview.css
@@ -1,0 +1,83 @@
+[data-plugin="exception-beautifier"] {
+    width: auto !important;
+}
+
+.markup_exception_container {
+    overflow-x: scroll;
+}
+
+.markup_exception_name {
+    display: block;
+    font-weight: bold;
+    font-size: 1.2em;
+}
+
+.markup_exception_message {
+    display: block;
+    padding: 15px 0;
+    font-size: 1.1em;
+}
+
+.markup_exception_stacktrace_title {
+    display: block;
+    margin-top: 15px;
+    font-weight: bold;
+}
+
+.markup_exception_stacktrace_line {
+    display: block;
+    white-space: nowrap;
+    padding: 8px 0;
+}
+
+.markup_exception_stacktrace_line:nth-child(even) {
+    background-color: #ffffff;
+}
+
+.markup_exception_stacktrace_line:nth-child(odd) {
+    background-color: #f2f2f2;
+}
+
+.markup_exception_stacktrace_line_number {
+    display: inline-block;
+    width: 40px;
+    margin-right: 15px;
+    text-align: right;
+    color: black;
+}
+
+.markup_exception_stacktrace_line_internal {
+    color: #2C3E50;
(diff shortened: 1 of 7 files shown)


daftspunk:
Looks great!

daftspunk:
There are some minor things that need addressing:

1. The file name should match the plugin name. Rename `preview.js` to `exception-beautifier.js`

1. We like to use dashes in CSS class names. Eg: `markup_exception_stacktrace_line_internal` -> `markup-exception-stacktrace-line-internal`

1. The output needs to be escaped**, if an argument contains HTML, the output is broken
    ```
    Backend\Classes\Controller->makeViewContent('\n\n<form ...')
    ```

1. The "Open & Remember for this session" button should probably be [a checkbox](http://octobercms.com/docs/ui/checkbox) instead with a label "Remember for this session"

1. The "Open with" dropdown should use the [custom select](http://octobercms.com/docs/ui/select) control.

1. The container (currently `markup_exception_container`) should use `overflow-x: auto` so the scrollbars are only enabled as needed

** Sample code to escape output:

```
function(string) {
    var htmlEscapes = {
            '&': '&amp;',
            '<': '&lt;',
            '>': '&gt;',
            '"': '&quot;',
            "'": '&#x27;',
            '/': '&#x2F;'
        },
        htmlEscaper = /[&<>"'\/]/g

    return ('' + string).replace(htmlEscaper, function(match) {
        return htmlEscapes[match];
    })
}
```

gabsource:
Thanks for the review, I'll work on it and create a new PR.

daftspunk:
Fantastic, thanks. I have [revived an escape helper in the backend](https://github.com/octobercms/october/commit/e20c0126ee5101fb7178022954999fb07a6f59c5), please use `$.oc.escapeHtmlString` for the escaping.

gabsource:
Great, I will use it to escape everything between markup tags. There could be HTML or reserved characters anywhere in the source.

I will rename the CSS class name to `exception-beautifier-*` to match the plugin name too.
Ara you ok with the plugin name or do you think of something more appropriate ?

daftspunk:
Ideally it would be `.plugin-exception-beautifier` to match `data-plugin="exception-beautifier"`. Then once within that scope, all child selectors can be prefixed with `beautifier-` as a short hand name. Eg:

```
.plugin-exception-beautifier .beautifier-function { }
.plugin-exception-beautifier .beautifier-sign { }
.plugin-exception-beautifier .beautifier-stacktrace-line-function { }
```

This is the pattern used throughout the other areas of October.

Flynsarmy:
This is amazing. Great work guys!

PR #2041, opened 2016-05-24
:fr: updated based on the :uk: version

MATsxm:
:fr: updated based on the :uk: version


MATsxm:
Closed (wrong branch)
Please see #2067

PR #2042, opened 2016-05-24
:fr: updated

MATsxm:
:fr: updated


MATsxm:
Closed (wrong branch) see #2066

PR #2044, opened 2016-05-24
Optimized brand settings layout for HD monitors

gergo85:
Maybe it is the better layout for high resolution monitors.

Before:
before

After:
after


daftspunk:
Looks nice, thanks

PR #2054, opened 2016-05-25
Improved exception beautifier plugin

gabsource:
Improves the backend event preview page : 
- Keep the raw message in another tab
- Added a button to show/hide the stacktrace
- Fixes Windows paths parsing
- Translate the page title

#### Initial state
image

#### Mouse hover
image

#### Deployed
image

#### Raw
image

### DIFF
--- a/modules/system/assets/css/eventlogs/exception-beautifier.css
+++ b/modules/system/assets/css/eventlogs/exception-beautifier.css
@@ -2,28 +2,38 @@
     width: auto !important;
 }
 
-.plugin-exception-beautifier .beautifier-name {
-    display: block;
-    font-weight: bold;
-    font-size: 1.2em;
-}
-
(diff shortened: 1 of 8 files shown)


daftspunk:
Did you notice the [Select the source code editor to use] popup appears twice when you click on a file path? One on top of the other.

gabsource:
Seems like I can't reproduce. Does a produce a duplicate in the DOM ? I have only one in mine, even if I click fast.

But I get a strange nesting though 
image

daftspunk:
Watch the animation and see the double box shadow. It is more than likely caused by the double up on `modal-content > modal-dialog > modal-content > modal dialog`

It should be `modal-content > modal-dialog`

gabsource:
Yes I added the `modal-dialog` in the content provided to `$.popup`, it should only start with `modal-content`. I open the PR to fix it.

PR #2058, opened 2016-05-26
Fix the double-popup animation effect due to bad nesting

gabsource:
Reported in #2054 comments

### DIFF
--- a/modules/system/assets/js/eventlogs/exception-beautifier.links.js
+++ b/modules/system/assets/js/eventlogs/exception-beautifier.links.js
@@ -37,31 +37,24 @@
                 open = $.oc.lang.get('eventlog.editor.open'),
                 cancel = $.oc.lang.get('eventlog.editor.cancel'),
                 popup = $('                                                                                            \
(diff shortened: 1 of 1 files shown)


daftspunk:
I don't think `<div class="modal-content">` is needed either.

gabsource:
I need a single root node content (that get disposed by the `$.popup`), I can remove the `class` and let the `<div>` alone if you prefer.

gabsource:
I've changed the code to use the jQuery object directly instead of the generated html in order to only define the real content with no wrapping tag.

daftspunk:
Uh oh, it looks to be functionally broken now - open the popup twice and the select dropdown fails. Perhaps it is being loaded twice?


            // Prevent duplicate loading
            if ($element.data('select2') != null) {
                return true; // Continue
            }

gabsource:
`data('select2')` is always undefined ... better revert and stick with the raw HTML (with the dummy root required for calling `.html()`) instead of the jQuery object...

PR #2070, opened 2016-05-29
Re-apply the crop-insert selection mode after the first use

### LINKED ISSUE #2022: Crop and insert does not remember options correctly
##### Expected behavior
The crop tool options should match the displayed options.

##### Actual behavior
The crop tool uses the default options. 

##### Reproduce steps
1. Select crop and insert
2. Set mode to fixed ratio or fixed size
3. Make a selection
4. Click crop and insert
5. Open crop and insert again
6. UI shows the mode selected on step 2 but the tool uses normal mode

##### October build
318

gabsource:
The crop-insert options are kept in session across requests. This fix applies the selection mode after the popup is shown to reflect the displayed options of the UI.

Fixes #2022


daftspunk:
Fixed by #2070

PR #2074, opened 2016-05-31
Command for changing configuration to env syntax

mplodowski:
Most Laravel developers uses env() syntax in configuration files, but OctoberCMS uses traditional default values approach.

Using env() has many advantage for developers, such as:

- allow for different configuration values based on the environment, 
- we don't push sensitive config informations to repository, 
- we can share configuration between team,
- we keep configuration in one place.

This command will change configuration files to use env() syntax for the most used values and create .env file with defaults.

**Use:**

    php artisan october:env

**Result .env file:** 

    APP_DEBUG=true
    APP_URL=http://localhost
    APP_KEY=anmYJlYVfA76hpRaLNjKgXvN2caC9dDU
    
    DB_CONNECTION=mysql
    DB_HOST=localhost
    DB_PORT=
    DB_DATABASE=database
    DB_USERNAME=root
    DB_PASSWORD=
    
    CACHE_DRIVER=file
    
    SESSION_DRIVER=file
    
    QUEUE_DRIVER=sync
    
    MAIL_DRIVER=mail
    MAIL_HOST=smtp.mailgun.org
    MAIL_PORT=587
    MAIL_USERNAME=null
    MAIL_PASSWORD=null
    MAIL_ENCRYPTION=tls
    
    ROUTES_CACHE=false
    ASSET_CACHE=false
    LINK_POLICY=detect
    ENABLE_CSRF=false

### DIFF
--- a/modules/system/ServiceProvider.php
+++ b/modules/system/ServiceProvider.php
@@ -219,6 +219,7 @@ class ServiceProvider extends ModuleServiceProvider
         $this->registerConsoleCommand('october.util', 'System\Console\OctoberUtil');
         $this->registerConsoleCommand('october.mirror', 'System\Console\OctoberMirror');
         $this->registerConsoleCommand('october.fresh', 'System\Console\OctoberFresh');
+        $this->registerConsoleCommand('october.env', 'System\Console\OctoberEnv');
 
         $this->registerConsoleCommand('plugin.install', 'System\Console\PluginInstall');
         $this->registerConsoleCommand('plugin.remove', 'System\Console\PluginRemove');
--- a/modules/system/console/OctoberEnv.php
+++ b/modules/system/console/OctoberEnv.php
@@ -16,6 +16,16 @@ class OctoberEnv extends Command
     protected $description = 'Creates .env file with default configuration values.';
 
     /**
+     * The current config cursor.
+     */
+    protected $config;
+
+    /**
+     * The current database connection cursor.
+     */
+    protected $connection;
+
+    /**
      * Create a new command instance.
      */
     public function __construct()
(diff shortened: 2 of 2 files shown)


gabsource:
OctoberCMS handles this in its own way already (kept from the time it was using L4)

https://octobercms.com/docs/setup/configuration#environment-config

mplodowski:
The idea is here that after every time you create project you must change each configuration file and replace static October values with env() function like so:

from:

        'mysql' => [
            'driver'    => 'mysql',
            'host' => 'localhost',
            'port' => '',
            'database' => 'database',
            'username' => 'root',
            'password' => '',
            'charset'   => 'utf8',
            'collation' => 'utf8_unicode_ci',
            'prefix'    => '',
        ],

to:

        'mysql' => [
            'driver'    => 'mysql',
            'host' => env('DB_HOST', 'localhost'),
            'port' => env('DB_PORT', ''),
            'database' => env('DB_DATABASE', 'database'),
            'username' => env('DB_USERNAME', 'root'),
            'password' => env('DB_PASSWORD', ''),
            'charset'   => 'utf8',
            'collation' => 'utf8_unicode_ci',
            'prefix'    => '',
        ],

And that take some time and it is tedious to do. With this command we can set up our environment in seconds, like we used to in Laravel. Laravel ships with configuration files including env() set up by default.

gabsource:
Yes I understand your problem as I often do the same thing. But OctoberCMS Team seems not to have chosen the way Laravel 5 handles configuration (through `.env`).

Don't get me wrong you idea is cool :-) There is the `october:mirror` to replicate the public folder path so why not a command to replicate `.env` config. Event if it's not integrated in the core you can publish it as as a dev plugin.

When you run the command, the `redis` config is assigned the `DB_USERNAME` key. It should be using a `REDIS_` I think.

mplodowski:
Yeah I read previous PRs similar to this one. I know the October team what they think about this, but this is a change which don't involve October itself and allow developers to do theirs way.

Yes, probably I will create a plugin if this PR will not be accepted, but I think it should do some good like `october:mirror`.

Good catch with REDIS_, I will fix this later, but now I am interested what October team think about this.

daftspunk:
Looks good, it needs some supporting documentation on how to use it.

mplodowski:
Just fixed database config parsing.

@daftspunk do I need to add documentation to october/docs repo?

Here https://octobercms.com/docs/setup/configuration#environment-config ?

daftspunk:
Or you can just write the docs here for us to review them.

mplodowski:
## Using .env configuration

Storing configuration in the environment is one of the tenets of a [twelve-factor app](http://12factor.net/config). Anything that is likely to change between deployment environments–such as resource handles for databases or credentials for external services–should be extracted from the code into environment variables.

To simplify using [DotEnv](https://github.com/vlucas/phpdotenv) syntax, run the following command:

    php artisan october:env
    
This will create .env file in project root directory and alter configuration files to use `env()` helper method.

If you now review the configuration files, you will notice several of the options using this helper:

    'debug' => env('APP_DEBUG', true),

The second value passed to the `env` function is the "default value". This value will be used if no environment variable exists for the given key.

Your .env file should not be committed to your application's source control, since each developer / server using your application could require a different environment configuration.

mplodowski:
@daftspunk what do you think?

daftspunk:
@mplodowski is this feature finished? Ready to be merged?

mplodowski:
Yes, I checked again on stable version and everything works fine.

daftspunk:
Rock 'n roll :metal:

This has been merged in 53b8df0721a3...7b1437d02697

PR #2087, opened 2016-06-03
Added double quote for the description of the theme

meysammahfouzi:
Because there is colon in the description, I added double quotes to it to eliminate this error in the backend's theme page:

> A colon cannot be used in an unquoted mapping value at line 2 (near "description: Demo OctoberCMS theme. Demonstrates the basic concepts of the front-end theming: layouts, pages, partials, components, content blocks, AJAX framework.").


daftspunk:
This is already done in `develop` branch.

PR #2090, opened 2016-06-04
New translation attachment_url

reg2005:
In Russian it will be so more correct


daftspunk:
Merged in b26fab725adb27413b0542c94f6956594db86c85

PR #2092, opened 2016-06-04
Fix searching in non text fields under Postgres

CSNWEB:
When using postgres the List Widgets threw an error for non text fields (e.g. integer). With this fix searching for an id or in other numerical columns works.

Please let me know if my solutions suits you, I might add it for the relation columns above too. I didn't yet because I have no example and would have to create on to test it.

### DIFF
--- a/modules/backend/widgets/Lists.php
+++ b/modules/backend/widgets/Lists.php
@@ -343,7 +343,6 @@ class Lists extends WidgetBase
                     $columnName = isset($column->sqlSelect)
                         ? DbDongle::raw($this->parseTableName($column->sqlSelect, $table))
                         : $table . '.' . $column->valueFrom;
-                    $columnName = DbDongle::cast($columnName, 'text');
                     $relationSearchable[$column->relation][] = $columnName;
                 }
                 /*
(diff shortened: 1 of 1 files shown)


daftspunk:
We should be using the [`\DbDongle`](https://github.com/octobercms/library/blob/develop/src/Database/Dongle.php) class for this.

CSNWEB:
I decided to cast every field regardless of the type for Postgres. It doesn't hurt and the search works like  expected when using mysql.

CSNWEB:
@daftspunk do you have any comments on the new version :)? As I'm using October with Postgres in several instances, it would help me a lot if this gets merged. Thanks

mplodowski:
The same issue exists for date column type. @CSNWEB can you also apply fix for this?

I think we should cast date to year only "YYYY".

CSNWEB:
@mplodowski the last commit already fix it for all types as everything is now casted to text. But I wouldn't cast any date to year only you can still search for -YYYY to find the year, but you could also be searching for an exact date when I leave it like it is now.

mplodowski:
That's fine by me. Thanks for this PR.

daftspunk:
Merged in 34bd318427135f8e3e19e7e1bc30d0629490f513
Thanks!

PR #2105, opened 2016-06-07
Fix fileupload image css on sidebar

triasrahman:
It's very annoying when I try to create image upload in form sidebar. We can't click the buttons like reorder and delete. So, I tweaked like this:

Before:
screen shot 2016-05-29 at 12 05 51 pm

After:
screen shot 2016-05-29 at 12 08 35 pm

Thanks,


daftspunk:
LGTM!

PR #2110, opened 2016-06-08
Update pt-br lang

acwoss:


daftspunk:
`Fatal error: syntax error, unexpected T_CONSTANT_ENCAPSED_STRING, expecting ']' in /home/travis/build/octobercms/october/modules/system/lang/pt-br/lang.php on line 13`

acwoss:
Fixed.

jfoliveira:
Hi @acwoss thanks for updating the translations for our lovely Portuguese language. :-)

Few questions:
1. Do you really think `esboço` is a better translation for `layout`? Everyone in this country uses/speaks the English word `layout`;
2. The same I believe applies to `template` that is very often used in English and you translated as `modelo`. Now `modelo` is the same translation used for `model`. Don't you think this might be misleading in some areas of October?
3. The same for `time zone` translated as `zona de horário` instead of `fuso horário`. Tem uma razão para isso?

Overall great job! Thank you for that!

acwoss:
@jfoliveira welcome, thanks for the feedback.

The translations mentioned in items 1 and 2 I made keep in mind that most who know the terms `layout` and `template`, knowing differentiate between the, are developers. Bout the end user, considering that you are developing a product for a client, may not be familiar with these terms. I don't see as a problem in keeping the terms in english, but I believe make the translation facilitates to understanding to lay. I confess I had not considered the ambiguity between translations for `modelo`. What do you think?

I agree, for item 3, that `fuso horário` is the correct translation. The term escaped from my mind at the time of translation and ended up doing a literal translation. I will include this item for the nest pull.

Sorry for possible mistakes in writing.

jfoliveira:
@acwoss thanks for the quick response.
I think `Templates` might be kept as `modelo` to keep it in Portuguese, perhaps not as a big deal as I originally thought, as they more likely not used in the same context as `models`.

`Layout` has been incorporated to Portuguese language with a new spelling `leiaute` and broader meaning. See [Michaelis](http://michaelis.uol.com.br/moderno/portugues/index.php?palavra=leiaute) and [Priberam](http://www.priberam.pt/dlpo/leiaute).

Perhaps translating `layout` to `leiaute` might help the non-developers to be able to easier search for that word on a dictionary if they don't know its meaning.
Does that make sense?

acwoss:
@jfoliveira I agree. I will include the translation for `layout` in the next pull also.

Issue #1848, opened 2016-03-16
Allow developers to change preview image


pedzed:
Related: http://octobercms.com/docs/themes/development#preview-image

The theme preview **MUST** be placed into `assets/images/theme-preview.png`. This is a very annoying limitation.
- This prevent the usage of JPG, GIF, SVG, etc.
- This forces you to create an images directory.

In our case, we're using an image minifier (with Gulp) which minifies everything from `assets/src/images` to `assets/dist/images`. It would be silly to create a new directory, just for the theme preview image.

daftspunk:
Fixed by #2426

PR #2117, opened 2016-06-12
Load the theme preview image from different folders and extensions

gergo85:


gabsource:
@daftspunk self assigned the issue related to this problem https://github.com/octobercms/october/issues/1848

Maybe he has thoughts on how to make it more generic / configurable.

gergo85:
New solution: #2426

LukeTowers:
If accepted, #2426 will be what we go with.

PR #2127, opened 2016-06-17
morphMany support

slipgrom:
RelationController with MorphMany relation support


LukeTowers:
@slipgrom Please add a test case to the [test plugin](https://github.com/daftspunk/oc-test-plugin) and also change the branch as requested. Let us know if you don't want to do that, we'll revisit this at a later date then.

daftspunk:
Pretty sure this was added in dfdd5f2d5613ccb1e0df8a6b73bae96a93a56f72 / #2128

Not even sure if this works...

PR #2139, opened 2016-06-20
Update the jQuery

gergo85:
Version: 2.1.3 -> 2.1.4
This is the latest version of 2.1.x thread.
The 2.2.x and 3.0.x versions maybe not compatible with older JS plugins.


daftspunk:
We should update this as we need it, otherwise we risk breaking something for no good reason.

PR #2140, opened 2016-06-20
Critical bugFix in themeData.php

reg2005:
Additional check is added, without it there is a critical error "Call to a member function hasCustomData() on null"

### DIFF
--- a/modules/cms/models/ThemeData.php
+++ b/modules/cms/models/ThemeData.php
@@ -207,6 +207,11 @@ class ThemeData extends Model
     public static function applyAssetVariablesToCombinerFilters($filters)
     {
         $theme = CmsTheme::getActiveTheme();
+        
+        if(!$theme){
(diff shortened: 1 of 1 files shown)


keiosweb:
Why exactly is this critical? In my case this pull request changed

```
/* Call to a member function hasCustomData() on a non-object (null) */
```

exception into 

```
/* The source file "/srv/www/october/themes/demo/assets/css/theme.css" does not exist. */
```

(on combined resources request, request to main domain shows "no active theme found" properly).

Still, empty **return** in my opinion is a bad way around this and there should be something like

```
if($theme){
throw new \Exception('no active theme found');
}
```

to exactly show what's wrong with the application.

daftspunk:
I'm more curious to know how applyAssetVariablesToCombinerFilters() is being called without an active theme.

PR #2141, opened 2016-06-20
Added path

reg2005:
During creation of a mirror, symlink on assets wasn't created. Many widgets use this way. For example: https://github.com/october-widgets/hasmany


daftspunk:
We need another way to address this. It is a non-standard path. It could be included by extension, using an event.

LukeTowers:
If anybody needs the suggested functionality to be included (extension of the mirrored / symlinked directories), please create a separate issue / pull request.

PR #2155, opened 2016-06-28
laravelcollective/html instead of illuminate/html

MarcoROG:
illuminate/html is abandoned


daftspunk:
October no longer uses this package. Thanks

PR #2156, opened 2016-06-28
Fix pt-br in $.oc.lang

brenodouglas:
Pt-br for filter, datepicker, alert and mediamanager components.


daftspunk:
This is a compiled file, it is most likely fixed by now. Thanks

PR #2157, opened 2016-06-28
error on saving attachment config when in pivot form of a relation

prasanth-darsan:
There is a error when you enter title and description of a attachment popup, if the attachment is put in a pivot form of a relation.

https://octobercms.com/forum/post/pivot-edit-form-and-image-upload

This is because the popup does not have manage_id and _related_field to pass back to the ajax handler once save is clicked form the popup.


daftspunk:
Great stuff, thanks!

PR #2166, opened 2016-07-04
Remove the old comments

gergo85:


daftspunk:
These comments show the Flat UI palette, October's colours stem from these. It should be kept for now.

PR #2188, opened 2016-07-07
Support table prefixes in new migrations

schniper:


daftspunk:
Might be better to [add this to the dongle itself](https://github.com/octobercms/library/blob/master/src/Database/Dongle.php#L188)

schniper:
I was reluctant to do it there so as not to break other scripts. But as I see that method is not used in any other way, I will do just that.

PR #2193, opened 2016-07-08
Make getFallbackLocale static

filipac:
$this is not accessible in static context in function applyConfigValues so the getFallbackLocale has to be declared as static and references to this function rewritten as static:: calls instead of using $this object


filipac:
thanks

PR #2205, opened 2016-07-12
Commit folder structure to fix october:mirror

JoelESvensson:
`october:mirror` won't mirror `storage/app/uploads`, `storage/app/media`, `storage/temp/public` if the folders don't exist.


daftspunk:
I bumped in to this issue too. Nice way to handle it. Thanks

PR #2206, opened 2016-07-12
Escapes exception output

### LINKED ISSUE #2198: Combine Assets Reflected XSS
One of the cool features of OctoberCMS is to minify assets (CSS and JS) so we use it all the time but one of our vulnerability assessment specialist report a vulnerability on the combine url that serve the  css and js.

POC:  
valid combined js -  https://mysite.com/combine/ea9f0b900ad64d2c2894fef4967643c3-1465376894
crafted xss link - https://mysite.com/combine/%3Cimg%20src=x%20onerror=prompt%28document.domain%29%3E
will produce like this 
image

Is this a valid Reflected XSS? I'm using Build 318.

JoelESvensson:
Fixes #2198

### DIFF
--- a/modules/system/classes/Controller.php
+++ b/modules/system/classes/Controller.php
@@ -35,7 +35,7 @@ class Controller extends ControllerBase
 
         }
         catch (Exception $ex) {
-            return '/* '.$ex->getMessage().' */';
+            return '/* '.e($ex->getMessage()).' */';
         }
     }
 }


JoelESvensson:
I suggest you get some kind of security policy in place so there is a process to quickly handle vulnerabilities. An email adress to contact will probably suffice. I think this is what Laravel does

PR #2207, opened 2016-07-13
Only public folder should be mirrored

JoelESvensson:
`storage/app/uploads` may also contain private data. Only `storage/app/uploads/public` is public


daftspunk:
Good pick up

PR #2210, opened 2016-07-13
Parse Twig code in mail view before parsing it into sections.

### LINKED ISSUE #2202: Twig code in settings sections of mails not working anymore since stable
##### Expected behavior

When using Twig syntax in the settings section of an email it should be processed correctly. This was the case up until the stable release.

##### Actual behavior

When using Twig syntax in the settings section of an email an ErrorException is thrown:

> ErrorException: syntax error, unexpected '{' in Unknown on line 1
 in /workdir/vendor/october/rain/src/Mail/MailParser.php:52

##### Reproduce steps

Create an email with the following contents:

```
{% if isAdmin %}
    subject = "This is the subject for admins"
{% else %}
    subject = "This is the subject for everyone else"
{% endif %}
==
Text content
==
HTML content
```

##### October build

348

tobias-kuendig:
This PR fixes issue #2202.

If a mail view contains Twig code in the header section the `parse_ini_string` call inside of [`MailParser::parse`](https://github.com/octobercms/library/blob/master/src/Mail/MailParser.php#L47) fails.

This PR parses all twig code before it gets passed to the `MailParser`.

### DIFF
--- a/modules/system/models/MailTemplate.php
+++ b/modules/system/models/MailTemplate.php
@@ -106,9 +106,9 @@ class MailTemplate extends Model
         }
     }
 
-    public function fillFromView()
+    public function fillFromView($data)
     {
-        $sections = self::getTemplateSections($this->code);
+        $sections = self::getTemplateSections($this->code, $data);
         $this->content_html = $sections['html'];
         $this->content_text = $sections['text'];
         $this->subject = array_get($sections, 'settings.subject', 'No subject');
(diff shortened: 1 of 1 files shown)


daftspunk:
The settings section is not supposed to support Twig in this way, eg:

```
{% if isAdmin %}
    subject = "This is the subject for admins"
{% else %}
    subject = "This is the subject for everyone else"
{% endif %}
==
Text content
==
HTML content
```

The intent is to parse the subject only with Twig, eg:

```
subject = "{{ customSubject }}"
==
Text content
==
HTML content
```

PR #2212, opened 2016-07-13
Show 404 page when directly visiting /error

Herlevsen:
Currently when visiting /error, the actual 500 error page is shown, with status code 200. Since the /error url is a hardcoded error page, it should not be visited directly, and should instead show the 404 page with 404 status code. #1723

### DIFF
--- a/modules/cms/classes/Controller.php
+++ b/modules/cms/classes/Controller.php
@@ -194,7 +194,7 @@ class Controller
         /*
          * If the page was not found, render the 404 page - either provided by the theme or the built-in one.
          */
-        if (!$page || $url === '404') {
+        if (!$page || $url == '404' || $url == 'error') {
             if (!Request::ajax()) {
                 $this->setStatusCode(404);
             }


daftspunk:
This prevents us from being able to easily test the `/error` page...

Herlevsen:
Good point.

Maybe we can check if the visitor is logged into the backend, and if not show the 404 page? Maybe even check if the user has the "cms.manage_pages" permission?

Herlevsen:
@daftspunk, what do you think, should i give that a go?

daftspunk:
Might be easier to just jack in to the `app.debug` setting. If the debug mode is disabled, then don't allow direct routing to the error page. Debug mode should be disabled in production environments.

This needs more thought and discussion. Should we even do this? Primarily, what is the main issue with allowing the `/error` page to exist?

Herlevsen:
I think it's a problem, that the /error page might possibly be indexed by search engines. It also just seems wrong that a user might browse to the error page by accident, even though nothing has went terribly wrong.

I think the app.debug settings is a good idea. Do you have any concerns with this approach, and could this possibly have any problems concerning backwards compatibility?

daftspunk:
I think this needs some discussion and thought, to discover the best solution. Will get back to you.

keiosweb:
well, we can all agree that direct browsing to error page is needed for debug purposes only. thereby +1 to app.debug idea.

LukeTowers:
@Herlevsen could you change this to also check the app.debug setting as well as using strict comparisons `===`?

LukeTowers:
Implemented in https://github.com/octobercms/october/commit/7c7ff31cd6adc94492af5cf919fc538bdff2060d

PR #2215, opened 2016-07-15
Pass the router reference on beforeRoute event

acasar:
I need to access the ``Router`` in the ``cms.router.beforeRoute`` event and this was the easiest change to enable that.

If there is any other way to access it please let me know.


acasar:
Wrong branch

PR #2219, opened 2016-07-17
Add 'normal' field preset type

Herlevsen:
Currently the preset field option does not allow to preset with the exact same value as the other field, but formats it as slug, url etc.

This PR adds a 'normal' preset type, that simply copies the value from the other field.

My use case was i needed to preset a pluralized word from the singular version, since they are the same in many cases.

Will add to documentation if merged.

### DIFF
--- a/modules/system/assets/ui/js/input.preset.js
+++ b/modules/system/assets/ui/js/input.preset.js
@@ -244,7 +244,10 @@
     }
 
     InputPreset.prototype.formatValue = function() {
-        if (this.options.inputPresetType == 'namespace') {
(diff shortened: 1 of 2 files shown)


daftspunk:
Thanks for the detailed notes. This should probably be called `exact` instead. This will be accepted, please go ahead and submit a PR to the docs repo to accompany this one.

Herlevsen:
Changed the type to `exact`. Will add documentation later today.

Btw, i rebased onto the latest commit and squashed the commit, to avoid unnecessary commit and make the build pass.

Herlevsen:
Forgot about the documentation, but it's there now

PR #2225, opened 2016-07-18
Add afterSort event to ReorderController's onReorder method

manogi:
I needed an event to listen to when a list has been reordered in the backend. So I added this event which fires every time a list has been reordered (either with the Nested Tree or the simple Sortable trait).

### DIFF
--- a/modules/backend/behaviors/ReorderController.php
+++ b/modules/backend/behaviors/ReorderController.php
@@ -136,6 +136,11 @@ class ReorderController extends ControllerBehavior
                     break;
             }
         }
+
+        $model->fireEvent('model.afterSort');
+        if ($model->methodExists('afterSort')) {
+            $model->afterSort();
+        }
     }
 
     //


daftspunk:
Thanks for this @manavo. The ReorderController mutates the logic of the controller, so this logic should be implemented in the controller. Here is how you can do this without modifying any code, define the following method in your controller:

    public function onReorder()
    {
        $this->asExtension('ReorderController')->onReorder();

        // Place the custom logic here
        // [...]

        // For example: the event code
        $model = $this->reorderGetModel();
        $model->fireEvent('model.afterSort');
        if ($model->methodExists('afterSort')) {
            $model->afterSort();
        }
    }

Have a great day!

manogi:
Thank you @daftspunk and @manavo for taking the time to look into this. Learned something again, didn't know I could "extend" the method this way in my own controller. Thanks!

PR #2231, opened 2016-07-21
bug fix in taglist widget with nameFrom option

enniel:
**Problem:**

If using taglist widget with mode relation, nameFrom is set as name by default and custom value don't worked.

```
   groups:
        label: Groups
        type: taglist
        mode: relation
        nameFrom: label
```


daftspunk:
Good pick up, thanks @enniel!

Merged in 37a5ae611b40e069c5c70bcbc61a72d6f33579f0

Issue #1679, opened 2016-01-03
Issues with repeater drag and drop sorting


darkwoods-:
Seems like there is a issue with dragging and dropping in the repeater.
If I drag and drop child-item it can get stuck in the parent-item and I can not take it back as child item and is forced to refresh the whole page to fix it back as child item..

thanks

jwilson8767:
duplicate of [rainlab/pages-plugin#109](https://github.com/rainlab/pages-plugin/issues/109)

daftspunk:
Is this when using nested repeaters?

darkwoods-:
@daftspunk yes when using nested repeaters the nested items gets stuck in the parent item when reordering them and makes the whole drag and drop reordering chaotic and unusable.

manogi:
I have the same issue. Reordering nested repeater items. The nesed items lands in the parent level when I try to reorder them.

manogi:
...in Build 318

daftspunk:
Nested repeaters aren't really supported at this stage (initial tests were not performed for a repeater in a repeater). The possible solution would be to give each repeater a unique group code in the sortable plugin, if not already. It seems like children can be moved to parents, but parents not to children. This is odd and it looks to be bugging out.

For replication: See Countries > States tab in [Test plugin](http://github.com/daftspunk/oc-test-plugin)

daftspunk:
Fixed by: #2233
Our hero: @acasar

![Up top](https://dl.dropboxusercontent.com/u/5381773/lel/up-top.jpg)

PR #2232, opened 2016-07-21
Fix nested repeater sorting

acasar:
This should fix https://github.com/octobercms/october/issues/1679

Basically I used an id to initialize a sortable plugin instead of a class. "Handles" now have unique classes as well.


acasar:
wrong branch :/

Issue #1891, opened 2016-03-31
dropdown options in repeater field doesn't update properly when depending on another field


fresswolf:
Let's assume the following configuration
```yaml
fields:
    one:
        label: One
    another:
        label: Tickets
        type: repeater
        dependsOn: one
        form:
            fields:
                two:
                    label: Two
                    type: dropdown
                    dependsOn: one
```

and further some dropdown options depending on the first field

```php
public function getTwoOptions()
    {
        if ($this->one) {
            return ['my new option: ' . $this->one];
        }
        return ['nothing'];
    }
```

Now, when I add one item and then change the value in "one", the dropdown options for "two" are refreshed as expected. However, if I first edit "one" and then add an item in the repeater, the options inside the repeater have not been refreshed

daftspunk:
This is a complex case. Please submit a Pull  #Request to the [test plugin for investigation](https://github.com/daftspunk/oc-test-plugin) that demonstrates the issue. Don't forget to include step by step instructions how how to replicate it *using the test plugin*.

daftspunk:
Fixed by: #2235
Our hero: @acasar

![Up top](https://dl.dropboxusercontent.com/u/5381773/lel/up-top.jpg)

PR #2235, opened 2016-07-21
DependsOn in repeater

acasar:
This fixes https://github.com/octobercms/october/issues/1891

I revisited this old issue, since I experienced the same problem myself. When we use ``DependsOn`` in a repeater field, dependent fields don't update properly.

To solve this issue I had to update the ``formwidget`` jQuery plugin to work in nested situations. With this solution, we can now use the following configuration:

```yaml
fields:
    Locations:
        label: Locations
        prompt: 'Add new location'
        span: full
        type: repeater
        form:
            fields:
                country:
                    label: 'Country'
                    span: auto
                    type: dropdown
                state:
                    label: 'State'
                    span: auto
                    dependsOn: country
                    type: dropdown
```

In the model we can define ``getStateOptions()`` as usual, but this time ``$this`` variable can be used to access repeater data:

```php
public function getStateOptions() {
    return State::where('country', $this->country)->lists('title', 'id');
}
```

### DIFF
--- a/modules/backend/formwidgets/Repeater.php
+++ b/modules/backend/formwidgets/Repeater.php
@@ -158,4 +158,12 @@ class Repeater extends FormWidgetBase
         // Useful for deleting relations
     }
 
+    public function onRefresh()
+    {
+        $index = post('_index');
+
+        $widget = $this->makeItemFormWidget($index);
+
+        return $widget->onRefresh();
+    }
 }
--- a/modules/backend/formwidgets/repeater/partials/_repeater.htm
+++ b/modules/backend/formwidgets/repeater/partials/_repeater.htm
@@ -13,7 +13,8 @@
         <a
             href="javascript:;"
             data-load-indicator
-            data-request="<?= $this->getEventHandler('onAddItem') ?>">
+            data-request="<?= $this->getEventHandler('onAddItem') ?>"
+            data-request-success="$('#<?= $this->getId('items') ?> .field-repeater-form:last').formWidget()">
             <?= e(trans($prompt)) ?>
         </a>
     </div>
--- a/modules/backend/formwidgets/repeater/partials/_repeater_item.htm
+++ b/modules/backend/formwidgets/repeater/partials/_repeater_item.htm
@@ -17,7 +17,10 @@
         </button>
     </div>
 
-    <div class="field-repeater-form">
+    <div class="field-repeater-form"
+         data-control="formwidget"
+         data-refresh-handler="<?= $this->getEventHandler('onRefresh') ?>"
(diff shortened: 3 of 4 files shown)


acasar:
One problem I still have with this solution is how to initialize the options. Right now it has to be done like this:

```php
public function getStateOptions($value) {
    $country = $this->country ?: State::find($value)->country;

    return State::where('country', $country)->lists('title', 'id');
}
```

That's because the ``country`` attribute doesn't really exist on this model, it's only a key in repeater. If anyone has any ideas on how to make this easier to use, let me know.

daftspunk:
Some great contributions here. Also some very advanced usage. The solution to the latter problem is to pass a static method call to the `options` option. This will give you greater access to the widget properties and obtain the `$data` values needed.

    [...]
    state:
        label: 'State'
        span: auto
        dependsOn: country
        type: dropdown
        options: ['\Path\To\Model', 'getStateOptionsSpecial']


    public static function getStateOptionsSpecial($widget, $field)
    {
        $countryId = empty($widget->data->country)
            ? key((new self)->getCountryOptions())
            : $widget->data->country;

        return State::where('country', $countryId)->lists('title', 'id');
    }

It is not ideal. Since we have come so far, let's see if we can improve on it.

daftspunk:
Let's see... the solution that jumps out the most: 
- Simply add the data set to the "options" method as the second argument.

    ```
    public function getStateOptions($currentValue, $dataSet)
    ```

This would allow for the following:

```
public static function getStateOptions($value, $data)
{
    $countryId = empty($data->country)
        ? key($this->getCountryOptions())
        : $data->country;

    return State::where('country', $countryId)->lists('title', 'id');
}
```

acasar:
Nice solution :+1: Thanks a lot!

zakir1929:
Hello @daftspunk  I used your solution but it does not worked and then I did some changes like this and its worked.

	public function getStateOptions($value, $attribute,$data){
	    $countryId = empty($data->country)
	        ? key($this->getCountryOptions())
	        : $data->country;
	    return \RainLab\Location\Models\State::where('country_id', $countryId)->lists('name', 'id');
	}

I know this issue is old and you did alot of modification to database table states but the main thing I need to ask is I passed 3 parameters here 

public function getStateOptions($value, $attribute,$repeater_object)

and 3rd parameters has the data object available.

so can you please help me to understand it?

Thanks and your CMS is awesome!!!

LukeTowers:
@zakir1929 see http://octobercms.com/docs/backend/forms#field-dropdown

zakir1929:
Thanks!!! @LukeTowers. I got it now.

PR #2241, opened 2016-07-25
Ability to sync data for *Many relations on Saving Form.

Upperfoot:
Fixes: Ability to sync data for *Many relations on Saving Form.

No idea why it was not there in the first place?

Missing pull request functionality (save relations after model create).

### DIFF
--- a/modules/backend/traits/FormModelSaver.php
+++ b/modules/backend/traits/FormModelSaver.php
@@ -51,16 +51,23 @@ trait FormModelSaver
         }
 
         $singularTypes = ['belongsTo', 'hasOne', 'morphOne'];
+        $multipleTypes = ['belongsToMany', 'hasMany'];
         foreach ($saveData as $attribute => $value) {
             $isNested = $attribute == 'pivot' || (
                 $model->hasRelation($attribute) &&
                 in_array($model->getRelationType($attribute), $singularTypes)
             );
+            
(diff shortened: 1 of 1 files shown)


daftspunk:
Hey @Upperfoot, it wasn't there in the first place because it doesn't need to be. Take a look at the [belongsToMany unit test](https://github.com/octobercms/october/blob/master/tests/unit/plugins/database/BelongsToManyModelTest.php#L39):

```
// Same as $author->roles()->sync(1,2,3,..);
$author->roles = [1, 2, 3, ...];
```

daftspunk:
Feel free to submit a Pull Request to the [test plugin for investigation](https://github.com/daftspunk/oc-test-plugin) that demonstrates the issue.

Upperfoot:
Just tried this against a fresh copy of Master branch of October, and a fresh copy of the Rainlab Plugin.

Assigning groups to a user fails to save, CTRL+F5 on a user edit page after assigning that user to a group shows that the groups have not been saved.

https://github.com/rainlab/user-plugin/blob/master/models/user/fields.yaml#L58-L62

daftspunk:
Hmm, I was unable to replicate this. 

1. I opened an existing user
1. Assigned the default group.
1. Clicked save
1. Refresh the page
1. The group was still assigned

PR #2245, opened 2016-07-26
Provides event to extend relation's configuration

LukeTowers:
Provides the `backend.relationController.modifyFieldConfig` event to extend the configuration of individual relation fields based on the model data.

### DIFF
--- a/modules/backend/behaviors/RelationController.php
+++ b/modules/backend/behaviors/RelationController.php
@@ -301,6 +301,11 @@ class RelationController extends ControllerBehavior
 
         $this->alias = camel_case('relation ' . $field);
         $this->config = $this->makeConfig($this->getConfig($field), $this->requiredRelationProperties);
+        
+        /*
+         * Extend the field's config
+         */
+        Event::fire('backend.relationController.modifyFieldConfig', [$field, $this->config, $model]);
 
         /*
          * Relationship details


daftspunk:
More detail is needed. Please provide the `Event::listen` code you are using to hook this event.

LukeTowers:
This is the `Event::listen` code I'm using in my controller's `__construct` method before calling `parent::__construct()`
```php
// Make the fields relation readOnly if the current model has been published
Event::listen('backend.relationController.modifyFieldConfig', function ($field, $config, $model) {
	if ($field === 'fields' && $model instanceof SurveyModel) {
		if (!is_null($model->published_at)) {
			$config->readOnly = true;
		}
	}
});
```

daftspunk:
Have you tried using the `filterFields` method to achieve this? https://octobercms.com/docs/backend/forms#filter-form-fields

LukeTowers:
I have, yes. I'm actually using it for modifying the normal fields for that update form in general, but AFAIK there is no way to influence the configuration of the relationController behaviour through filterFields as the 'field' in question that houses the relation controller is just a partial that calls `relationRender()`.

```php
public function filterFields($fields, $context = null) {
    if ($context === 'update') {
	    if ($this->published_at) {
		    $fields->is_published->value = true;
		    $fields->name->disabled = true;
		    
		    // Disable unpublishing when responses are recorded
		    if (!is_null($this->responses) && !$this->responses->isEmpty()) {
			    $fields->is_published->disabled = true;	    
		    }
		    
			$descFieldId = 'options[description]';
			$fields->{$descFieldId}->disabled = true;
	    }
    }
}
```

I tried setting disabled on the field holding the `relationRender` call but that didn't do anything because it's just a partial. I also tried thinking of some solution where I'd pass `readOnly = false` through the call to `relationRender` but then I realized that it wouldn't take effect for AJAX requests. After some deliberation adding this event seems like the cleanest way to achieve my desired results, being able to modify the configuration of the relationController behaviour on the fly based on the model data, field id, and existing configuration.

daftspunk:
It should be applied to both instances because `filterFields` is called from the Form widget, shared by both. If I understand correctly, you want different rules to apply under different circumstances. This is where the `$context` variable comes in handy. Tell the relation controller to use a different context and you should be able to check for it in the second argument of filterFields.

LukeTowers:
If I understand you correctly, that still doesn't enable me to change the `readOnly` property of the relationController though. My relationController manage view has different fields that are loaded dynamically so I would prefer not to iterate over all of fields in the manage view and manually set `$field->disabled = true;` on each of them. On top of that, I need the toolbar presented by the relationController's record listing view to not include the modification buttons (create, delete) based on the value of a field attached to the main controller view; and as far as I can tell, there isn't a way to set that dynamically through `filterFields()`. 

I apologize for any confusion in my last response, the way I worded it made it sound like I was trying to modify the fields that appear in the relationController's manage view with this new event. What I meant to say was that I am already using `filterFields()` in addition to the new event. I was coming across limitations of `filterFields()` (namely, I can't set the `readOnly` property of the relationController), and that is why I'm proposing adding this event in. 

Are there any concerns over the addition of this event? I tried to follow the naming convention I've seen in the repo so far for events but if you have any suggestions in that regard I would love to hear them

daftspunk:
No concerns, just need to be able to justify it with a simple use case that has no solution already. If you wouldn't mind, please submit a Pull Request to the [test plugin for investigation](https://github.com/daftspunk/oc-test-plugin) that demonstrates the issue. Don't forget to include step by step instructions how how to replicate it *using the test plugin*.

LukeTowers:
@daftspunk so you'd like me to submit a pull request to the test plugin that utilizes this event?

LukeTowers:
Whoops, pressed the wrong button for commenting :)

daftspunk:
Just demonstrate a functional test that demonstrates what you want to happen, that isn't happening. Don't need a demonstration of the solution at this stage.

LukeTowers:
@daftspunk added the pull request to the test plugin (daftspunk/oc-test-plugin#18), let me know if you need anything else

daftspunk:
Wow, quite a long journey here to discover out what you are wanting to do.

#### Issue

I would like to transfer the `disabled` property of a form field to the Relation Controller, when rendered in as a `partial` type.

#### Prototype code

Form field definition

    comments:
        label: Comments
        type: partial
        disabled: true

Form field partial - `comments.htm`

```
/* I want to make this render in read only mode if the field is disabled */
<?= $this->relationRender('comments') ?>
```

daftspunk:
#### Solution

Form field partials - `comments.htm`

```
<?= $this->relationRender('comments', ['readOnly' => $field->disabled]) ?>
```

LukeTowers:
@daftspunk This doesn't seem to change the relation form that pops up, while the checkboxes and toolbar buttons for the relation list widget have been disabled; your solution only takes effect on initial render of the relationController list widget and thus does not effect the AJAX events that occur for the displaying of the update form for related records. Are we able to revisit adding this event into october? 

Additional info: octobercms/october#2294

daftspunk:
Looks like this is a bug in the code, the read only mode should persist across these requests. We need this feature to exist (and work!) because it is a common requirement.

We can explore the event again if there is a use case for it. Events are tricky because they are easy and cheap to create, but they are tricky to remove or modify and can end up polluting the code too much.

LukeTowers:
@daftspunk that's true. I didn't realize it was going to be as easy a fix as @gabsource created! Looking forward to that getting merged in

PR #2246, opened 2016-07-26
Enable refresh on form controller requests

LukeTowers:
This adds support for a redirect to the current page (a refresh) to be returned by the form controller. An example use case is adding `data-request-data="refresh:1"` to the save button on an update form. I'm using it for an update form that disables certain fields (including a relationRender partial) based on values submitted on that page. Refreshing the view after saving allows my disabling rules to take effect.


daftspunk:
LGTM, wrong branch though.

LukeTowers:
Whoops, I'll resubmit to the correct branch. Thanks for catching that!

PR #2180, opened 2016-07-06
asset cache fix

christianWilling:
Hi,
this changes will make it possible to use a `HashableInterface` for the cache generation.

So you can make a hash with the modification dates from the included files.

And than changes in the included files will be displayed without saving the parent asset.

PR #2248, opened 2016-07-27
Fix asset cache

christianWilling:
see https://github.com/octobercms/october/pull/2180 for more infos.


daftspunk:
I can't accept this with the 5 spaces used on the `prepareCombiner` method. Please fix.

christianWilling:
now it is fixed

PR #2250, opened 2016-07-27
Reverse proxy support

JoelESvensson:
This one is a bit tricky. In order for October to handle reverse proxies correctly then this package needs to be added to October. https://github.com/octobercms/library also needs a new global middleware to handle it. I'm not sure if this is the best way to do it, but it works. I'm open to suggestions for how to implement this change in a better way.


JoelESvensson:
Closing this in favour of a plugin-space solution. See https://github.com/octobercms/library/pull/226

PR #2260, opened 2016-07-30
Fix the background for flex-layout panel like in RainLab.Builder

gabsource:
The custom background transparency CSS was not set for panels like in Builder plugin.
Did not dig to see if it applies elsewhere too.

image


daftspunk:
Managed to get this working in 42a702f70a035555c994faae84c54964be3a00d3

PR #2261, opened 2016-07-30
fix hardcoded sort order name

cesarzagonel:
Use getSortOrderColumn from Sortable trait to determinate what is the name of the sort order column.


acasar:
@cesarzagonel @daftspunk This PR broke reordering with ``NestedTree``.

```
Call to undefined method October\Rain\Database\QueryBuilder::getSortOrderColumn()
```

daftspunk:
Thanks, reverted in #2273

PR #2267, opened 2016-07-31
Added ROMANIAN_MAP to modules/system/assets/ui/js/input.preset.js

cdarken:
Used for generating correct slugs in form widgets of type slug when using Romanian diacritics in the source field.


cdarken:
Hi. This is not working well. I see my commit in the subtree here: https://github.com/octoberrain/system/tree/master/assets/ui but the storm-min.js file is not updated, and I understood that I only have to modify the sources, the compiling of the assets  is done automatically.
Sorry if I should have commented elsewhere, but here seems more appropriate.
Thanks.

daftspunk:
The compiling isn't really handled automatically, more like in batches to make changes easier to read. I have recompiled them for you in 0b0e1580094f6ac19818b869b138a7e974115e3d.

PR #2277, opened 2016-08-02
Add scrolling to JS TreeView when dragging sortable elements

jimcottrell:
##### Expected behavior
When interacting with UI using core's JS TreeView (nothing in core that I'm aware of, but Pages, Sitemap, possibly other plugins), when I drag an element, I expect the list to scroll if necessary so I can place my element where I want it to be.

##### Actual behavior
The list does not scroll, requiring dragging, dropping, scrolling, dragging again, etc. if the element's desired placement is significantly away from its existing location.

##### Reproduce steps
- Install Pages plugin
- Create a few dummy pages to at least partially fill up the sidebar
- Shrink the browser height (if necessary) to where not all pages can be displayed in the sidebar without scrolling
- Drag a page entry past the visible list but where elements exist above or below what's visible.

#### Implementation
With the TreeView/Scrollbar/Sortable mix, it's not an entirely straightforward issue, so I understand if there's a better way to do this, or if maybe you'd like to do something like this at a different layer so that it's closer to the Sortable level, benefiting more than just TreeView.  I saw the DragScroll plugin, which I had hoped would be what I needed, but that seemed closer to Scrollbar functionality.

I chose to access Scrollbar and Sortable objects directly for this initial pull request.  I saw at least one other instance where that had been done elsewhere, but I'd understand if that's frowned on.  Particularly updating Sortable's cursor adjustment from TreeView could be seen as breaking scope.  If you'd rather access functionality through the public interface of the jQuery plugin, I believe there will need to be edits to both Scrollbar and Sortable plugin functions, as there are a couple of issues with context and argument lists.

Let me know if there's anything I should do to make this a better fit.  Hopefully this code ends up being helpful in some form.

### DIFF
--- a/modules/backend/assets/js/october.treeview.js
+++ b/modules/backend/assets/js/october.treeview.js
@@ -71,6 +71,7 @@
 
     TreeView.prototype.dispose = function() {
         this.unregisterHandlers()
+        this.clearScrollTimeout()
 
         this.options = null
         this.$el.removeData('oc.treeView')
@@ -210,6 +211,7 @@
             useAnimation: false,
             usePlaceholderClone: true,
             handle: 'span.drag-handle',
+            onDrag: this.proxy(this.onDrag),
             tolerance: -20 // Give 20px of carry between containers
         })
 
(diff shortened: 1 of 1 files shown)


daftspunk:
You are a star! Thanks heaps

daftspunk:
Hey @jimcottrell, LTNS 😁 

Took this a step further by supporting the mouse scroll wheel event too. So when you click and drag, you can also scroll and it moves the placeholder with it.

Here is the code change if you'd like to review it ed2eec6afb97141003c443586e6370d379e2dc5c

jimcottrell:
@daftspunk That's a much better experience when scrolling while dragging.

You forced me to remember how my code works, so that's fun. I've looked through ed2eec6, and it looks good to me. Fairly straightforward once I refreshed my memory, so I'm all for it.

Great to see you more active on October.

PR #2285, opened 2016-08-05
Added instantiation of relationships if they do not exist on save

tschallacka:
When a model has a belongsTo relationship, and a formfield is named

`SomeModelName[relationshipname][field]` and the relationship does not exist on the model the following warning appears:

"Call to a member function hasRelation() on null" on line 74 of F:\xxxxxxxxx\modules\backend\traits\FormModelSaver.php"

This code hopes to address that non existing relation ship and initialise it.

### DIFF
--- a/modules/backend/traits/FormModelSaver.php
+++ b/modules/backend/traits/FormModelSaver.php
@@ -37,6 +37,25 @@ trait FormModelSaver
     }
 
     /**
+     * Initialise a singular relation object if it doesn't exist yet.
+     * @param Model $model
+     * @param string $attribute
+     */
+    protected function initSingularRelationObject($model, $attribute) {
+
+        $relation = $model->makeRelation($attribute);
+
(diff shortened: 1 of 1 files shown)


tschallacka:
Okay... I simply realised you cannot make belongsTo relationships without saving them first, and them have them automagically add their id if they are new.

For that I need to modify the BelongsTo and Model have to be modified to check all belongsTo relationships, and if the ->exists but key is not set, that the key gets updated...

All I wanted was a simple adding of relationships... guess it's going to be a bit more complicated. 
@daftspunk should I further investigate to solve this foreign key not getting transposed? I have a solution in mind, but if you are against it, I won't pursue it.

daftspunk:
This solution duplicates the existing behavior of model relationships (setSimpleValue). Before you do anything we need a functional test that demonstrates the issue you want to fix.

PR #2296, opened 2016-08-08
Add model defined constraints to the List query for the relation view list

### LINKED ISSUE #2292: Relation Controller view (list) mode and scope constraint
##### Expected behavior
The output of `<?= $this->relationRender('relation_with_scope') ?>` should constraints the visible records to the scope definition. Sorting and searching (if enabled) should be done only on the subset of these elements.

##### Actual behavior
All the values from the table are displayed.

Forum related post [https://octobercms.com/forum/post/problem-with-scope](https://octobercms.com/forum/post/problem-with-scope)

##### Reproduce steps
1. Add a relation with a scope on a model
2. Add a RelationController behavior to the backend model controller and render the relation widget for the relation inside the form.

##### October build
355

=====

### Possible fix

Changing [line 637](https://github.com/octobercms/october/blob/develop/modules/backend/behaviors/RelationController.php#L637) of `RelationController` from : 

`$this->relationObject->addConstraints();`

to : 

`$this->relationObject->addDefinedConstraintsToQuery($query);`

seems to fix the problem. Search and sorting are OK. 

The same principle has been applied to the manage mode (list). Is there any reason to not apply the same logic for the view (list) mode ?

gabsource:
It fixes #2292 

I can't figure out if there are side effects.

### DIFF
--- a/modules/backend/behaviors/RelationController.php
+++ b/modules/backend/behaviors/RelationController.php
@@ -634,7 +634,7 @@ class RelationController extends ControllerBehavior
                     $this->relationObject->withDeferred($sessionKey);
                 }
                 elseif ($this->model->exists) {
-                    $this->relationObject->addConstraints();
+                    $this->relationObject->addDefinedConstraintsToQuery($query);
                 }
 
                 /*


daftspunk:
Superseded by 618a03561643b4c24f99bff9838d9ef197bb37dc

PR #2299, opened 2016-08-08
Default value for datepicker

serega-kasyanow:
Default value for datepicker widget in strtotime() compatible format.


modules/backend/formwidgets/DatePicker.php
```diff
@@ -89,7 +96,11 @@ public function prepareVars()
 
 
         if ($value = $this->getLoadValue()) {
-            $value = $value instanceof Carbon ? $value->toDateTimeString() : $value;
+            if ($value instanceof Carbon) {
+                $value = $value->toDateTimeString();
+            } elseif ($value == $this->default) {
+                $value = Carbon::createFromTimestamp(strtotime($this->default));
```
daftspunk: Consider using instead:
```
$value = Carbon::parse($this->default);
```

modules/backend/formwidgets/DatePicker.php
```diff
@@ -89,7 +96,11 @@ public function prepareVars()
 
 
         if ($value = $this->getLoadValue()) {
-            $value = $value instanceof Carbon ? $value->toDateTimeString() : $value;
+            if ($value instanceof Carbon) {
+                $value = $value->toDateTimeString();
+            } elseif ($value == $this->default) {
```
daftspunk: What is the rationale behind the logic `$value == $default`?

serega-kasyanow: $value == $default when field is empty

daftspunk: This doesn't make much sense because this expression contradicts the actual feature. You want to specify a default, right?

```
default: "tomorrow + 3 hours 15 minutes"
```

Then

```
if (null == "tomorrow + 3 hours 15 minutes")
```

How will the value ever be applied?

daftspunk:
Please confirm that you have tested this code.

modules/backend/formwidgets/DatePicker.php
daftspunk: It would appear `$this->default` is not even a valid property, it will always return null...

daftspunk:
All sorted now, thanks @serega-kasyanow 

Refs 1c3631bede881e1e60b4422b091f228bc156ac2d

PR #2308, opened 2016-08-13
Fix readOnly initialization for RelationController when using options

gabsource:
The proposed PR, injects the `readOnly` property inside the original configuration for the given relation field. The `initRelation` method will clone this config and read the readOnly property from this config.

The `makeReadOnly` is called at the end of the `initRelation` method, using the post value and defaulting to the config value.

====

While testing after #2295 has been merged I figured out that the `readOnly` option is set to `true` after the widgets are rendered (so the links are rendered with `readOnly = false` ).
So the first relation rendered is always editable (links in the list open the form) and all the others are readOnly (links open the preview).

```
 public function relationRender($field, $options = [])
    {
        $field = $this->validateField($field);           // THIS RENDERS THE WIDGETS
    ...
        /*
         * Read only mode
         */
        if (isset($options['readOnly']) && $options['readOnly']) {
            $this->makeReadOnly();                         // THIS SETS THE READONLY
        }
    }
```

It requires to set the `$this->readOnly = true` before calling the widgets render method while `makeReadOnly` must still be called after the widgets are rendered (for the toolbar and view widgets to be available).

There is also a `makeReadOnly` called inside the `initRelation` to detect `$_POST` vars. Maybe it can all be centralized in the `initRelation` method.

Also the `readOnly` attribute is set in the RelationController instance so shared across all the widgets for the edited model. This implies that with the current version we can't make some of the widgets "not read only" after one has been set to read only using the `$options`. With this PR the readOnly config is read per widget and not globally.

Maybe there is a nicer way to inject the value into the original config.

### DIFF
--- a/modules/backend/behaviors/RelationController.php
+++ b/modules/backend/behaviors/RelationController.php
@@ -380,6 +380,13 @@ class RelationController extends ControllerBehavior
      */
     public function relationRender($field, $options = [])
     {
+        /*
+         * Read only mode
+         */
+        if (isset($options['readOnly']) && $options['readOnly'] && property_exists($this->originalConfig, $field)) {
+            $this->originalConfig->{$field}['readOnly'] = true;
+        }
+
         $field = $this->validateField($field);
 
         if (is_string($options)) {
@@ -391,13 +398,6 @@ class RelationController extends ControllerBehavior
          */
         if (isset($options['sessionKey'])) {
             $this->sessionKey = $options['sessionKey'];
-        }
-
-        /*
-         * Read only mode
-         */
-        if (isset($options['readOnly']) && $options['readOnly']) {
-            $this->makeReadOnly();
         }
 
         $this->prepareVars();
@@ -603,7 +603,7 @@ class RelationController extends ControllerBehavior
                 $this->relationModel->getKeyName(),
                 $this->field,
                 $this->relationGetSessionKey(),
-                $this->readOnly ? 1 : 0
+                $this->getConfig('readOnly') ? 1 : 0
             );
 
             if ($config->recordUrl) {


daftspunk:
Ah yes I see the problem. I think it will never be possible to overwrite the config/originalConfig because everything has already initialized by the time `relationRender` is called. So read only mode must be applied post configuration, post initialization. In other words, we need to rewrite `recordOnClick` option on the list widget inside the `makeReadOnly` method. 

Perhaps the logic that defines `$defaultOnClick` can be moved to a helper method, then called twice. Once at init and again when read only is activated.

LukeTowers:
Is there a way to dynamically control the value of readOnly via the server instead of passing it through the client? If it's being passed through the client, then the client can always change any configuration values that are being passed, which doesn't seem secure to me. I'd love to hear your input on this @daftspunk

daftspunk:
You can define an override method in your controller

    /**
     * Provides an opportunity to manipulate the field configuration.
     * @param object $config
     * @param string $field
     */
    public function relationExtendConfig($config, $field)
    {
    }

LukeTowers:
@daftspunk Am I able to access the parent model within this method? My current setup involves looking at an attribute on the model to determine if the relation controller needs to be read only, which I need access to the parent model of the relation in question in order to do so. If not, should I submit a pull request to add that functionality?

daftspunk:
@LukeTowers Good point. Done @ 535dc78db190ca299281745fcaab46c7d3d07eee

daftspunk:
Extra notes

```
8:42 AM <spunky> Usually the model is available via the widget, I've always found this kind of awkward
8:42 AM <spunky> So I'm tossing up between passing the model as the 3rd arg
8:42 AM <spunky> Or introducing a relationGetModel() method, akin to formGetModel() in FormController
8:43 AM <spunky> formGetModel is reliable, it will always return a single model (1 form / 1 controller)
8:43 AM <spunky> relationGetModel, not so much, it will return whatever model is being used in that context (N relations / 1 controller)
8:43 AM <luketowers> Yeah, I was just thinking that, which is why it'd be nice to pass it as a third argument
8:44 AM <spunky> formGetModel() is normally used inside an AJAX handler or page action, so I'm not sure I can see the use for relationGetModel()
8:44 AM <spunky> Another vote for 3rd arg, sweet
```

PR #2312, opened 2016-08-13
Fixed double escaping

verlok:
This double escaping, due to the escape filter + the <pre> tag, was
visualizing escaped HTML in the main demo page. Fixed it.


verlok:
What branch do I need to use?

verlok:
Develop, I guess :) Will do

daftspunk:
Yes `develop` please

PR #2320, opened 2016-08-16
Reorder partial should use the model key name attribute to display the ID

gabsource:
##### Expected behavior
Reorder controller should use the model primary key attribute to fill the records IDs.

##### Actual behavior
Reorder controller uses a hardcoded `->id` reference in it's partial.

##### Reproduce steps
Set the primary key name of a model used by a controller with `ReorderController` behavior to something different than `id`.

```
public $primaryKey = 'uuid';
```

##### October build
develop


daftspunk:
LGTM! Thanks

PR #2328, opened 2016-08-23
Fixed caching concurrency issues

MarcoROG:
After a series of tests we found out that the issue was related to two concurrent thread writing in the same resource.
In order to avoid such problem, the destination file is used as a file lock, so that only one thread at a time can enter the problematic region of code.
Different files lock different resources, so that regenerating a specific page doesn't slow down the access to others.
The presence of the resource in the global cache is checked both before and after the locking (double-checked locking) in case the cache has been rebuilt by a concurrent thread while the lock was being acquired.

We extrapolated the file rebuilding logic to the "rebuild" function.
We also optimized the process by avoiding full file regeneration if the cache is stale but the file hasn't been modified.
In that case we only lookup the class name with a regex (lines 114 to 123) and refresh the cache.

The caching is handled in the following way (if a check is negative then the next line is executed):
Check if local cache entry exists.
Check if global cache entry exists.
[Acquire lock]
Check if global cache entry exists.

At this point the rebuild / refresh takes place:
If a cached file doesn't exist or the original file has been modified : rebuild
If a cached file exists, the original file is untouched, and only the cache is stale: refresh

The solution has been tested for a couple of days, running both PHP 5.6 and 7.0 and using file cache.

### DIFF
--- a/modules/cms/classes/CodeParser.php
+++ b/modules/cms/classes/CodeParser.php
@@ -1,6 +1,7 @@
 <?php namespace Cms\Classes;
 
 use File;
+use Log;
 use Lang;
 use Cache;
 use Config;
@@ -41,7 +42,7 @@ class CodeParser
     public function __construct(CmsCompoundObject $object)
     {
         $this->object = $object;
-        $this->filePath = $object->getFullPath();
+        $this->filePath = $object->getFilePath();
     }
 
     /**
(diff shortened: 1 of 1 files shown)


daftspunk:
Thanks for your insights here, they were very useful in finding the solution.

PR #2334, opened 2016-08-28
Update the Hungarian translate

gergo85:


daftspunk:
So many commits! No worries, I will squash them

Regarding the changes to file `modules/backend/formwidgets/richeditor/assets/vendor/froala/js/languages/hu.js`, you should consider contacting Froala with these changes to ensure they are not lost.

Issue #2337, opened 2016-08-29
Access the App name & description settings within Twig


LukeTowers:
I would like to be able to access the App Name and Tagline settings within my theme's Twig. As it stands right now, I have to add 

```
$this->page['app_name']        = BrandSetting::get('app_name');
$this->page['app_description'] = BrandSetting::get('app_tagline');
```

to a custom component that I must include for every layout in order to be able to access these variables within my theme. I believe it would be helpful if they were accessible to the themes through Twig by default, similar to the other general variables available under the `this` object. Perhaps something like `{{ this.app.name }}` and `{{ this.app.tagline }}`?

daftspunk:
There is performance considerations here. The `$this->vars['this']` object called in CmsController is providing access to objects that we have already initialized anyway so there is no additional overhead.

If we add a `BrandSettings::get` call, it will force every front-end page to initialize the BrandSettings model, which is exclusively used in the backend at the moment.

Consider creating a utility component that injects the values you need, or alternatively [extending twig](http://octobercms.com/docs/plugin/registration#extending-twig)] to introduce the functionality as a filter or function.

Some other options that we can consider:

1) Adding a type called `variable` to `registerMarkupTags` that would allow you to register custom variables to the `this` object. Eg to make available `{{ this.app_name }}` use

```
public function registerMarkupTags()
{
    return [
        'variables' => [
            'app_name => BrandSetting::get('app_name')
        ],
        [...]
    ];
}
```

2) Seeding the `$vars` property with the `View` facade shared variables

```
$this->vars = (array) View::getShared();
```

Allowing you to share any variable across the entire front end, as we can do with Content and Mail Templates already. To make `{{ app_name }}` available use

```
View::share('app_name', BrandSetting::get('app_name'));
```

3) Both

LukeTowers:
I know I could always just continue to have a utility component / plugin that injects those variables, however I was thinking that it's a reasonably common use case to want to use those values in themes for OctoberCMS. 

A parallel to those settings would be the site name and tagline used in WordPress sites, utilized by themes in creating title tags for pages and copyright notices, as well as just general branding of the theme specific to the site its being used on. 

This way, themes wouldn't have to essentially duplicate these options in their own settings pages, they could just use the common values that would already be assigned for the backend.

What are your thoughts on that?

seanthepottingshed:
+1

daftspunk:
[...]

PR #2338, opened 2016-08-29
Add this.app.name & this.app.tagline to the default theme variables

LukeTowers:
Issue: #2337

This enables theme authors to access the name and tagline of the app as set in the backend settings within their themes, allowing for a more standardized way of accessing these settings within themes.


daftspunk:
NACK See #2337

PR #2339, opened 2016-08-29
The sorting error in preview context

AlexEsaulov:
The error "A widget with class name ':name' has not been bound to the controller" is fixed in preview context


daftspunk:
As per the contribution guidelines, please provide the exact steps you are taking for the issue to occur. This will be used for peer review.

PR #2355, opened 2016-09-07
Add type checking to resolveModelAttribute()

### LINKED ISSUE #2354: Backend\Classes\FormField->resolveModelAttribute() does not support associative array attributes
### Expected behavior
Fields that require getting dynamic options (checklist, radio, dropdown) should support being a part of an attribute's associative array just like more simple fields do (i.e. `type: text`).

### Actual behavior
`Backend\Classes\FormField->resolveModelAttribute()` (called by `Backend\Widgets\Form->getOptionsFromModel()` for fields that require dynamic options (`type: dropdown | radio | checklist`)) does not implement type checking with its iterative 'drill down' method of resolving the specified attribute. As a result of that, `Trying to get property of non-object` is thrown when the current implementation attempts to drill down through an attribute that is an associative array.

### Reproduce steps
Create a model that has a JSON attribute that is capable of storing an associative array of data. 

**MIGRATION:**
```
$table->json('options')->nullable();
```

**MODEL:**
```php
protected $jsonable = ['options'];

public function getLevel2Options($keyValue)
{
    return [
        'option1' => 'Option 1',
        'option2' => 'Option 2',
    ];
}
```

**FIELD CONFIG:**
```yaml
options[level1][level2]
    label: 'Test field'
    type: dropdown
```
Now attempt to load the page containing that form, October will throw an exception that you're `Trying to get property of non-object` in `\Backend\Classes\FormField->resolveModelAttribute()`. This occurs because in order to resolve the model attribute, the code is attempting to 'drill down' through multiple levels of model attributes when the field name is an array.

However, because this array represents an associative array stored within a JSONable attribute, and not a series of relations, this code fails on its assumption that each sub level of the attributes that it goes through will contain a Model class, and attempts to access the nonexistent property of the associative array instead of attempting to access the child element of that array.

##### October build
dev-develop (Latest commit as of this issue)

LukeTowers:
Fixes #2354

### DIFF
--- a/modules/backend/classes/FormField.php
+++ b/modules/backend/classes/FormField.php
@@ -605,9 +605,20 @@ class FormField
 
         $parts = is_array($attribute) ? $attribute : HtmlHelper::nameToArray($attribute);
         $last = array_pop($parts);
+        $currentAttr = $model;
 
         foreach ($parts as $part) {
-            $model = $model->{$part};
+            
+            if (is_object($currentAttr)) {
+                $currentAttr = $currentAttr->{$part};
+                if (is_object($currentAttr)) {
+                    $model = $currentAttr;
+                }
+            }
+            elseif (is_array($currentAttr)) {
+                $currentAttr = $currentAttr[$part];
+            }
+            
         }
 
         return [$model, $last];


daftspunk:
When an attribute resolves to an array, theoretically it is no longer and attribute. So this method currently functions as it should.

PR #2364, opened 2016-09-13
Filter lists by switch or group query scopes

alvaro-canepa:
Add the ability to filter by multiples scopes using `group` filter or switching between two scopes, using `switch` filter.

```yaml

    # EXAMPLE 1
    # Just choose by multiple scopes at a time
    status:
        label: Status
        type: group
        scope: :filtered
        options:
            pending: Pending
            active: Active
            closed: Closed

    # EXAMPLE 2
    # On this case, a Blog post, can be published on Facebook has a related model to store the Facebook link and published date
    # The method scopeIsFbPublished run $query->has('related_model') and scopeIsNotFbPublished run the opposite $query->doesntHave('related_model') 
    fb_published:
        label: Published on Facebook
        type: switch
        scope:
            - isFbPublished
            - isNotFbPublished
```

Sorry my english.

### DIFF
--- a/modules/backend/widgets/Filter.php
+++ b/modules/backend/widgets/Filter.php
@@ -616,6 +616,20 @@ class Filter extends WidgetBase
                  * Scope
                  */
                 elseif ($scopeMethod = $scope->scope) {
+                    /*
+                     * Switch scope: multiple scopes
+                     */
+                    if(is_array($scopeMethod)) {
+                        $scopeNum = is_array($value) ? 0 : $value - 1;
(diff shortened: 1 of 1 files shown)


daftspunk:
This is not required. The filtered value is passed to the scope, so a parent "controller" can be used to apply child scopes:

```
public function scopeProxyFilter($query, $filtered)
{
    if (in_array('pending', $filtered) {
        $this->scopeSomething($query);
    }
   // ...
}
```

alvaro-canepa:
Thanks, I don't know that!

PR #2392, opened 2016-09-27
Remember current page in Lists

adrenth:
Originally the current page was not stored in session.

Situation:
- Navigation to page 3
- Edit a record by clicking row
- Press 'Cancel'
- List displays page 1

New situation:
- Navigation to page 3
- Edit a record by clicking row
- Press 'Cancel'
- List displays page 3


daftspunk:
When we looked at this last time, the page needs to be reset on the first entrance. i.e

Navigation to page 3
Edit a record by clicking row
Press 'Cancel'
List displays page 3
<the next day>
Navigate to list
List displays page 3 on entrance (BAD)

adrenth:
I see that "List displays page 3 on entrance (BAD)" is really a bad point. I didn't thought about that! Session lifetime can be variable in each different project.

When using sessions, there must be one trigger that resets the session for that specific ListView instance.

I think it is really a challenge to fix this isse with URL parameters; since you can apply a filter or search query on the list view which results in a 1 page list for example.

When I got some time, I'll look into some more different solutions for this issue.

daftspunk:
We discussed this internally, the best approach is to leverage the URL to store the current page. This is the most natural and obvious UX. It also keeps the native browser back/forward button behavior.

Since the list is AJAX'ified, it might be best to use the `#hash` in the URL, as opposed to a `?query` string which could be reserved for some other use, which would create a logic conflict.

seanthepottingshed:
Yes please!!!

daftspunk:
This solution is not accepted. We need a solution that stores the current page inside the URL.

PR #2399, opened 2016-09-29
Fix filter search function under postgres

CSNWEB:
Like in some other places before, the columns have to be casted to text under postgres.


daftspunk:
This should probably be baked in to the [searchWhere internals](https://github.com/octobercms/library/blob/762bb0a048d103db4d659d3749a02ea4877ba48f/src/Database/Builder.php#L71) instead.

LukeTowers:
This will be handled by octobercms/library#231

PR #2400, opened 2016-09-30
Hide pagination, when the number of records is zero

gergo85:


modules/backend/widgets/lists/partials/_list_pagination.htm
```diff
@@ -1,16 +1,18 @@
...
+        <?php endif ?>
         <?php if ($pageCurrent > 1): ?>
-            <a
-                href="javascript:;"
-                class="page-back"
-                data-request="<?= $this->getEventHandler('onPaginate') ?>"
-                data-request-data="page: <?= $pageCurrent-1 ?>"
-                data-load-indicator="<?= e(trans('backend::lang.list.loading')) ?>"
-                title="<?= e(trans('backend::lang.list.prev_page')) ?>"></a>
+        <a
```
LukeTowers: Why'd you remove the indentation here?

gergo85: Because it has same indentation, than the Next link.

PR #2403, opened 2016-10-02
Hiding empty block, if there are no buttons

PopcornPHP:


daftspunk:
It is possible for a toolbar to have search with no buttons. This would cause it to break?

PopcornPHP:
If no buttons - displays an empty block above the table. This change - solely visual effect. I doubt that this can break system

daftspunk:
Is this relation controller? We can't accept this PR because it will break the toolbar search (no buttons + search).

It's difficult to discuss a solution without first understanding the problem. Please, explain the problem itself and not the solution that you have in mind. For more info on presenting your problem see http://mywiki.wooledge.org/XyProblem. Thanks!

daftspunk:
If using `ListController` you can set this in the config_list.yaml and will do the same thing:

```
toolbar: null
```

If you are using `RelationController` you can use a similar solution:

```
myrelation:
    view:
        showSearch: false
        toolbarButtons: false
```

PR #2407, opened 2016-10-04
fix - spaces in |media filter

nerijunior:
This PR is a fix for spaces in `|media` url.

Steps to reproduce:
In a layout put some inline style like this:

``` html
<div style="background: url({{ var | media }})"></div>
```

If the path contains a space the image was not showed. 

This simple fix change spaces by `%20` and everything works.


LukeTowers:
Shouldn't this Twig filter be running paths through `urlencode()` to eliminate any of these sorts of issues?

daftspunk:
Unable to replicate this. Works fine:

```
{% set var = 'lol whut.jpg' %}
<div style="width: 100px; height: 100px; background: url('{{ var|media }}')"></div>
```

Issue #2417, opened 2016-10-11
AJAX framework request's content type causing issues with JSON request data


LukeTowers:
##### Expected behavior
Data sent to the server via the AJAX framework gets properly decoded as the correct types.

##### Actual behavior
When sending JSON to the server via the AJAX framework (specifically within the `data-request-data` attribute), the request is made with a `Content-Type` of `application/x-www-form-urlencoded`. This causes issues when the contents of the the `data-request-data` attribute being sent is a JSON representation of data as the server does not properly decode this data upon receipt. For an example of this in practice, look at [this recent commit for extending the relationController](https://github.com/octobercms/october/commit/0d336151d2e83e25e6738d07d550f2102dc173cb#diff-9a1dc597b1fb67808fd720ea4e688068R3). The value of `$relationExtraConfig` is properly encoded as JSON in the `data-request-data` attribute, but when the server receives that AJAX request, it is being decoded as if it was `application/x-www-form-urlencoded`. 

The biggest issue with this comes when passing booleans with a value of false. These booleans get properly represented as **(bool)** `false` in the JSON, however the server receives them as **(string)** `'false'`. This is an issue of course due to how PHP evaluates strings as booleans. The string `'false'` evaluates as a `true` boolean, which leads to the problem with the relationController.

##### Reproduce steps
Create a partial that will render a relation and include the extra configuration value of `readOnly` in the call to `$this->relationRender()` set to the value of `$field->disabled`.
```
<?= $this->relationRender('myRelation', ['readOnly' => $field->disabled]) ?>
```
This will cause the relation controller to be rendered with 
```
data-request-data="_relation_field: 'fields', _relation_extra_config: {"readOnly":false}"
```
Now, when you click on a related record to open its manage form, the form will be in preview mode (read only mode), because `$relationController->readOnly` is equal to the **string** `'false'`, and that value is then passed on to the form widget being rendered as its `previewMode` property. Now, when every system field partial gets rendered, it checks `if ($this->previewMode)` before rendering the preview / read only version of itself. However, that will evaluate to `true` because `$this->previewMode` is equal to the **(string)** `'false'` which in PHP evaluates as `true`. 

##### October build
Running off of branch `develop`, latest commits were:
october/cms (dev-develop 6f4ae6a)
october/system (dev-develop a9cc524)
october/backend (dev-develop bbbef87)

daftspunk:
[...]

PR #2418, opened 2016-10-11
Support readOnly extra config as string

LukeTowers:
Supports the readOnly extra config value being provided as a string of 'false' due to incorrect content encoding passed via the AJAX framework. Temporary solution in relation to octobercms/october#2417


daftspunk:
Fixed by #2417 (8eafcaa090a05b7499cce90d8a9614151060c35c)

PR #2423, opened 2016-10-14
Cache flush problem

FrancescoPassanti:
I noticed that when you login in the backend, this function delete all the cache stored. I was trying to create a plugin to store in Redis all the visited pages. After the login in the backend i saw that all element in the cache was gone.


daftspunk:
This will only occur when there are pending updates, not every time, Clearing the cache is necessary as part of the update process and should not be relied upon for data storage.

As per the contribution guidelines, please provide the exact steps you are taking for the issue to occur. This will be used for peer review.

PR #2426, opened 2016-10-17
Set the path of theme preview image manually

gergo85:


daftspunk:
Documented as 

**previewImage** | custom preview image, path relative to the theme directory, eg: `assets/images/preview.png`, optional.

Issue #2450, opened 2016-10-26
RFE: Empty option for backend dropdown 


leoquijano:
Hi,

I'm building a backend form, like this:

```
  interval_type:
      label: Foo
      type: dropdown
      showSearch: false
      span: left
```

It would be useful to have an option to specify an empty option:

```
      emptyOption: Select a Foo
```

This is similar to what Laravel does, and to what the frontend does. Importantly, it should save NULL as the model value for that field, if the field is not specified. This is important for numeric dropdowns (such as an enum), since adding "" => "something" is not the same, and can bring a database error.

Is this feasible?

mplodowski:
Try to use `placeholder` property.

leoquijano:
Thanks, that works. Maybe the documentation could use an example in the ```dropdown``` section. Even though the ```placeholder``` property is mentioned in the general field options, it is not clear it can be used in the ```dropdown``` field, since HTML5 doesn't support placeholders for SELECTs.

leoquijano:
There's an additional issue with the usage of the ```placeholder``` property. Typically, a lot of dropdowns like this have integer keys. For example, consider the following possible values for a "status" dropdown: 

```php
[ 
  1 => "Pending",
  2 => "Processed",
  3 => "Completed"
]
```

But if I pick the placeholder only, the SELECT will return an empty string ''. A database, like PostgreSQL, won't accept that for a numeric value:

```
"SQLSTATE[22P02]: Invalid text representation: 7 ERROR: invalid input syntax for integer: "" (SQL: insert into "plan_types" ("recurring", "code", "name", "description", "currency_code", "amount", "recurring_amount", "interval_type", "interval_count", "updated_at", "created_at") values (0, test_plan, Test Plan, Single $50 fee, USD, 500, , , 1, 2016-10-28 18:42:40, 2016-10-28 18:42:40) returning "code")" 
```

So I'm stuck when using the placeholder property. Note that FormController doesn't allow me an easy way to filter values, since the ```update_onSave``` and ```create_onSave``` methods are a bit convoluted.

It would be nice to make dropdowns default to NULL when the placeholder is selected. Something like this in ```modules/backend/widgets/Form.php``` line 976:

```php
/*
 * Number fields should be converted to integers
 */
if ($field->type === 'number') {
    $value = !strlen(trim($value)) ? null : (float) $value;
}
/*
 * Empty dropdown fields should be converted to NULL
 */
elseif ($field->type === 'dropdown') {
[...]

PR #2459, opened 2016-11-04
Fix for #2450

Patroklo:
1. Added the placeholder behavior of the Select2 library into october's dropdown and relation widgets.
2. Added `allowClear` option into dropdown and relation widgets.
Now it's possible to add the `allowClear` option into relation and dropdown, adding the Select2 clear behavior into those widgets. Since select2 library forces to have a defined placeholder, also added a default placeholder text in case there is not an user defined one.

How to use it:

```yaml
        Foo:
            label: Mofu!
            type: dropdown
            placeholder: My little placeholder text
            span: left
            allowClear: true
            options:
              1: Value 1
              2: Value 2
```


daftspunk:
Thanks for this. Note the "allowClear" option was removed, see 4aca2d40ff5b2c125e41823e1d3a8a3ba43b5d43 for details.

PR #2460, opened 2016-11-05
Typo in RelationController.php

alexeydemin:


LukeTowers:
Going with #2635 instead due to lack of response from @alexeydemin

PR #2467, opened 2016-11-07
Translate the Page link popup

gergo85:
and update the Hungarian translation


daftspunk:
Should consider submitting the changes back to Froala as these will most likely be lost next time it upgrades.

gergo85:
Thank you for your advice. I sent the updated translation to Froala's author.

PR #2469, opened 2016-11-08
Update Filter.php

joseph-d:
There is an addScopes function for extensibility but no way to remove scopes which are already there.

### DIFF
--- a/modules/backend/widgets/Filter.php
+++ b/modules/backend/widgets/Filter.php
@@ -477,6 +477,17 @@ class Filter extends WidgetBase
             $this->allScopes[$name] = $scopeObj;
         }
     }
+    
+    /**
+     * Programatically remove a scope, used for extensibility.
+     * @param string $scopeName Scope name
+     */
+    public function removeScope($scopeName)
+    {
+        if (isset($this->allScopes[$scopeName])) {
+            unset($this->allScopes[$scopeName]);
+        }
+    }
 
     /**
      * Creates a filter scope object from name and configuration.


daftspunk:
This could be problematic, we need to decide if it is better to "hide" scopes instead of destroying them. Can you give your use case for needing this? It will help with making a decision.

joseph-d:
I have been extending the Rainlab Blogs plugin. As part of what I'm doing, I wanted to remove (or hide) the categories feature as it is not required.

I was able to do the following:

        Event::listen('backend.menu.extendItems', function($manager) {
           $manager->removeSideMenuItem('RainLab.Blog', 'blog','categories');
        });

        PostsController::extendListColumns(function($list) {
            $list->removeColumn('categories');
        });

        PostsController::extendFormFields(function($form) {
            $form->removeField('categories');
        });

The only thing remaining is the category filter in on the posts list view which there doesn't appear to be a method for.

To be honest, hiding the scope would be just as good as destroying the scope in this particular case, just as hiding the above column, field and side menu item would be acceptable too.

daftspunk:
You cited precedence (Lists / Form widget have removeX methods), that's all I need to see. This will be accepted.

LukeTowers:
@daftspunk The requested changes have been made, could you merge this?

LukeTowers:
@daftspunk The requested changes have been made, could you merge this?

daftspunk:
Strange, I wasn't notified about the base branch change. Thanks for picking it up @LukeTowers!

PR #2487, opened 2016-11-14
Added new configuration for the text in switch form field

SebastiaanKloos:
You can now add “onText” and “offText” to your switch field YAML
config to change the default on and off state text.


daftspunk:
This is a slippery slope, the switch can only support a few characters in the CSS. Can you give an example of what you might update this to?

SebastiaanKloos:
For a question with possible answers Yes and No I would like to use Yes and No and not On and Off. 

On another form in the same plugin I would like to use the On and Off text. So it would be really nice to make this configurable on field level.

daftspunk:
This PR looks quite broken, please submit and make reference to this PR #2487.

PR #2491, opened 2016-11-14
Fix list searching with additional scope

alxy:
Just becaue the ``searchScope`` is present, it shouldnt ignore the usual searching implementation at all. This fixes the mentionned issue.

### DIFF
--- a/modules/backend/widgets/Lists.php
+++ b/modules/backend/widgets/Lists.php
@@ -1185,10 +1185,9 @@ class Lists extends WidgetBase
                 $q->$scopeMethod($term);
             });
         }
-        else {
-            $searchMethod = $boolean == 'and' ? 'searchWhere' : 'orSearchWhere';
-            $query->$searchMethod($term, $columns, $this->searchMode);
-        }
+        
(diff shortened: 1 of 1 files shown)


daftspunk:
Need more information on this one. I think the original intent was for the search scope to supplement the search query.

Can you provide more detail about how you are using this and how it came about.

LukeTowers:
Closed as per discussion in IRC:
```
[16:20:07] +alxy:	If you define a custom search scope
[16:20:21] +alxy:	the default search features of october no longer work
[16:20:29] +alxy:	(searching all searchable columns)
[16:20:39] +alxy:	I have discussed this with spunky already and we fixed it
[16:20:47] +alxy:	the $term is now passed to the scope
[16:21:14] spunky:	So he can close it right?
[16:21:49] +alxy:	right
[16:21:58] spunky:	^^
[16:23:45] +alxy:	but still I think it would be nicer to apply both, the default behavior and the scope :p as it is really some kind of pain to get it working the way it works in october if you have many searchable columns and searchable relations
[16:23:53] +alxy:	however, I agree this is breaking
```

PR #2497, opened 2016-11-16
Allow combine scss files from command.

gpasztor87:


LukeTowers:
@daftspunk This cleans up some docblocks and adds support for compiling scss via command (support for compiling less via command already exists). Looks good to me

PR #2501, opened 2016-11-20
Support Bootstrap field error state

dshoreman:
Toggles the `has-error` class on the surrounding `.form-group` when a field has errors, allowing us to style the actual input as well as the error text.

### DIFF
--- a/modules/system/assets/js/framework.extras.js
+++ b/modules/system/assets/js/framework.extras.js
@@ -41,6 +41,8 @@
             $container = $('[data-validate-error]', $this),
             messages = [],
             $field
+        
+        $('.form-group.has-error', $this).removeClass('has-error')
 
         $.each(fields, function(fieldName, fieldMessages) {
             $field = $('[data-validate-for='+fieldName+']', $this)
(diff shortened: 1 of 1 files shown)


daftspunk:
While Bootstrap is the reference framework for the front-end, this assumption is too low level to be included in the framework.js, which needs to remain completely neutral.

For example, [Semantic UI uses `.grouped.fields.error` as its error classes](https://github.com/daftspunk/dry-ui/blob/master/js/form-tools.js#L8) and so we would need a way to support this as well.

The current implementation is simple enough to get by on, anything more complicated should probably use some simple JavaScript to make it happen. Eg:

```
$(window).on('ajaxInvalidField', function(event, fieldElement, fieldName, errorMsg, isFirst) {
    $(fieldElement).closest('.form-group').addClass('has-error');
});

$(document).on('ajaxPromise', '[data-request]', function() {
    $(this).closest('form').find('.form-group.has-error').removeClass('has-error');
});
```

Thoughts?

LukeTowers:
@daftspunk We should close this but add a note to the documentation about using such a method to accomplish the original goal of the pull request. Where do you think it'd be best located within the docs?

PR #2507, opened 2016-11-24
Use the otherKey to update a relation

oliverpool:
If the `otherKey` of the relationship was set, the widget was considering the wrong column name.

## To reproduce:
Create a model with a custom `otherKey`
```php
public $belongsTo = [
    'library' => ['Library', 'key' => 'library_id', 'otherKey' => 'national_id']
];
```
Add a widget with this relation in some backend form.

Look in the database after edition: the `library_id` took the `id` of the Library (and not the `national_id`)

### DIFF
--- a/modules/backend/formwidgets/Relation.php
+++ b/modules/backend/formwidgets/Relation.php
@@ -140,8 +140,8 @@ class Relation extends FormWidgetBase
             }
 
             $field->options = $usesTree
-                ? $result->listsNested($nameFrom, $relationModel->getKeyName())
-                : $result->lists($nameFrom, $relationModel->getKeyName());
+                ? $result->listsNested($nameFrom, $relationObject->getOtherKey())
+                : $result->lists($nameFrom, $relationObject->getOtherKey());
 
             return $field;
         });


daftspunk:
This is a real shame because it is a nice improvement, however Eloquent uses different methods

#### getOtherKey

- belongsTo
- belongsToMany
- morphToMany
- morphedByMany

#### getLocalKey

- hasMany
- hasOne

This code will cause hasMany/hasOne to break

```
Call to undefined method October\Rain\Database\QueryBuilder::getOtherKey()
```

daftspunk:
October calls it "otherKey" so we should have a method to match. This has been added to the library in https://github.com/octobercms/library/commit/afa1985ed282b3bddad054c625d0184b4f6a7e19 so this PR can be accepted. Thanks!

oliverpool:
There is a [similar call](https://github.com/octobercms/october/pull/2507/files#diff-94995028565920ead02101783020b5acR134) a couple of lines higher. 
```php
$selectColumn = $usesTree ? '*' : $relationModel->getKeyName(); // change to getOtherKey() ?
```
Since I don't know what it is used for, I preferred not to edit it.

daftspunk:
I think it is much of a muchness, if we change that code or leave it.

PR #2509, opened 2016-11-24
Add support for hasManyThrough in the RelationController

### LINKED ISSUE #2508: RelationController does not support hasManyThrough
##### Expected behavior
The relation controller should support the hasManyThrough relation as it does with the hasMany relation.

##### Actual behavior
Attempting to setup a hasManyThrough relation to use the relation controller will result in the error  `Call to render() on null $relationViewWidget` from `relationcontroller/partials/_view.htm`. This is caused because `$this->viewWidget` in the RelationController is set to null due to the `evalViewMode()` not returning a valid view mode for relationships of type hasManyThrough which further causes the `makeViewWidget` method to return null as well.

##### Reproduce steps
Setup a hasManyThrough relationship on your model, and then set it up to use the relation controller using standard practice. Render the relationcontroller through `$this->relationRender('name_of_your_hasManyThrough_relationship')` in a partial field and then you will see the error occur. As a side note, do not use the name of the relationship for the name of the field or else you will run into a separate issue, that of the `HasManyThrough` not having a `getSimpleValue` method.

##### October build
october dev-develop

LukeTowers:
Fixes #2508 by setting the view mode for `hasManyThrough` relationships to `multi` just like that of `hasMany` relationships.

### DIFF
--- a/modules/backend/behaviors/RelationController.php
+++ b/modules/backend/behaviors/RelationController.php
@@ -1402,6 +1402,7 @@ class RelationController extends ControllerBehavior
             case 'morphToMany':
             case 'morphedByMany':
             case 'belongsToMany':
+            case 'hasManyThrough':
                 return 'multi';
 
             case 'hasOne':


daftspunk:
Looks good, however it will need a usage example in the [test plugin](https://github.com/daftspunk/oc-test-plugin) to prove it works.

LukeTowers:
After some initial testing with just my use case, it appears that it works fine for listing the records, however it fails on creating / updating the related records. I'll do some further testing and development to see if I can add full support for hasManyThrough relations.

daftspunk:
I'm guessing it fails due to the simple value getter/setter not being available in the relation class. We'll need a test scenario to exist to remedy this.

LukeTowers:
In this case it's erroring in `onRelationManageCreate()` specifically when it calls `add()` on the `relationObject` as HasManyThrough relation objects don't have an add method being more of a readable relationship instead of a read/write relationship

robinbonnes:
Can we merge this in?

LukeTowers:
@robinbonnes no, unless you can propose a solution to https://github.com/octobercms/october/pull/2509#issuecomment-262842858

robinbonnes:
For my use case I also only need to show the data. I hotfixed it in my plugin by implementing a custom RelationController.

```
class CustomRelationController extends \Backend\Behaviors\RelationController
{
    protected $forceViewMode = 'multi';
	
    public function relationMakePartial($partial, $params = [])
    {
        $prefix = base_path('modules/backend/behaviors/relationcontroller/partials/');
        $contents = $this->controller->makePartial($prefix . 'relation_'.$partial, $params + $this->vars, false);
        if (!$contents) {
            $contents = $this->makePartial($prefix . $partial, $params);
        }

        return $contents;
    }
}

```

chocolata:
Hi @robinbonnes thanks for this. Could you elaborate a bit on how to use this to get a hasManyThrough relationship working?

robinbonnes:
Hey @maartenmachiels,

I use this to show the related `hasManyThrough` models as a list in the preview context of the parent form. 
For my specific case I extended the `RainLab.User` plugin to show related models. Added the following in my plugin to get it working:

/plugins/your/plugin/behaviors/CustomRelationController.php:
```
<?php namespace Your\Plugin\Behaviors;

class CustomRelationController extends \Backend\Behaviors\RelationController
{
    // Add multi support.
    protected $forceViewMode = 'multi';
    
    public function __construct($controller)
    {
        parent::__construct($controller);
        
        // Fix correct asset paths.
        $this->assetPath = '/modules/backend/behaviors/relationcontroller/assets';
        $this->viewPath = base_path().'/modules/backend/behaviors/relationcontroller/partials';
    }
}
```

/plugins/your/plugin/controllers/users/config_relation.yaml:
```
related:
    label: Related models
    readOnly: true
    view:
        showSearch: false
        toolbarButtons: []
        list: $/your/plugin/models/related/columns.yaml
        recordsPerPage: 20
```

/plugins/your/plugin/controllers/user/_related.htm:
```
<?= $this->relationRender('related') ?>
```

/plugins/your/plugin/plugin.php:
```
public function boot()
{
        // Extend model to add relation.
        \RainLab\User\Models\User::extend(
            function ($model) {
                
                $model->hasManyThrough['related'] = [
                    'Your\Plugin\Models\Related',
                    'key'        => 'user_id',
                    'through'    => 'Your\Plugin\Models\Parent',
                    'throughKey' => 'parent_id'
                ];
            }
        );
        
        // Extend controller to add custom relation controller and override the relation config.
        \RainLab\User\Controllers\Users::extend(
            function ($controller) {
            
                $customRelationController = 'Your\Plugin\Behaviors\CustomRelationController';
                if (is_array($controller->implement) && !in_array($customRelationController, $controller->implement)) {
                    $controller->implement[] = $customRelationController;
                }
                
                $controller->relationConfig  =  '~/plugins/your/plugin/controllers/users/config_relation.yaml';
            }
        );
        
        // Extend form fields to display related models in a tab.
        Event::listen(
            'backend.form.extendFields',
            function ($widget) {
				
                if (!$widget->model instanceof User || $widget->isNested) {
                    return;
                }

                $widget->addFields([
                        'user_related' => [
                            'context' => 'preview',
                            'tab' => 'Related',
                            'type' => 'partial',
                            'path' => '~/plugins/your/plugin/controllers/users/_related.htm'
                        ]
                    ],
                    'primary'
                );
            }
        );
}        
```

chocolata:
Hi @robinbonnes , thanks so much for explaining this in so much detail. I'm trying to display a list of bought products per user (and the reverse - have a list of users that bought a specific product) in the OFFLINE Mall plugin.

So following your logic, I did the following in my boot method (and I did follow your other instructions).

image


```
        // Extend model to add relation.
        \OFFLINE\Mall\Models\Product::extend(
            function ($model) {

                $model->hasManyThrough['related'] = [
                    'RainLab\User\Models\User',
                    'key'        => 'id',
                    'through'    => 'OFFLINE\Mall\Models\Order',
                    'throughKey' => 'customer_id'
                ];
            }
        );

        // Extend controller to add custom relation controller and override the relation config.
        \OFFLINE\Mall\Controllers\Products::extend(
            function ($controller) {

                $customRelationController = 'Chocolata\MallEvents\Behaviors\CustomRelationController';
                if (is_array($controller->implement) && !in_array($customRelationController, $controller->implement)) {
                    $controller->implement[] = $customRelationController;
                }

                $controller->relationConfig  =  '~/plugins/chocolata/mallevents/controllers/users/config_relation.yaml';
            }
        );

        // Extend form fields to display related models in a tab.
        \Event::listen(
            'backend.form.extendFields',
            function ($widget) {

                if (!$widget->model instanceof \OFFLINE\Mall\Models\Product || $widget->isNested) {
                    return;
                }

                $widget->addFields([
                    'user_related' => [
                        'context' => 'preview',
                        'tab' => 'Related',
                        'type' => 'partial',
                        'path' => '~/plugins/chocolata/mallevents/controllers/users/_related.htm'
                    ]
                ],
                    'primary'
                );
            }
        );
```

Now I'm getting this weird error message which gives me the impression that our custom relation controller isn't being fired...

```
Relation behavior does not contain a definition for 'brand'.
E:\xxx\htdocs\xxx\modules\backend\behaviors\RelationController.php line 340
```

Do you maybe have an idea what might be wrong here?

robinbonnes:
Probably you have to merge the original relation config with the new one, could you try replacing:

```
$controller->relationConfig  =  '~/plugins/chocolata/mallevents/controllers/users/config_relation.yaml';
```

By:

```
$controller->relationConfig = $controller->mergeConfig(
    $controller->relationConfig,
     '~/plugins/chocolata/mallevents/controllers/users/config_relation.yaml'
);
```

chocolata:
Hi @robinbonnes thank you so much for your answer... I almost got it working. I saw that the path didn't seem to work, so I changed it to the following (note the dollar sign):

```
                $controller->relationConfig = $controller->mergeConfig(
                    $controller->relationConfig,
                    '$/chocolata/mallevents/controllers/purchases/config_relation.yaml'
                );
```
Now I'm getting very very close, as I'm now seeing the following error message:

```
Missing configuration for view.list in RelationController definition brand
E:\Dropbox\Projects\_xampp\htdocs\xxx\modules\backend\behaviors\RelationController.php line 1767
```

Do you have the final answer :-) ? Thanks again for all your insights. I managed to solve multiple issues already because of your intervention.

daftspunk:
>  I keep getting the following error now:
> "Call to undefined method October\Rain\Database\QueryBuilder::getSimpleValue()" on line 2483 of E:\Dropbox\Projects\_xampp\htdocs\xxx\vendor\laravel\framework\src\Illuminate\Database\Query\Builder.php 

Sorry to interrupt

> I'm guessing it fails due to the simple value getter/setter not being available in the relation class.

@maartenmachiels you've arrived at the point where the implementation is non-existent. Are you using v2 by any chance? We could look at introducing native support for it

chocolata:
Hi @daftspunk - thanks for your interjection. I'm not yet using v2, but am very willing to (I bought a bronze subscription). It would be very nice if there would be native support for the hasManyThrough relationship, even if it would be limited to a simple non-editable list view. Do you think that support for this feature would be feasible in the short term, or do you feel I'd better find a workaround?

Strange though that the custom relation controller won't "take" in this situation. Do you have an idea why it doesn't work @robinbonnes  ?

daftspunk:
Ok, we will add it to the list of things to add when RelationController gains attention. Although keep in mind, you would need your site to be running v2 to benefit from the latest features.

daftspunk:
> limited to a simple non-editable list view

Yes, this relationship does appear to be a read-only type. Combing through the relation class, there is no clear method for associating a relationship and this makes sense because it involves touching multiple models.

Anyhow, support is added for this in v2.1.5

Thanks @maartenmachiels + @robinbonnes

hauzmann:
Can I display hasManyThrough relation count value in backend list by setting useRelationCount: true?

daftspunk:
In theory, setting `relationCount: true` should work, yes.

Issue #2125, opened 2016-06-15
UI bug on Forgot your password page


gergo85:
In English, the layout is right:
![ui-en](https://cloud.githubusercontent.com/assets/2959112/16073628/2c76f83a-32e7-11e6-9fac-50d945243c3e.jpg)

If the text of button is too long:
![ui-hu](https://cloud.githubusercontent.com/assets/2959112/16073669/5c1f4858-32e7-11e6-8bcb-b7878cd54890.jpg)

daftspunk:
Fixed by #2513 

This will be available in Build 385+

PR #2513, opened 2016-11-27
Using flexbox on login and reset page.

gpasztor87:
The reset button is not always displayed in a row with the input field. 
Flexbox is solves this problem.


daftspunk:
Hey @gpasztor87. Thanks for this. Would you mind sharing some screenshots that demonstrate a comparison of the old VS. new proposed interface?

gpasztor87:
Of course :)

Before:
before-hungary
before-deutsch

After:
after-hungary
after-deutsch

daftspunk:
This fixes #2125... oh yeah!

Great job!

Issue #1235, opened 2015-06-24
Make partial parameters accessable in the PHP section of the partial template


PMK:
Hi,

Along with other developers I'm unable to access the parameters of a partial. These parameters are not accessible inside the PHP section of the template of the partial, but are in the HTML.

```
{% partial "my-partial-name" paramName="value" %}
```
- https://octobercms.com/forum/post/accessing-partial-variables-in-onstart
- https://octobercms.com/forum/post/acces-partial-parameter-in-code-tab

In class Cms\Classes\Controller.php in function renderPartial() I do see the parameters with the values, including the name of the partial.

If I make an "else" state on line number 825, and putting the partial stuff inside the $this->vars array, then I can access it.

``` PHP
if (($partial = Partial::loadCached($this->theme, $name)) === null) {
    // ... existing code here ...

} else {
    // Obviously this is not correct...
    $this->vars[$name] = $parameters;
}
```

I think this should be done inside the Controller class via a getter, but if you paste this code here below inside your partial template in the PHP part, it will fix the problem.

``` PHP
$controller = $this->getController();
$partialParams = $controller->vars["your-partial-name-here"];
// var_dump($partialParams);
```

If you have access to the parameters in the PHP section, you can load your own class to enrich the partial with extra functionality. I am not sure if this is a bug or a "new feature". I did assume that this was possible because I was able to use the parameters in the HTML section.

nsrosenqvist:
+1

triasrahman:
+1

estudiogenius:
+1

vebam:
+1

merodeador:
+1

danilobrinu:
+1

fresswolf:
+1

snipiba:
+1

Asinox:
And... any news about this?

Beaudinn:
+1

cracki:
+1

Kilgaloon:
I worked a little bit on this issue and enabled parameters passed to partials to be available through method ``getParam( $param )`` and to get all parameters with ``getParams()``

daftspunk:
Being looked at as part of #2517

daftspunk:
This was fixed in Feb as part of ec03b7b2f85cbe89082eda9fe7e457534bdc522d

PR #2517, opened 2016-11-29
Make partial parameters accessable in the PHP section of the partial template #1235

Kilgaloon:
Enabled parameters passed to partial to be used in Code segment of partial


daftspunk:
This looks good, can you provide an example of how it is used.

Kilgaloon:
When you include partial in template like this ``{% partial "some" testParam=testValue %}`` you can get parameter inside code block of partial like this ``$this->getParam('testParam')`` or you can get all parameters with  ``$this->getParams()``. Also will change code like said.

daftspunk:
Thanks, this relates to #1235. The method `$this->param()` will access the URL parameters currently, so this method would create some confusion. 

We'll need to find a more suitable way. I would expect that the values are available simply with `$this`. (`$this->controller->vars`) or some other approach.

Kilgaloon:
Will look at this in next few days.

Kilgaloon:
@daftspunk Instead of adding parameter that is passed to `$this->vars["this"]["param"]`` i think its good idea to add them ``$this->vars["this"]["partial"]["param"]``. And get them like previously set in pull request. I wanted to make parameters passed to partial available in ``$this`` scope, but better is to separete partial from anything else and make section partial inside ``$this->vars["this"]`` so other things can be set for partial and that can be accessed.

daftspunk:
@Kilgaloon, it turns out the variables are available. It's possible this was implemented recently through general maintenance but #1235 was never closed. Am I missing something here?

#### Partial contents

```
==
function onStart()
{
    traceLog($this->foo);
}
==
{{ foo }}
```

#### Calling the partial

```
{% partial 'mypartial' foo='bar' %}
```

daftspunk:
This was fixed in Feb as part of ec03b7b2f85cbe89082eda9fe7e457534bdc522d

PR #2519, opened 2016-11-29
Added new event listener

SebastiaanKloos:
It would be really nice to manipulate the partials before they render. I added the cms.partial.beforeRender event.


LukeTowers:
Adding myself to update my event lists if this gets merged

SebastiaanKloos:
I like your list. I hope this will get through. And maybe your list could be added to the docs.

LukeTowers:
At some point I'll make a PR to the docs for it, once I have more time for that sort of thing :)

SebastiaanKloos:
@daftspunk Is it possible that this will be included in the next update?

daftspunk:
This is an appropriate event, however it lacks consistency with `cms.page.renderContent` and `cms.page.beforeRenderContent` in both naming and signature. Additionally, need to understand the "why" behind this. Can you share the code you are using to hook it so we can see how it's used.

SebastiaanKloos:
@daftspunk It looks like the we can not manipulate the partials before they are being rendered. When you want to change the content of the partial before they are rendered, we need a new event. I have tested all the other events but I was not able to change the content. 

I use this event to get the content of the partial. Then I can change the public content variable of the partial object so I can change the content before they are being rendered in the view.

daftspunk:
A code paste says a thousand words. (Paste your code)

SebastiaanKloos:
Event::listen('cms.partial.beforeRender', function($controller, $partial) {
	$partial->content = 'Hello World!';
});

Changing the content of the partial before they are being rendered. 

In my case I want to replace a few character combinations with regex.

PR #2562, opened 2016-12-15
Respect database tables config.

sw-double:
This does not alter default behaviour and allows proper configuration

### DIFF
--- a/modules/system/classes/UpdateManager.php
+++ b/modules/system/classes/UpdateManager.php
@@ -119,7 +119,7 @@ class UpdateManager
      */
     public function update()
     {
-        $firstUp = !Schema::hasTable(config('database.migrations'));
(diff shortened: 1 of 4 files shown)


LukeTowers:
@daftspunk Looks good to me. As a side note, should October introduce support for fully configurable system table names so that all of the table names (not just the few already supported) would support configured table names?

sw-double:
LGTM? Otherwise what's the point of having configuration options that are actually unusable without hacking the core?

Laravel supports them, I don't see why october should not.

LukeTowers:
Not sure if there's a bit of confusion here or not, just telling @daftspunk that your PR is easily mergeable.

I also added a comment about potential future improvements to core, which admittedly was poorly worded. Apologies for any confusion.

sw-double:
Ready

sw-double:
Fixed.

PR #2572, opened 2016-12-20
add meta_keywords

kideny:
add meta_keywords for seo

### DIFF
--- a/modules/cms/classes/Page.php
+++ b/modules/cms/classes/Page.php
@@ -28,6 +28,7 @@ class Page extends CmsCompoundObject
         'description',
         'is_hidden',
         'meta_title',
+        'meta_keywords',
         'meta_description',
         'markup',
         'settings',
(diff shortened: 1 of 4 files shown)


LukeTowers:
Just something to consider with this:

At best Google & Bing completely ignore the meta keywords tag, at worst they penalize you for including it.

http://searchengineland.com/the-meta-keywords-tag-lives-at-bing-why-only-spammers-should-use-it-96874
https://chrisedwards.me/seo/keyword-meta-tag-google/
http://www.hobo-web.co.uk/definitive-guide-to-using-important-meta-tags/#meta-keywords-seo-best-practice

kideny:
@LukeTowers meta_keywords tag have usefully for some seo tools ,  we should support meta_keywords setting  for some user and they can empty it .

kideny:
meta_keywords have usefully for some other country's search engine .  octoberCMS is a  international cms . we should let user choice  itself .

LukeTowers:
I don't disagree with you, just thought I should bring those articles up.

Honestly, in my opinion meta tags shouldn't have been part of the core in the first place; they're more within the realm of plugins.

daftspunk:
We intentionally didn't include `keywords` because it has become redundant in many schools of thought, including our own. This could easily be injected by a plugin so we won't need to merge this.

PR #2574, opened 2016-12-21
Keep change event intact

alxy:
This should allow to watch for the change event on a presetted field. Not tested yet.

### DIFF
--- a/modules/system/assets/ui/js/input.preset.js
+++ b/modules/system/assets/ui/js/input.preset.js
@@ -257,7 +257,7 @@
             $el.val(prefix + self.formatValue()).trigger('oc.inputPreset.afterUpdate')
         })
 
-        this.$el.on('change', function(event) {
+        this.$el.on('change', function() {
             self.cancelled = true
         })
     }
--- a/modules/system/assets/ui/js/input.preset.js
+++ b/modules/system/assets/ui/js/input.preset.js
@@ -247,19 +247,18 @@
         if ($el.val().length && $el.val() != prefix)
             return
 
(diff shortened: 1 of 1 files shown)


daftspunk:
I appreciate the honesty, but why submitting untested code? Please get back to us on this!

alxy:
@daftspunk This is now tested and works. However, it obviously behaves a bit different than the regular change event, as it triggers on every key stroke/character inserted. As opposed to the regular event, that triggers when the input loses focus. But I don't think this is a big problem, just notable.
This will be active with the next asset recompilation.

daftspunk:
I see. It is not accurate in this case, shouldn't the replicated event be called `keyup` instead of `change`?  Please give more detail when posting issues -- why is this even needed? Thanks!

alxy:
If for whatever reason you need to watch a preseted field for change and want to trigger some other action based on the data entered (or automatically entered), this is needed or only works with an ugly workaround. However, I can definetly change it for ``keyup`` if this will be accepted.

In my concrete case I tried to implement a SEO analyzer, that also analyzes the url/slug. This one was not easily possible to be tracked, once it is changed by ``input.preset.js``

daftspunk:
Native events probably isn't the best approach here, since the internals rely on them and it has potential for recursion.

[Using input.trigger as a basis](https://github.com/octobercms/october/blob/master/modules/system/assets/ui/js/input.trigger.js#L44) please create two new events

- `oc.inputPreset.update` - before the update occurs
- `oc.inputPreset.afterUpdate` - after the update occurs

PR #2575, opened 2016-12-21
Enhance usability of list pagination.

triasrahman:
Hi, I want to improve the usability of list pagination. We found the problem when the list has so many data. Our admins were frustrated to use this.

**Previous behavior**
image

Problems:
- No current page indicator
- Can't move to any page
- Not mobile friendly

**Proposed behavior**
image

Using select2:
image

Mobile friendly (centered):
image

The new behavior is user can easily navigate between pages and even jump to any page.

Hopefully it can make OctoberCMS backend better :)


LukeTowers:
This looks pretty nice! @daftspunk thoughts?

daftspunk:
Yes do it, please! This annoys me too.

PR #2582, opened 2016-12-27
Unescaped quote fix

tschallacka:
Fixes an unescaped quote in ukrain language file.


LukeTowers:
@daftspunk Is there any way that we could create unit tests for language files to validate that syntax issues like this don't happen again?

daftspunk:
@LukeTowers we have one already, not sure how this made it through: https://github.com/octobercms/october/blob/master/tests/unit/system/classes/CoreLangTest.php

tschallacka:
Well, maybe the locale isnt in the list of locales being tested? It seems to pull a list of locales at https://github.com/octobercms/october/blob/master/tests/unit/system/classes/CoreLangTest.php#L30 if its not in there, it wont get tested.

I only noticed it because eclipse threw a hissy fit

daftspunk:
So you're saying `uk` wasn't in the list?

tschallacka:
Well, i just looked at a possible explanation on mobile why uk is skipped in the unit test...

LukeTowers:
@tschallacka Why is that?

PR #2594, opened 2017-01-06
Improved `getConfigured` method logic

cx-emerge:
Improved `getConfigured` method logic

### DIFF
--- a/modules/backend/models/EditorSetting.php
+++ b/modules/backend/models/EditorSetting.php
@@ -115,7 +115,7 @@ class EditorSetting extends Model
     }
 
     /**
-     * Returns the value only if it differs from the default value.
+     * Returns the configured value.
      * @return mixed
      */
     public static function getConfigured($key, $default = null)
@@ -124,9 +124,9 @@ class EditorSetting extends Model
 
         $value = $instance->get($key);
 
-        $defaultValue = $instance->getDefaultValue($key);
+        $defaultValue = $default ?: $instance->getDefaultValue($key);
 
-        return $value != $defaultValue ? $value : $default;
+        return $value ?: $defaultValue;
     }
 
     public function getDefaultValue($attribute)


LukeTowers:
What improvements does this make besides switching around ternary operators? Code style changes without an actual benefit to the framework are unlikely to be accepted.

daftspunk:
Not sure why this is needed. There might be an error in the logic: If the value is 0 or empty string, it will return default.

cx-emerge:
```
return $value != $defaultValue ? $value : $default;
```
can't return `$defaultValue`, it will return null.
I think when `$default` is `null`, this function return `$defaultValue`.

LukeTowers:
@cx-emerge Yeah, that what it looks like to me as well. Do you have an example of this behaviour causing issues in projects though?

cx-emerge:
@LukeTowers 
```php
...
use Backend\Models\EditorSetting;
...
$htmlAllowTags = EditorSetting::getConfigured('html_allow_tags');
...
```

LukeTowers:
@cx-emerge Thanks, that's very helpful :) 

It makes sense to me now @daftspunk I'll test this sometime this afternoon and then I think this should be good to merge.

daftspunk:
Thanks @LukeTowers. 

This issue lacks enough detail or the replication instructions cannot be actioned easily. As per [the contribution guidelines](https://github.com/octobercms/october/blob/master/CONTRIBUTING.md#reporting-a-bug-with-octobercms), please provide the exact steps you are taking for the issue to occur. This will be used for peer review.

Here is an example of a **bug summary**:

> The CMS page URL is not being set automatically from the title value.

Here is an example of some **reproduce steps**:

> 1. In the back-end, open the CMS > Pages section
> 1. Click the "Add" button
> 1. Enter [Hello] for the Title field
> 1. Click the Save button 

Here is an example of **expected behavior**:

> - The Title should be set to [Hello]
> - The URL should be set to [/hello]

Here is an example of **actual behavior**:

> - The Title is set to [Hello] correctly
> - The URL is left empty

cx-emerge:
Here is an example of a bug summary:
> `EditorSetting::getConfigured` can't return the correct value

Here is an example of some reproduce steps:
```php
...
use Backend\Models\EditorSetting;
...
$htmlAllowTags = EditorSetting::getConfigured('html_allow_tags');
...
```
> will return null.

Here is an example of expected behavior:
> The `$htmlAllowTags` should be `$defaultHtmlAllowTags` value

daftspunk:
As above.

LukeTowers:
@cx-emerge Please do not make additional comments as edits to previous comments, they **will** be overlooked and ignored.

cx-emerge:
Here is an example of a bug summary:
> `EditorSetting::getConfigured` can't return the correct value

Here is an example of some reproduce steps:
In [AnandPatel.WysiwygEditors](https://github.com/anand-patel/oc-wysiwyg-editor) plugin
`formwidgets/Editor.php`
```php
...
use Backend\Models\EditorSetting;
...
public function prepareVars()
{
...
        $this->vars['valid_children'] =
            '+body[' .
                str_replace([',', ' '], ['|', ''],
                    EditorSetting::getConfigured('html_allow_tags')
                ) .
            ']' .
            ',-body[' .
                str_replace([',', ' '], ['|', ''],
                    EditorSetting::getConfigured('html_remove_tags')
                ) .
            ']'
        ;
}
...
```
`formwidgets/editor/partials/_tinymce.htm`
```html
...
function tinymceInit() {
...
    tinymce.init({
        ....
        valid_children: '<?= $valid_children ?>'
    });
...
}
...
```
> `$this->vars['valid_children']` will return null.

Here is an example of expected behavior:
> The `$this->vars['valid_children']` should be `$defaultHtmlAllowTags` value

LukeTowers:
@tschallacka You're going to have to provide an example of how you are using this code and how it causes you an problem. The mentioned plugin does not reference this class anymore, it talks to its own Settings model now: https://github.com/anand-patel/oc-wysiwyg-editors/blob/master/formwidgets/Editor.php

daftspunk:
@cx-emerge thanks for the example. The current behavior is intentional, it wants to provide the value *only if it differs from the default value*, as per the method comment. This is because Froala will use its own default values internally when the setting is NULL.

```
/**
 * Returns the value only if it differs from the default value.
 * @return mixed
 */
public static function getConfigured($key, $default = null)
```

In your case you want a value every time, so you should use code like this:

```
EditorSetting::getConfigured('html_remove_tags', EditorSetting::getDefaultValue('html_remove_tags'));
```

This will force a returned result every time.

cx-emerge:
@daftspunk thank, This provides me with some help

tschallacka:
@LukeTowers ehh.. what?

LukeTowers:
@tschallacka :face_palm: The auto @ failed me, sorry :)

PR #2596, opened 2017-01-06
Add keyboard shortcut to close tabs

matteotrubini:
Add "ctrl/cmd+q" keyboard shortcut to close the active tab.
Currently "ctrl+w" (it would be my first choice) cannot be used due to browser's reserved shortcuts, see https://bugs.chromium.org/p/chromium/issues/detail?id=119881

### DIFF
--- a/modules/system/assets/ui/js/tab.js
+++ b/modules/system/assets/ui/js/tab.js
@@ -109,6 +109,12 @@
             $(li).append($('<span class="tab-close"><i>&times;</i></span>').click(function(){
                 $(this).trigger('close.oc.tab')
                 return false
+            }).hotKey({
+                hotkey: 'ctrl+q, cmd+q',
+                callback: function(element){
+                    $(element).trigger('close.oc.tab');
+                    return false
+                }
             }))
         }
 
--- a/modules/system/assets/ui/js/tab.js
+++ b/modules/system/assets/ui/js/tab.js
@@ -110,7 +110,7 @@
                 $(this).trigger('close.oc.tab')
                 return false
             }).hotKey({
-                hotkey: 'ctrl+q, cmd+q',
+                hotkey: 'alt+w',
                 callback: function(element){
                     $(element).trigger('close.oc.tab');
                     return false


SebastiaanKloos:
This does not seem to be a good "shortcut" because cmd+q is a reserved command on the Mac.

matteotrubini:
Ouch! I didn't know it, I'm not a Mac user. Which could be an alternative?

sw-double:
I'd suggest using `ctrl+w` and `alt+w`

daftspunk:
This is a good idea, however some hotkeys are reserved by the browser and OS. For example `ctrl+w` will not work on Windows because it is reserved for closing a tab.

matteotrubini:
What do you think about middle click? I use it often to close browsers / editors tabs and using jQuery's event.which would be a breeze.

daftspunk:
Let's go with: `ALT+W` and middle click to close the tab.

ctf0:
we can implement the short cut in the same way the current save file works, as it overrides the browser keybindings.

for the windows/unix issue, we can detect it http://stackoverflow.com/questions/9514179/how-to-find-the-operating-system-version-using-javascript and add keybindings accorindgly

matteotrubini:
I'm sorry for the delay, I'm really busy in these weeks. I had made some test to add the middle click but I had not found any way to obtain the correct behavior without getting errors in console: in each case the event seems to be binded to all tabs, not only to the active one.
Actually I'm away from office, I can share the code as soon as I come back.

SebastiaanKloos:
@ctf0 cmd+w is a browsers shortkey I use very often...

ctf0:
@SebastiaanKloos yes i understand, but so the `cmd+s` to save the browser current page.

honestly there is no universal solution for this, u basicly want to have the same shortcut for the browser and the editor, and using the mouse to close the tab especially the opened tabs doesnt get closed unless u do it is a killer

SebastiaanKloos:
I personally think that this feature is better placed in a plugin.

PR #2601, opened 2017-01-10
Belarusian Localization

GinoPane:
Added Belarusian translations of backend, cms and system modules (all lines covered basically). Also a list of locales was changed a little, so locales are always written in their native language and the user can choose a language he actually knows and understands.


daftspunk:
Just concerned this is a destructive change. Is this standard practice to show locales like in the native tongue regardless of the active locale? Need evidence to support this.

LukeTowers:
@daftspunk Here's a couple of resources on the topic:

https://www.sitepoint.com/representing-language-on-the-web/
http://www.456bereastreet.com/archive/200604/indicating_language_choice_flags_text_both_neither/
http://ux.stackexchange.com/questions/11891/should-i-use-country-flags-in-language-selection-fields
http://ux.stackexchange.com/questions/2472/how-to-graphically-represent-a-language


Also, the BBC seem to represent each language choice in their native tongue, but then provide a list in English as well.
http://www.bbc.co.uk/ws/languages

Facebook displays language options in the native tongue of each language being present as an option:
https://www.facebook.com/

Overall, general consensus seems to be displaying the language option in the language that it represents; occasionally optionally prefixed / suffixed with the name of the language in the current language being used.

GinoPane:
I also kept in mind Youtube practices:

image

Gmail practices:
image


And even Windows practices (though they use duplication):

image

daftspunk:
Thank-you both. Let's do this!

mahony0:
Sorry for commenting to the merged PR. As you can see in these [old lines of the turkish lang file](https://github.com/octobercms/october/commit/fe4d593519b4e76838fbe89ee1f250a906bb33dc#diff-28cbb2c89919dcc69157bf4134174f80L8), what about prepending the english version of the language names? Yes, that's true native name is a good idea at all. But it seems pretty easy to understand languages by prepending english versions of them. Btw, what is the source of this native locales? For example for turkish language, it should be "Türkçe" for language name, not "Türk", and "Türk" means local people of turkey (idk other locales are all true)..

I'm not master about languages but maybe it should be checked before merging.

LukeTowers:
We're not masters of all languages either, so if anybody has any corrections to make to the locale names, we're open to it :) @mahony0

PR #2604, opened 2017-01-10
If CMS module is loaded (and MediaManager exists) bind it

dubcanada:
I am using OctoberCMS without the CMS module (removed from modules directory and turned off in cms.php) and I am getting an error in the backend regarding missing `MediaManager` class. I wrapped this in a class_exists to check that MediaManager exists before binding it.

Let me know your thoughts.

### DIFF
--- a/modules/backend/classes/Controller.php
+++ b/modules/backend/classes/Controller.php
@@ -153,8 +153,10 @@ class Controller extends Extendable
         /*
          * Media Manager widget is available on all back-end pages
          */
-        $manager = new MediaManager($this, 'ocmediamanager');
-        $manager->bindToController();
+        if (class_exists('\\Cms\\Widgets\\MediaManager')) {
+            $manager = new MediaManager($this, 'ocmediamanager');
+            $manager->bindToController();
+        }
     }
 
     /**


LukeTowers:
If we truly wish to support that use case of October, then maybe we need to consider whether the MediaManager widget being within the Cms module is the best idea in the first place. If it's used within both modules, then perhaps it's better to place it within the system module.
@daftspunk any thoughts on this?

dubcanada:
I don't see why you wouldn't want too support it, turning off the CMS module in cms.php and removing the entire CMS directory causes this error, and one more then you load plugins. But that's it. The rest works perfectly fine.

This would allow people to use it not only as a CMS, but as a strong bases for building applications.

LukeTowers:
Valid point, I honestly hadn't considered using October without the CMS (was still thinking about the backend as cms - which isn't what the cms module is - aspect of it); but now that I think about it that's not a bad idea for some projects.

What's the issue when loading plugins?

daftspunk:
The very definition of "module" means you should be able to remove any of them and it still functions.  We have plans on addressing this at some point, i.e

- CMS without Backend
- Backend without CMS

Seems like a step in the right direction.

@LukeTowers the media manager does have a minor identity crisis going on here. It probably should be moved to the System module, since functionality is shared. 

@dubcanada Would this be a problem for you? If Media Manager was moved to System, keeping in mind the System module cannot be disabled.

dubcanada:
Nope, I use the backend and system module. The one thing I am wondering is the media manager is directly related too the CMS, no? If it is related to the CMS why isn't it just part of the CMS?

Does it need to be initialized before the CMS is or something?

LukeTowers:
@dubcanada The media manager is used by a couple of different formwidgets (backend module) off the top of my head; and the CMS module uses it in its `|media` twig filter as a quick example of the CMS module using it. 

So like @daftspunk was saying, it probably should be moved to the System module instead since both the CMS and Backend modules rely on it for various functionality.

daftspunk:
Moving the MM to the system module represents a breaking change for [some](http://octobercms.com/plugin/manogi-mediathumb) [plugins](http://octobercms.com/plugin/nexxa-enhancedmedia). We can accept this as a architectural quirk for now.

PR #2605, opened 2017-01-11
If the relationModel is empty it still is "true"

dubcanada:
Currently the check is if

    !$this->relationModel

And all that does too check if it is true/false. That doesn't work for Collections as they are always true, regardless if they are full or not.

We need too check using `isEmpty()` or a similar function too determine if there are results.

### DIFF
--- a/modules/backend/formwidgets/RecordFinder.php
+++ b/modules/backend/formwidgets/RecordFinder.php
@@ -209,7 +209,7 @@ class RecordFinder extends FormWidgetBase
 
     public function getKeyValue()
     {
-        if (!$this->relationModel) {
+        if ($this->relationModel->isEmpty()) {
             return null;
         }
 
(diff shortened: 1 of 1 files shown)


dubcanada:
Basically this an error happens when the database is empty. It's rather easy to replicate, just create a plugin with a relation, two empty db tables, and have a form with a recordfinder widget on it. If you load the form and click on the recordfinder, it breaks (error attempting too access null value `$this->relationModel->{$this->keyFrom}` for example).

So either we check that `$this->relationModel->{$this->keyFrom}` exists or we check that `$this->relationModel` isn't empty. I assume the second way would be easier.

LukeTowers:
@daftspunk is probably going to ask for a PR made to the [test plugin](https://github.com/daftspunk/oc-test-plugin) for demonstration of the issue, so if you could do that for this issue, that'd be great. 

Did you have any thoughts on using `count($this->relationModel)` instead of `$this->relationModel-isEmpty()`?

daftspunk:
The record finder only supports "singular" relationship types, so it is a concern if a collection is being returned, it should result in further errors with the logic. @LukeTowers is correct, we will need a valid test case to fully understand the problem and implement a suitable fix.

PR #2606, opened 2017-01-11
Update lang.php

kubis:
typo. Error delete file is already in line 202. This is for folder (probably).


LukeTowers:
This was merged in #2625

Issue #2609, opened 2017-01-13
File Upload forms does not work with Relation Manager and $attachOne


lukaszbanas-extremecoding:
##### Expected behavior
When i upload file by relation manager, i expect that file will be uploaded to related model

##### Actual behavior
File is added to deferred_bindings table but not to model

##### Reproduce steps
main model has $hasOne binding to another model. Atother model has $attachOne relation to file.

##### October build
382

Suggested fix (modules/backend/behaviors/RelationController.php): 
```php
 public function onRelationManageUpdate()
    {
        $this->forceManageMode = 'form';
        $this->beforeAjax();
        $saveData = $this->manageWidget->getSaveData();

        if ($this->viewMode == 'multi') {
            $model = $this->relationModel->find($this->manageId);
            $modelsToSave = $this->prepareModelsToSave($model, $saveData);
            foreach ($modelsToSave as $modelToSave) {
                $modelToSave->save(null, $this->manageWidget->getSessionKey());
            }
        }
        elseif ($this->viewMode == 'single') {
            $this->viewWidget->setFormValues($saveData);
            $this->viewModel->save(null, $this->manageWidget->getSessionKey());
        }

        return ['#'.$this->relationGetId('view') => $this->relationRenderView()];
    }
```

lukaszbanas-extremecoding:
https://github.com/octobercms/october/pull/2610

LukeTowers:
@lukaszbanas-extremecoding Are you able to submit a PR to the [test plugin](https://github.com/daftspunk/oc-test-plugin/) for replication so that we can test your issue and your fix?

daftspunk:
Fixed by #2610, thanks @lukaszbanas-extremecoding!

PR #2610, opened 2017-01-13
Support deferred binding for single view modes in RelationController

lukaszbanas-extremecoding:
https://github.com/octobercms/october/issues/2609


daftspunk:
This makes sense, I would like to merge it -- however, I can't without a test case to prove something has been fixed.

PR #2617, opened 2017-01-19
Don't search relations with empty relationObjects

### LINKED ISSUE #2616: Improve error handling when attempting to search or sort by unsearchable / unsortable columns
#### Expected behavior
When performing a search on a list widget where a column with a relation defined; the search should proceed as normal. If the relation type is not supported, then it should fail gracefully with an appropriate message.

#### Actual behavior
Searches are unable to be performed on polymorphic relations, specifically the `morphTo` relation. This is because `\October\Rain\Database\Model->makeRelation($name)` returns null for the morphTo relation type. Unfortunately, because the List widget assumes that if a model hasRelation it will also return a valid value from makeRelation (which is not the case when it comes to `morphTo` relations), this causes it to throw an exception when it attempts to call `getTable()` on the null value that it has received from `Model->makeRelation()`. 

#### Reproduce steps
Create a model with a `morphTo` relation. In the columns.yaml for the model, specify a column that utilizes that relation, and then in the list view try searching for anything via the search box.

#### Discussion
@daftspunk Should this be resolved by silently ignoring `morphTo` relations in `Backed\Widgets\List->isColumnRelated()`, or should we provide an exception message to let developers know what's wrong?

LukeTowers:
Fixes octobercms/october#2616

### DIFF
--- a/modules/backend/widgets/Lists.php
+++ b/modules/backend/widgets/Lists.php
@@ -348,12 +348,15 @@ class Lists extends WidgetBase
                  * Related
                  */
                 if ($this->isColumnRelated($column)) {
-                    $table = $this->model->makeRelation($column->relation)->getTable();
-                    $columnName = isset($column->sqlSelect)
-                        ? DbDongle::raw($this->parseTableName($column->sqlSelect, $table))
-                        : $table . '.' . $column->valueFrom;
+                    $relationObject = $this->model->makeRelation($column->relation);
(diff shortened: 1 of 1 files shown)


daftspunk:
Blocked by discussion in #2616

LukeTowers:
Closing because I don't really run into this issue any more, if someone does and wants to come up with a better solution at some point then feel free to comment / open a new PR.

PR #2622, opened 2017-01-23
customViewPath in relationConfig

joebuonocore:
Only way I was able to edit the view partials in a relation was by adding `_relation_container.htm` or `_relation_view.htm` to my plugin controller, but that didn't provide the same ability as `customViewPath` or passed the data to view despite what I found in docs.

Couldn't use `customViewPath` within `relationConfig` as I was able to in `listConfig`, so I ended up adding the line from the ListController in backend behaviors to the RelationController.


daftspunk:
So this is a bug of some kind? We'll need to replicate that first to get a better understanding. 

As per [the contribution guidelines](https://github.com/octobercms/october/blob/master/CONTRIBUTING.md#reporting-a-bug-with-octobercms), please provide the exact steps you are taking for the issue to occur. This will be used for peer review.

LukeTowers:
Due to lack of response from @joebuonocore this has been closed.

PR #2628, opened 2017-01-24
Add events to Media Manager widget

Klaasie:
This pull request adds before and after events to the MediaManager widget actions (upload, move, rename & delete).

Adding these events will allow plugins to hook into them and act on the event.

For example:
If a reference to a media file is found in the a blog plugin, and the media file is moved or renamed, the blog plugin could listen to that event and change the reference in the database.

### DIFF
--- a/modules/cms/widgets/MediaManager.php
+++ b/modules/cms/widgets/MediaManager.php
@@ -240,12 +240,22 @@ class MediaManager extends WidgetBase
                 $filesToDelete[] = $pathInfo['path'];
             }
             else if ($pathInfo['type'] == 'folder') {
+
+                $this->fireSystemEvent('media.folder.beforeDelete', [$pathInfo['path']]);
+
                 $library->deleteFolder($pathInfo['path']);
+
+                $this->fireSystemEvent('media.folder.afterDelete', [$pathInfo['path']]);
             }
         }
 
         if (count($filesToDelete) > 0) {
+
+            $this->fireSystemEvent('media.files.beforeDelete', [$filesToDelete]);
+
             $library->deleteFiles($filesToDelete);
+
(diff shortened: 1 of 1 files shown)


LukeTowers:
Couple questions:
1. Do these events allow for the modification of the values passed to them or just the recording of them? If it's just recording and modifying them isn't supported, then I don't see why there should be before and after events of everything.
2. All current calls to `fireSystemEvent()` have their event id prefixed with either `backend` or `cms`. Currently, the MediaManager widget is within the `CMS` module, but should at some point in the future be moved to the `System` module with an alias left behind in the `CMS` module for backwards compatibility. This raises the question, should we prefix these events with `system.media` to indicate the (eventual) location or are we fine with leaving it prefixed with `media`

I'd also like to get @daftspunk's input on this

daftspunk:
Media should really be it's own module but then we'd be getting too granular, so the event prefix "media" should be fine.

daftspunk:
Note some events have been renamed to suit the [dev guide](http://octobercms.com/help/guidelines/developer) (no "after" events) and `media.files.delete` has been renamed to `media.file.delete`.

See 15209a957520860fd5cc989b12be06a94c9d314e

daftspunk:
> Do these events allow for the modification of the values passed to them or just the recording of them? 

These events *can* support modification of the values by passing them as references, eg:

```
$this->fireSystemEvent('media.folder.beforeDelete', [&$path]);
```

We can cross that bridge when we come to it. It is preferred to add code based on a real need, instead of trying to cover all bases. It's a good point, the "before" events do not serve the specified requirement 

> If a reference to a media file is found in the a blog plugin, and the media file is moved or renamed, the blog plugin could listen to that event and change the reference in the database.

Removed in 8b71385481cba11d1c96674c3e0f96c8f4baa81c

PR #2642, opened 2017-02-01
Update RelationController::onRelationManageUpdate

norotaro:
As in `onRelationManageCreate()`, `onRelationManageDelete()`, `onRelationManageAdd()`, etc, it needs to return `$this->relationRefresh()` for `relationExtendRefreshResults()` to be executed

### DIFF
--- a/modules/backend/behaviors/RelationController.php
+++ b/modules/backend/behaviors/RelationController.php
@@ -1061,7 +1061,7 @@ class RelationController extends ControllerBehavior
             $this->viewModel->save();
         }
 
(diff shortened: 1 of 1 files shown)


LukeTowers:
@daftspunk LGTM, unless you have input on why it wasn't using that method like the rest of the relation AJAX handlers.

daftspunk:
Looks good, except wrong branch.

LukeTowers:
Whoops, missed that. I blame just waking up :/

norotaro:
It's my fault (it's my first pull request) branch changed to develop

PR #2647, opened 2017-02-02
Added blacklist rules for nginx

jarek-odidigital:
The rules block access to the same folders as specified in .htaccess. To include them in nginx add the below line to your nginx config file:

` include /var/www/your-site/*.nginx;
`


LukeTowers:
I feel like this would be better located within the documentation instead of within the October project repository itself as nginx rules can't be applied via a directory level configuration like .htaccess can. I'd like to get @daftspunk's input on this though

jarek-odidigital:
They can be applied with `include` directive, not automatically like .htaccess though. I though it would be good to have it in the code like some other projects include it. I think it's easier to find it in the code than in the documentation.

daftspunk:
This needs to go here: https://github.com/octobercms/docs/blob/master/setup-configuration.md#nginx-configuration

LukeTowers:
On further review, the rules provided in the default configuration example in the docs are better as they hand control of denied routes to the application instead of denying at the server level.

```
    location / {
        try_files $uri /index.php$is_args$args;
    }

    rewrite ^themes/.*/(layouts|pages|partials)/.*.htm /index.php break;
    rewrite ^bootstrap/.* /index.php break;
    rewrite ^config/.* /index.php break;
    rewrite ^vendor/.* /index.php break;
    rewrite ^storage/cms/.* /index.php break;
    rewrite ^storage/logs/.* /index.php break;
    rewrite ^storage/framework/.* /index.php break;
    rewrite ^storage/temp/protected/.* /index.php break;
    rewrite ^storage/app/uploads/protected/.* /index.php break;
```

jarek-odidigital:
From performance point of view it's better to handle it at server level since nginx can easily handle **2000 requests/second** and October might not even be able to handle 200 but these rules are not great any way as they return empty 404 page.

LukeTowers:
True, but I can't envision a case where 2000 requests a second are coming in to those routes. Even a DDOS could always just make requests to another part of the application if these were being handled by nginx. 

Irregardless, I still think that using a public directory is the best solution anyways :)

daftspunk:
I feel like someone who is aware of this distinction is probably capable of creating a custom solution for themselves anyway.

jarek-odidigital:
Basically to make your website fast and efficient you want to handle as many requests as possible at load balancer, cache (Varnish, Squid etc.) or web server level and only pass through requests to application when necessary. Of course it's irrelevant for low traffic websites.

LukeTowers:
Again, I don't think that comes into play for these routes as they shouldn't be accessed anyways. As @daftspunk says, anybody that's using load balancers is going to know enough to configure the rules how they like, so I'd rather retain the default rules as they currently are in the documentation

PR #2655, opened 2017-02-05
Add color button to richeditor

gpasztor87:


LukeTowers:
Please explain the need for this PR. Unexplained PRs will not be merged.

gpasztor87:
I think it's useful if we can change the color of the text. This plugin is already in the widget.

LukeTowers:
@gpasztor87 You already can by specifying the `toolbarButtons` value on the `richeditor` field: http://octobercms.com/docs/backend/forms#widget-richeditor

daftspunk:
I'd prefer to add color by using CSS and the definitions Editor Settings. Gives more control that way, say, if you wanted to use a different shade of red in future versions of the site.

PR #2656, opened 2017-02-05
Typo

gpasztor87:


LukeTowers:
@daftspunk Any concerns over backwards compatibility here? It's a `protected` method so it shouldn't affect anything but I would still like your input.

Issue #2662, opened 2017-02-08
Richeditor/Markdown editor within MLRepeater doesn't load correct value when switching language


canthis:
Created this issue here, because looks like problem is with Richeditor formwidget itself, not Translate Plugin's MLRepeater widget.

##### Expected behavior
Richeditor within MLRepeater should be populated with corresponding value per language, when switching MLRepeater language to another language and then back to previous language.

##### Actual behavior
Richeditor is not being populated with corresponding language's value at all, or only with default language's value, when switching MLRepeater language.

##### Reproduce steps

1. Install Translate Plugin and create 2 languages.
(Let's assume you have 2 languages -  default: EN and secondary: RU)
2. Install [Test plugin](https://github.com/canthis/mlrepeatertest) I've created to demonstrate this issue to `/plugins/canthis/mlrepeatertest` folder.
3. Go to MLRepeaterTest plugin, and press "Create" button.
4. Fill repeater form fields with content for EN language.
5. Switch language to RU and fill form fields with content for RU language.
6. Now try switching languages and you should see, that Richeditor is not being populated with correct values, although values where successfully saved to DB.

##### Proposed fix
Changing [line 74 in october/modules/backend/formwidgets/RichEditor.php](https://github.com/octobercms/october/blob/master/modules/backend/formwidgets/RichEditor.php#L74) to
```
$this->vars['value'] = $this->formField->value;
```
makes richeditor load values as expected. 

##### October build
398 (Edge updates turned on)

##### Translate Plugin version
1.2.10

Refs: rainlab/translate-plugin#218

daftspunk:
Fixed in https://github.com/octobercms/october/pull/2663

PR #2663, opened 2017-02-08
Ensure Richeditor loads correct value when it's within MLRepeater

canthis:
Fixes octobercms/october#2662
Fixes rainlab/translate-plugin#218

### DIFF
--- a/modules/backend/formwidgets/RichEditor.php
+++ b/modules/backend/formwidgets/RichEditor.php
@@ -71,7 +71,7 @@ class RichEditor extends FormWidgetBase
         $this->vars['stretch'] = $this->formField->stretch;
         $this->vars['size'] = $this->formField->size;
         $this->vars['name'] = $this->getFieldName();
-        $this->vars['value'] = $this->getLoadValue();
+        $this->vars['value'] = $this->formField->value;
         $this->vars['toolbarButtons'] = $this->evalToolbarButtons();
 
         $this->vars['allowEmptyTags'] = EditorSetting::getConfigured('html_allow_empty_tags');


LukeTowers:
To me this indicates a deeper issue with `getLoadValue()`. Are you able to do some digging into that and explain why the RichEditor widget doesn't receive the correct value from that method?

canthis:
@LukeTowers, I've tried digging into it, tested `getLoadValue()'s` behavior with almost each type of form widget, and yeah, it loads wrong value for any type of form widget, which is within MLRepeater.
I've ended up confused, and now I can't find a reason, why we should use all this code at all:
```
public function getLoadValue()
{
    return $this->formField->value;
    $defaultValue = !$this->model->exists
        ? $this->formField->getDefaultFromData($this->data ?: $this->model)
        : null;

    return $this->formField->getValueFromData($this->data ?: $this->model, $defaultValue);
}
```
when going like this works just fine for any type of form widget within MLRepeater or standalone form widget:
```
public function getLoadValue()
{
    return $this->formField->value;
}
```
P.S. Just to remind, test plugin, which demonstrates issue is here [canthis/mlrepeatertest](https://github.com/canthis/mlrepeatertest)

I't would be nice to have @daftspunk thoughts on this.

LukeTowers:
@canthis Have you tested with providing default values to the defined form fields? On a side note, I was planning on revisiting that method at some point as it only supports defaults on non-existent models when it really should support default values for fields that are dynamically added to an existing model.

canthis:
@LukeTowers, Tested original code, and my piece of code of `getLoadValue()` method with providing default values for defined form fields.

##### Config
- Here's my form fields definitions [fields.yaml config file](https://github.com/canthis/mlrepeatertest/blob/master/models/product/fields.yaml)
- [Here's](https://github.com/canthis/mlrepeatertest/blob/master/Plugin.php#L29) my plugin's `boot()` method, from where I extend form with additional richeditor field, which comes with `default` value property provided.

##### Note
Basic form fields like [textarea](https://github.com/octobercms/october/blob/master/modules/backend/widgets/form/partials/_field_textarea.htm#L11) and others are loading their values by accessing `$field->value` which is similar to what I've changed `getLoadValue()` method:
```
public function getLoadValue()
{
    return $this->formField->value;
}
```

##### Testing results

Both types of `getLoadValue()` method successfully loads `defaults` for non-existing models
and fails (loads empty value) for dynamically added fields to an existing model.

Same behavior applies to a basic form fields, non-formwidgets as per `getFieldValue($field)` [method of october/modules/backend/widgets/Form.php](https://github.com/octobercms/october/blob/master/modules/backend/widgets/Form.php#L897)

I'm not sure if there's something to do with dynamically added fields with default values in `FormWidgetBase` and `Form` classes, because this could be easily handled by providing or even extending model with, for example, `getAnyFieldAttribute()` method like so:
```
TestModel::extend(function($model) {
    $model->addDynamicMethod('getAnyFieldAttribute', function() use ($model) {
        // Return everything based on your needs
    });
});
```
P.S. Unfortunately it seems that all this is getting out of my scope.

LukeTowers:
I'll try to take some time in the next couple of days to replicate this locally and see if I can't drill down to the root cause and also solve the other issue with default values for uninitialized fields. We can't just switch a single FormWidget to pull directly from the `value` property, whatever the issue is it effects all of the FormWidgets so we need to implement a solution that fixes it for all of them. 

Thank you for your effort so far in tracking this down, I'll keep you updated on a fix.

Fl0Cri:
I encountered this problem too and @canthis fix seems to solve the problem in my case (I used the initial suggestion to replace `getLoadValue()` to limit side effects).

I tried to drill a litle bit into this `getLoadValue()` method so I come here to share what I found hoping it will help:

Here are the variables I defined into a test layout:
```
{repeater name="pageblocks" prompt="Add a block"}
    {variable name="text" type="richeditor" label="Content"}{/variable}
{/repeater}
```

And after saving a new static page, here are the generated files:
static-pages/test.htm
```
[viewBag]
title = "Test"
url = "/test"
layout = "test_layout"
is_hidden = 0
navigation_hidden = 0
pageblocks[1][text] = "<p>Text EN</p>"
==
```
static-pages-fr/test.htm
```
[viewBag]
title = "Test"
pageblocks[2][text] = "<p>Texte FR</p>"
==
```

One thing that surprised me at first but seems to be wanted is that my `pageblocks` repeater has an item indexed as `1` in the default language version and `2` in the translated page.

Now, digging from the richeditor's `getLoadValue()` call, I found something that seems me wrong in https://github.com/octobercms/october/blob/master/modules/backend/formwidgets/Repeater.php

The following piece of code is executed twice when I try to switch the language of the repeater from my static page:
+ first time without caring of the language (default language seems to be used)
+ a second time when running this https://github.com/octobercms/october/blob/master/modules/backend/classes/Controller.php#L521 with `$handlerName` being set to `"onSwitchItemLocale"`

I put some comments to quote the values of some variables durring both executions (start reading from a call to `processExistingItems()`)
```PHP
protected function processExistingItems()
{
	$loadValue = $this->getLoadValue();    // Both calls: $loadValue == array('1' => array('text' => '<p>Text EN</p>'))
	if (is_array($loadValue)) {
		$loadValue = array_keys($loadValue);
	}

	$itemIndexes = post(self::INDEX_PREFIX.$this->formField->getName(false), $loadValue);
	// At first call: $itemIndexes == array('0' => '1')
	// At second call: $itemIndexes == array('0' => 2)

	if (!is_array($itemIndexes)) {
		return;
	}

	foreach ($itemIndexes as $itemIndex) {
		$this->makeItemFormWidget($itemIndex);
		$this->indexCount = max((int) $itemIndex, $this->indexCount);
	}
}

protected function makeItemFormWidget($index = 0)
{
	$loadValue = $this->getLoadValue();    // Both calls: same as in processExistingItems()
	if (!is_array($loadValue)) {
		$loadValue = [];
	}

	$config = $this->makeConfig($this->form);
	$config->model = $this->model;

	// Here's the intresting part: at first call:
	// - $loadValue == array('1' => array('text' => '<p>Text EN</p>'))
	// - $index == '1'
	// And at second call
	// - $loadValue == array('1' => array('text' => '<p>Text EN</p>'))
	// - $index == 2
	// So obviously, array_get() will return an empty array instead of the translated data

	$config->data = array_get($loadValue, $index, []);
	$config->alias = $this->alias . 'Form'.$index;
	$config->arrayName = $this->getFieldName().'['.$index.']';
	$config->isNested = true;

	$widget = $this->makeWidget('Backend\Widgets\Form', $config);
	$widget->bindToController();

	return $this->formWidgets[$index] = $widget;
}
```

So @canthis fix works because when using `$this->formField->value` we get the value directly from the translated page, but when calling `$this->getLoadValue()` we get the value that was set in the `$config->data` of the widget. That said I have no idea of what side effects this can lead to.

I have not enough knowledge in October internals to try to fix this, so @LukeTowers I let you with these findings hoping those are corrects and that you haven't already found them ;-)

LukeTowers:
Thanks for the extra information @Fl0Cri, that helps! Haven't had time to look further into this yet, so your information will help when I've got some more time to look into this.

amdad:
@LukeTowers @daftspunk Hey guys, I can confirm that bug exist even in latest edge update (419). And this commit fix it. Thanks @canthis

LukeTowers:
The support demand for October is increasing, while the available time the founders have to give remains the same. I'd love to review this ASAP, but I'm pretty busy right now. I will bump this up in my internal list to get to though.

amdad:
#YouLukeAwseome!

daftspunk:
Form widgets should use `getLoadValue` as a way to mutate values, although, it seems like the base getLoadValue should be looking at the form field before trying to figure out the value.

Fixed in edcd899a794caa78e2489a18987d59fd7acf83f2 (Build 420+)

anastasiauz:
The problem with Translate/Static pages still there. And not only with Richeditor but all Repeater fields. 

I'm pretty sure Richeditor problem was fixed after adding lines to ...classes/FormWidgetBase.php as suggested here: https://github.com/octoberrain/backend/commit/e4d44e5d6da07638ef08a2026b230399eae126d4 But then the second languages'  values from ALL fields including content of Static pages were gone. Just blank values under second language variants, after saving - completely deleted from the site. I upgraded and this problem was gone. But - 

after upgrading October to 431 values in the second languages' repeater fields are missing. Just blank (right values on the site still present, I don't save the page in admin panel not to kill them). How can I fix that?

LukeTowers:
I'm not sure exactly what you're saying @anastasiauz

anastasiauz:
I have Static Pages, plugin Translate, and use Repeater fields. I edited the file ...classes/FormWidgetBase.php as suggested here: octoberrain/backend@e4d44e5. Then I upgraded to 431 October and after that the values in second language (EN for me is second, RU first) were gone, just blank.

LukeTowers:
@amdad can you confirm the issue reported by @anastasiauz?

amdad:
Hm.. give me a minute.

amdad:
419 was long time ago for me...  As far as I can remember, I patched my sites manually to have working version of repeater properly saving data then. So.. now after updates things staying the same. But I cannot exactly remember if there was problem after updates. Last time when I lost translations was when I change default language in Translate plugin settings for a playground purpose. This is way to lost data for sure. Because translation is stored separately to main language.

Now using 432, and data are stored nicely.  So I cannot confirm this bug from my experience.

anastasiauz:
I never changed default language. For test purpose I've upgraded to 431 another site with the same setting (Static Pages+Multilingual Repeater) - the values in EN (default is RU) disappear too - here the fields are not blank - the EN values are replaced with values in default lang, I try to fill in again and save new values, close page, reopen - still no second lang...

amdad:
Make a clean demo, Fresh install, Static Pages and Translate plugin only. Clean theme. Recreate scenerio in this minimal environment on old versions. Do update. And then... we know for sure @anastasiauz

anastasiauz:
Ok: I've made test with fresh October install v1.0.426 from here https://github.com/octobercms/october/releases/tag/v1.0.426 (that's the version of the last mentioned site before upgrade to 431), only Static Pages + Translate and 1 Repeater in almost clean theme (1 layout, 1 page). I cannot even save the fields - as soon as I change the lang and then back - I see default lang's values. First lang is RU (default), second EN (I tried changing vice versa with same result...) 
My repeater is first thing in main layout (I also tried to insert after the <head> with same result).
{repeater name="testimonials" prompt="Review" tab="Reviews"}
    {variable type="text" name="name" label="Name" tab="Reviews"}{/variable}
    {variable type="richeditor" name="text" label="Text" tab="Reviews"}{/variable}
    {variable type="mediafinder" name="image" label="Photo" tab="Reviews"}{/variable}
    {/repeater}
What am I doing wrong if I'm the only one experiencing this problem on multiple installs?

I can upload this site to test domain and provide access to admin panel if needed. I have 4 websites with this combination of plugins and the problem is quite nasty (( Great thanks you for the support!!

munxar:
@anastasiauz basically you do nothing wrong, and I can reproduce your bug with the 'stable' version (1.3.3) of the translate plugin.
We fixed the repeater bug in version 1.3.5. To get this version simply enable
```yaml
    'edgeUpdates' => true,
```
option in the file *config/cms.php*

reminder @LukeTowers @daftspunk we should mark the current version as stable, because this repeater bug was really a show stopper for lots of our projects. The fix worked so far very well.

amdad:
Yea, probably that's why I didn't encounter issue. In many cases edge updates are more stable than stable ;) Because of critical bugfixes.  Stable version of Translate plugin didn't work for me few days ago because of middleware.

anastasiauz:
Thank you, it works!

LukeTowers:
We've pushed it stable now by the way for anyone else that's interested.

PR #2669, opened 2017-02-11
Add 'logging' language variable for logging setting tab.

angelside:
- Add 'logging' language variable for logging setting tab.
- "defaultTab: Logging" defination is hard coded. I am change it with "system::lang.log.logging", language variable is "Logging".

### DIFF
--- a/modules/system/lang/en/lang.php
+++ b/modules/system/lang/en/lang.php
@@ -353,7 +353,7 @@ return [
     'log' => [
         'menu_label' => 'Log settings',
         'menu_description' => 'Specify which areas should use logging.',
-        'logging' => 'Logging',
+        'default_tab' => 'Logging',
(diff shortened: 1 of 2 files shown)


modules/system/lang/en/lang.php
```diff
@@ -353,6 +353,7 @@
     'log' => [
         'menu_label' => 'Log settings',
         'menu_description' => 'Specify which areas should use logging.',
+        'logging' => 'Logging',
```
LukeTowers: Change the key to `default_tab` instead and I'll merge this

angelside:
I did something, is it true I am not sure? If is wrong, I make clean PR again (with "default_tab").

LukeTowers:
Perfect. Thanks!

PR #2679, opened 2017-02-15
add helper info for the util cmnd

ctf0:


ctf0:
u mean like adding the available cmnds instead ?

LukeTowers:
Yes, like `php artisan create` would do

ctf0:
screen shot 2017-02-15 at 6 40 25 pm

LukeTowers:
Instead of a hardcoded list, I'd like to see you get the methods in the class, and then based on the methods available starting with 'util' (which the fire method uses to figure out what command to actually call), generate the list of available commands dynamically from that instead.

For example:
```php
public function utilCompileLang() {}
public function utilPurgeThumbs() {}
```
Would generate the list:
```
Did you mean one of these?
    october:util compile lang
    october:util purge thumbs
```

ctf0:
screen shot 2017-02-16 at 9 19 53 am

LukeTowers:
One last change and it should be good to go, please prefix the listed commands with the previous `october:` so that they follow the format `october:util compile lang`.

Thanks again for your work on this!

ctf0:
to confirm
> note that some methods are still un-finished ex.`purge uploads`

LukeTowers:
@ctf0 That's fine with me. If you want, you can add a list of stub methods that aren't finished yet in order to ignore and have those hardcoded, but I'm perfectly fine with leaving it as it is right now.

As a note, just noticed now, you have made the PR to the master branch, you need to make it to the develop branch instead. Please change that and then hopefully we can get this merged.

LukeTowers:
As soon as the check completes, I'll merge this.

PR #2687, opened 2017-02-18
some changes/enhancments

ctf0:
i apologize for opening more than one PR as i don't know how to make a PR for each change, so here are all of them.

- add helper info for the util cmnd https://github.com/octobercms/october/pull/2679
- remove wildcards from `oct:mirror` as it shouldnt be listed under public
- update htaccess
- add `driver & builder` plugins choice to the installer
- use regex for media ignore instead of file names
- move static pages menu reorder btn to left

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -17,24 +17,25 @@
     ##
     RewriteRule ^bootstrap/.* index.php [L,NC]
     RewriteRule ^config/.* index.php [L,NC]
-    RewriteRule ^vendor/.* index.php [L,NC]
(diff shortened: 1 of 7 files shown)


ctf0:
maybe u can review them and add what u like :+1:

LukeTowers:
Sorry, don't have the time. If you want them added, you'd have to create an issue / pull request for each one describing in detail what it is that is being changed and why it should be changed.

daftspunk:
Yes please, one concern per issue/PR makes it easier for us to see what is happening.

PR #2691, opened 2017-02-20
remove wildcards as it shouldnt be listed under public

ctf0:


LukeTowers:
You've just broken the entire purpose of the `october:mirror public` command. This will not be accepted unless you can clearly explain why.

ctf0:
all it does is that it doesnt sym-link the `plugin/module/theme` folders to public folder anymore because the outside world shouldnt get access to any of those.

also this doesnt break any of the cms functionality, u can still do everything as normal.

LukeTowers:
It does break that functionality, because it's not linking `plugin/module/theme` folders, it's linking the `assets/resources` directories within those folders, which contain client assets like fonts, images, JS, and CSS. Without linking those directories, a blacklist approach will block access to those resources.

ctf0:
in the matter of fact i have made a theme with some assets already and i call it through
```twig
<link href="{{ 'assets/css/app.css'|theme }}" rel="stylesheet">
```

daftspunk:
@ctf0 Please take the time to familiarize yourself with the platform before submitting any more PRs. Not to discourage your enthusiasm because it's great!

PR #2692, opened 2017-02-20
use regex to hide files under media manger

ctf0:


LukeTowers:
Please explain why this is necessary.

ctf0:
under media manager u have to specify each new file in order to ignore it, which with regex all the dot files will be ignored by default

daftspunk:
I like ignoring dot files, can't think of any reason why the config might be needed. An event hook might be better but that can be omitted for now.

daftspunk:
Wasn't aware config already existed, my bad. This looks good.

PR #2694, opened 2017-02-20
add driver & builder to installer

ctf0:


LukeTowers:
Please explain why this is necessary.

ctf0:
because the driver is needed if u gonna use anything other than local disk which 90% will do.
for the builder because again a very big percentage of users will add a new functionality to the cms.

in both the user have the choice whether to install them or not. so its not enforced.

LukeTowers:
I'll have to get @daftspunk's opinion on this. I personally don't think it's required as the user can install those plugins after October is installed.

ctf0:
thats why they are kept as choice rather than enforcing them.

LukeTowers:
Yes, but you may as well just add all possible plugins as a choice here. I don't know that it's worth the extra effort of maintaining this code to simply install the plugins desired a couple steps ahead of time.

ctf0:
at the end of the day its up to u, but i thought that everything else is optional except those two which

1- allow u to use other services rather than `local & file` which most users will
2- another that allow u to build ur app, which without it u dont have much of an option

LukeTowers:
I have to say I disagree with you. October ships with scaffolding commands. For example, `artisan create:component Vendor.Plugin MyComponent`

October's focus is "going back to basics", so shipping without the drivers plugin is a good thing because many casual users are not going to use it. I think we're going to have to see input from other members in the community on this before we can make a decision on whether or not to provide this option to install it by default.

ctf0:
yeah yeah sure, plz note its a choice cmnd, so it wont get installed if the user answered with **NO**

daftspunk:
I don't mind asking for `October.Drivers` so long as it appears under the advanced settings.

```
if ($this->confirm('Configure advanced options?', false)) {
```

This logic also needs to be encapsulated in its own method.

LukeTowers:
@ctf0 If you implement @daftspunk's suggestion then I think we can probably merge this

daftspunk:
Still need to prompt the user if they want Drivers plugin installed or not...

LukeTowers:
@ctf0 again, see @daftspunk's latest comment

daftspunk:
Damn, this was so close to being merge-able. Might leave this open in case someone else wants to pick it up.

PR #2695, opened 2017-02-20
move static menu item order btn to left

ctf0:


LukeTowers:
First problem, don't edit compiled files, edit the LESS source. Secondly, I'm going to auto-close any PRs without descriptions from now on.

ctf0:
i didnt add desc as the title define the purpose, also u can keep it open and will push the less edits later

LukeTowers:
You haven't explained why it is necessary.

ctf0:
necessary because if u tried to reorder the menu item on wide screen u will feel lost between checking and re-checking if it is the right item to move , while now u have the icon right next to the item name which will make ur work much easir

daftspunk:
This issue lacks enough detail or the replication instructions cannot be actioned easily. As per [the contribution guidelines](https://github.com/octobercms/october/blob/master/CONTRIBUTING.md#reporting-a-bug-with-octobercms), please provide the exact steps you are taking for the issue to occur. This will be used for peer review.

Here is an example of a **bug summary**:

> The CMS page URL is not being set automatically from the title value.

Here is an example of some **reproduce steps**:

> 1. In the back-end, open the CMS > Pages section
> 1. Click the "Add" button
> 1. Enter [Hello] for the Title field
> 1. Click the Save button 

Here is an example of **expected behavior**:

> - The Title should be set to [Hello]
> - The URL should be set to [/hello]

Here is an example of **actual behavior**:

> - The Title is set to [Hello] correctly
> - The URL is left empty

PR #2697, opened 2017-02-21
Show multi-line version details from plugin changelog

petehalverson:
Version details displayed under the `Changelog` tab on the plugin page are currently truncated to the first entry if they are defined as an array in the plugin `updates/version.yaml`.

This pull request adds support for rendering those details as additional lines in the changelog.

### DIFF
--- a/modules/system/controllers/Updates.php
+++ b/modules/system/controllers/Updates.php
@@ -171,7 +171,7 @@ class Updates extends Controller
 
             foreach ($updates as $version => $details) {
                 $contents[$version] = is_array($details)
-                    ? array_shift($details)
+                    ? implode($details, "\n")
                     : $details;
             }
         }
--- a/modules/system/controllers/updates/details.htm
+++ b/modules/system/controllers/updates/details.htm
@@ -78,7 +78,7 @@
                         <dl>
                             <?php foreach ($changelog as $version => $comment): ?>
                                 <dt><?= e($version) ?></dt>
-                                <dd><?= e($comment) ?></dd>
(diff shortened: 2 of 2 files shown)


daftspunk:
Please support this by providing some replication instructions. What does your version.yaml file look like that would prompt this change?

petehalverson:
The original intent was to allow plugin authors to use YAML arrays to describe extensive updates.

In hindsight, I don't think it's a good idea to introduce mixed use cases for such arrays as we'd have to purge filenames when [migrations/seeds are used](http://octobercms.com/docs/plugin/updates#version-file):

```
1.0.1: Added some upgrade
2.0.1:
    - Extensive update with lots to describe
    - More detail on everything
    - Even more detail
2.0.2:
    - Create settings table
    - create_settings_table.php
```

screen shot 2017-04-05 at 4 45 59 pm

I'll close this because I'm over it. Re-open if you feel like it's worth the added logic.

daftspunk:
Ah I see. This was never really supported, updates only support 1 line of comments.

```
    - Extensive update with lots to describe. More detail on everything. Even more detail
```

petehalverson:
Yeah, sorry for the knee jerk PR.

LukeTowers:
Thanks for the effort anyways!

daftspunk:
I was just thinking more about this, I think internally we handle extensive updates as separate versions. So we just bump the version multiple times. It could be a workable alternative.

```
1.0.1: Extensive update with lots to describe.
1.0.2: More detail on everything.
1.0.3: Even more detail
```

PR #2703, opened 2017-02-22
Adding an option to make the Search widget fire on enter key

manogi:
The Search widget fires on every key stroke, which is fine with small data sets, but soon becomes problematic with medium to large data sets. 

A customer of mine has a shop with products. He starts entering a word into the search field, and the search widget starts to search for example for "am " when he is actually entering "am hofgarten". The search for "am hofgarten" returns the result sooner than the search for "am " (because there are less entries), so my customer sees the results for "am hofgarten" for a very short time, then the results for "am " come in and replace the other results. Now my customer sees "am hofgarten" in the search input, but has the wrong results. And phones me and asks me to fix it ;-)

So I changed the Search.php and the _search.htm file to look for an `everykey` option, which needs to be true (which is the default, so nothing needs to be changed if the old behavior is wanted) to have the `data-track-input` be inserted into the partial. 

In your config files you now can add an `everykey` option (next to the others like `prompt`, `mode` and  `scope` and set it to false if you don't want the old behavior. 

I hope I made myself clear... 

Thanks for OctoberCMS!

### DIFF
--- a/modules/backend/widgets/search/partials/_search.htm
+++ b/modules/backend/widgets/search/partials/_search.htm
@@ -5,7 +5,7 @@
         name="<?= $this->getName() ?>"
         value="<?= $value ?>"
         data-request="<?= $this->getEventHandler('onSubmit') ?>"
-        data-track-input
+        <?= $everykey ? 'data-track-input' : '' ?>
         data-load-indicator
         data-load-indicator-opaque
         class="form-control <?= $cssClasses ?>"


manogi:
By the way, I looked into the JS files and saw that you programmed a 300 ms "waiting period" for the actual submitting after each key stroke, so that someone with a quick typing speed (and no interruptions) will probably not have these problems. My customer is a slow typer though, and he probably pauses and looks for the right way o write something in the middle of typing. I also saw that one could change that period by setting for example `data-track-input="1000"`. So maybe a setting in the config files to enter the milliseconds would also be an idea?

daftspunk:
This is indeed an interesting scenario that will require some research for the best solution.

modules/backend/widgets/Search.php
```diff
@@ -41,6 +41,11 @@ class Search extends WidgetBase
      */
     public $scope;
 
+    /**
+     * @var bool Search on every key stroke.
+     */
+    public $everykey = true;
```
GinoPane: Maybe would be better to rename it to "onEveryKey" or "fireOnEveryKey".

manogi: Cool - I will amend my pull request.

manogi: Actually it was backwards compatible - the `everykey` option defaulted to true. But I also think that making it the other way around "sounds" more backward compatible, so I made a new Pull Request #3524

manogi:
Please find the amended PR here: #3524

LukeTowers:
@manogi in the future please don't close and reopen pull requests, just make the changes to the original request so that the discussion history is preserved.

manogi:
yeah I'm sorry - the PR was over a year old, I thought maybe the original files had changed more in the mean time - but that would have shown in the "Files Changed", now that I think of it, right?
I'm a bit new to the PR business...

LukeTowers:
No worries, just something to keep in mind for next time

PR #2710, opened 2017-02-25
Limit for the repeater items

panakour:
Option for the client to use `$maxItems` property to set a limit for repeater items. Default value(no-limit == 0) of the repeater items


modules/backend/formwidgets/repeater/partials/_repeater.htm
```diff
@@ -14,9 +15,10 @@
 
     <div class="field-repeater-add-item loading-indicator-container indicator-center">
         <a
-            href="javascript:;"
-            data-request="<?= $this->getEventHandler('onAddItem') ?>"
-            data-load-indicator>
+                href="javascript:;"
+                data-request="<?= $this->getEventHandler('onAddItem') ?>"
+                data-request-complete="($('.field-repeater').attr('data-max-items') > $('.field-repeater-item').length || $('.field-repeater').attr('data-max-items') == 0) ? $('.field-repeater-add-item').show() : $('.field-repeater-add-item').hide()"
```
LukeTowers: I feel like this could be handled cleaner, within the `repeater.js` logic, instead of on the element itself. I'd like you to try hooking into the AJAX framework events http://octobercms.com/docs/ajax/javascript-api#global-events

modules/backend/formwidgets/repeater/assets/js/repeater.js
```diff
@@ -44,6 +44,7 @@
         this.$el.one('dispose-control', this.proxy(this.dispose))
 
         this.hidePrompt()
+        $(document).on('ajaxSuccess', this.proxy(this.hidePrompt))
```
LukeTowers: Does this event need to be bound to `document`? Why not bind it more specifically to `this.$el`?

modules/backend/formwidgets/repeater/assets/js/repeater.js
```diff
@@ -134,6 +137,16 @@
         return defaultText
     }
 
+    Repeater.prototype.hidePrompt = function () {
+        var maxItems = $('.field-repeater').attr('data-max-items')
```
LukeTowers: All direct references to jQuery selectors within this method need to be changed, they are too broad. Try testing with multiple repeaters on the same form and you'll see what problems these broad selectors will cause.

Try using `this.$el` to be more specific.

Also, please restructure to store the maxItems data in the widget JS options property. See https://github.com/octobercms/october/blob/master/modules/backend/formwidgets/repeater/assets/js/repeater.js#L33-L37 and https://github.com/octobercms/october/blob/master/modules/backend/formwidgets/repeater/assets/js/repeater.js#L147 for reference. Look into what `$.data('oc.repeater')` returns and see how you can best integrate passing the option from the FormWidget options in PHP to the JS that runs this.

Thank you for your work on this so far, keep it up!

modules/backend/formwidgets/repeater/partials/_repeater_item.htm
```diff
@@ -12,8 +12,7 @@
                 data-request="<?= $this->getEventHandler('onRemoveItem') ?>"
                 data-request-data="'index': '<?= $indexValue ?>'"
                 data-request-success="$(this).closest('.field-repeater-item').remove()"
-                data-request-confirm="<?= e(trans('backend::lang.form.action_confirm')) ?>"
-                data-request-complete="($('.field-repeater').attr('data-max-items') > $('.field-repeater-item').length || $('.field-repeater').attr('data-max-items') == 0) ? $('.field-repeater-add-item').show() : $('.field-repeater-add-item').hide()">
+                data-request-confirm="<?= e(trans('backend::lang.form.action_confirm')) ?>">
```
panakour: I try to listen from repeater.js when this event get triggered using global events http://octobercms.com/docs/ajax/javascript-api#global-events but it does not work

panakour: Ok. Thank you very much.

LukeTowers:
Thanks @panakour for your work, I've made some changes and merged the functionality in #2723

PR #2733, opened 2017-03-04
Fixed undefined property when setting minmaxDate

alexdevmotion:
Fixed this error: https://octobercms.com/forum/post/set-daterange-mindate-to-report-the-bug
Trying to set the minDate or maxDate property on the daterange filter throws Exception.


daftspunk:
Please change the branch to `develop`.  This also needs replication instructions, how does one produce this error?

LukeTowers:

PR #2736, opened 2017-03-06
Minor language sign/mark corrections

angelside:
- Remove a blank space at backend/lang.php
- Change unnecessary double quotes to single quotes at backend/lang.php
- Add dot mark of sentence end for skipped/forgotten "*_comment" lines.


modules/backend/lang/en/lang.php
```diff
@@ -141,7 +141,7 @@
             'name_field' => 'Name',
             'description_field' => 'Description',
             'is_new_user_default_field_label' => 'Default group',
-            'is_new_user_default_field_comment' => 'Add new administrators to this group by default',
+            'is_new_user_default_field_comment' => 'Add new administrators to this group by default.',
```
LukeTowers: The rest of the changes are fine with me, however I just don't think we're consistent enough with which translation strings end with a period vs which don't to make those changes. I'm thinking either we come up with a consistent approach to that and implement it for all the strings within the English language files or we just don't bother and leave it as is for now. What are your thoughts on that?

angelside: I know that such changes are insignificant. I will try not to bother you again for these.

I am actually translating in Turkish. Current Turkish is very bad, and could not catch the standards. At this moment I encounter small things in English files. 

For now I only see missing single quotes on variables like :id, :class, :model, some have some not. It's okay if you say "stay like this".

Another thing is on/off buttons, my Turkish on/off is a little more lengthy, I modify css with plugin for on/off button width.

And for time format, why we not have language variables for that. We can define it in own yaml, but I think we need a language variable for system wide. I know some time variable is come from moment.js language.

angelside: I have no idea, if I need do anything about review. I don't understant how it works.

***

I know that such changes are insignificant. I will try not to bother you again for these.

I am actually translating in Turkish. Current Turkish is very bad, and could not catch the standards. At this moment I encounter small things in English files. 

For now I only see missing single quotes on variables like :id, :class, :model, some have some not. It's okay if you say "stay like this".

Another thing is on/off buttons, my Turkish on/off is a little more lengthy, I modify css with plugin for on/off button width.

And for time format, why we not have language variables for that. We can define it in own yaml, but I think we need a language variable for system wide. I know some time variable is come from moment.js language.

LukeTowers: I don't mind the minor formatting changes (only in language files), I'm just not going to go out of my way to make them. I'll merge them if PRs are made for them though; just not as sure about the changes to the periods.

As far as the on / off button width, I'd suggest making your translation changes and then creating an issue for the width of the buttons so that we can deal with it in the most efficient way possible after reviewing the issue.

Not sure what you're asking about the time format language variables

angelside: Time format sample from phpBB:

```php
'DATE_FORMAT'		=> '|d M Y|',	// 01 Jan 2007 (with Relative days enabled)
'DATETIME_FORMAT'	=> '|d M Y, H:i|',	// 01 Jan 2007, 13:37 (with Relative days enabled)

// The value is only an example and will get replaced by the current time on view
'dateformats'	=> array(
    'd M Y, H:i'			=> '01 Jan 2007, 13:37',
    'd M Y H:i'				=> '01 Jan 2007 13:37',
    'M jS, \'y, H:i'		=> 'Jan 1st, \'07, 13:37',
    'D M d, Y g:i a'		=> 'Mon Jan 01, 2007 1:37 pm',
    'F jS, Y, g:i a'		=> 'January 1st, 2007, 1:37 pm',
    '|d M Y|, H:i'			=> 'Today, 13:37 / 01 Jan 2007, 13:37',
    '|F jS, Y|, g:i a'		=> 'Today, 1:37 pm / January 1st, 2007, 1:37 pm',
),
// The default dateformat which will be used on new installs in this language
// Translators should change this if a the usual date format is different
'default_dateformat'	=> 'D M d, Y g:i a', // Mon Jan 01, 2007 1:37 pm
```

LukeTowers: That would be a separate issue, we'd have to audit everywhere date formats are used within October, assign them keys based on commonalities and then reference those keys to change the datetime formats associated with them. I imagine something like that would be integrated as a backend setting. 

I do agree with the premise though, I would also like more control over how October displays dates by default across the system. I know you can change the options for specific instances but it would be convenient to be able to modify the defaults system wide.

LukeTowers:
You can't approve them, you can only make the changes requested in them and then I would approve or deny them. 

As far as this pull request goes, I don't think we've got a convincing enough reason to change whether we use periods or not. 

As for the other issue, the one about the time variables, make a separate issue for it and then we can discuss it there.

daftspunk:
Was this merged somewhere or just closed?

LukeTowers:
Just closed, didn't see a good enough reason to merge it.

daftspunk:
It's the PRs that Luke Towers rejects that makes Luke Towers, the best.

PR #2742, opened 2017-03-10
Allows Settings to be defined in YAML

firemankurt:
Allow settings array to be defined in the plugin.yaml file instead of just in php file.

### DIFF
--- a/modules/system/classes/PluginBase.php
+++ b/modules/system/classes/PluginBase.php
@@ -138,7 +138,10 @@ class PluginBase extends ServiceProviderBase
      */
     public function registerSettings()
     {
-        return [];
+		$configuration = $this->getConfigurationFromYaml();
+		if (array_key_exists('settings', $configuration)) {
+			return $configuration['settings'];
+		}
     }
 
     /**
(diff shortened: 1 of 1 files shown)


LukeTowers:
Don't forget to only make PRs to `develop`. I'll change it for you this time but keep it in mind for the future.

modules/system/classes/PluginBase.php
```diff
@@ -138,7 +138,10 @@ public function registerPermissions()
      */
     public function registerSettings()
     {
-        return [];
+		$configuration = $this->getConfigurationFromYaml();
```
LukeTowers: Project standard is spaces, not tabs. Please fix and then we can merge this.

firemankurt:
Converted all indents to spaces as requested

LukeTowers:
Perfect, thanks!

Now that settings can be initialized in the plugin.yaml, feel up to the task of extending the RainLab.Builder plugin to allow for managing the settings of a plugin through that?

firemankurt:
I was afraid you might ask that.   I will take a look at it when I get
time.  Thanks for your hard work.

LukeTowers:
Gotta keep moving forward man :) Thanks for your work too!

PR #2745, opened 2017-03-11
update phpunit version to 5.*

panakour:


LukeTowers:
October officially supports PHP >= 5.5.9. PHPUnit only supports PHP >= 5.6. This will most likely get taken care of when October moves to utilizing Laravel 5.5 LTS.

Issue #2739, opened 2017-03-07
Backend list adding switch clickable buttons


ChVuagniaux:
It would be nice if we can have switch buttons in backend list for can switch boolean status of a model.

By example publish unpublish item without need to open the update form.

~My idea is to implement it with a Backend Behaviors (`Backend\Behaviors\SwitchController`) that use the switch column type ( https://octobercms.com/docs/backend/lists#column-switch )~

This is a right way for you ? 

See PR https://github.com/octobercms/october/pull/2748

LukeTowers:
That's a neat idea. You can implement this yourself in your plugin, and then submit a PR for a built-in behavior to integrate it into core.

daftspunk:
Moving discussion to #2748

PR #2748, opened 2017-03-13
Backend List - make switch columns clickable for reverse the value

ChVuagniaux:
This PR add the possibility to replace Backend list switch by clickable buttons that reverse the value of the field : That allow you to create a publish / unpublish button on the lists backend (see bellow for demo).

These changes introduced will not change something on existing switch, for apply this behavior you need to edit the `columns.yaml` field for add `switchable: true` see bellow.

This PR add the following configuration parameters for `columns.yaml` 
```yaml
    your_field:
        label: 'Your Label'
        type: switch
        switchable: true # This is that parameter that convert a switch column to a button
        icon: true # If true replace Yes/No text by ✓/✗ icons

        # With the two following params, you can override 
        # the default title displayed on the button hover
        switchLinkTitleTrue: "Unpublish item" # by default : Click to switch to No
        switchLinkTitleFalse: "Publish item" # by default : Click to switch to Yes
```

## Demo
*Behavior with ` icon: true`*
switch-icon

*Behavior with ` icon: false` (or without icon field)*
switch-text

*With custom titles*
switch-icon-custom-titles

I have just one problem, after have edited the value, when I update the list (`return $this->controller->listRefresh($this->primaryDefinition);`)  I can't redirect to the rights pagination page.
That make the following behavior : 
switch-paginate-error

related to issue #2739

### DIFF
--- a/modules/backend/behaviors/ListController.php
+++ b/modules/backend/behaviors/ListController.php
@@ -327,6 +327,30 @@ class ListController extends ControllerBehavior
     }
 
     /**
+     * Switch a boolean value of a model field
+     * @return void
+     */
+    public function index_onSwitchField()
+    {
+        $field = post('field');
+        $id = post('id');
+        $modelClass = post('model');
+
+        if (empty($field) || empty($id) || empty($modelClass)) {
+            Flash::error('Following parameters are required : id, field, model');
+            return;
+        }
+
+        $model = new $modelClass;
+        $item = $model::find($id);
+        $item->{$field} = !$item->{$field};
+
+        $item->save();
+
+        return $this->controller->listRefresh($this->primaryDefinition);
+    }
+
+    /**
      * Renders the widget collection.
      * @param  string $definition Optional list definition.
      * @return string Rendered HTML for the list.
--- a/modules/backend/classes/ListColumn.php
+++ b/modules/backend/classes/ListColumn.php
@@ -92,6 +92,26 @@ class ListColumn
     public $path;
 
     /**
+     * @var bool If set to yes convert switch column into a button
+     */
+    public $switchable = false;
+
+    /**
+     * @var bool If set to yes display tick/cross instead of yes/no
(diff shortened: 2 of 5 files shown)


pvullioud:
For Pagination in list look at this PR : https://github.com/octobercms/october/pull/2392
Not implemented yet

SebastiaanKloos:
I really like this idea. Hope it will be merged soon!

daftspunk:
The list is usually seen as a part of the read function in CRUD. Just to clarify, is this proposal seen as an improvement to core? Is it possible to achieve this functionality without this PR?

LukeTowers:
See #2739 for context @daftspunk. I think it's a neat idea, not sold on this implementation with all the options though.

seanthepottingshed:
This would be lovely

daftspunk:
This seems like an over engineered way to do the same thing as a `type: partial` with

```
<button data-request="onToggleStatus" data-request-data="id: <?= $record->id ?>">...</button>
```

With

```
function onToggleStatus() {
    // Toggle some status

    return $this->listRefresh();
}
```

It is unlikely this will be accepted as to keep the core light and simple to use. Consider leveraging the `registerListColumnTypes` plugin registration method to introduce this pattern inside a plugin.

daftspunk:
This is not accepted for reasons outlined:

- Complex implementation for what is a relatively simple feature
- Toggling the status could be handled in bulk, using the checkboxes and a toolbar button
- The UX is not great, a single misclick could result in a corruption of data, where the user is unaware
- The logic for toggling is hard coded to modify a model attribute, in some cases custom logic may be needed, such as a custom method on the model

ChVuagniaux:
If you are interested by this feature, I have released a free plugin that do this job
https://octobercms.com/plugin/inetis-listswitch

SebastiaanKloos:
Thankyou :)

PR #2760, opened 2017-03-16
Fix inflexible relation toolbar

tschallacka:
Not all models use 'id'  as primary key.

This way the models that use $primaryKey = 'foobar' can work too :-)


LukeTowers:
I'll take a look to see if this happens anywhere else before merging, thanks for spotting this though!

LukeTowers:
That took far longer than I intended it too, whoops.

PR #2764, opened 2017-03-18
Update _recordfinder_form hard coded button "Cancel" text

angelside:
Update _recordfinder_form hard coded button "Cancel" text

***

We have relation.cancel, form.cancel, account.cancel but we not have recordfinder.cancel. I decided to add "recordfinder.cancel" language variable instead of use form.cancel.

### DIFF
--- a/modules/backend/formwidgets/recordfinder/partials/_recordfinder_form.htm
+++ b/modules/backend/formwidgets/recordfinder/partials/_recordfinder_form.htm
@@ -15,7 +15,7 @@
                 type="button"
                 class="btn btn-default"
                 data-dismiss="popup">
-                Cancel
+                <?= e(trans('backend::lang.recordfinder.cancel')) ?>
             </button>
         </div>
     <?= Form::close() ?>
--- a/modules/backend/lang/en/lang.php
+++ b/modules/backend/lang/en/lang.php
@@ -256,7 +256,8 @@ return [
         'return_to_list' => 'Return to the list'
     ],
     'recordfinder' => [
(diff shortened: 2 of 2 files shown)


LukeTowers:
I'm not convinced this is the right way to go, perhaps we should have a `general.cancel` instead? Plus, anyone who updates is now going to have the English version of `cancel` showing up instead of their language like normal because their language hasn't implemented the key yet. If we're going to do this, make sure that you copy over the key values from all the other language files as well please, otherwise people will be confused why all of the sudden certain "Cancel" buttons in their language switched languages.

angelside:
I know it's effected all language files, the first thing that came to mind was this. First I think use form.cancel to avoid this. After I saw like relation.cancel more than one "cancel" variables is seperated and I thought it was a rule and it's for customization.

Ok, I think you say edit all language files, they already have a translated "cancel" for referance. It's not a problem for those who maintain language files, right?

Please, do not leave me this kind of choice, just tell me what need I do or which is should to be? Not a problem at all. I'm new comer to October cms. I am avoid to made a non-standart changes.

1. "general.cancel" -> Add this in all language files and use it in "_recordfinder_form" ? I have no idea where add this variable in language file.
2. "recordfinder.cancel" -> Add this in all language files and use it in "_recordfinder_form" ?

I think I need create new PR for this.

LukeTowers:
Ah, just took a closer look at this. I didn't realize you were replacing a hardcoded variable, thought you were changing a reference from `form.cancel` to `recordfinder.cancel`. This is fine, and I'll accept the PR, no need for another one.

PR #2776, opened 2017-03-24
line-height from .status-icon to .status-icon > i

lrs:
Fix alignment issue with dashboard status icons on Firefox in /modules/system/assets/ui/less/chart.less.

Moving the line-height from .status-icon to .status-icon > i re-centers the icon in Firefox.

Tested on firefox 52.0.1, chrome 57.0.2987.110, IE 11.953.14393.0, Edge 38.14393.0.0.

Untested in Safari, Opera.

Before:
before_1

After:
after_1


LukeTowers:
Unable to reproduce reported alignment issue in Firefox 47 on macOS 10.12, will wait to see if it pops up after updating.

lrs:
Perhaps it's an issue with Firefox on Windows 10? I get the same misalignment with Firefox 47.

LukeTowers:
Perhaps @daftspunk can chime in when he gets back from his few days off, he uses Windows 10 AFAIK.

PR #2779, opened 2017-03-25
Add 'iconSvg' to $purgeable

alxy:
This is needed for the Updates controller to function properly. Otherwise, on plugin enable/freeze, it tries to actually set that column value and leads to a column not found error.

Not sure why this has not been spotted earlier, as the interface really no longer works: [Screenshot](https://www.dropbox.com/s/ga14k2e8wfo36qr/Screenshot%202017-03-25%2021.51.53.png?dl=0)

### DIFF
--- a/modules/system/models/PluginVersion.php
+++ b/modules/system/models/PluginVersion.php
@@ -25,7 +25,7 @@ class PluginVersion extends Model
     /**
      * @var array List of attribute names which should not be saved to the database.
      */
-    protected $purgeable = ['name', 'description', 'orphaned', 'author', 'icon', 'homepage'];
+    protected $purgeable = ['name', 'description', 'orphaned', 'author', 'icon', 'iconSvg', 'homepage'];
 
     public $timestamps = false;
 


LukeTowers:
I can't reproduce this locally, what plugin do you have that throws this error on freeze / disabling?

alxy:
@LukeTowers Any plugin that also defines an [SVG icon](http://octobercms.com/docs/plugin/registration#basic-plugin-information) in the ``pluginDetails()`` method. However, I wonder where this is ever used...

LukeTowers:
Yeah, it's not even used by RainLab.Blog or RainLab.pages. Do you have a plugin that I could test it with?

alxy:
You can simply add ``'iconSvg'     => 'plugins/rainlab/blog/assets/images/blog-icon.svg',`` to the RainLab.Blog plugin and then try to freeze it. It should break. But its really weird that this option exists, since it isn't really used anywhere (I can tell of)...

LukeTowers:
Yeah, I can't tell where it's used either. It's not used on the plugin details page where the icon property normally is. 

@daftspunk Should we implement support for the plugin iconSvg property on the plugin details page in the backend or just remove it entirely? I'm leaning for proper support personally.

LukeTowers:
I can reproduce this, leaving it unmerged and open for now so that we remember to look into actually utilizing that property - don't want to forget about it and not take care of that problem either one way or the other.

daftspunk:
The same issue was true with foo => bar

```
    public function pluginDetails()
    {
        return [
            'name'        => 'rainlab.blog::lang.plugin.name',
            'description' => 'rainlab.blog::lang.plugin.description',
            'author'      => 'Alexey Bobkov, Samuel Georges',
            'icon'        => 'icon-pencil',
            'homepage'    => 'https://github.com/rainlab/blog-plugin',
            'foo' => 'bar'
        ];
    }
```

Alternative fix used fe13d8a1d1d697fdd3ae012dc8f5b0b1aa63038b

Now you can put whatever you like in there, although I'm not sure why you would =)

PR #2784, opened 2017-03-28
Fix translatable URL caching

acasar:
The same fix as https://github.com/rainlab/pages-plugin/pull/244 but for CMS pages.


daftspunk:
Nice find

PR #2798, opened 2017-04-06
Fix field dependencies for RecordFinder

alxy:
This fixed a bug where the RecordFinder field is the master field.
The problem is, that the change event needs to be triggered on the container (div): https://github.com/octobercms/october/blob/2659ae708f0c4cd45b48e4eaaac9e042d1a06649/modules/backend/widgets/form/assets/js/october.form.js#L126
The current implementation triggers the change on the input element (dataLocker), which does not trigger this behavior.
There might be a better implementation, however this is a working fix.

### DIFF
--- a/modules/backend/formwidgets/recordfinder/assets/js/recordfinder.js
+++ b/modules/backend/formwidgets/recordfinder/assets/js/recordfinder.js
@@ -63,6 +63,7 @@
     RecordFinder.prototype.updateRecord = function(linkEl, recordId) {
         if (!this.options.dataLocker) return
         var $locker = $(this.options.dataLocker)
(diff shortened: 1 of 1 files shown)


daftspunk:
Are you sure? The change event should bubble to the container.

alxy:
@daftspunk I know and this is what made me wonder as well. However, on one of my installations it did work well, and broke from this update on. I see there is some major refactoring to make the RecordFinder disposable, maybe this is something. I can provide you a test installation so you can have a look at this issue.

daftspunk:
Oh yeah I see, the control is getting disposed once a item is selected and the element is getting removed from the DOM, replaced by another version of itself.

This'll be fixed in Build 412+

PR #2799, opened 2017-04-06
Ability to add field after another field when extending a backend form

toughdeveloper:
Hi, 

This was posted a while ago on the forum - http://octobercms.com/forum/post/ability-to-add-field-after-another-field-when-extend-form

We regularly find ourselves removing many fields to re-add them (KurtJensen suggestion). So would like to see this as a built in feature.

My first pull request, please advise if I have done anything wrong.

Thanks,
Matt

### DIFF
--- a/modules/backend/classes/FormField.php
+++ b/modules/backend/classes/FormField.php
@@ -171,6 +171,11 @@ class FormField
     public $preset;
 
     /**
+     * @var string Specifies another field this field should be positioned directly after
+     */
+    public $after;
+
+    /**
      * Constructor.
      * @param string $fieldName
      * @param string $label
@@ -281,6 +286,7 @@ class FormField
             'trigger',
             'preset',
             'path',
+            'after',
         ];
 
         foreach ($applyConfigValues as $value) {
--- a/modules/backend/widgets/Form.php
+++ b/modules/backend/widgets/Form.php
@@ -460,18 +460,24 @@ class Form extends WidgetBase
         $this->fireSystemEvent('backend.form.extendFields', [$this->allFields]);
 
         /*
-         * Convert automatic spanned fields
+         * Process after definitions and convert automatic spanned fields
          */
-        foreach ($this->allTabs->outside->getFields() as $fields) {
+        foreach ($this->allTabs->outside->getFields() as $key => $fields) {
+            $this->processAfterDefinition($fields);
             $this->processAutoSpan($fields);
+            $this->allTabs->outside->fields[$key] = $fields;
         }
 
-        foreach ($this->allTabs->primary->getFields() as $fields) {
+        foreach ($this->allTabs->primary->getFields() as $key => $fields) {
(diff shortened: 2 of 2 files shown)


LukeTowers:
While I think that positioning of dynamic fields is a valuable tool to provide developers with, I don't think this particular implementation is the right way to go about it. I'd rather see something where the developer perhaps specifies an `order` property for their fields, similar to menu items. 

I'd like to get @daftspunk's input on it first though.

toughdeveloper:
Thank for feedback. My rational for the approach was for it to be similar to how you may go about positioning a new column in a migration. However I'm open to feedback to get this inline with other areas of October. 

When you mention menu items, is this within the static pages plugin ? Could you send me a link to code if possible? Thanks

LukeTowers:
I'm referring to the backend navigation menus: https://octobercms.com/docs/plugin/registration#navigation-menus

daftspunk:
This is indeed a tricky one and probably why it hasn't been implemented already. Let's explore the interaction between both proposals. Consider I want `brother` to go in between father and sister. 

```
mother
father
sister
```

Using the `$order` approach, applying a incremental default of +100

- 100: mother
- 200: father
- 300: sister

Developer somehow needs to know the auto generated range of 200-300 and come up with

- 250: brother

Now let's say the definition is modified to include two new form fields, since the default ordering is malleable, we end up with

- 100: grandmother
- 200: grandfather
- 250: brother
- 300: mother
- 400: father
- 500: sister

Developer somehow needs to know the new auto generated range of 400-500 and modify order to 450. Now consider that **250** was produced by a third-party plugin, not so easy to change to the required 450. So explicit order definition will probably never exist when defining fields.

This proves, to me at least, that @toughdeveloper is on the right track by using an API that says:

- *Insert brother after father* or *Insert brother before sister*

However the implementation is a bit too dirty, we prefer code that can be understood easily at a glance (one of October's core values). So in psuedocode should read something like this

As `insertFieldBefore()` function with arguments `$fieldDefinition`, `$targetFieldName` and `$tolerance`
1. Declare `$tolerance` with a default value of `1`
1. Define `$targetFieldName` internal `sortOrder` property as `$targetSortOrder`
1. Define `$fieldDefinition` internal `sortOrder` property as (`$targetSortOrder` minus `$tolerance`)
1. Add `$fieldDefinition` to array of form fields

The above would require the inclusion of an internal sortOrder property, used only as an internal marker for managing the sort order, not available via the public API (at this stage). The system would use this marker to sort the fields before rendering.

Hope this helps!

toughdeveloper:
Thanks for looking into this. I can have a look at making these changes. 

Just so I'm clear, where were you imaging `insertFieldBefore` being called from?

daftspunk:
It would be called by [extending code as per the documentation](http://octobercms.com/docs/backend/forms#extend-form-fields).

LukeTowers:

PR #2801, opened 2017-04-07
Prevent preloading files in advance

amdad:
Files should be loaded on user click. Huge amount of data is loaded on page load, especially if many audio/VIDEO instances are on page. This is temporary solution. Ultimately widget should support params to match html5 audio/video tags.

### DIFF
--- a/modules/cms/classes/MediaViewHelper.php
+++ b/modules/cms/classes/MediaViewHelper.php
@@ -94,11 +94,11 @@ class MediaViewHelper
     {
         switch ($type) {
             case 'video':
-                return '<video src="'.e($src).'" controls></video>';
+                return '<video src="'.e($src).'" controls preload="metadata"></video>';
             break;
 
             case 'audio':
-                return '<audio src="'.e($src).'" controls></audio>';
+                return '<audio src="'.e($src).'" controls preload="metadata"></audio>';
             break;
         }
     }


LukeTowers:
I have no idea what the problem is nor how the proposed solution solves that problem from the current description. Please expand on your explanation and provide examples of what the problem is and how this fixes it.

amdad:
When you add audio or video element to content though media manger you get snippet in content which render to HTML outpt on page as: 

```html 
<audio src="/storage/app/media/file.mp3" controls></audio> 
```
With output like this every video or audio are preloaded/prebuffered by browser by default. Which couse a lot of wasted transfer. When you have many audio or video files on page this make impact on both, server and user side.

This fix [adding tag](https://www.w3schools.com/tags/att_audio_preload.asp) to load only info data:
```html 
<audio src="/storage/app/media/file.mp3" controls preload="metadata"></audio> 
```

LukeTowers:
This is something that can be [done on a theme level](http://octobercms.com/docs/cms/mediamanager#audio-and-video-players). @daftspunk what do you think? Is this a worthwhile addition to core? My initial feeling is that if this is desired behaviour then it should be done in each theme that wants to use it instead of by default across the board.

amdad:
If I thought it will be useful only for special occasions I wouldn't try to commit this to core.
In those days where we fight for every KB of data, responsive images, small JS, CDN, etc.  Preloading whole sets of media in background by default is probably not best idea. From user/viewer perspective it may be a small disaster. 

Simple scenario. You are on mobile with slow connection. You got 5 media files on page. You read the list, ignore first 4 and click on fifth. But, browser already waste a lot of transfer on rest of items which you don't need. What if list is longer? If someone want's to allow to preloading like this THEN he can override partial. But I think by default we should think about economy.  As you see in default markup 'controls' are in tag. Then by default left in user control, to decide what he want to see/hear (and load).

Enough said. Final decision is always on your side.

daftspunk:
Wow. How is this not the default behavior? I always just assumed it was.

Issue #1866, opened 2016-03-22
Backend form relation disable


shoguniphicus:
I am not sure whether this has been reported before or not but a search in the issues didn't come up anything for this.

In the models' field.yaml, I define one of the field as disable = true

categories:
        label: Categories
        type: relation
        nameFrom: title
        disabled: true

However, the checkboxes can still be checked and unchecked.

shoguniphicus:
To clarify, this happens on checkboxlist formwidget when I have a type:relation in field.yaml

LukeTowers:
Fixed by #2803

PR #2803, opened 2017-04-07
Add readonly / disabled support for checkboxlist (relation form widget)

CSNWEB:
Relations rendered with the relation formwidget using checkboxlist are now correctly rendered like in the previewMode. This issue was described here #1866.


daftspunk:
Please DRY up this code by moving it to an `$readOnly` variable. Then

`if ($readOnly) { ...`

CSNWEB:
Sure you are right, did it :)!

PR #2816, opened 2017-04-12
Unnecessary static call directly to phpunit

panakour:


LukeTowers:
Unless you can clearly explain why this change is required, it will not be merged. Changes for the sake of style by themselves are not accepted; they must accomplish something.

panakour:
This is not only for the style. It's because it will be break in the next update of phpunit

LukeTowers:
Thank you for providing a description of what you're changing. That is necessary for us to understand why so we can clearly consider the change. @daftspunk do you have any input on this?

daftspunk:
Looks fine. In future please include notes with pull requests. Doesn't have to be a novel, @panakour, your subsequent comment is adequate. It just needs to help us all get on the same page. The most value actually comes from the infinite future where we might need to roll something back.

panakour:
Thanks for your information @daftspunk. Have a nice day.

PR #2825, opened 2017-04-18
Translate the "by" word

gergo85:


LukeTowers:
Changes simply for the sake of code formatting will not be accepted. Code formatting changes may be accepted within the same file when other functional changes are being applied, but not by themselves.

Resubmit this with just your translation changes applied, none of the extra formatting changes in unrelated files please.

PR #2827, opened 2017-04-19
Translate the "by" word

gergo85:


modules/cms/lang/en/lang.php
```diff
@@ -77,7 +78,7 @@
         'export_folders_label' => 'Folders',
         'export_folders_comment' => 'Please select the theme folders you would like to export',
         'delete_button' => 'Delete',
-        'delete_confirm' => 'Are you sure you want to delete this theme? It cannot be undone!',
```
LukeTowers: Why did you change this?

gergo85: Because of: https://github.com/octobercms/october/commit/d6a7fdc74c8726149358545aff4e360a01bdf4fb#diff-73ae1299d2c352fefb63eb86c178f48dL114

LukeTowers: Perfect, thanks for the explanation!

modules/system/controllers/updates/_install_themes.htm
```diff
@@ -48,7 +48,7 @@ <h4 class="section-header">
                                 </div>
                                 <div class="details">
                                     <h4><?= $theme['name'] ?></h4>
-                                    <p>by <?= $theme['author'] ?></p>
+                                    <p><?= e(trans('cms::lang.theme.by')) ?> <?= $theme['author'] ?></p>
```
daftspunk: Bad practice here. Some languages may include "by" after the author name. This should really be `by_author` with value `by :name`.

gergo85: I changed the translations: #2849

PR #2829, opened 2017-04-19
Update ListController.php

priitperna:
Added conditions and scope support for the list configuration.
The code is copied from the relation controller.


priitperna:
This was not as straightforward as I thought and it need a bit more work.

Issue #2834, opened 2017-04-20
Simple Backend Sidenav Layout exception


interworks-morr:
##### Expected behavior
No exception

##### Actual behavior
On only one of the many october sites I've set up, I get the following exception:
exception 'Symfony\Component\Debug\Exception\FatalErrorException' with message 'Class 'HTML' not found' in C:\inetpub\wwwroot\modules\backend\layouts\_sidenav.htm:21

In looking at the rest of the code, it appears that line 21 should be "Html" not "HTML".  I'm not sure why it's only affecting one site and not all of the others, but if I change it to "Html" the exception goes away.

##### Reproduce steps
Use the backend sidenav layout

##### October build
413

PR #2835, opened 2017-04-20
Fixing case-sensitive issue with Backend Sidenav Layout - Issue #2834

interworks-morr:
A fix for https://github.com/octobercms/october/issues/2834


austinderrick:
I'm having this issue on several sites as well. +1 for merging. :)

Weird that it seems to have just recently popped up though, it's been like that forever.

LukeTowers:
Weird every other reference uses `Html`. Huh, I'll merge this then.

PR #2847, opened 2017-04-27
AssetMaker Improvements

LukeTowers:
Improvements to the System\Traits\AssetMaker trait to support combining assets passed to the `$this->addCss()` and `$this->addJs()` methods. As a result of this, LESS and SCSS files will now also be able to be passed to the the `addCss()` method and combined on the fly.

Assets can now be combined when using these methods by passing an array of assets:
```php
$pathToCss = plugins_path('myvendor/myplugin/assets/css');

$this->addCss([
    "$pathToCss/base.css",
    "$pathToCss/special.css",
]);
```
As usual with the AssetMaker trait, you can specify `$this->assetPath` to provide a path to prepend to all the assets passed in to simplify your code. The following example is from a backend controller class where the CSS files are stored in **plugins/myvendor/myplugin/controllers/mycontroller/assets**

```php
class MyController extends Controller
{
    public function __construct()
    {
        parent::__construct();
        $this->assetPath = $this->guessViewPath('/assets', true);
        $this->addCss(['base.css', 'special.less']);
     }
}
```
Finally, using the combiner grants the ability to compile LESS and SCSS files on the fly, just like you can do with the `|theme` filter in Twig:
```php
$this->addCss(['base.less']);
```

### DIFF
--- a/modules/system/classes/CombineAssets.php
+++ b/modules/system/classes/CombineAssets.php
@@ -404,7 +404,7 @@ class CombineAssets
         $filesSalt = null;
         foreach ($assets as $asset) {
             $filters = $this->getFilters(File::extension($asset)) ?: [];
-            $path = File::symbolizePath($asset, null) ?: $this->localPath . $asset;
+            $path = file_exists($asset) ? $asset : File::symbolizePath($asset, null) ?: $this->localPath . $asset;
             $files[] = new FileAsset($path, $filters, public_path());
             $filesSalt .= $this->localPath . $asset;
         }
@@ -458,7 +458,7 @@ class CombineAssets
         $key = '';
 
         $assetFiles = array_map(function ($file) {
-            return File::symbolizePath($file, null) ?: $this->localPath . $file;
+            return file_exists($file) ? $file : File::symbolizePath($file, null) ?: $this->localPath . $file;
         }, $assets);
 
         foreach ($assetFiles as $file) {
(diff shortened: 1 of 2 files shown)


jBOKA:
While at it, is there a reason why components cannot inject their assets to one centralized asset bundle that might be added to the scripts tag?

jBOKA:
Using plugins you always end up having too many script and style links in the markup. Might be an optional feature or depend on debug mode, but would improve performance of the delivered frontends pretty much

LukeTowers:
The biggest hurdle that I can see is that you can't combine the various assets added via the `addCss` and `addJs` methods after the fact as they are all transformed into absolute URLs instead of the file paths that the asset combiner requires to do it's work.

I hear the "improve performance" argument a lot, if you were able to provide detailed examples with performance testing showing the improvement in performance in providing a method to combine as many assets from different sources into one link as possible then we could consider it. 

Although, we'd still need to figure out the best way to go about making what you're suggesting technically possible.

seanthepottingshed:
@JBOKA I totally agree

daftspunk:
I tend to disagree. This leads us down a path that is magical. An asset stream would become a black box, making it difficult to find the originating source. It actually would be worse for performance because each page would potentially have a unique composition, as opposed to sharing cached bundles across pages. It is better to have the combined code neatly bundled according to its owner/plugin (separation of concerns). Think of each component as providing a CDN-like resource that is cached across all pages when used, improving performance and saving on bandwidth.

Also consider that [HTTP/2](https://en.wikipedia.org/wiki/HTTP/2) should make our obsession with combining assets a thing of the past. If you are super concerned about performance and want a single JS + single CSS file, then you should probably compile this yourself.

PR #2849, opened 2017-04-28
Improve translations

gergo85:


daftspunk:
Like this please @gergo85 

```
trans('cms::lang.theme.by_author', ['name' => ...])
```

Do not use `str_replace` since it is built in to the `trans` function.

gergo85:
I made it.

modules/system/controllers/updates/_install_plugins.htm
```diff
@@ -46,7 +45,7 @@ <h4 class="section-header">
                                 </div>
                                 <div class="details">
                                     <h4><?= $plugin['name'] ?></h4>
-                                    <p><?= e(trans('cms::lang.theme.by')) ?> <?= $plugin['author'] ?></p>
+                                    <p><?= trans('cms::lang.theme.by_author', ['name' => $plugin['author']]) ?></p>
```
daftspunk: It should still be escaped with `e()`

modules/cms/controllers/themes/_theme_list_item.htm
```diff
@@ -8,7 +8,7 @@
 <div class="layout-cell min-height theme-description">
     <h3><?= e($theme->getConfigValue('name', $theme->getDirName())) ?></h3>
     <?php if (strlen($author)): ?>
-        <p class="author"><?= str_replace(':name', '<a href="'.e($theme->getConfigValue('homepage', '#')).'">'.e($author).'</a>', e(trans('cms::lang.theme.by_author'))) ?></p>
+        <p class="author"><?= trans('cms::lang.theme.by_author', ['name' => '<a href="'.e($theme->getConfigValue('homepage', '#')).'">'.e($author).'</a>']) ?></p>
```
LukeTowers: Don't escape everything individually, do it all at once.
Example:
```php
<?= e(trans('cms::lang.theme.by_author', ['name' => "<a href='" . $theme->getConfigValue('homepage', '#')) . "'>$author</a>"]); ?>
```
Make sure you apply that style to everything in this PR. Don't individually escape data, just do it once.

daftspunk: Has to be individual because of the HTML markup I think.

LukeTowers: Hmm, you know what, you're right. My bad.

gergo85:
Finally what should I change?

daftspunk:
Thanks for that and great work

PR #2855, opened 2017-05-03
changes to add a number range filter like the date range filter to oc…

purposebuiltscott:
This pull requests adds a filter type called numberrange that is like the daterange filter type.  The numberrange filter type accepts a min, or a min and max, and filters the result to a scope or condition like the daterange filter type.

### DIFF
--- a/modules/backend/widgets/Filter.php
+++ b/modules/backend/widgets/Filter.php
@@ -1,11 +1,14 @@
 <?php namespace Backend\Widgets;
 
+use Backend;
+use Backend\Classes\FormField;
+use Backend\FormWidgets\DatePicker;
+use Carbon\Carbon;
 use Db;
 use Str;
 use Lang;
-use Backend;
+use Event;
 use DbDongle;
-use Carbon\Carbon;
 use Backend\Classes\WidgetBase;
 use Backend\Classes\FilterScope;
 use ApplicationException;
@@ -39,7 +42,7 @@ class Filter extends WidgetBase
     //
 
     /**
-     * @inheritDoc
+     * {@inheritDoc}
      */
     protected $defaultAlias = 'filter';
 
@@ -135,6 +138,32 @@ class Filter extends WidgetBase
                     }
                 }
 
+            case 'numberrange':
+                if ($scope->value && is_array($scope->value) && count($scope->value) === 2 &&
+                    $scope->value[0] &&
+                    $scope->value[1]
+                ) {
+                    $min = $scope->value[0];
+                    $max = $scope->value[1];
+
+                    if($min) {
+                        $params['minStr'] = $min;
+                        $params['min']    = $min;
+                    }
+                    else {
+                        $params['minStr'] = '';
+                        $params['min']    = null;
+                    }
+
+                    if($max) {
+                        $params['maxStr'] = $max;
+                        $params['max']    = $max;
+                    }
+                    else {
+                        $params['maxStr'] = '∞';
+                        $params['max']    = null;
+                    }
+                }
                 break;
         }
 
@@ -202,15 +231,26 @@ class Filter extends WidgetBase
 
                 $this->setScopeValue($scope, $dates);
                 break;
+            case 'numberrange':
+                $numbers = $this->numbersFromAjax(post('options.numbers'));
+                if (!empty($numbers)) {
+                    list($min, $max) = $numbers;
+
+                    $numbers = [$min, $max];
+                }
+                else {
+                    $numbers = null;
+                }
+
+                $this->setScopeValue($scope, $numbers);
+                break;
         }
 
         /*
          * Trigger class event, merge results as viewable array
          */
         $params = func_get_args();
-
         $result = $this->fireEvent('filter.update', [$params]);
-
         if ($result && is_array($result)) {
             return call_user_func_array('array_merge', $result);
         }
@@ -305,7 +345,8 @@ class Filter extends WidgetBase
         /*
          * Extensibility
          */
-        $this->fireSystemEvent('backend.filter.extendQuery', [$query, $scope]);
+        Event::fire('backend.filter.extendQuery', [$this, $query, $scope]);
+        $this->fireEvent('filter.extendQuery', [$query, $scope]);
 
         if (!$searchQuery) {
             return $query->get();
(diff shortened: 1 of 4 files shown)


modules/backend/widgets/Filter.php
```diff
@@ -1,11 +1,14 @@
 <?php namespace Backend\Widgets;
 
+use Backend;
+use Backend\Classes\FormField;
+use Backend\FormWidgets\DatePicker;
+use Carbon\Carbon;
 use Db;
 use Str;
 use Lang;
-use Backend;
```
LukeTowers: Don't move stuff around just because you feel like it. October isn't super consistent with the order of it's use statements but it usually goes by length of the `use X` statement.

modules/backend/widgets/Filter.php
```diff
@@ -39,7 +42,7 @@ class Filter extends WidgetBase
     //
 
     /**
-     * @inheritDoc
+     * {@inheritDoc}
```
LukeTowers: The `{}` have been stripped for a reason, the [API docs](https://octobercms.com/docs/api/) don't support them.

modules/backend/widgets/Filter.php
```diff
@@ -305,7 +345,8 @@ protected function getOptionsFromModel($scope, $searchQuery = null)
         /*
          * Extensibility
          */
-        $this->fireSystemEvent('backend.filter.extendQuery', [$query, $scope]);
```
LukeTowers: Why did you replace this with the split version? `fireSystemEvent()` does exactly the same thing as calling both global and local events separately here.

modules/backend/widgets/Filter.php
```diff
@@ -365,7 +406,7 @@ protected function filterOptionsBySearch($options, $query)
     {
         $filteredOptions = [];
 
-        $optionMatchesSearch = function ($words, $option) {
+        $optionMatchesSearch = function($words, $option) {
```
LukeTowers: Avoid making inconsistent style changes for the sake of style changes alone please. Same goes for your earlier changes to whitespace between lines.

modules/backend/widgets/Filter.php
```diff
@@ -470,20 +513,16 @@ public function addScopes(array $scopes)
                 $scopeObj->maxDate = '2099-12-31';
             }
 
+            /*
+             * Ensure dates options are set
```
LukeTowers: Comment incorrect

modules/backend/widgets/Filter.php
```diff
@@ -470,20 +513,16 @@ public function addScopes(array $scopes)
...
+             */
+            if (!isset($config['minNumber'])) {
+                $scopeObj->minNumber = '0';
+                $scopeObj->maxNumber = '999999999';
```
LukeTowers: `$config['maxNumber']` could be set but that would be ignored if `$config['minNumber']` wasn't; the opposite is also true; i.e. minNumber could be set while maxNumber isn't. Those cases don't appear to be handled here.

modules/backend/widgets/Filter.php
```diff
@@ -470,20 +513,16 @@ public function addScopes(array $scopes)
...
+            }
             $this->allScopes[$name] = $scopeObj;
         }
     }
-    
-    /**
-     * Programatically remove a scope, used for extensibility.
-     * @param string $scopeName Scope name
-     */
-    public function removeScope($scopeName)
```
LukeTowers: Why did you remove this? It's a public function, this is a breaking change.

modules/backend/widgets/Filter.php
```diff
@@ -593,6 +632,33 @@ public function applyScopeToQuery($scope, $query)
 
                 break;
 
+            case 'numberrange':
+            \Log::info(print_r($scope->value, true));
```
LukeTowers: Don't leave debugging statements in your pull requests.

modules/backend/widgets/Filter.php
```diff
@@ -593,6 +632,33 @@ public function applyScopeToQuery($scope, $query)
...
+                if (is_array($scope->value) && count($scope->value) > 1) {
+                    list($min, $max) = array_values($scope->value);
+
+                    if ($min && $max) {
+
```
LukeTowers: Random whitespace?

modules/backend/widgets/Filter.php
```diff
@@ -593,6 +632,33 @@ public function applyScopeToQuery($scope, $query)
...
+                        /*
+                         * Condition
+                         *
```
LukeTowers: Random extra comment line?

modules/backend/widgets/Filter.php
```diff
@@ -750,6 +816,7 @@ protected function optionsFromAjax($options)
         return $processed;
     }
 
+
```
LukeTowers: Again, extra line of whitespace?

modules/backend/widgets/Filter.php
```diff
@@ -789,6 +856,46 @@ protected function datesFromAjax($ajaxDates)
...
+     * @return array
+     */
+    protected function numbersFromAjax($ajaxNumbers)
+    {
+        $numbers = [];
+        $dateRegex = '/\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}/';
+
+        if (null !== $ajaxNumbers) {
+            if (!is_array($ajaxNumbers)) {
+        //        if(preg_match($dateRegex, $ajaxDates)) {
```
LukeTowers: Commented out code is not acceptable unless it is part of the inline documentation.

modules/system/assets/ui/storm.js
```diff
@@ -29,7 +29,6 @@
 =require js/foundation.event.js
 =require js/foundation.controlutils.js
 =require js/flashmessage.js
-=require js/autocomplete.js
```
LukeTowers: Why is this removed?

modules/system/assets/ui/storm.js
```diff
@@ -39,6 +38,7 @@
 =require js/toolbar.js
 =require js/filter.js
 =require js/filter.dates.js
+=require js/filter.numbers.js
```
LukeTowers: Conversely, what does this JS file actually do in detail?

PR #2856, opened 2017-05-03
Add Number Range Filter

purposebuiltscott:
Hi, 

This PR adds a number range scope type to OctoberCMS List filters.  In filtering lists, it can be useful to filter the list on a numeric attribute, using a range from min to max.

The number range scope displays a form for two numbers to be selected as the number range.  The conditions parameters are passed as :min and :max.

numberrange1

numberrange2

numberrange3

### DIFF
--- a/modules/backend/widgets/Filter.php
+++ b/modules/backend/widgets/Filter.php
@@ -136,6 +136,33 @@ class Filter extends WidgetBase
                 }
 
                 break;
+            case 'numberrange':
+                if ($scope->value && is_array($scope->value) && count($scope->value) === 2 &&
+                    $scope->value[0] &&
+                    $scope->value[1]
+                ) {
+                    $min = $scope->value[0];
+                    $max = $scope->value[1];
+
+                    if($min) {
+                        $params['minStr'] = $min;
+                        $params['min']    = $min;
+                    }
+                    else {
+                        $params['minStr'] = '';
+                        $params['min']    = null;
+                    }
+
+                    if($max) {
+                        $params['maxStr'] = $max;
+                        $params['max']    = $max;
+                    }
+                    else {
+                        $params['maxStr'] = '∞';
+                        $params['max']    = null;
+                    }
+                }
+                break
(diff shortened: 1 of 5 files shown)


LukeTowers:
Out of curiosity, why did this require a whole new JS and LESS file? Why couldn't the existing stuff from the daterange filter widget be used as a base?

purposebuiltscott:
@LukeTowers being perfectly honest, I don't have the best front end skills and did my best to match the current conventions.  If there is a more elegant way to implement the js and less portions, let me know, and I'll do my best to refactor them.

LukeTowers:
I'll try to take a closer look soon, I'm not super familiar with the implementation of the date range filter right now but I'll look closer and see if I can make some recommendations. If I don't respond within a week that means I've gotten distracted from it by other stuff so you'll need to ping me again :)

daftspunk:
Looks good. Needs a supporting PR to the docs repo.

purposebuiltscott:
Just added a PR for the docs

LukeTowers:
Thanks! We'll hold off on merging this just yet, we want to have a super stable version as the last version that supports PHP 5.5.9 and before we bump the Laravel version to 5.5 LTS

CSNWEB:
@LukeTowers now that 5.5 is merged can this be merged as well?

LukeTowers:
@CSNWEB I'll do some testing of it and then merge it once I'm done. There are some minor things that I'd like to change (such as max int being set to an arbitrary 9999 number by default, want to make that a bit more realistic first), so I'll test it and then merge it and make the necessary minor changes.

LukeTowers:
@CSNWEB @purposebuiltscott I've merged this into develop and finished implementing it, this PR was missing some key things like support for a number filter by itself and the templates to render the numberrange filter, along with the `numbersFromAJAX` method. 

It should be all good to go now, if you both could test it, that would be fantastic. Thanks for your initial work on this @purposebuiltscott!

PR #2857, opened 2017-05-04
Use FormData in AJAX framework to allow file uploads

ghost:
Updates framework.js to use FormData to allow AJAX based file uploads

### DIFF
--- a/modules/system/assets/js/framework.js
+++ b/modules/system/assets/js/framework.js
@@ -52,7 +52,7 @@ if (window.jQuery.request !== undefined) {
         })
 
         if ($el.is(':input') && !$form.length) {
-            appendElementToFormData(element, data)
+            appendElementToFormData($el, data)
         }
 
         if (options.data !== undefined && !$.isEmptyObject(options.data)) {
@@ -448,8 +448,7 @@ if (window.jQuery.request !== undefined) {
 
     // Appends an element to the form data
     // ========================
-    function appendElementToFormData(el, formdata) {
-        var $el = $(el)
+    function appendElementToFormData($el, formdata) {
         var name = $el.attr('name')
 
         if (!$el.is(':input') || name === undefined) return;
--- a/modules/system/assets/js/framework.js
+++ b/modules/system/assets/js/framework.js
@@ -460,7 +460,7 @@ if (window.jQuery.request !== undefined) {
 
         var val = $el.is(':file') ? $el[0].files : $el.val()
 
-        if (typeof val === 'object') {
+        if (typeof val === 'object' && val !== null) {
             Object.keys(val).forEach(function (key) {
                 formdata.append(name, val[key])
             })
--- a/modules/system/assets/js/framework.js
+++ b/modules/system/assets/js/framework.js
@@ -47,7 +47,7 @@ if (window.jQuery.request !== undefined) {
             useFlash = options.flash !== undefined
 
         $.each($el.parents('[data-request-data]').toArray().reverse(), function extendRequest() {
-            appenObjectToFormData(paramToObj('data-request-data', $(this).data('request-data')), data)
+            appendObjectToFormData(paramToObj('data-request-data', $(this).data('request-data')), data)
         })
 
         if ($el.is(':input') && !$form.length) {
--- a/modules/system/assets/js/framework.js
+++ b/modules/system/assets/js/framework.js
@@ -418,13 +418,13 @@ if (window.jQuery.request !== undefined) {
     }
 
     // Appends an object to the form data
-    // =======================
+    // ==================================
     function appendObjectToFormData(obj, formdata, namespace) {
         var formKey
 
         if (typeof obj !== 'object' || obj === null) return
 
-        if (formdata instanceof Array) {
+        if ($.isArray(formdata)) {
             formdata.push($.param(obj))
             return
         }
@@ -447,25 +447,29 @@ if (window.jQuery.request !== undefined) {
     }
 
     // Appends an element to the form data
-    // ========================
-    function appendElementToFormData($el, formdata) {
-        var name = $el.attr('name')
+    // ===================================
+    function appendElementToFormData(el, formdata) {
+        var $el = $(el),
+            name = $el.attr("name")
 
-        if (!$el.is(':input') || name === undefined) return;
+        if (name === undefined || $el.is(':checkbox, :radio') && !$el.is(':checked')) return
 
-        if (formdata instanceof Array) {
+        var val = $el.is(':file') ? $el[0].files : $el.val()
+
+        if (val === null) return
+
+        if ($.isArray(formdata)) {
             formdata.push($el.serialize())
             return
         }
 
-        var val = $el.is(':file') ? $el[0].files : $el.val()
-
-        if (typeof val === 'object' && val !== null) {
-            Object.keys(val).forEach(function (key) {
+        if (typeof val === "object") {
+            Object.keys(val).forEach(function(key) {
                 formdata.append(name, val[key])
             })
-        } else {
-            formdata.append(name, val)
+        }
+        else {
+            formdata.append(name, val);
         }
     }
 
--- a/modules/system/assets/js/framework.js
+++ b/modules/system/assets/js/framework.js
@@ -44,7 +44,7 @@ if (window.jQuery.request !== undefined) {
         var loading = options.loading !== undefined ? options.loading : null,
             isRedirect = options.redirect !== undefined && options.redirect.length,
             useFlash = options.flash !== undefined,
-            sendFiles = $form.attr('enctype') == 'multipart/form-data' && ($el[0] == $form[0] || $el.is(':file') || options.sendFiles),
+            sendFiles = $el.is(':file')  || ($el[0] == $form[0] && $form.attr('enctype') == 'multipart/form-data') || options.sendFiles,
             data = sendFiles ? new FormData($form[0]) : [$form.serialize()]
 
         $.each($el.parents('[data-request-data]').toArray().reverse(), function extendRequest() {
@@ -54,7 +54,7 @@ if (window.jQuery.request !== undefined) {
         if ($el.is(':input') && !$form.length) {
             var inputName = $el.attr('name')
             if (inputName !== undefined && options.data[inputName] === undefined)
-                options.data[inputName] = $el.val()
+                options.data[inputName] = $el.is(':file') ? $el[0].files : $el.val()
         }
 
         if (options.data !== undefined && !$.isEmptyObject(options.data)) {
@@ -434,13 +434,21 @@ if (window.jQuery.request !== undefined) {
         Object.keys(obj).forEach(function (property) {
             if (namespace) {
                 formKey = namespace + '[' + property + ']'
-            } else {
+            }
+            else {
                 formKey = property
             }   
             
-            if (typeof obj[property] === 'object' && !(obj[property] instanceof File)) {
+            if (typeof obj[property] === 'object' && !obj[property] instanceof File && !obj[property] instanceof FileList) {
                 appendObjectToFormData(obj[property], formdata, formKey)
-            } else {
+            }
+            else if (obj[property] instanceof FileList) {
+                var files = obj[property]
+            
+                for (var index = 0; index < files.length; index++)
+                    formdata.append(formKey, files[index])
+            }
+            else {
                 formdata.append(formKey, obj[property])
             }
         });
--- a/modules/system/assets/js/framework.js
+++ b/modules/system/assets/js/framework.js
@@ -47,7 +47,7 @@ if (window.jQuery.request !== undefined) {
             useFlash = options.flash !== undefined
 
         $.each($el.parents('[data-request-data]').toArray().reverse(), function extendRequest() {
-            appenObjectToFormData(paramToObj('data-request-data', $(this).data('request-data'), data)
+            appenObjectToFormData(paramToObj('data-request-data', $(this).data('request-data')), data)
         })
 
         if ($el.is(':input') && !$form.length) {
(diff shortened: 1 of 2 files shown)


LukeTowers:
You're going to have to explain why this is necessary and why you can't use the methods utilized by existing solutions to this problem: http://octobercms.com/plugin/netsti-uploader

Nyorai:
@LukeTowers the NetSTI uploader only works with a model, I'm working on something that passes the uploaded files to another REST service in the form handler.

daftspunk:
This implementation looks clean. Can you provide some usage examples so we can test it? I'm kinda familiar with XHR uploads but never seen anything like this before 🍻

vannut:
Did you look into the issue of IE not fully supporting formData? 
I can't see the implications of this PR as a regression to other use cases. 

https://caniuse.com/#search=FormData

Nyorai:
I'll get some usage examples together and test it out with IE11 and get back.

pat-och:
i test the framework.js Nyorai's modfication with egde, ff and chrome and my globale $_FILES is correctly filled.

Here my code

in my form

```
{{ form_ajax('onUpload', { files: 'true', class: 'repeater', 'data-request-validate': true }) }}
....
<input type="hidden" name="MAX_FILE_SIZE" value="30000" />
<input type="file" id="upload_file" name="upload_file" />
....
<input type="submit" value="OK">
{{ form_close() }}
```
and in my ajax handler function onUpload() `debugbar::info($_FILES);` display expected values

LukeTowers:
@pat-och were you able to test in IE 10 & 11?

pat-och:
IE 11, have the same result, expected values in $_FILE

here the `debugbar::info($_FILES);` result (i use jquery-repeater, this is why i have group-a structured data)

```
array:1 [
  "group-a" => array:5 [
    "name" => array:2 [
      0 => array:1 [
        "piece_fichier" => ""
      ]
      1 => array:1 [
        "piece_fichier" => "test_upload_file.txt"
      ]
    ]
    "type" => array:2 [
      0 => array:1 [
        "piece_fichier" => ""
      ]
      1 => array:1 [
        "piece_fichier" => "text/plain"
      ]
    ]
    "tmp_name" => array:2 [
      0 => array:1 [
        "piece_fichier" => ""
      ]
      1 => array:1 [
        "piece_fichier" => "C:\wamp64\tmp\php49E9.tmp"
      ]
    ]
    "error" => array:2 [
      0 => array:1 [
        "piece_fichier" => 4
      ]
      1 => array:1 [
        "piece_fichier" => 0
      ]
    ]
    "size" => array:2 [
      0 => array:1 [
        "piece_fichier" => 0
      ]
      1 => array:1 [
        "piece_fichier" => 4984
      ]
    ]
  ]
]
```


but the file php49E9.tmp is not in the folder C:\wamp64\tmp\ 
so i don't know if it is good, i will try to
i will try to implement filesystem/cdn to see if that works

pat-och:
ok my $_FILE have expected values but i can't find the uploaded file anywhere so .... perhaps i'm doing something wrong but doesn't work at the moment

Nyorai:
I've picked up on one issue, and that is that the AJAX framework always submits all form fields, so if there is a file field and say an autocomplete field, the file would be submitted with each keystroke. This could obviously cause performance issues. What are the thoughts on having a data-request-with-files=true/false? This would default to false, so no backward compatability issues and the request would take the value from the element that triggered the event, so individual elements could have a different setting from the main form.

LukeTowers:
@Nyorai generally for flags like that, we default to false and if the flagging attribute is present, then we enable the featured. See `data-request-flash` and `data-request-validate`. I would suggest that it be named something like `data-request-send-files`. Thoughts?

Nyorai:
@LukeTowers that's exactly how I see it, this should behave as an enhancement so without the attribute it would be false and behave exactly as the current framework. Should a form which contains file inputs always submit with files? Is there any use case where you wouldn't want that? I was thinking of ways to limit the need to always put the attribute, for example, `data-request` on a file input would imply you want to send the files. What do you think?

modules/system/assets/js/framework.js
```diff
@@ -41,13 +41,14 @@ if (window.jQuery.request !== undefined) {
         $triggerEl.trigger(_event, context)
         if (_event.isDefaultPrevented()) return
 
-        var data = [$form.serialize()],
-            loading = options.loading !== undefined ? options.loading : null,
+        var loading = options.loading !== undefined ? options.loading : null,
             isRedirect = options.redirect !== undefined && options.redirect.length,
-            useFlash = options.flash !== undefined
+            useFlash = options.flash !== undefined,
+            sendFiles = $form.attr('enctype') == 'multipart/form-data' && ($el[0] == $form[0] || $el.is(':file') || options.sendFiles !== undefined),
```
LukeTowers: Isn't this going to set `sendFiles` to true if `options.sendFiles = false`?

LukeTowers: Isn't this going to set `sendFiles` to true if `options.sendFiles = false`?

Nyorai: Thanks for spotting that @LukeTowers, it would. In all honesty I just copied how the `useFlash` was implemented, which has the same problem.

LukeTowers:
@Nyorai this is looking great, I wish I had more time to review it so that I could merge it! I'll try to review it as soon as possible, but perhaps @daftspunk can look into it. 

@daftspunk Should we look at reworking the fileupload widget to take advantage of this improvement?

Nyorai:
@LukeTowers Need to fix a few issues in the last update, found some things I broke :(

LukeTowers:
It looks like this code has gotten fairly complex, it will take some time to review it properly. If you were able to comment it to provide clear explanations as to what and why each change is doing what it does (even just commenting on it on Github, not in the code itself), that would be very helpful.

Nyorai:
@LukeTowers no problem.

https://github.com/Nyorai/october/blob/af4e72b4c5257ac43f251e54f93ff5a1fd574dcd/modules/system/assets/js/framework.js#L48

This serializes the form to an array of objects with name and value properties. Rather than serializing the form directly, it serializes the child inputs of `$triggerEl` which is either the form or the actual element if not using a form. Doing it this way also allows the `$triggerEl` to be a `div` when not using a form, for example:

```
<div id="login">
  <input type="text" name="username">
  <input type="password" name="password">
  <button type="button" onClick="$('#login').request('onLogin')">Login</button>
</div>
```

In this example, `$triggerEl` would be the `div` and the child inputs `username` and `password` would be submitted. This only works when the inputs are not part of a form.

https://github.com/Nyorai/october/blob/af4e72b4c5257ac43f251e54f93ff5a1fd574dcd/modules/system/assets/js/framework.js#L58

`serializeArray` doesn't serialize file inputs, so if we have requested them we need to add them to our AJAX data, which must be an instance of `FormData`. We simply append our current data to the `FormData` instance, we then find all the file inputs and add the files. This is an array of files, so we need to loop over this and append each one to the `FormData`.

https://github.com/Nyorai/october/blob/af4e72b4c5257ac43f251e54f93ff5a1fd574dcd/modules/system/assets/js/framework.js#L436

This serializes a deep object to form name and value pairs. If the index is numeric and the value is not an object then the name is appended with [], otherwise it is appended with [index] and `appendObjectToFormData` is called recursively on the value.

I hope that explains the main parts of the code, if you have any questions just ask.

modules/system/assets/js/framework.js
```diff
@@ -41,23 +41,36 @@ if (window.jQuery.request !== undefined) {
...
         if (_event.isDefaultPrevented()) return
 
-        var data = [$form.serialize()],
-            loading = options.loading !== undefined ? options.loading : null,
+        var loading = options.loading !== undefined ? options.loading : null,
             isRedirect = options.redirect !== undefined && options.redirect.length,
-            useFlash = options.flash !== undefined
+            useFlash = options.flash !== undefined,
+            sendFiles = $el.is(':file') || ($el[0] == $form[0] && $form.attr('enctype') == 'multipart/form-data') || options.sendFiles,
+            data = $triggerEl.find(':input').addBack(':input').serializeArray()
```
LukeTowers: This serializes the form to an array of objects with name and value properties. Rather than serializing the form directly, it serializes the child inputs of $triggerEl which is either the form or the actual element if not using a form. Doing it this way also allows the $triggerEl to be a div when not using a form, for example:
```html
<div id="login">
  <input type="text" name="username">
  <input type="password" name="password">
  <button type="button" onClick="$('#login').request('onLogin')">Login</button>
</div>
```
In this example, $triggerEl would be the div and the child inputs username and password would be submitted. This only works when the inputs are not part of a form.

modules/system/assets/js/framework.js
```diff
@@ -41,23 +41,36 @@ if (window.jQuery.request !== undefined) {
...
+
+            $triggerEl.find(':file').addBack(':file').each(function () {
+                var $el = $(this),
+                    name = $el.attr('name'),
+                    files = $el.prop('files')
+
+                $.each(files, function (index, file) {
+                    formdata.append(name, file)
+                })
+            })
```
LukeTowers: `serializeArray` doesn't serialize file inputs, so if we have requested them we need to add them to our AJAX data, which must be an instance of FormData. We simply append our current data to the `FormData` instance, we then find all the file inputs and add the files. This is an array of files, so we need to loop over this and append each one to the `FormData`.

modules/system/assets/js/framework.js
```diff
@@ -415,7 +431,24 @@ if (window.jQuery.request !== undefined) {
         }
     }
 
-    $(document).on('change', 'select[data-request], input[type=radio][data-request], input[type=checkbox][data-request]', function documentOnChange() {
+    // Appends an object to the form data
```
LukeTowers: This serializes a deep object to form name and value pairs. If the index is numeric and the value is not an object then the name is appended with [], otherwise it is appended with [index] and `appendObjectToFormData` is called recursively on the value.

modules/system/assets/js/framework.js
```diff
@@ -306,7 +319,9 @@ if (window.jQuery.request !== undefined) {
         context.error = requestOptions.error
         context.complete = requestOptions.complete
         requestOptions = $.extend(requestOptions, options)
-        requestOptions.data = data.join('&')
+        requestOptions.data = sendFiles ? formdata : data
```
LukeTowers: Out of curiosity, why is `data.join('&')` no longer required?

LukeTowers (changes requested):

LukeTowers:
@Nyorai copied your comments to add them inline on the pull request itself for easier viewing. Also added a question about `data.join('&')`

modules/system/assets/js/framework.js
Nyorai: As the data is now an array of objects with name and value properties join('&') is no longer needed as $.ajax accepts data in this format.

LukeTowers: Perfect, thanks for the explanation! Have you tested this latest refactor thoroughly with and without using fileuploads? Does the old method of uploading files over ajax used by the mediamanager and fileupload widgets still work? Is it worthwhile to also convert those widgets to use this new functionality?

LukeTowers (approved):

LukeTowers:
Have you tested this latest refactor thoroughly with and without using fileuploads? Does the old method of uploading files over ajax used by the mediamanager and fileupload widgets still work? Is it worthwhile to also convert those widgets to use this new functionality?

Nyorai:
@LukeTowers I've tested this with mediamanager and fileuploads with the builder plugin and all seems to be working as expected.

modules/system/assets/js/framework.js
Nyorai: Think I've tried to be too clever here. If using `$.request` it will find all the elements in the document and submit them, pretty sure this isn't desired behaviour.

LukeTowers: Not sure what you mean by that latest comment. Care to expand?

SeriousKen: When using pure JS you can call `$.request('onEvent')` rather than `$('#form').request('onEvent')` (see https://octobercms.com/docs/ajax/javascript-api#javascript-examples for usage without a form). My changes ended up collecting all the form elements in the page and sending them in the request which is not current AJAX API behaviour.

Nyorai: When using pure JS you can call `$.request('onEvent')` rather than `$('#form').request('onEvent')` (see https://octobercms.com/docs/ajax/javascript-api#javascript-examples for usage without a form). My changes ended up collecting all the form elements in the page and sending them in the request which is not current AJAX API behaviour.

LukeTowers: Ah, gotcha. Glad you caught that!

LukeTowers:
Also would you be able to add some documentation to the octobercms/docs repository on this new functionality? Perhaps under the AJAX section, or really any section that makes sense. Thanks again for your hard work on this, I'm looking forwards to when we can merge this!

daftspunk:
Some nice work here. Peer review complete. This code has been adapted to be less invasive. See 72eb48cf2ee0bfb7aa01d70c00ebadd8e0f2fc16

Some areas of note:

- The functionality is enabled by `data-request-files`
- Support should be added to `{{ form_ajax('onDoSomething', { files: 1 }) }}`
- This still needs to be documented
- No support for using `input:file` without a form yet

daftspunk:
The aforementioned areas of note have now been address.

daftspunk:
Gist created for awesome example: https://gist.github.com/daftspunk/7491eccdb1986c2435f8fe64d9714921

adc91:
I still have php 5.5 for my client hosting issues and this solution comes in handy. Thank you.

adc91:
Nice!!

LukeTowers:
@adc91 just a warning for you, October will not support anything below PHP7 when Build 420 is pushed to stable.

adc91:
@LukeTowers yes I know. I'll be staying for a while even in build 419 since many of my clients' things do not have support for php 7, unfortunately.

meysammahfouzi:
Does this support `progress` for the file being uploaded?

Issue #814, opened 2014-12-03
.AppleDouble directories showing in the backend


gpunx:
In the backend, while listing pages and partials, .AppleDouble files are included in the list. Also, while editing pages, .AppleDirectories appear in the layout dropdown list
![screenshot 2014-12-03 16 49 42](https://cloud.githubusercontent.com/assets/275590/5289529/8c4fe51c-7b0c-11e4-9191-e46c201cce81.png)
![screenshot 2014-12-03 16 50 10](https://cloud.githubusercontent.com/assets/275590/5289533/924af04c-7b0c-11e4-99f9-f48379de07ae.png)

daftspunk:
Confirmed, the same is true for `.git` folders. This should be handled by checking the `October\Rain\Filesystem\Definitions::ignoreFiles` collection and ignore those folders.

LukeTowers:
Fixed by #2858.

PR #2858, opened 2017-05-04
Filter layout by ignored file names

TimFoerster:
The Layouts get filtered by the filesystem definition of ignored files.
#814

### DIFF
--- a/modules/cms/classes/Page.php
+++ b/modules/cms/classes/Page.php
@@ -4,6 +4,7 @@ use Lang;
 use Cms\Classes\Theme;
 use Cms\Classes\Layout;
 use ApplicationException;
+use October\Rain\Filesystem\Definitions as FileDefinitions;
 
 /**
  * The CMS page class.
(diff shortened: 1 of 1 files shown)


modules/cms/classes/Page.php
```diff
@@ -88,8 +89,24 @@ public function getLayoutOptions()
         $layouts = Layout::listInTheme($theme, true);
         $result = [];
         $result[null] = Lang::get('cms::lang.page.no_layout');
+        $ignoreNames = FileDefinitions::get('ignoreFiles');
+
         foreach ($layouts as $layout) {
             $baseName = $layout->getBaseFileName();
+
```
LukeTowers: Please update this logic to mirror the logic used in the [Cms\Classes\MediaLibrary](https://github.com/octobercms/october/blob/master/modules/cms/classes/MediaLibrary.php#L515) class. We like to keep things simple in October :)

LukeTowers:
Also this pull request has been made to the wrong branch. Please review [the contributing guidelines](https://github.com/octobercms/october/blob/master/CONTRIBUTING.md#pull-requests) as all PRs need to be made to the develop branch.

modules/cms/classes/Page.php
TimFoerster: The problem is that it just looks for the exact file name and not for patterns (=> dosn't resolve problems with filtering folders). However ignore patterns doesn't exists for [October\Rain\Filesystem\Definitions](https://github.com/octoberrain/filesystem/blob/master/Definitions.php#L45).
Idea 1.: Adding ignore file patterns to filesystem definitions and using it
Idea 2.: Check folder names of files and filter every folder with ignore files
What do you prefer?

LukeTowers: I would prefer adding ignorePatterns to the filesystem definitions to make it more consistent with the rest of October, but I'll double check with @daftspunk.

daftspunk:
Thanks for this!

PR #2864, opened 2017-05-11
Change the sorting icon

gergo85:
The default sorting type is asc so I changed the icon of button.


LukeTowers:
That doesn't make sense having it be an `asc` button if the default direction is `asc`. The button is supposed to indicate what will happen when you click on it, in this case switch to the `desc` sorting direction.

daftspunk:
I'm not sure the icon is supposed to be accurate, more like indicative. Which one looks better? Probably the one we already have, since Alexey chose it.

PR #2866, opened 2017-05-13
Add method to easily extend filter scopes

CSNWEB:
Hello,

I thought an interface similar to the one provided for extending list columns would be useful for filter scopes as well. At least I need that feature currently to add some dynamic filters.

### DIFF
--- a/modules/backend/behaviors/ListController.php
+++ b/modules/backend/behaviors/ListController.php
@@ -572,7 +572,7 @@ class ListController extends ControllerBehavior
             if (!is_a($widget->getController(), $calledClass)) {
                 return;
             }
-            call_user_func_array($callback, [$widget, $widget->model]);
+            call_user_func_array($callback, [$widget]);
         });
     }
 }
--- a/modules/backend/behaviors/ListController.php
+++ b/modules/backend/behaviors/ListController.php
@@ -543,4 +543,20 @@ class ListController extends ControllerBehavior
             call_user_func_array($callback, [$widget, $widget->model]);
         });
     }
+    
+     /**
+     * Static helper for extending filter scopes.
+     * @param  callable $callback
+     * @return void
+     */
+    public static function extendFilterScopes($callback)
+    {
+        $calledClass = self::getCalledExtensionClass();
+        Event::listen('backend.filter.extendScopes', function ($widget) use ($calledClass, $callback) {
+            if (!is_a($widget->getController(), $calledClass)) {
+                return;
+            }
+            call_user_func_array($callback, [$widget, $widget->model]);
+        });
+    }
 }
--- a/modules/backend/behaviors/ListController.php
+++ b/modules/backend/behaviors/ListController.php
@@ -240,6 +240,13 @@ class ListController extends ControllerBehavior
             $filterWidget->bindEvent('filter.update', function () use ($widget, $filterWidget) {
                 return $widget->onRefresh();
             });
+            
+            /*
+             * Filter Widget with extensibility
+             */
+            $filterWidget->bindEvent('filter.extendScopes', function () use ($filterWidget) {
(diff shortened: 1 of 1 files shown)


LukeTowers:
Please provide your real-world example of how you will be using this. We need that for peer review. Also, why doesn't `listExtendQuery` work for you?

LukeTowers:
After discussion and further examination this merely provides a similar method to access `backend.filter.extendScopes` as `listExtendColumns` and `extendListColumns` does for `backend.list.extendColumns`. Will merge it when the Travis build completes.

CSNWEB:
Just in case anybody is interested here is a simple real world exmaple:

```PHP
public function filterExtendScopes($filter) {
        $id = last(Request::segments());
        $account =  Account::find($id);
        $filter->addScopes([
          'hasSameGender' => [
            'label' => 'Has Same Gender',
            'type' => 'checkbox',
            'conditions' => 'gender like \''.$account->gender.'\''
          ]
        ]);
}
```

Or as a static extension:

```PHP
Buddys::extendFilterScopes(function($filter){
        $filter->addScopes([
          'filter' => [
            'label' => 'label',
            'type' => 'checkbox',
            'conditions' => 'condition'
          ]
        ]);
});
```

iboved:
This does not work if you try to add a list filter to the controller and this controller does not have any filters before. It seems that the filter widget is not bind to the controller.

This works only if the controller has at least one filter.

Maybe you know how to add list filter to the controller if this controller does not have any filters?
Thank you in advance.

CSNWEB:
Why not add an empty filter to config?

iboved:
I am extending an external plugin. I don't want to edit this plugin directly.
There is `addViewPath()` method to override controller views. But I cannot override `$configPath` property, because this propery is protected.

CSNWEB:
I think you should be able to extend the controller and add the filter widget

iboved:
Ok, thanks for the answers. I will ask on Slack.

CSNWEB:
@iboved you can try to overwrite listGetConfig: https://github.com/octobercms/october/blob/a3b93f1a0b84c14b238c464a1eb93801efea06d6/modules/backend/behaviors/ListController.php#L435

Toghether with this: http://octobercms.com/docs/backend/lists#overriding-action
and this:
http://octobercms.com/docs/services/behaviors

```PHP
ControllerThatShouldBeExtended::extend(function($controller) {
    $controller->addDynamicMethod('listGetConfig', function() use ($controller) {
        $config = $controller->asExtension('ListController')->listGetConfig();
      //add filter to config 
      return $config
    });
});
```
No clue if that would actually work though

iboved:
It would be amazing if it worked. But for some reason my dynamic `listGetConfig` method is not called. I just checked it.
Thank you very much for your advice.

iboved:
@CSNWEB I have found the solution.
In my Plugin.php file in the `boot()` method I specified the full path to the `config_list.yaml` file:

```
        ControllerThatShouldBeExtended::extend(function ($controller) {
            $controller->listConfig = '$/acme/demo/controllers/mycontroller/config_list.yaml';
        });
```

And in the `config_list.yaml` file I pecified the full path to the `config_filter.yaml` file:

```
filter: $/acme/demo/controllers/mycontroller/config_filter.yaml
```

LukeTowers:
@iboved @CSNWEB you can use the `mergeConfig()` method to merge two configs, so do:
```php
ControllerThatShouldBeExtended::extend(function ($controller) {
    $controller->listConfig = $controller->mergeConfig('/original/config.yaml', 'overriding/config.yaml');
});
```

iboved:
@LukeTowers thanks! It works really cool.

LukeTowers:
@iboved, no problem, happy to help! Are you able to suggest any improvements to the documentation in the form of a PR to octobercms/docs?

PR #2881, opened 2017-05-20
Typo

gergo85:
It is low priority changes. Yesterday I found them too, when I fixed the array index name issue.


daftspunk:
Where is typo?

gergo85:
I only added some extra spaces.

PR #2885, opened 2017-05-22
Closing div tag matches nothing

PaulGwamanda:
Removing redundant closing div tag on line 64


daftspunk:
Does it not close `<div id="layout-canvas">`?

PaulGwamanda:
Quite right,  the 'flyout-container' div is left open on line 18 for the php tag and closed in line 24 :+1: 
My phpstorm did not account for this

Issue #1233, opened 2015-06-23
Media Finder: add imageWidth and imageHeight parameters


gergo85:
It would be great, if the Media finder widget could support the imageWidth and imageHeight parameters like File upload widget.

matthewerskine:
+1

Almusamim:
+1

Almusamim:
1+

it would also be nice to have auto crop popup opening with image upload option.. something like cropModal: true

khoatran:
+1 Exactly what I need

LukeTowers:
The support demand for October is increasing, while the available time the founders have to give remains the same. These issues will be handled in time, however there are ways you can help:

- Submit a pull request that introduces this feature and wait for it to be reviewed
- [Post a bug bounty](https://www.bountysource.com/teams/october) for this specific issue so that more people can become involved with creating a fix
- [Make a financial contribution to the project](http://octobercms.com/fundraising) so we can allocate more time towards making improvements like these that benefit everyone

daftspunk:
Fixed by #2894 and the great @gergo85!

daftspunk:
This should be available in Build 420+

PR #2894, opened 2017-05-25
Add imageWidth and imageHeight parameters

gergo85:
#1233

### DIFF
--- a/modules/cms/formwidgets/mediafinder/assets/js/mediafinder.js
+++ b/modules/cms/formwidgets/mediafinder/assets/js/mediafinder.js
@@ -43,9 +43,18 @@
 
         this.$el.one('dispose-control', this.proxy(this.dispose))
 
+        if (this.options.thumbnailWidth > 0) {
+            this.$el.css('maxWidth', this.options.thumbnailWidth + 'px')
+        }
+
+        else if (this.options.thumbnailHeight > 0) {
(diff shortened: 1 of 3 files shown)


modules/cms/formwidgets/mediafinder/assets/js/mediafinder.js
```diff
@@ -43,9 +43,18 @@
 
         this.$el.one('dispose-control', this.proxy(this.dispose))
 
+        if (this.options.thumbnailWidth > 0) {
+            this.$el.css('maxWidth', this.options.thumbnailWidth + 'px')
+        }
+
+        else if (this.options.thumbnailHeight > 0) {
```
LukeTowers: Why is this an elseif instead of a separate if statement?

gergo85:
The imageWidth and imageHeight parameters of Media finder formwidget are not equivalent to Fileupload formwidget parameters.
In some cases the preview image is too big, therefore I want to set the maximum width or height of preview image.

LukeTowers:
If they are not directly equivalent, then I think they should be named differently. Perhaps `previewImageWidth` and `previewImageHeight`?

daftspunk:
LGTM

daftspunk:
Please add documentation for this https://github.com/octobercms/docs

PR #2898, opened 2017-05-30
Added Extensibility to ReportWidgets

arzola:
Added an extensibility Event to the WidgetManager class to be able to "extends" or "remove" a *report widget*

Now we can hide default report widgets without touch nothing on the backend

```php
\Event::listen('system.reportwidgets.extendItems', function ($manager) {
            $manager->removeReportWidgetItem('Cms\ReportWidgets\ActiveTheme');
            $manager->removeReportWidgetItem('System\ReportWidgets\Status');
        });
```

PS.
I added two Unit Tests

### DIFF
--- a/modules/backend/classes/WidgetManager.php
+++ b/modules/backend/classes/WidgetManager.php
@@ -223,9 +223,11 @@ class WidgetManager
     }
 
     /**
-     * Removes a single report widget item
+     * Remove a registered ReportWidget.
+     * @param string $className Widget class name.
+     * @return void
      */
-    public function removeReportWidgetItem($className)
+    public function removeReportWidget($className)
     {
         if (!$this->reportWidgets) {
             throw new SystemException('Unable to remove a widget before widgets are loaded.');
--- a/modules/backend/classes/WidgetManager.php
+++ b/modules/backend/classes/WidgetManager.php
@@ -2,6 +2,7 @@
 
 use Str;
 use System\Classes\PluginManager;
+use Event;
 
 /**
  * Widget manager
@@ -188,6 +189,11 @@ class WidgetManager
             }
         }
 
+        /*
+         * Extensibility
+         */
+        Event::fire('system.reportwidgets.extendItems', [$this]);
+
         return $this->reportWidgets;
     }
 
@@ -215,4 +221,17 @@ class WidgetManager
     {
         $this->reportWidgetCallbacks[] = $definitions;
     }
+
+    /**
+     * Removes a single report widget item
+     */
+    public function removeReportWidgetItem($className)
+    {
+        if (!$this->reportWidgets) {
+            throw new SystemException('Unable to remove a widget before widgets are loaded.');
+        }
+
+        unset($this->reportWidgets[$className]);
+    }
+
 }
--- a/tests/unit/backend/classes/WidgetManagerTest.php
+++ b/tests/unit/backend/classes/WidgetManagerTest.php
@@ -13,4 +13,36 @@ class WidgetManagerTest extends TestCase
         $this->assertArrayHasKey('TestVendor\Test\FormWidgets\Sample', $widgets);
         $this->assertArrayHasKey('October\Tester\FormWidgets\Preview', $widgets);
     }
+
+    public function testIfWidgetsCanBeExtended()
+    {
+        $manager = WidgetManager::instance();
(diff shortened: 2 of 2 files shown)


arzola:
Whoops, Travis HHVM it's broken
`HHVM is no longer supported on Ubuntu Precise. Please consider using Trusty with `dist: trusty`.`

LukeTowers:
I've notified @daftspunk, thanks @arzola!

seanthepottingshed:
@arzola Excellent

arzola:
@seanthepottingshed Thanks, i hope October Team approve the PR 🤙

daftspunk:
This is a solution without a problem. What's the problem we are trying to solve here? You want to modify the [default dashboard](https://github.com/octobercms/october/blob/master/modules/backend/controllers/index/config_dashboard.yaml)? Is that correct?

arzola:
Thanks for your answer

Yeah, i know that i can modify the config_dashboard.yaml but i don't wan't to touch that and even i don't want that the user use those widgets because doesn't works for my customer, i think it's a nice option to give the the developer the option to remove them

daftspunk:
> i don't want that the user use those widgets because doesn't works for my customer

Located the problem. So we need a way to extend the default dash widgets (config file) in the code. Allowing new ones to be provided and existing ones to be removed. For example: Google Analytics plugin might like to include its visitor line chart in the default configuration.

Am I correct in saying this Pull Request will remove the widgets completely, the same as de-registering them, not even allowing the user to add them through customization? Is that your intention?

To clarify, do you want to:

(a) Completely destroy a widget and prevent it from being used at all
(b) Modify the default dashboard widgets given by the system
(c) Both

arzola:
Hi 👋🏻 

I need both, consider this escenario:

Sometimes the end user doesn't know anything about HTML, themes or even plugin installation if the user has permissions to view the desktop but not system updates, that will be confusing that the user can append those system/cms reportwidgets and be notified about the system updates, i think this PR allows to the developer remove those widgets securely and transparent while we will rethink the default dashboard to the end user

modules/backend/classes/WidgetManager.php
```diff
@@ -215,4 +221,17 @@ public function registerReportWidgets(callable $definitions)
...
+    public function removeReportWidgetItem($className)
+    {
+        if (!$this->reportWidgets) {
+            throw new SystemException('Unable to remove a widget before widgets are loaded.');
+        }
+
+        unset($this->reportWidgets[$className]);
+    }
+
+}
```
daftspunk: Place a newline here.

modules/backend/classes/WidgetManager.php
```diff
@@ -215,4 +221,17 @@ public function registerReportWidgets(callable $definitions)
     {
         $this->reportWidgetCallbacks[] = $definitions;
     }
-}
+
+    /**
+     * Removes a single report widget item
+     */
+    public function removeReportWidgetItem($className)
```
daftspunk: Rename to `removeReportWidget` and include docBlock

```
* @param string $className
* @return void
```

tests/unit/backend/classes/WidgetManagerTest.php
```diff
@@ -13,4 +13,36 @@ public function testListFormWidgets()
...
+        $manager->registerReportWidget('Acme\Fake\ReportWidget\HelloWorld', [
+            'name' => 'Hello World Test',
+            'context' => 'dashboard'
+        ]);
+        $manager->registerReportWidget('Acme\Fake\ReportWidget\ByeWorld', [
+            'name' => 'Hello World Bye',
+            'context' => 'dashboard'
+        ]);
+
+        $manager->removeReportWidgetItem('Acme\Fake\ReportWidget\ByeWorld');
```
daftspunk: Ding.

arzola:
Thanks buddy

It's done

LukeTowers:
@arzola Could you please add a docblock to the event like this: https://github.com/octobercms/october/blob/master/modules/cms/classes/Theme.php#L174-L183?

It will be used to generate the API docs here: http://octobercms.com/docs/api/cms/theme/getactivetheme

arzola:
For sure @LukeTowers

daftspunk:
Good pick up my dude

PR #2913, opened 2017-06-07
Fixed the message variable issue

gergo85:
##### Reproduce steps
1. Create a new mail template (Settings > Mail templates).
2. Add the {{message}} variable for the HTML content of template.
3. Send a test message.

The October will show the following error message:
error

### DIFF
--- a/modules/system/models/MailLayout.php
+++ b/modules/system/models/MailLayout.php
@@ -19,9 +19,9 @@ class MailLayout extends Model
     protected $table = 'system_mail_layouts';
 
     public $rules = [
-        'code'                  => 'required|unique:system_mail_layouts',
-        'name'                  => 'required',
-        'content_html'          => 'required',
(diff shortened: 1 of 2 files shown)


modules/system/models/MailTemplate.php
```diff
@@ -161,11 +164,15 @@ public static function addContentToMailer($message, $code, $data)
         /*
          * HTML contents
          */
+        if (preg_match('{{(\s*message*\s*)}}', $template->content_html)) {
+            throw new ApplicationException('Do not use the \'message\' variable.');
```
LukeTowers: This is not a very good solution, it's a bandaid at best. Simply telling people that they can't use `{{ message }}` is not good enough, we'll need to better examine the problem and come up with a more robust solution that properly isolates the Message object from being used as a variable within the template.

LukeTowers:

PR #2922, opened 2017-06-14
Implement api-module

khoatran:
This PR is for creating a separated module for API development using dingo/api. 
Actually, you can implement the API inside your own plugin. However, the bootstrap process of the the whole system will affect to the performance of the API. 

With this implementation, if you want just a separated API host, what you need to do is only changing the loadModules in cms.php configuration to load only the api module.
This module also helps to take advantage of developing API inside OctoberCMS easier following guideline at dingo/api library.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,8 @@
         "october/backend": "~1.0",
         "october/cms": "~1.0",
         "laravel/framework": "5.1.*",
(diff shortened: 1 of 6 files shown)


SebastiaanKloos:
This looks interesting!👍

daftspunk:
This can easily be a plugin.

Refs https://github.com/rainlab/user-plugin/issues/271

Issue #2394, opened 2016-09-28
Backend session expiration


mosid:
##### Expected behavior

Backend session expires when setting the 'expire_on_close' to true or setting lifetime to zero or whatever.
##### Actual behavior

Session options 'expire_on_close' and 'lifetime' are bypassed even after setting them. session never expires.
##### Reproduce steps
- open `session.php` in config folder
- set `expire_on_close=>true` and `lifetime=>0`  
##### October build

363

Nyorai:
I've come across this and was trying to find a solution but this behaviour is hardcoded into `October\Rain\Auth\Manager` and `Backend\Controllers\Auth`. The controller calls the `authenticate` method of the manager with `$remember` set to `true`, which then creates a forever cookie in the `login` method (line 420).

```
if ($remember) {
    Cookie::queue(Cookie::forever($this->sessionKey, $toPersist));
}
```

The `expire_on_close` and `lifetime` settings only apply to the `october_session` cookie, not the `admin_auth` cookie which is used on the backend or the `user_auth` which is user by the Rainlab User plugin.

Nyorai:
Forgive me, I'm not fully up to speed with Github and pull-requests, but I've made some changes to the backend module to deal with this issue.

https://github.com/Nyorai/october/pull/1

daftspunk:
I'm seeing two different issues here.  This issue lacks enough detail or the replication instructions cannot be actioned easily. As per [the contribution guidelines](https://github.com/octobercms/october/blob/master/CONTRIBUTING.md#reporting-a-bug-with-octobercms), please provide the exact steps you are taking for the issue to occur. This will be used for peer review.

Here is an example of a **bug summary**:

> The CMS page URL is not being set automatically from the title value.

Here is an example of some **reproduce steps**:

> 1. In the back-end, open the CMS > Pages section
> 1. Click the "Add" button
> 1. Enter [Hello] for the Title field
> 1. Click the Save button 

Here is an example of **expected behavior**:

> - The Title should be set to [Hello]
> - The URL should be set to [/hello]

Here is an example of **actual behavior**:

> - The Title is set to [Hello] correctly
> - The URL is left empty

LukeTowers:
Fixed by #2924

PR #2924, opened 2017-06-14
Backend login allow user to have not persistent connection

ChVuagniaux:
Currently all backend connection are persistent ( the lifetime of the cookie is 5 year ), this is perfect for many case. But for project that require more security this is not very good.

This PR add a config parameter `cms.backendLoginRemember` that allow administrators to choose the wanted behavior of this remember me feature.

value | description
-- | :---
true (by default) | like now : session not expire
false | session life time is the value of `session.lifetime` (2 hours by default)
null | display a checkbox on the login page that allow user to choose 

This PR can fix the following issue #2394

*This commit is inspired from the PR of @Nyorai ( https://github.com/Nyorai/october/pull/1)*

### DIFF
--- a/config/cms.php
+++ b/config/cms.php
@@ -53,6 +53,24 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Back-end login remember
+    |--------------------------------------------------------------------------
+    |
+    | Define live duration of backend sessions :
+    |
+    | true  - session never expire (cookie expiration in 5 years)
+    |
+    | false - session have a limited time (see session.lifetime)
+    |
+    | null  - The form login display a checkbox that allow user to choose
+    |         wanted behavior
+    |
+    */
+
+    'backendForceRemember' => true,
+
+    /*
+    |--------------------------------------------------------------------------
     | Back-end timezone
     |--------------------------------------------------------------------------
     |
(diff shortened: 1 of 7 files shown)


modules/backend/lang/en/lang.php
```diff
@@ -39,6 +39,7 @@
         'restore' => 'Restore',
         'login_placeholder' => 'login',
         'password_placeholder' => 'password',
+        'remember_me' => 'Keep me logged in?',
```
LukeTowers: Change this string to "Stay logged in" please

config/cms.php
```diff
@@ -329,4 +329,21 @@
...
+    | true  - session never expire (cookie expiration in 5 years)
+    |
+    | false - session have a limited time (see session.lifetime)
+    |
+    | null  - The form login display a checkbox that allow user to choose
+    |         wanted behavior
+    |
+    */
+
+    'backendLoginRemember' => true,
```
LukeTowers: Please rename to `backendForceRemember`, and move this to just below https://github.com/octobercms/october/blob/master/config/cms.php#L52

ChVuagniaux:
@LukeTowers I have applied requested changes

modules/backend/controllers/auth/signin.htm
```diff
@@ -32,6 +32,21 @@
             </button>
         </div>
 
+        <?php if (is_null(config('cms.backendForceRemember', true))) : ?>
+        <!-- Remember checkbox -->
```
LukeTowers: Could you please indent the contents of this if block?

LukeTowers:
@ChVuagniaux as soon as you make the one other minor formatting change, I'm fine with merging this, assuming that @daftspunk doesn't have any other input on it.

Thanks for your work, it's a nice and simple solution to a problem that's been on my backburner for months now :)

ChVuagniaux:
@LukeTowers change done

Happy to contribute one more time to this AWESOME CMS :)

What do you think of a similar approach for the [rainlab/user-plugin](https://github.com/rainlab/user-plugin) with a settings in backend configuration ?

LukeTowers:
I like the idea, perhaps also enabling defaulting session length to the config value but being able to override it with a plugin setting?

PR #2940, opened 2017-06-23
Added firstDay option to datepicker

tobias-kuendig:
To my knowledge there is currently no easy way to change the first day of the week in the datepicker to something else than Sunday. 
This PR adds a `firstDay` option to the datepicker form widget so the first day of the week can be set via the yaml definition files. 

```yaml
        delete_at:
            label: Delete at this date
            mode: datetime
            format: d.m.Y
            firstDay: 1
```

See https://github.com/octobercms/docs/pull/243 for the updated docs.


modules/backend/formwidgets/DatePicker.php
```diff
@@ -46,6 +46,12 @@ class DatePicker extends FormWidgetBase
      */
     public $yearRange = null;
 
+    /**
+     * @var int first day of the week
+     * eg: 0 (= Sunday) or 1 (= Monday)
```
LukeTowers: This is a very minor change, but please change this comment to read `eg: 0 (Sunday), 1 (Monday), 2 (Tuesday), etc.` to make it more clear that this isn't just a boolean flag, it is an integer option that can support any day being the first day of the week.

tobias-kuendig: Makes sense! Changed it.

LukeTowers:
Awesome, thanks for your work on this @tobias-kuendig!

PR #2942, opened 2017-06-26
Translate the value of the text field

gergo85:
It is a useful feature, if you like to add translatable value in the yaml file.


daftspunk:
I have to roll this back because there is no use case for it. Saying something is useful is not good enough without evidence to support it. There are probably much better ways to handle this, the text input value should be explicit, not mutated, ever.

PR #2945, opened 2017-06-29
Added the backend.user.created event

priitperna:
The event is fired after a new backend user is created.
This makes backend users a bit more extendable.


daftspunk:
Can't you get to this already using

```
Backend\Models\User::creating(function($model) { });
```

Or even

```
Backend\Models\User::extend(function($model) {
    $model->bindEvent('model.afterCreate', function() { });
});
```

priitperna:
Indeed, you are correct. I was looking at the User model and missed the general model events. Sry, my bad.

PR #2948, opened 2017-06-30
update modules/backend/lang/zh-cn/lang.php file

buuug7:
this file can't keep up with the `en/lang.php` step, i update `modules/backend/lang/zh-cn/lang.php` file, keep synchronized with  `en/lang.php`

PR #2952, opened 2017-07-03
update modules/backend/lang/zh-cn/lang.php

buuug7:
compared with `en/lang.php`, the `zh-cn/lang.php` is outdate ,so i update `modules/backend/lang/zh-cn/lang.php`


LukeTowers:
Already merged in #2948

PR #2958, opened 2017-07-07
add KEY taglist

blubcow:
Add the ability to make a taglist formwidget, but use the key instead of the value to save data.
See my merge request on "oc-test-plugin" for examples.

Many times we want to save a huge list of tags into a field, but we can't use the name (or value). We want a list of IDs or keys to create very fast queries.
______________________________________________________
Sorry,
there was something missing in my commit.
Updated my commit:
https://github.com/sewagolubkow/october/commit/59a66fec60348c6782d634770cb6116993307675

(In case the link doesn't work - it was a forced commit update)
______________________________________________________
And here is the updated "oc-test-plugin":
https://github.com/daftspunk/oc-test-plugin/pull/40

### DIFF
--- a/modules/backend/formwidgets/TagList.php
+++ b/modules/backend/formwidgets/TagList.php
@@ -41,6 +41,11 @@ class TagList extends FormWidgetBase
      * @var string If mode is relation, model column to use for the name reference.
      */
     public $nameFrom = 'name';
+	
+	/**
+     * @var bool Use the key instead of value for saving and reading data
+     */
+    public $useKey = false;
 
     //
     // Object properties
@@ -62,6 +67,7 @@ class TagList extends FormWidgetBase
             'options',
             'mode',
             'nameFrom',
+            'useKey'
         ]);
     }
 
@@ -71,7 +77,12 @@ class TagList extends FormWidgetBase
     public function render()
     {
         $this->prepareVars();
-        return $this->makePartial('taglist');
+		
+		if (($this->mode !== static::MODE_RELATION) && ($this->useKey === true)) {
+			return $this->makePartial('taglist_key');
+		}else{
+			return $this->makePartial('taglist');
+		}
     }
 
     /**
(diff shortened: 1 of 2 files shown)


sewagolubkow:
Update test to work with KEY-taglist:
https://github.com/daftspunk/oc-test-plugin/pull/40

sewagolubkow:
Sorry,
there was something missing in my commit.
Updated my commit: https://github.com/octobercms/october/pull/2958/commits/fc2e1faff8450a6f5787582e5718dd39e4005f8c

(In case the link doesn't work - it was a forced commit update)

LukeTowers:

LukeTowers:
I'm not sure I fully understand the problem here, could you please post what data is currently saved / used by the existing taglist widget and what data you need it to be saving, along with the fields.yaml configuration of the widget that you use to accomplish your goals?

sewagolubkow:
Ok, sorry, I will follow the guidelines strictly next time.

## Currently, we would setup a taglist like that:

#### `models/model/fields.yaml`
```
tags:
    type: taglist
    mode: string
    options: true
```
#### `models/model.php`
```
public function getTagsOptions($value, $formData){
    return Tag::all()->lists('name');
}
```
#### And the saved data would look like that
```
tag1,tag2,tag3,tag4,tag5
```

## But I like to setup a taglist like that
> What this changes are achieving, as you can test out in (https://github.com/daftspunk/oc-test-plugin/pull/40/commits/ff2cdff6b68573004355bb3f78218e7428f70708)
is saving the IDs, and not the values of specific tags.

#### `models/model/fields.yaml`
```
tags:
    type: taglist
    mode: string
    options: true
```
#### `models/model.php`
```
public function getTagsOptions($value, $formData){
    return Tag::all()->pluck('name', 'id')->toArray();
}
```
#### And the saved data would look like that
```
1,2,3,4,5
```

> In this case we can display all kinds of different information in our taglist field. Many problems are solved, such as displaying a comma or a space, when we are using them as separators for our data. Or just plain translations. Or displaying the ID and the name in one. Or think about translating the tags.
Also, we can make much faster queries when saving many tags.

daftspunk:
LGTM but rewritten as 45f1714a039cee87f324e76877bee5d7f0143a76

Cheers! 🍻

Issue #2953, opened 2017-07-03
No 404 status code set when visiting the /404 page when you aren't logged in


maxibanki:
##### Expected behavior
http 404 status code

##### Actual behavior
http 200 status code

##### Reproduce steps
1. Set content to the 404 page
2. Visit the /404 page

##### October build
419

maxibanki:
Sorry, was my fault. It's working^^

maxibanki:
Sorry for reopening the issue, but I can reproduce the bug. When you are logged in via the backend and have so the cookie set the correct http status code (404) will be shown. When you aren't logged in (removed the cookie) like a anonymous user, a http 200 status code will be returned.
I hope I can help you with these information.

maxibanki:
When you are not logged in, the $response will be recognized as a string. Ref to:
https://github.com/octobercms/october/blob/master/modules/cms/classes/Controller.php#L227-L229

LukeTowers:
Merged #2959

PR #2959, opened 2017-07-07
Adding getStatusCode() in Cms\Classes\Controller

maxibanki:
Referring to the issue #2953 which is no fault of the October CMS.
The problem was that a plugin which I'm using [xeor-minify](https://octobercms.com/plugin/xeor-minify) sets the http status code hardcoded. So I want to improve it and add the ability to get the current http status code of the page (which can use the plugin above in the next version).


LukeTowers:
LGTM, thanks!

Issue #2949, opened 2017-07-01
Improved october:env behaviour


SaifurRahmanMohsin:
Running october:env moves the environment values to the .env file but also keeps a copy of that value as default value in the env format i.e. as `env('ENV_VARIABLE_NAME', 'value');`. This isn't good because when I create a new git repo with the files except .env it still has all the database credentials and the application key inside the files as the default values now.

##### Expected behavior
It should remove the values completely and leave the default value as some common default such as 'localhost', 'root', '127.0.0.1', 'defaultKey', etc.

##### Actual behavior
It extracts the value and replaces it with the env function as `env('ENV_VARIABLE_NAME', 'value');`.

##### Reproduce steps
* Create a new OctoberCMS project. Install with any workflow (From project ID, scratch, theme)
* Regardless of the workflow, the installer would generate application key and store DB credentials in various files (database.php, app.php).
* Initialize a git repo with a .gitignore having the .env to prevent credentials going upstream and other stuff such as /uploads, etc.
* Make first commit. The repo is now infested with the application key as well as the database credentials used to setup the project.

##### October build
419

daftspunk:
Feel free to submit a PR that addresses this.

LukeTowers:
https://github.com/octobercms/october/pull/2963 in progress to fix this.

LukeTowers:
Implemented by https://github.com/octobercms/october/commit/0ac7f340be7fb8aba8427b318d62ed9bce189653

PR #2963, opened 2017-07-13
Replace sensitive config values with defaults for october:env command

BenjaVR:
My suggestion to fix issue #2949 

Added an option (`--protect`) to the `october:env` command, to replace some config values by a default, generic, value.

### DIFF
--- a/modules/system/console/OctoberEnv.php
+++ b/modules/system/console/OctoberEnv.php
@@ -31,6 +31,11 @@ class OctoberEnv extends Command
     protected $config;
 
     /**
+     * Default configuration.
+     */
+    protected $defaultConfig = $this->protectedConfig();
+
+    /**
      * The current database connection cursor.
      */
     protected $connection;
@@ -63,7 +68,7 @@ class OctoberEnv extends Command
     protected function getOptions()
     {
         return [
-            ['protect', null, InputOption::VALUE_NONE, 'Make sure to replace sensitive config values by the default ones.'],
+            ['protect', null, InputOption::VALUE_NONE, 'Replaces sensitive config values in the config files with the default values.'],
         ];
     }
 
--- a/modules/system/console/OctoberEnv.php
+++ b/modules/system/console/OctoberEnv.php
@@ -1,6 +1,7 @@
 <?php namespace System\Console;
 
 use Illuminate\Console\Command;
+use Symfony\Component\Console\Input\InputOption;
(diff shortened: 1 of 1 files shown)


modules/system/console/OctoberEnv.php
```diff
@@ -394,4 +409,31 @@ private function dbConfig()
         ];
     }
 
+    /**
+     * Config (env) keys of which the value should not be included in orginal config files anymore.
+     *
+     * @return array
+     */
+    private function protectedConfig()
```
LukeTowers: This should either be named `getDefaultConfig()` or made available as a `defaultConfig` property on the class.

modules/system/console/OctoberEnv.php
```diff
@@ -57,6 +58,16 @@ public function fire()
     }
 
     /**
+     * Get the console command options.
+     */
+    protected function getOptions()
+    {
+        return [
+            ['protect', null, InputOption::VALUE_NONE, 'Make sure to replace sensitive config values by the default ones.'],
```
LukeTowers: Description of option should be "Replaces sensitive config values in the config files with the default values"

LukeTowers (changes requested):

modules/system/console/OctoberEnv.php
LukeTowers: I didn't mean `$defaultConfig = $this->protectedConfig`. I meant literally
```php
$defaultConfig = [
    'APP_KEY' => 'CHANGE_ME!!!!!!!!!!!!!!!!!!!!!!!',
];
```

BenjaVR: My bad. Do you mean like I did it now?

modules/system/console/OctoberEnv.php
```diff
@@ -184,6 +217,10 @@ private function buildCallback($envKey, $configKey)
 
             $this->saveEnvSettings($envKey, $value);
 
+            if ($this->option('protect') && array_key_exists($envKey, $defaultConfig)) {
```
LukeTowers: Does this actually work as it is right now? I don't see where you are actually pulling the $defaultConfig variable into this function for using.

LukeTowers:
Implemented by https://github.com/octobercms/october/commit/0ac7f340be7fb8aba8427b318d62ed9bce189653

Issue #2900, opened 2017-05-31
"php artisan october:up" fails when CMS module is disabled and removed.


Upperfoot:
##### Expected behavior
Removing CMS Module disables CMS specific functionality but does not affect anything else.

##### Actual behavior
Running the command `php artisan october:up` fails because Cms\Classes\ThemeManager does not exist.

##### Reproduce steps

- Remove the CMS module from the codebase
- Remove the entry in loadModules in config/cms.php
- Run the command `php artisan october:up`
- Command throws an error because Cms\Classes\ThemeManager does not exist when called at https://github.com/octobercms/october/blob/master/modules/system/classes/UpdateManager.php#L90

##### October build
419

##### Suggested Fix
I would suggest that we need to check if the CMS module exists and whether `Cms\Classes\ThemeManager` exists before trying to utilise it, if the ability is given to disable these modules we should make use of a few checks to check whether modules have been enabled/disabled.

LukeTowers:
The support demand for October is increasing, while the available time the founders have to give remains the same. These issues will be handled in time, however there are ways you can help:

- Submit a pull request that introduces this feature and wait for it to be reviewed
- [Post a bug bounty](https://www.bountysource.com/teams/october) for this specific issue so that more people can become involved with creating a fix
- [Make a financial contribution to the project](http://octobercms.com/fundraising) so we can allocate more time towards making improvements like these that benefit everyone

LukeTowers:
This is fixed in 81b28e3...e1629a9

PR #2970, opened 2017-07-14
Safely remove CMS module

BenjaVR:
This is a small fix for issue #2900 
It solves the specific use case of the issue, but it's not complete at all. The CMS module is probably used at other places as well. @LukeTowers already pointed out _"the MediaManager is located within the CMS module and is used within the backend and system modules"_.

My fix might be a dirty one, so other suggestions are welcome!


daftspunk:
This is fixed in 81b28e393872...e1629a941364

PR #2971, opened 2017-07-14
Add support for hasManyThrough relationships in backend lists

purposebuiltscott:
When trying to show a list of hasManyThrough relationships, the backend is unable to create the widget because it does not have the hasManyThrough relationship type identified in the evalViewMode method.  By adding this value in the check, the backend is able to successfully render the relationship list.


LukeTowers:
Duplicate of https://github.com/octobercms/october/pull/2509

PR #2973, opened 2017-07-14
Do not render the Theme Widget if CMS not loaded

filipac:
So, if the CMS module is not loaded, the default Theme report widget is still shown on the dashboard. This looks nasty, because translations are not resolved.

Also it makes no sense to render this widget. No CMS => no themes at all.

This PR returns false in the render function for the widget if the CMS module is not loaded.
screen shot 2017-07-15 at 00 46 33

### DIFF
--- a/modules/cms/reportwidgets/ActiveTheme.php
+++ b/modules/cms/reportwidgets/ActiveTheme.php
@@ -2,6 +2,7 @@
 
 use Lang;
 use Cms\Classes\Theme;
+use Cms\ServiceProvider;
 use Cms\Models\MaintenanceSetting;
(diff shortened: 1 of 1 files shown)


LukeTowers:
I don't think this is the correct solution to the problem. If the CMS module isn't loaded, then this widget shouldn't even exist. It would be better to do the check before even registering the widget to the dashboard. I think we need to look at getting a cleaner method in general for checking if various modules are disabled or not given the relatively large number of sections we need to check for modules being enabled.

filipac:
@LukeTowers where do you suggest to have this method? Maybe like a method in `October\Rain\Foundation\Application` class?

LukeTowers:

PR #2976, opened 2017-07-15
Added `october:installed` artisan command to clean files from installer

BenjaVR:
When the installer is done it says you should delete the installer files. This command should take care of that.

Don't know if this is prefered as plugin instead, or just considered unnecessary.


LukeTowers:
@BenjaVR couldn't we just make the installer itself do the cleaning? That would make the most sense to me.

Issue #2990, opened 2017-07-24
[Suggestion] Data from jsonable field into BackendList


Samorai:
Hi to everyone!

##### Expected behavior
I have model with jsonable field named **foo** with values `{"bar":"baz"}`.
When I try to render it into Backend List, I'have wrote this code in columns.yml
```
foo.sss:
    type: text
```
And I wanted to see in list "baz".

##### Actual behavior
I had saw nothing

##### Resolving
I have added custom list column type named "array"
```
public function registerListColumnTypes()
    {
        return [
            'array' => function($value, $column, $record) {
                list($key, $values) = explode('.', $column->columnName);
                return array_get($record->$key, $values);
            }
        ];
    }
```
And change data into columns.yml file to 
```
foo.sss:
    type: array
```

##### Suggestion
I suggest to add evalArrayTypeValue to Backend\Widgets\Lists.

If it's OK, I can create PR with this feature.

LukeTowers:
If this were to be added, the syntax for accessing a jsonable property should be the same as in `fields.yaml`: `property[key]`

Samorai:
@LukeTowers 
The syntax for accessing jsonable field does not matters for me. 
But if you suggest the syntax `property[key]`, how you can access for lower properties like `property[key][key2][key3]` without errors if key does not exists?
Function array_get has resolve this problem.

LukeTowers:
Well, either way you can resolve the issue right now by using a column type `partial` and accessing the value you wish to display directly through `<?= $record->property['key'] ?>`

Samorai:
I know. I have resolved the issue using custom list column type.
But I don't really want to move this part of code between the projects. 
And I have suggest to discuss the syntax, what would I add the PR with resolving the issue globally for everyone.

LukeTowers:
Just a couple side comments:

1. Support would not include sorting or searching support by default, anybody that wanted to utilize this on their columns would need to set `searchable: false` and `sortable: false` on the columns that utilized this to prevent the list controller's search or sort functionality from breaking.

2. I don't think that adding a new `type: array` is the correct way of solving this, since any type of data can be stored in a jsonable property, so a solution should implement it at the data retrieval level, not the rendering level.

Samorai:
When I have jsonable value `{"foo":"foo_value", "bar": "bar_value"}` and I try to render with 
`type: text`. I have `foo_value, bar_value` in output.

LukeTowers:
Not really sure what you mean by that @Samorai

[...]

PR #2994, opened 2017-07-28
#2990 Getting data from jsonable fields into backend lists

Samorai:
Example: 

Foo jsonable property: {"date":"2017-01-01 20:10:32", "info": "some text"}

YAML Column config:
foo[date]:
    type: datetime
foo[info]:
    type: text


Samorai:
Resolving #2990

Samorai:
@LukeTowers  There are one fail with hhvm
HHVM is no longer supported on Ubuntu Precise. Please consider using Trusty with `dist: trusty`.

PR #2995, opened 2017-07-28
Getting data from jsonable fields into backend lists

Samorai:
Example:

Foo jsonable property: {"date":"2017-01-01 20:10:32", "info": "some text"}

YAML Column config:
foo[date]:
type: datetime
foo[info]:
type: text

Resolving #2990

### DIFF
--- a/modules/backend/classes/ListColumn.php
+++ b/modules/backend/classes/ListColumn.php
@@ -232,6 +232,9 @@ class ListColumn
                 $result = $result->{$key};
             }
             else {
+                if (is_array($result) && array_key_exists($key, $result)) {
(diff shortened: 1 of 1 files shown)


LukeTowers:
@Samorai does this cause any problems when a `foo[date]` column attempts to be sorted by or searched within? Do we need to add additional logic to disable searching and sorting on array values in backend lists?

Samorai:
@LukeTowers there are no any problems. Because `searchable` and `sortable` attributes disabled for array values by default

LukeTowers:
Awesome, thanks for this @Samorai!

Issue #3002, opened 2017-08-01
[L5.5] MailBrandSettings controller has hardcoded texts


jan-vince:
##### Expected behavior
Translations should work for texts in *Customize mail appearance part* of administration.

##### Actual behavior
All text are hardcoded.

##### October build
420

daftspunk:
Fixed by #3004

PR #3004, opened 2017-08-01
Added translation for pageTitle into MailBrandSettings

Samorai:
Resolving #3002

### DIFF
--- a/modules/system/controllers/MailBrandSettings.php
+++ b/modules/system/controllers/MailBrandSettings.php
@@ -54,7 +54,7 @@ class MailBrandSettings extends Controller
     {
         parent::__construct();
 
-        $this->pageTitle = 'Customize mail appearance';
+        $this->pageTitle = 'system::lang.mail_brand.page_title';
 
         BackendMenu::setContext('October.System', 'system', 'settings');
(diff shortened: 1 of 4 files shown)


jan-vince:
Hi, this is a good start - but there is a whole config file in ````/modules/system/models/mailbrandsetting/fields.yaml```` that needs to be translated.

And we also have to deal with ````/modules/system/models/mailbrandsetting/samle_template.html````. Texts inside should be translated with ````<?= e( trans(...) ) ?>````.

Have you time to include this in your PR?

Samorai:
@jan-vince Hi, thanks.
I'll include that to my PR tonight. (EEST GMT +3)

jan-vince:
It's not a race :) 

Thanks!

LukeTowers:
I'll let @daftspunk review this since he's the one that has been putting a lot of work into the mail system recently

jan-vince:
Hi @Samorai, I have tested this and it works form me. Great job!

*I just can not figure out how to align a 'Right-Aligned' cell to right side. It looks like MD table doesn't allow to change forced alignment. If so, we should probably change the text.*

daftspunk:
Looks great! <3 you guys

PR #3012, opened 2017-08-03
Do not change behavior: pass request data back to the options object

alxy:
Current changes to the ``framework.js`` led to a breaking change due to the fact, that ``options.data`` was no longer populated with the request data. The ``options`` object is part of the ``context`` object, which is later beeing passed to ``ajaxBeforeSend``
The two relevant versions of the file show the change of behavior:
https://github.com/octobercms/october/blob/b6fedfb688e2d99461fd62b99bc8412f38502180/modules/system/assets/js/framework.js#L56 => https://github.com/octobercms/october/blob/develop/modules/system/assets/js/framework.js#L88
The change is required to remain consistent with the old behavior. This code is tested and appears to be working as it did formerly.

### DIFF
--- a/modules/system/assets/js/framework.js
+++ b/modules/system/assets/js/framework.js
@@ -78,15 +78,15 @@ if (window.jQuery.request !== undefined) {
             $.extend(data, paramToObj('data-request-data', $(this).data('request-data')))
         })
 
-        if (options.data !== undefined && !$.isEmptyObject(options.data)) {
-            $.extend(data, options.data)
-        }
-
         if ($el.is(':input') && !$form.length) {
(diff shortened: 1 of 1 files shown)


tschallacka:
This also works with checkboxes? it seems to me that it would only leave the last checkbox value set and not if multiple are set with the same name.

LukeTowers:
@tschallacka that would be a separate issue. This just moves the order that the data.options is set so that the same information is available as was previously available before the AJAX file upload improvements. If you would like to dig deeper into your theoretical problem and then provide a detailed demonstration of the problem and a solution for it, that would be fantastic.

PR #3014, opened 2017-08-04
Fixed the slow query when Files are attached

### LINKED ISSUE #2877: System file and deferred_bindings query too slow
##### Actual behavior
I'm developing a store plugin, when added some products(more than one thousand), the system files and deferred bindings query is very slow in backend product update page. see below image:

Sometimes the query time will be up to a few seconds, I must to run the `DeferredBinding::cleanUp(1);` command, that query time will come down(but query time still very long, as shown in the picture above).

How to optimize this query? it looks very complicated.

##### October build
All build.

MatheusNaldi:
Created a migration that fixes this adding an index to session_key and changed slave_id to unsigned int in deferred_bindings

This fixes #2877 

I did it on master because it's a majority fix and it has to merged as soon as possible, if you don't accept, I can do it again on develop branch.

### DIFF
--- a/modules/system/database/migrations/2017_08_04_121309_Db_Deferred_Bindings_Add_Index_Session.php
+++ b/modules/system/database/migrations/2017_08_04_121309_Db_Deferred_Bindings_Add_Index_Session.php
@@ -3,7 +3,7 @@
 use Illuminate\Database\Schema\Blueprint;
 use Illuminate\Database\Migrations\Migration;
 
-class DbDeferredBindingsAlterSlaveToIntAndAddUniqueSession extends Migration
+class DbDeferredBindingsAddIndexSession extends Migration
 {
     /**
      * Run the migrations.
--- a/modules/system/database/migrations/2017_08_04_121309_Db_Deferred_Bindings_Add_Index_Session.php
+++ b/modules/system/database/migrations/2017_08_04_121309_Db_Deferred_Bindings_Add_Index_Session.php
@@ -13,7 +13,6 @@ class DbDeferredBindingsAlterSlaveToIntAndAddUniqueSession extends Migration
     public function up()
     {
         Schema::table('deferred_bindings', function (Blueprint $table) {
-            $table->integer('slave_id')->unsigned()->change();
             $table->index('session_key');
         });
     }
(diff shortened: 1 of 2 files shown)


LukeTowers:
@MatheusNaldi thanks for this, however pull requests will not be merged direct to master, they must go through develop. The only commit that has been pushed direct to master was a case of a composer dependency `~1.3` pulling in a breaking change version of that dependency, so people were no longer able to use `compose create project october/october myoctober` because it would be broken from the start with the incorrect version of the dependency.

This issue however is a performance issue, and while this change will improve it, it is not something that I would consider site-breaking. Thank you again for submitting this, I'll review it when I've got the time! :)

LukeTowers:
Now, my only concern with this is that this change will no longer support deferred bindings for slaves where slave models do not use an integer key. This is a breaking change, at least in the theory of it, I have no idea if anyone actually uses deferred bindings with slaves that don't have integer primary keys, so we'll have to get @daftspunk to weigh in on this.

Personally, I think that if we were to add it to the Build 420 Release Note that we have changed this and it will no longer support non-integer primary keys on the slave models, then that should be fine.

MatheusNaldi:
@LukeTowers alright, I understand you to not merge in master directly...it's not really a site-breaking however.

The main of this fix was in the index on session_key, this solved my problem instantly, the unsigned int I have changed for your tip, if it's not right to change it to int, just keep the index of session_key and it will works fine.

LukeTowers:
Are you able to compare the performance of the following and post the numbers from your system?

No Change, deferred_bindings remains as it is in master
Only slave id column type changed to int
Only index added on session key
Index added on session key and column type changed to int

MatheusNaldi:
I can't rollback the fix on our system to take screenshots for you, but here's what I remember.

We have 35k registers in 'system_files' and 1.5k in 'deferred_bindings', the query running for some session_key that do not have rows in deferred_bindings takes 20~30 sec, for someone that have a lot of registers in deferred_bindings takes 2 minutes. (The query that I'm talking is in the issue #2877)

In htop the 8 cores of processor stands almost all the time above 90% (we have a lot of accesses), now 7 cores stands near 1% and 1 of these reaches 15% or less

SSH took 1 minute to change folders, now it's instantly.

It have affected the whole server performance, everything is beautiful again.

LukeTowers:
I was hoping to get a better understanding of the performance impact of the slave_id type change vs the indexing of the session_key, but that's fine if you can't obtain that. 

If you could change this migration to only add the index, I could merge that without requiring review from @daftspunk

MatheusNaldi:
@LukeTowers done!

LukeTowers:
@MatheusNaldi Thanks for your work on this!

PR #3027, opened 2017-08-09
Laravel 5.5 Engine directory change

Teranode:
They moved the Engine interface to Contracts so this is a small fix.


LukeTowers:
I'll need @daftspunk to review this since he's been working with the mail system so presumably he would have encountered this issue before

PR #3039, opened 2017-08-16
Popover filter bug

vakopian:
When popover filters are used in a modal (e.g. a search list for a relation), the date pickers don't open properly (they open and close quickly since the input loses focus).
This change adds the `popup-allow-focus` class to the inputs for the filters to allow for the date picker to stay open.


LukeTowers:
@vakopian does this effect the number and numberrange filters as well?

PR #3041, opened 2017-08-17
Fixed bug caused by not existing use statement

tschallacka:
Caused an error on line 95 where this class was used

### DIFF
--- a/modules/cms/twig/DebugExtension.php
+++ b/modules/cms/twig/DebugExtension.php
@@ -1,5 +1,6 @@
 <?php namespace Cms\Twig;
 
+use Twig_Template;
 use Twig_Extension;
 use Twig_Environment;
 use Twig_SimpleFunction;


LukeTowers:

LukeTowers:
As per [the contribution guidelines](https://github.com/octobercms/october/blob/master/CONTRIBUTING.md#reporting-a-bug-with-octobercms), please provide the exact steps you are taking for the issue to occur. This will be used for peer review.

tschallacka:
steps to reproduce:

Open the october as provided by installer in eclipse without any folders marked as library folders.

it will complain about the not existing class in the use statement. The class is not present anywhere in the library folders, also it's nowhere used in the class itself.

LukeTowers:
Can you trigger PHP itself complaining about the class not existing rather than just your IDE? I understand how it wouldn't work in theory, but I'm curious as to what needs to happen to actually trigger that code path.

tschallacka:
This script checks for unused statements. Maybe you could use something like that? 

https://stackoverflow.com/questions/9895502/easiest-way-to-detect-remove-unused-use-statements-from-php-codebase#9896557

then one could use something like:

    $unused = collect($sanitizer->getUnused());
    if($unused->count()) {
        throw new Exception("Unused use statements :".$unused->implode(', ');
    }

LukeTowers:
That comment only makes sense when removing `use` statements that aren't actually used. This PR is *adding* a `use Twig_Template;` statement, I'd like to know why.

tschallacka:
Sorry, I got commits mixed up in my head. sorry for that.

I added the use because of https://github.com/tschallacka/october/blob/dd769e3688fd7324dcae4006e186ec7f7261a405/modules/cms/twig/DebugExtension.php#L96

PR #3042, opened 2017-08-17
Dynamic plugin dependencies

pmhou:
This allows plugins the ability to declare their dependencies at run time and solves an issue where Plugin A can optionally extend Plugin B and Plugin C. 

When running `october:up` - the dependency order won't be set correctly for Plugin A as Plugin B and C are optional and can't be added to the usual `$require` property. However there may be conditional migrations in Plugin A that depend on the base tables being present from either Plugin B or C.

`setRequiredPlugins()` should return either a string or array in the same way as `$require`. See below for usage example.

```php
public function setRequiredPlugins()
{
    $requires = [];
    if (PluginManager::instance()->hasPlugin('RainLab.Blog')) {
        $requires[] = 'RainLab.Blog';
    }
		
    return $requires;
}
```

### DIFF
--- a/modules/system/classes/PluginManager.php
+++ b/modules/system/classes/PluginManager.php
@@ -635,6 +635,10 @@ class PluginManager
         if (is_string($plugin) && (!$plugin = $this->findByIdentifier($plugin))) {
             return false;
         }
+        
+        if (method_exists($plugin, 'setRequiredPlugins')) {
+		    $plugin->require = $plugin->setRequiredPlugins();
+	    }
 
         if (!isset($plugin->require) || !$plugin->require) {
             return null;


tschallacka:
Uhm... that kinda seems counter productive to the use case of requires...

why can't you just use 

if(PluginManager::instance()->hasPlugin('RainLab.Blog')) {
    \AuthorName\MyPlugin\RainlabBlogExtender::run();
}

That way your class will only be run when that plugin is loaded. and when it's not loaded it won't be called and won't throw errors.

pmhou:
@tschallacka Can you explain how that will resolve the dependency issue order when running october:up?

tschallacka:
This returns true even when they haven't run yet their october:up queueu otherwise it would be useless to add this to your requires statement.

This means you can also use it whenever you want. in your queue. Ideally in your extension you just add a hasOne relationship with your own models, and the model extensions don't get reloaded, and you can modify the forms etc with your extension code, all that will work.

Only use case for you to wish the other plugins to be processed is when you add fields yourself to the tables of the models of the plugins you're extending. Still I would not recommend that because of the same reason why one should not modify javascript default objects with own functions. It might cause conflicts in the future with the plugin updates.

LukeTowers:
I would suggest that you check to see if the necessary table exists before running your migration and if it doesn't (but the plugin it belongs to does exist), then hook into an event that would run after the migration process as completed by `october:up` and recheck your migration.

LukeTowers:
Also:

LukeTowers:

PR #3043, opened 2017-08-17
Attribute values for < and > and " should be written als &lt; &gt; and &quot;

tschallacka:
Attribute values for < and > and " should be written als &lt; &gt; and &quot;

This way you can be sure that   

  A) browsers don't execute rogue html/throw parsing exceptions when they behave stupid, and  
  B) IDE's won't throw invalid html warnings.

### DIFF
--- a/modules/cms/widgets/mediamanager/partials/_upload-progress.htm
+++ b/modules/cms/widgets/mediamanager/partials/_upload-progress.htm
@@ -3,7 +3,7 @@
         <div class="upload-progress">
             <h5 
                 data-label="file-number-and-progress"
-                data-message-template="<?= e(trans('cms::lang.media.uploading_file_num')) ?> <span>:percents</span>"
+                data-message-template="<?= e(trans('cms::lang.media.uploading_file_num')) ?> &lt;span&gt;:percents&lt;/span&gt;"
                 data-success-template="<?= e(trans('cms::lang.media.uploading_complete')) ?>"
                 data-error-template="<?= e(trans('cms::lang.media.uploading_error')) ?>"
             ></h5>


LukeTowers:

LukeTowers:
Also, in what browsers and / or IDEs does this currently cause a problem? We would like to deal with specifics if possible.

tschallacka:
It causes a problem in eclipse with the php editor set to html files. 
It's invalid syntax and the editor can't handle it it thinks a tag starts there.

Also, technically it's invalid html.  Most browsers are smart enough to catch it, but why not take the sure for the unsure and escape it with the appropriate html entity.

https://www.w3.org/TR/html5/syntax.html#attributes-0

> Unquoted attribute value syntax
    The attribute name, followed by zero or more space characters, followed by a single U+003D EQUALS 
    SIGN character, followed by zero or more space characters, followed by the attribute value, which, in 
    addition to the requirements given above for attribute values, must not contain any literal space 
    characters, any U+0022 QUOTATION MARK characters ("), U+0027 APOSTROPHE characters ('), "=" 
    (U+003D) characters, **["<" (U+003C)](url)** characters, **[">" (U+003E)](url)** characters, or U+0060 GRAVE ACCENT 
    characters (`), and must not be the empty string.
>   Double-quoted attribute value syntax
    The attribute name, followed by zero or more space characters, followed by a single U+003D EQUALS 
    SIGN character, followed by zero or more space characters, followed by a single """ (U+0022) 
    character, followed by the attribute value, which, **[in addition to the requirements given above for 
    attribute values](url)**, must not contain any literal U+0022 QUOTATION MARK characters ("), and finally 
   followed by a second single """ (U+0022) character.****

LukeTowers:
And does the JS still parse those escaped `<` and `>` characters correctly?

tschallacka:
JS will see the escaped characters as we, humans would see them when they are unescaped.

data-value=`"foo:&quot;&lt;bar&gt;&quot;"` will with  `return $(el).data('value')` turn for javascript to `foo:"<bar>"`

LukeTowers:
Alright, looks good to me. Thanks!

PR #3050, opened 2017-08-21
Add temp folder path into config

gaotangshen:
The temp folder is not configurable currently, but I think this should be a feature of October's CMS configuration as the media, uploads, sessions, cache, and theme directories are already configurable.

**Scenario**
When configuring the media directory path to a different location, or onto a shared mount, this can place `storage/app/media` and `storage/app/uploads` in a separate path from `storage/temp`. This causes issues when cropping or resizing a media file to not be available in the shared mount.

```
/var/www/storage/app --> /var/share/storage/app
    /var/www/storage/app/media   --> /var/share/storage/app/media
    /var/www/storage/app/uploads --> /var/share/storage/app/uploads
```

If we could just configure the temp directory path, even say to be `storage/app/temp` then we could mount a single shared directory at `storage/app`.

```
/var/www/storage/app --> /var/share/storage/app
    /var/www/storage/app/media   --> /var/share/storage/app/media
    /var/www/storage/app/uploads --> /var/share/storage/app/uploads
    /var/www/storage/app/temp    --> /var/share/storage/app/temp
```

I also did some research and figured out that the config will not work unless we overwrite the `temp_path` function as so:
```
     app()->instance('path.temp', $this->tempPath());

    public function tempPath()
    {
        return base_path().config('cms.storage.temp');
    }
```
This could also be added in private plugins, and not use the `temp_path()` function.

### DIFF
--- a/config/cms.php
+++ b/config/cms.php
@@ -255,6 +255,7 @@ return [
     */
 
     'storage' => [
+        'temp' => '/storage/temp',
 
         'uploads' => [
             'disk'   => 'local',
--- a/modules/cms/widgets/MediaManager.php
+++ b/modules/cms/widgets/MediaManager.php
@@ -870,7 +870,7 @@ class MediaManager extends WidgetBase
 
     protected function getThumbnailImageUrl($imagePath)
     {
-        return Url::to('/storage/temp'.$imagePath);
+        return Url::to(Config::get('cms.storage.temp', '/storage/temp').$imagePath);
     }
 
     protected function thumbnailExists($thumbnailParams, $itemPath, $lastModified)


LukeTowers:
Would it make sense to support different disks for the temp path than just the default disk? Similar to how the uploads and media config options exist currently?

gaotangshen:
@LukeTowers Thank you for your response. Nice suggestion. We can do that and it will take some time for coding and testing.
And we should only do that for `/storage/temp/public` folder since the `/storage/temp/media` folder is created only for storing media files temporarily. 
If you agree with this, I will try to do this when I am free.

LukeTowers:
That sounds good, although if we're making `/storage/temp/public` configurable then shouldn't `/storage/temp/media` be as well?

gaotangshen:
@LukeTowers The reason I thought `/storage/temp/media` shouldn't be there because it only store temp files and most time will be an empty folder. But it seems benefits. I will consider adding that.

LukeTowers:
People that configure the media storage folder would probably also want control over where the temp files are saved as well; at least that's my rationale behind that.

LukeTowers:

PR #3057, opened 2017-08-23
Make media URLs in accordance with the Link Policy

adsa95:
Since the Link Policy controls how URLs are generated "throughout the application", I thought it should also apply to media URLs. While I was at it, I made sure that it always returns valid URLs (before it kept spaces etc).


LukeTowers:

LukeTowers:
Are you able to provide an example of how the current behaviour is broken and the output generated by the new behaviour? Thanks for your work on this!

adsa95:
Sorry for the branch error, thanks for fixing! Here's an example:

Assume these settings:
`LINK_POLICY=force`
`APP_URL=http://example.dev`

And this path saved in the database from the media finder: 
`/tester/Volvo/V60 D6 TE/thumb.jpg`

Before my PR the result was the following while using the |media filter: 
`/storage/app/media/tester/Volvo/V60 D6 TE/thumb.jpg`

This does not comply with the link policy, and also contains spaces and is therefore an invalid URL.

This would be the result with my PR:
`http://example.dev/storage/app/media/tester/Volvo/V60%20D6%20TE/thumb.jpg`

When using S3 etc the `cms.storage.media.path` config variable should be a valid URL and the Url::to() method will return the path directly, so that shouldn't cause issues. Below is an excerpt from the UrlGenerator::to() method:

````php
// First we will check if the URL is already a valid URL. If it is we will not
// try to generate a new one but will simply return the URL as is, which is
// convenient since developers do not always have to check if it's valid.
    return $path;
}
````

However it might be more efficient to move the Url::to() call to where the MediaManager->storagePath variable is set in the MediaManager init() function? Then it would only be called once and not every time. That would also produce URLs according to the link policy while using the Markdown editor and probably some other places, which I guess is a good thing but would make it harder to migrate content (but not harder than if you're using S3 etc).

LukeTowers:
@adsa95 I think you messed up this PR by merging those changes in

adsa95:
Yup, that wasn't intentional. I'll create a new PR with these changes from another branch than my master, sorry about that, I hope it's not to much trouble!

PR #3074, opened 2017-08-30
Configure APP_URL in testing environment

### LINKED ISSUE #2939: ControllerTest::testThemeUrl failure
##### Expected behavior
Test should retrieve the correct theme path.

##### Actual behavior
The theme path is tested against a localhost path.

##### Reproduce steps
Run *phpunit*

##### October build
419

##### Generated output
`There was 1 failure:

1) ControllerTest::testThemeUrl
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'http://localhost/themes/test'
+'https://example.com.it/themes/test'

/var/www/html/meccMilano/tests/unit/cms/classes/ControllerTest.php:23`

harmenjanssen:
This fixes #2939, where `ControllerTest::testThemeUrl` fails when the configured url is not `localhost`.

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -20,6 +20,7 @@
     </testsuites>
     <php>
         <env name="APP_ENV" value="testing" />
+        <env name="APP_URL" value="http://localhost" />
         <env name="CACHE_DRIVER" value="array" />
         <env name="SESSION_DRIVER" value="array" />
     </php>


LukeTowers:
@harmenjanssen @Jamesking56 wouldn't it be better for the test to be refactored to not rely on a hardcoded URL? https://github.com/octobercms/october/blob/master/tests/unit/cms/classes/ControllerTest.php#L30

LukeTowers:
@alen12345 what do you think about my last comment?

alen12345:
Sorry @LukeTowers for the late reply; I personally do not like this solution since we rely on an harcoded url or try to set env vars...

I will think about a refactoring and propose some ideas!

alen12345:
@LukeTowers  what about this? We just retrieve the url from config and concatenate the correct path:

```php
$theme = Theme::load('test');
$controller = new Controller($theme);

$url = $controller->themeUrl();
$this->assertEquals(config('app.url') . '/themes/test', $url);

$url = $controller->themeUrl('foo/bar.css');
$this->assertEquals(config('app.url') . '/themes/test/foo/bar.css', $url);
```

LukeTowers:
I would accept a PR to merge in https://github.com/octobercms/october/pull/3074#issuecomment-365393146

PR #3075, opened 2017-08-30
Added cms.maintenance.listpages Event

arzola:
With this PR the developer will can filter the pages that are shown on the Maintenance settings dropdown.

On a Big OctoberCMS site this comes very handy.

### Example:

Show to the end user only the 'support' page that has the url `/support` to put as the maintenance page.

You can put a listener right on your `register` method at your Plugin.php file

```php
Event::listen('cms.maintenance.listpages', function($pages) {
    return $pages->filter(function($page){
        return $page->url == '/support'; 
    });
});
```

screen shot 2017-08-30 at 2 31 11 pm

### DIFF
--- a/modules/cms/models/MaintenanceSetting.php
+++ b/modules/cms/models/MaintenanceSetting.php
@@ -4,6 +4,7 @@ use Model;
 use Cms\Classes\Page;
 use Cms\Classes\Theme;
 use ApplicationException;
+use Event;
 
 /**
  * Maintenance mode settings
@@ -37,9 +38,11 @@ class MaintenanceSetting extends Model
      */
     public $rules = [];
 
+
     /**
      * Initialize the seed data for this model. This only executes when the
      * model is first created or reset to default.
+     *
      * @return void
      */
     public function initSettingsData()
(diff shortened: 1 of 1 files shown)


jan-vince:
Hi @arzola - good work!

I have tested this and it works as described :)

Maybe @LukeTowers can merge this.

*And @LukeTowers - will this be available in 420 version ? I have no luck with updating with edgeUpdates enabled. Downloaded version 420 doesn't have recent commits. Maybe I am doing something wrong?*

LukeTowers:
@jan-vince it's probably because the 420 build is locked to the commit it was generated on: https://github.com/octobercms/october/releases/tag/v1.0.420. In order to be guaranteed to be running the latest commits, I suggest using composer instead. `edgeUpdates` is more for users that want new features, but still some modicum of stability. Composer will guarantee that you always pull the latest commits

jan-vince:
@LukeTowers I understand that.

I wanted to test 420 update process locally on some of my webs - just to be ready for problems.

And I believe current status of 420 build is much worse than it would be with all commits included.

What do you think?

LukeTowers:
You're probably not wrong, @daftspunk just doesn't have the time to be continually updating the tag to the latest commit though and I'm not sure if the web based updater will pull new commits with the same build number. We'd have to get his input.

LukeTowers:
@arzola @jan-vince I've made some slight changes to the implementation with #3076, please test it if you have the time.

jan-vince:
I will look at this later today and let you know!

arzola:
I like the proposal at #3076 , but that requiere a big refactor, also if we try to filter the pages only for the Maintenance settings it also will be filtered the pages on the cms section and this is not the desired behaviour, and the solution to this PR becomes more tricky

LukeTowers:
Closing this PR in favour of #3076 as this is too specific to the CMS maintenance setting page. When adding new events, we have to be able to support them going forward for a very long time. As such, we need them to be as useful as possible in as many use cases as possible.

PR #3076, opened 2017-08-31
Add the cms.object.listInTheme event

LukeTowers:
Inspired by #3075.

@arzola @jan-vince please test this alternative to #3075 and let me know if it replicates the existing behaviour while also providing the necessary hooks for the changes proposed in #3075.

### DIFF
--- a/modules/cms/classes/CmsObject.php
+++ b/modules/cms/classes/CmsObject.php
@@ -144,7 +144,7 @@ class CmsObject extends HalcyonModel implements CmsObjectContract
                 $loadedItems[] = static::loadCached($theme, $item);
             }
 
-            $result = $intance->newCollection($result);
+            $result = $instance->newCollection($loadedItems);
         }
 
         /**
--- a/modules/cms/classes/CmsObject.php
+++ b/modules/cms/classes/CmsObject.php
@@ -165,7 +165,7 @@ class CmsObject extends HalcyonModel implements CmsObjectContract
          *     });
          *
          */
-        Event::fire('cms.object.listInTheme', [$this, $result]);
+        Event::fire('cms.object.listInTheme', [$instance, $result]);
 
         return $result;
     }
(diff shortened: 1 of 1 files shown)


arzola:
I like this, but this requiere a big refactor, also if we try to filter the pages only for the Maintenance settings it also will be filtered the pages on the cms section and this is not the desired behaviour, and the solution to the original PR #3075  comes more tricky

modules/cms/classes/CmsObject.php
```diff
@@ -126,24 +127,47 @@ public static function loadCached($theme, $fileName)
...
+         *         if ($cmsObject instanceof \Cms\Classes\Page) {
+         *             $objectList->filter(function($page){
+         *                 // Filter out all pages except the one with the '/support' URL
+         *                 return $page->url == '/support';
+         *             });
+         *         }
+         *     });
+         *
+         */
+        Event::fire('cms.object.listInTheme', [$this, $result]);
```
arzola: I think you can't pass `$this` in a static context, maybe `$instance` ?

LukeTowers:
@arzola no rewrite required (besides minor typos).

Below is a working example of how your use case could utilize this event:

```php
// Extend only the Settings Controller
SettingsController::extend(function($controller) {

    // Listen for the cms.object.listInTheme event
    Event::listen('cms.object.listInTheme', function ($cmsObject, $objectList) {

        // Get the current context of the Settings Manager to ensure we only affect what we need to affect
        $context = SettingsManager::instance()->getContext();
        if ($context->owner === 'october.cms' && $context->itemCode === 'maintenance_settings') {

            // Double check that this is a Page List that we're modifying
            if ($cmsObject instanceof \Cms\Classes\Page) {

                // Perform filtering with an original-object modifying method as $objectList is passed by reference (being that it's an object)
                foreach ($objectList as $index => $page) {
                    if ($page->url !== '/404') {
                        $objectList->forget($index);
                    }
                }

            }

        }

    });

});
```

Issue #2788, opened 2017-03-31
Loading of 20 million records takes 6 minutes


ElijahGM:
##### Expected behavior
I have a list of 20 million records i want it to load faster in milliseconds and allow filtering (in list time possible) and export reords based on filters
##### Actual behavior
Trying to load the List it is taking close to 6 minutes to load the  page, when i try to to export i get a 500 error

##### Reproduce steps
Click on controller index action to list all records (Loads after 6 minutes)
Click on Export (Application crushes)

##### October build
(October build number)

daftspunk:
This is more of a DevOps issue. I assume the error 500 is caused by a lack of server resources to process the action. Again with the 6 minute load time. How much RAM do you have allocated for the server and for the PHP and MySQL processes? The bottleneck could be anywhere really. Consult your server administrator to pinpoint exactly where this is.

ElijahGM:
I have 32GB of ram , when I load 3k records it works perfectly, the problem
is when I have 20m records,
For the error 500, when I restart the server it resumes perfectly, when I
try to export again, takes time then when I refresh the page I get
permanent error 500, until I restart again, all other application works
well throughout, when I try to create a Php raw or yii2 application with
the same records they work well.
 Please advice

On Mar 31, 2017 7:34 AM, "Samuel Georges" <notifications@github.com> wrote:

This is more of a DevOps issue. I assume the error 500 is caused by a lack
of server resources to process the action. Again with the 6 minute load
time. How much RAM do you have allocated for the server and for the PHP and
MySQL processes? The bottleneck could be anywhere really. Consult your
server administrator to pinpoint exactly where this is.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/octobercms/october/issues/2788#issuecomment-290613632>,
or mute the thread

daftspunk:
Ok, we need someone to benchmark the ListController with 20m records and tell us where the bottleneck is. Sounds like it could be a fun exercise for someone.

LukeTowers:
With that many records, wouldn't it be better to use `DB::raw()` statements instead of fully instantiated Collection and Model objects for all the records?

daftspunk:
@LukeTowers I don't think it would matter too much because the Collection would only contain the number of records shown on a given page. Unless pagination is disabled, this wouldn't be a cause for concern.

LukeTowers:
```
Click on controller index action to list all records (Loads after 6 minutes)
[...]

PR #3084, opened 2017-09-05
Allow simple pagination option on ListControllers

jofrysutanto:
In response from #2788 this adds additional `showPageNumbers` option on Lists widget (consequently on list_config.yaml) to deal with tables with millions of records. This improves performance in general and also prevent HTML render stalled by populating page dropdown `<option></option>` in pagination partial.

If this is okay, then we should also proceed with making changes to the list_config.yaml stub in October/Rain repo, and also the documentation repo to introduce this additional feature.

### DIFF
--- a/modules/backend/widgets/lists/partials/_list.htm
+++ b/modules/backend/widgets/lists/partials/_list.htm
@@ -18,7 +18,11 @@
     <?php if ($showPagination): ?>
         <div class="list-footer">
             <div class="list-pagination">
+                <?php if ($showPageNumbers): ?>
                 <?= $this->makePartial('list_pagination') ?>
+                <?php else: ?>
+                <?= $this->makePartial('list_pagination_simple') ?>
+                <?php endif ?>
             </div>
         </div>
     <?php endif ?>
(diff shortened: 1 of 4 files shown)


modules/backend/widgets/lists/partials/_list.htm
```diff
@@ -18,7 +18,11 @@
     <?php if ($showPagination): ?>
         <div class="list-footer">
             <div class="list-pagination">
+                <?php if ($showPageNumbers): ?>
                 <?= $this->makePartial('list_pagination') ?>
```
LukeTowers: Fix the indentation on this please :)

LukeTowers:
This looks great! Go ahead and make the appropriate PRs to the library and docs repos.

Thanks a bunch for your work on this!

daftspunk:
Not sure I like this, it's overcomplicated. If there are millions of records, and hence lots of page numbers, we should show the first 10 pages and last 10 pages instead.

daftspunk:
Actually, scratch that, I can see this is using the `simplePaginate` internal method, thus the performance improvement. Very good.

Issue #3064, opened 2017-08-26
Richeditor choose link not working properly 


anand-patel:
Hi, in richeditor choose link function not working properly, it always add link at the start of the content if we add link from choose link(select page dropdown)

##### Expected behavior
- Go the static page
- Type text as link name in richeditor(after any content not at the start)
- Select that text
- Click on insert link from toolbar
- It will open form with URL and Text(text field is auto populated from the selected text)
- Click on search icon(choose link)
- It will open page selection dropdown 
- Select static page or CMS page
- It open form with URL and text
- Text field should be selected text if user has selected text, or it should be page name
- Affter clicking on insert, it should insert link where we have selected text or a cursor points.

##### Actual behavior
- Go the static page
- Type text as link name in richeditor(after any content not at the start)
- Select that text
- Click on insert link from toolbar
- It will open form with URL and Text(text field is auto populated from the selected text)
- Click on search icon(choose link)
- It will open page selection dropdown 
- Select static page or CMS page
- It will again open form with URL and Text 
- Text field is changed with the Page name
- When we click on insert, it alwasy add at the start of that content, not where we selected that text or a cursor

##### October build
419

joltmode:
Was about to report the same issue. Will just add my notes here. First off, I made a GIF:

![OctoberCMS Froala Link Bug](https://media.giphy.com/media/26n7b43HLKcr2fOJa/giphy.gif)

There are actually multiple issues I have noticed there, therefore I added numbers and highlights to the GIF:

1. The label is supposed to be hidden(?), but there's something wrong with the technique over there, that also obscures the input field. This also happens for Image/Video/Audio insertion popups.  
![Label obscures input field](https://user-images.githubusercontent.com/1030080/29872450-5b092116-8d98-11e7-9399-3385572c6d37.png)  
The element highlight in web view shows the overlap.
2. The URL field is not populated at all. I, at least, expect it to show the URL of the selected page.
3. The Text field is unexpectedly updated, even though, I had selected a text to be linked before insertion. I expect it to be left at the selected value.
4. Well, after inserting the link, you can see how it's prepended to the content, not updated in place of selected text.

Didn't notice this while recording the GIF — the href for the freshly inserted link is actually left empty (probably that's why it didn't show the URL for `#2`). As seen in code view:  
![image](https://user-images.githubusercontent.com/1030080/29876052-65350ee6-8da4-11e7-9309-08248094c08b.png)

---

There are also problems when editing links.  
[...]

PR #3086, opened 2017-09-05
RichEditor Page Links, save and restore selections accordingly.

joltmode:
Fixes reported #3064, explicitly, fourth bullet point issue from https://github.com/octobercms/october/issues/3064#issuecomment-326008148

I had to add this extra event, because, before the popup is shown, editor seems to have lost it's track of selection.

Somehow (couldn't actually find out why) the selection is restored once the popup has been shown (not on "_show_", but explicitly "_shown_").  
I tracked it down visually and by accident, even by debugging I couldn't find the actual code where and why the selection gets restored back once the popup is shown. While debugging I did trace it down to [these lines](https://github.com/octobercms/october/blob/4074ad8cc9c20a3196f058e908d98c5efca9ad1b/modules/system/assets/js/framework.js#L304): before triggering the events no selection was active, after the events, selection is back on track. I did a search over the code to find any listeners to `ajaxUpdateComplete` but I could find any that would resemble anything about selections...

So yeah, after the popup was shown, I noticed how at that point clicking anywhere on the page (which includes the pageLinks select field) removed the selection from editor, and possibly focus too, though, Froala didn't emit blur event then<sup>1</sup>. The selection never got restored back after that.

My first attempt was to use "_show_" event, but it was fired before the selection got restored, so I attempted "_shown_" and voila, that's where I save the selection.

And then, knowing that while the popup is active the selection will be forgotten, I restore selection only after the actual link values are being set back on Froala's `links.popup`.

I'm not really happy about this fix, but I'm providing this PR so that we can continue to investigate this further, knowing that you guys might know what's going on there. Or, if we don't find out proper solution, then this could stay as a fix.

I chose `one` instead of `on`, because popup is cached internally and `on` did bind multiple events.

<sup>1</sup> I actually found the `blur` event quite unreliable through my debugging. There are plenty of times when `blur` isn't called even though Froala has lost focus. That, obviously, feeds back to focus not being fired when Froala gets the actual focus back, but that's a whole other issue. Maybe internally Froala remains focused at all times, but if that's so, then whole focus/blur system is totally unreliable. I really expect it to lose focus when I activate an actual input field... _Yeah, froala's blur event wasn't called even when I clicked the pageLinks select field._

### DIFF
--- a/modules/backend/formwidgets/richeditor/assets/js/plugins/pagelinks.js
+++ b/modules/backend/formwidgets/richeditor/assets/js/plugins/pagelinks.js
@@ -54,6 +54,9 @@ $.FroalaEditor.DEFAULTS.key = 'HHMDUGENKACTMXQL==';
                 $input = $(check_inputs[i]);
                 $input.prop('checked', $input.data('checked') == link[$input.attr('name')]);
             }
+
+            // Restore selection, so that the link gets inserted properly.
+            editor.selection.restore();
         }
 
         function insertLink() {
@@ -61,6 +64,9 @@ $.FroalaEditor.DEFAULTS.key = 'HHMDUGENKACTMXQL==';
 
             editor.$el.popup({
                 handler: editor.opts.pageLinksHandler
+            }).one('shown.oc.popup.pageLinks', function () {
+              // I still have no idea why selection is only restored after popup is actually shown.
+              editor.selection.save()
             })
         }
 


LukeTowers:
Thanks a ton for your work on this and the other PRs fixing the issues! (#3085, #3087) This is incredibly helpful :)

Why are you not happy with the proposed fix?

joltmode:
I'm not happy because I couldn't find out what is causing the re-selection once the popup is shown. :smile:.

Therefore, listening to the _shown_ event to save selection feels like a loose end. There is no place documented why even I could use _shown_ event to get back selection info. In other words, I cannot trace the logic, but programming is all about chaining everything together.

Maybe that's just me... I'm quite critical and have high standards, especially for myself.

LukeTowers:
That makes sense, it would bug me too. On the topic of development mysteries, see if you could figure this one out:
```php
/*
    NOTE: This shenanigan is caused by a PHP bug that occassionaly rears it's ugly head where accessing a StdClass grand-property directly the first time
    will fail with "Creating default object from empty value". However, it will not fail if you pass it to a function to test it
    (i.e. dd($stdClass->stupidPropertyThatDoesExist->subProperty) will work fine, but $stdClass->stupidPropertyThatDoesExist->subProperty will fail).
    Somehow, just forcing PHP to attempt the action twice will actually cause it to work on the second time. Could be any sort of race condition, improper initialization
    or any other weird-ass bug that you could think of that causes this, but I can't track down the root cause. As a side note, this doesn't happen in the almost-identical
    \IACEA\Exchanges\Models\Programme->filterFields() method, only in this class. If at some point someone somewhere manages to fix this, I would much
    appreciate being sent an explanation of the bug and the fix (octobercms@luketowers.ca).
*/
try {
    $fields->nation->disabled = true;
} catch (\Exception $e) {
    $fields->nation->disabled = true;
}
```

w20k:
@LukeTowers not relevant anymore. Fixed in the latest build.

PR #3087, opened 2017-09-05
RichEditor Page Links, don't update Froala's link.popup text if it is already filled.

joltmode:
Fixes third bullet point issue from https://github.com/octobercms/october/issues/3064#issuecomment-326008148.

I found the previous _if checks_ quite ambiguous, hence introduced a new one:

1. Change value of text field only if current value was not set, as to not override.
2. Change values of matching input fields to provided values. Seems to me, to be compliant with previous routine.


tim0991:
i agree this,but why not merge this pr

LukeTowers:
@tim0991 because it needs to be reviewed, along with every other open PR and issue on the repository.

w20k:
@LukeTowers not relevant anymore. Fixed in the latest build.

PR #3091, opened 2017-09-06
Fix subject escaped twice in mail template

Fl0Cri:
Before this PR, when sending a mail with the default template and setting the subject as `Nouvelle demande d'adhésion`, it get displayed like this:
```
Nouvelle demande d&#039;adhésion
```

This is because the raw html value is as:
```
Nouvelle demande d&amp;#039;adhésion
```

With this PR we avoid to escape twice the html entity, so it will rendered properly like that by the mail client:
```
Nouvelle demande d'adhésion
```

### DIFF
--- a/modules/system/views/mail/layout-default.htm
+++ b/modules/system/views/mail/layout-default.htm
@@ -18,7 +18,7 @@ name = "Default layout"
 
         <!-- Header -->
         {% partial 'header' body %}
-            {{ subject }}
+            {{ subject|raw }}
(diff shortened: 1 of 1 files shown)


Fl0Cri:
Oops, wrong branch, this is the correct PR: https://github.com/octobercms/october/pull/3092

LukeTowers:
@Fl0Cri you can simply edit the Pull Request to change the base branch, you don't have to create a new one.

PR #3092, opened 2017-09-06
Fix subject escaped twice in mail template

Fl0Cri:
Before this PR, when sending a mail with the default template and setting the subject as `Nouvelle demande d'adhésion`, it get displayed like this:
```
Nouvelle demande d&#039;adhésion
```

This is because the raw html value is as:
```
Nouvelle demande d&amp;#039;adhésion
```

With this PR we avoid to escape twice the html entity, so it will rendered properly like that by the mail client:
```
Nouvelle demande d'adhésion
```

### DIFF
--- a/modules/system/views/mail/layout-default.htm
+++ b/modules/system/views/mail/layout-default.htm
@@ -18,7 +18,7 @@ name = "Default layout"
 
         <!-- Header -->
         {% partial 'header' body %}
-            {{ subject }}
+            {{ subject|raw }}
         {% endpartial %}
 
         <tr>


LukeTowers:
What is causing it to be escaped twice? Are you able to look into that? I would prefer if we could figure that out and fix that instead of just using `|raw` as a fix.

Fl0Cri:
This is because the subject is escaped a first time here: https://github.com/octobercms/october/blob/master/modules/system/views/mail/layout-default.htm#L21

Then, the escaped string is passed to the `header` partial, and inside this partial it is escaped a second time: https://github.com/octobercms/october/blob/master/modules/system/views/mail/partial-header.htm#L13

The only solution I see is to allow a raw value to be passed in one of these places. After looking at other partials, I noticed that most of them are using the `|md_safe` filter. Not sure if it allows html entities, but maybe the `|raw` (or any other filter that don't encode html entities) can be in the partial instead of the layout.
It would however require to add it on all 3 occurences of the `body` variable.

What do you think would be best?

LukeTowers:
Given the other instances of `|raw` in the template I think this change is fine as is on second consideration. Thanks!

PR #3106, opened 2017-09-12
Develop

datune:
This PR enables basic SVG previews in the mediamanager grid views and the sidebar preview.

### DIFF
--- a/modules/cms/widgets/MediaManager.php
+++ b/modules/cms/widgets/MediaManager.php
@@ -142,6 +142,9 @@ class MediaManager extends WidgetBase
     {
         $path = Input::get('path');
         $lastModified = Input::get('lastModified');
+        
+        $isSVG = $this->isSVG($path);
+        if ($isSVG) return $isSVG;        
 
         $thumbnailParams = $this->getThumbnailParams();
         $thumbnailParams['width'] = 300;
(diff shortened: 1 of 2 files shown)


LukeTowers:
You need to make these changes in separate branches, I'm not going to be merging in the mediamanager changes right now because I'm in the middle of making some changes there. 

Please create a new PR with only the ListController improvements. Thanks!

datune:
@LukeTowers Yes I just noticed, the same is true for the Lists.php update. How can I remove the commits from this PR?

LukeTowers:
Make each group of commits to a new branch on your forked repo and then make a separate PR for each branch

datune:
Will resolve this once https://github.com/octobercms/october/pull/3170 is available in the next build.

LukeTowers:
@datune I'm not sure @daftspunk wants October to support this, SVG has been removed from the list of valid image types: https://github.com/octobercms/library/commit/3bbbbf3da469f457881b5af902eb0b89b95189a2

datune:
@LukeTowers  LOL, that's not funny, not even remotely funny! So what about product and brand logo's? I have clients who need to maintain hundreds of SVG's (images!). Now that we finally have SVG support all over, we're going to take ten steps backwards? 

Also:

[SVG files displayed with <image> are treated as an image: external resources aren't loaded, :visited styles aren't applied, and they cannot be interactive. ](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/image)

But seriously, if you want this to be a Media Manager, it needs to support SVG previews. Disabled by default, but deny it or completely remove support...I don't know.

LukeTowers:
I'm not opposed to supporting it still, just letting you know that SVG is going to be disabled by default from now on, unless better solutions to the security risks that SVG files present can be found. Perhaps sanitizing on upload?

datune:
Yes, I will look into it when time permits. However, displaying previews of SVG's in the Media Manager within image tags does not propose any risk, since, as I quoted above, none of the XSS risks are possible. 

I think it's not that hard to write a parser to sanitize them on upload. I'll look into this once I have time, I'd be surprised if this doesn't already exist!

Perhaps put this PR on hold or something for the time being?

Issue #3111, opened 2017-09-14
Change of richeditor z-index broke popup-ovelay


Fl0Cri:
Changing the z-index of the richeditor in https://github.com/octobercms/october/pull/3099 caused a side effect with popover-overlay when the richeditor is not inside a popup, like in static pages plugin when using a page snippet.

##### Expected behavior
The following capture shows the behaviour of a static page snippet before changing the richeditor's z-index (like in build 422).
![backend_expected](https://user-images.githubusercontent.com/16371551/30417236-516cb1de-992f-11e7-8075-21f913052733.png)

##### Actual behavior
Currently (build 423), the richeditor's z-index is higher than both `.popover-overlay` (597) and `.control-popover` (599):
![backend_actual](https://user-images.githubusercontent.com/16371551/30417308-8ade5c9c-992f-11e7-87ff-1795788cc8c9.png)

##### Reproduce steps
1. From a clean install, install the static pages plugin.
1. Create a partial that will be usable as a snippet:
```
[viewBag]
snippetCode = "testSnippet"
snippetName = "Test"
snippetProperties[titleText][title] = "Test"
snippetProperties[titleText][type] = "string"
snippetProperties[titleText][default] = ""
snippetProperties[titleText][options][] = ""
==
<span>{{ titleText }}</span>
```
3. Create a new static page
3. Include the snippet in the page and try clicking on it to show the edition popup.

##### October build
since 423

##### Workaround
I managed to fix it in my case by putting the following css in my backend styles, but it's kind of a z-index war and I don't know what side-effects it could cause in other situations.
```CSS
div.popover-overlay {
    z-index: 1001;
}
div.control-popover {
    z-index: 1002;
}
/* Fix user settings menu */
nav#layout-mainmenu ul li .mainmenu-accountmenu {
    z-index: 1002;
}
```

LukeTowers:
Could you please look into reverting the PR that made the issue and the fixing the original issue with CSS instead?

datune:
@Fl0Cri 1. Thanks for a supreme example of how to submit a bug report, chapeau! 2. It is indeed a z-index war, since the layout-side-panel is also affected when unpinned, it's z-index is 600.

So to append your workaround for the time being I've added:
```css
body.display-side-panel #layout-side-panel {
    z-index: 1001;
}
```

LukeTowers:
[...]

Issue #3098, opened 2017-09-07
Editor: Table formatter popup is hiding behind another popup


TimFoerster:
##### Expected behavior
The table formatter popup should be infront of the popup

##### Actual behavior
The table formatter popup is behind the popup

##### Reproduce steps
Use a plugin with a popup (for example leaflet maps).
Create a table in the popup editor and click on a cell.

##### October build
current

![untitled](https://user-images.githubusercontent.com/7849550/30159620-7bb69dee-93c9-11e7-85cb-eecad36ff628.jpg)
![untitled2](https://user-images.githubusercontent.com/7849550/30159624-7f2f3314-93c9-11e7-904a-88b44393c618.jpg)

TimFoerster:
with the PR i get 
![untitled3](https://user-images.githubusercontent.com/7849550/30161357-71c5c2c2-93d0-11e7-838f-0d3d5f033dd7.jpg)

LukeTowers:
Fixed in https://github.com/octobercms/october/commit/ead70c43f31d65516fc98da92766dc0513d91ec5. Thanks for your help!

PR #3113, opened 2017-09-14
Fixes broken richeditor z-index inside and outside of a popup

Fl0Cri:
Fixes both https://github.com/octobercms/october/issues/3098 and https://github.com/octobercms/october/issues/3111

By default Froala has a z-index of 1 (it's this setting that was raised to 1000 in a previous PR). Other components generated by the editor (including the popups) are rendered with an inline style computed based on this root z-index. For example, the popup is `[root z-index] + 4` but it doesn't directly needs the z-index of the editor itself.

Looking at what css rules are applied to the popup, we can see that there is a z-index of 9997 coming from here: https://github.com/octobercms/october/blob/master/modules/backend/formwidgets/richeditor/assets/css/richeditor.css#L30

The problem is the inline z-index added by Froala that overrides the one in the css.
An `!important` should do the trick.

Note that I cannot build the richeditor assets because it needs a `froala_drm` directory in its vendors.
I tried to simulate the behaviour by tweaking the compiled files and I can confirm that it fixes both issues.


LukeTowers:
Awesome, thanks for doing this @Fl0Cri!

TimFoerster:
Well it brakes  #3098 ...
screen shot 2017-10-23 at 02 28 57

LukeTowers:
@TimFoerster is there a way that we can reset where the table popup gets it's z-index base from? I.e. make the richeditor position relative or something? This constant z-index fighting is getting ridiculous.

Fl0Cri:
I have made a new PR that should fix the problem: https://github.com/octobercms/october/pull/3194
@TimFoerster can you have a try?

TimFoerster:
@Fl0Cri I'll but I'm on the road right now. Will come back around Friday.

TimFoerster:
nvm, i checked it quickly, looks good! thanks for your work
screen shot 2017-10-23 at 20 48 06

Issue #3124, opened 2017-09-20
[FeatureRequest]: Warn users when uploading duplicate file names to the media manager


buuug7:
## description
if upload two same name files into media manager, the files uploaded later will  auto overwrite the previous file without prompt.   what did i do for prevent  auto overwrite? i have many files in one directory,about 5000, when i upload new files into media manager, i must check the file name first, but this become more difficulty when the numbers of files getting more and more .  

did there have any good solutions?

korneliuskristianr:
@buuug7  Just Create a folder then upload the same file into that folder

buuug7:
@korneliuskristianr if you editor a article and add  images  with richeditor with 
![image](https://user-images.githubusercontent.com/12119289/30629442-691db092-9e0e-11e7-8f47-36917b909ba8.png)
it will auto upload image into `uploaded-files` directory, when you add a second article , also with a same name image, it will overwrite the previours images. when you see your article, first article and second article have a sample image. but people need it different.   

want i need do?

buuug7:
if uploaded files generate a random string name , it will prevent this problem. for same image ,user can manager manually, like below, just delete the same images. 
![ef](https://user-images.githubusercontent.com/12119289/30629872-abfb172c-9e10-11e7-9f1d-8e1ad177dbb7.png)

tgz-cz:
I have same problem with the media manager.

I override important file (that was uploaded in past) with newer file with same filename.
None prompt or error or warning was displayed.

I have few questions:

1. Is there any configuration for change this behaviour to raise error in this case or auto-rename newly uploaded file?

2. Or is it possible change this behaviour via events/extensions/plugins of MediaManager?

3. Another way is using my own model with related "File Attachments" ('System\Models\File') instead of MediaManager. There is not problem with overriding existing files, but if user downloads the file then original filename is lost and the filename is 558ab0f0aa231581256800.xls instead.
It is possible download this file with original filename?

Thank you for answers

buuug7:
@tgz-cz i have not best solutions, it seems only by modifying the source code of `modules/cms/widgets/MediaManager.php` line 1098 ,as follows 
```
    protected function checkUploadPostback()
    {
        $fileName = null;
        $quickMode = false;

        if (
            (!($uniqueId = Request::header('X-OCTOBER-FILEUPLOAD')) || $uniqueId != $this->getId()) &&
            (!$quickMode = post('X_OCTOBER_MEDIA_MANAGER_QUICK_UPLOAD'))
[...]

PR #3128, opened 2017-09-22
warn users when uploading duplicate file names to the media manager

buuug7:
for issue #3124 
when uploading duplicate file names to the media manager, throw excepiton for hint. if quickMode enable, auto rename filename.


buuug7:
since the MediaNanager moved from CMS module to Backend Module, this PR is outdate  and should not merge, i will Give a new PR

PR #3135, opened 2017-09-27
Alter `reserved` column to nullable when rolling back migration

kuloed:
We should add column `reserved` to nullable to prevent this error,
```General error: 1 Cannot add a NOT NULL column with default value NULL (SQL: alter table "jobs" add column "reserved" integer not null)```

### DIFF
--- a/modules/system/database/migrations/2017_10_01_000022_Db_Jobs_FailedJobs_Update.php
+++ b/modules/system/database/migrations/2017_10_01_000022_Db_Jobs_FailedJobs_Update.php
@@ -30,7 +30,7 @@ class DbJobsFailedJobsUpdate extends Migration
     public function down()
     {
         Schema::table($this->getTableName(), function (Blueprint $table) {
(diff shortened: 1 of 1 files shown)


LukeTowers:
Why is this required? I have no idea what bug is being solved from your brief description, please explain in more detail.

kuloed:
if we already have some data in jobs table, we cannot alter the table to add 'reserved' not null column. so the command will throw the exeption i said before.

LukeTowers:
That still doesn't tell me exactly what you are doing that is causing this bug and why it is something that needs to happen to the project as a whole instead of a just specific to you thing.

kuloed:
like when i'm using october as testing application, and add some jobs queue, then after i'm finished i run october:down to reset october's table. If you see this is not problem then we can close this.

LukeTowers:
So does october throw an error running `october down` when there is content in the jobs table?

PR #3143, opened 2017-10-01
Account for NestedTree Trait setting

datune:
See https://github.com/rainlab/blog-plugin/issues/338. The Lists Widget did not account for the NestedTree (showTree) setting which would result in Lists being rendered using sortOrders and not "Tree" orders.

### DIFF
--- a/modules/backend/widgets/Lists.php
+++ b/modules/backend/widgets/Lists.php
@@ -486,7 +486,7 @@ class Lists extends WidgetBase
         /*
          * Apply sorting
          */
-        if ($sortColumn = $this->getSortColumn()) {
+        if ($sortColumn = $this->getSortColumn() && !$this->showTree) {
             if (($column = array_get($this->allColumns, $sortColumn)) && $column->valueFrom) {
                 $sortColumn = $this->isColumnPivot($column)
                     ? 'pivot_' . $column->valueFrom
--- a/modules/backend/widgets/Lists.php
+++ b/modules/backend/widgets/Lists.php
@@ -486,7 +486,7 @@ class Lists extends WidgetBase
         /*
          * Apply sorting
          */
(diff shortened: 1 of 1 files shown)


LukeTowers:
@datune if we could get someone else to verify that this solves the problem I'll merge it. I don't have time to test it myself right now unfortunately but if someone else can verify that it works as a cross check then I'd be happy to merge it.

Thanks for your work on this!

graker:
I've tried this one and it does fix the problem with categories list at `backend/rainlab/blog/categories` - the list is sorted as it is set on reorder page which is an expected behavior. As for `categories` component of the blog plugin ([this issue](https://github.com/rainlab/blog-plugin/issues/338)), it seems that it doesn't help and the categories are still ordered by name.

LukeTowers:
@graker that's to be expected, this only affects the Backend Lists Widget. Perhaps you can try applying the fix the categories listing component and make a PR for that.

graker:
@LukeTowers I don't see how to apply it yet. The order-by clause in that component is like this:
```
$categories = BlogCategory::orderBy('name');
```
So since it was working before upgrades, this order-by is probably overridden somewhere else. Unfortunately, I also can't make time to see it through right now and to find the exact source of the problem.

LukeTowers:
Well, we'll leave that issue open for now and hopefully someone can find time to dig into it deeper. @graker

datune:
@LukeTowers @graker I digged a bit and think that [this PR](https://github.com/rainlab/blog-plugin/pull/342) fixes it. @graker Please test and review so we can move forward. 

I'm pretty convinced that nested tree sorting has never worked in the categories component, it was always sorted by name. I think the reason is historical (like there is ever any other reason :-P). Nesting has been introduced after the fact, which is why this has slipped through.

PR #3147, opened 2017-10-02
Translated data-group-code attribute

gpasztor87:


LukeTowers:
Why is this required?

gpasztor87:
It's not required, but untranslated strings don't look good in the html template.

LukeTowers:
You're going to have to provide a better reason than "it doesn't look good". If this is only visible in the markup and not to the user, then why should we change it? Don't fix what isn't broken, you're likely to break something else.

PR #3149, opened 2017-10-03
Allow language string for repeater group item title

Floriskoch:
Allows using language files for group titles.


Floriskoch:
@LukeTowers Ah sorry, didn't pay enough attention. Thanks!

PR #3150, opened 2017-10-03
Support form field commentHtml flag for section, checkbox, and switch types

jimcottrell:
**Expected behavior:**

Setting a backend form field's `commentHtml` value to true allows direct HTML display in the field's comment.

**Actual behavior:**

The field types that don't run through the standard `_field.htm` partial-- `section`, `checkbox`, and `switch`-- ignore the setting and display escaped comments regardless. 

I can't see why that would be intentional, so this commit corrects that behavior.


LukeTowers:
Good catch, thanks!

Issue #2592, opened 2017-01-05
Plugin test can't boot up boot method


surahmans:
##### Expected behavior
I want to test my plugin using `TestCase` instead of `PluginTestCase`. I add dynamic method using `addDynamicMethod` on plugin boot method, and hope it works by calling `$user->subscibe($plan)`

##### Actual behavior
I get this error:
```
1) Indovisual\Membership\Tests\Unit\SubscriptionTest::user_can_subrcribe_a_plan
BadMethodCallException: Call to undefined method October\Rain\Database\QueryBuilder::subscribe()
```

But when I adding the `subscribe` method directly to model it works.

##### Reproduce steps
Here's my code to extend the `User model`:
```
            $model->addDynamicMethod('subscribe', function($plan) use ($model) {
                if ($plan instanceof \Indovisual\Membership\Models\Plan) {
                    return $model->subscribes()->save($plan);
                }
            });
```

##### October build
v350

daftspunk:
You may need to boot the plugin manually, or add the dynamic method as part of your tests.

surahmans:
Thank you, @daftspunk for your great advice! 😄 

I digged into your awesome code and find helpful code to solve my case by adding `PluginManager::instance()->bootAll();` into `setUp` method. Now my tests code looks like;

```
<?php

use RainLab\User\Models\User;
use System\Classes\PluginManager;

class UserTest extends TestCase
{
    public function setUp()
    {
        parent::setUp();

        PluginManager::instance()->bootAll();
    }

    public function test_call_dynamic_method()
    {
        $user = User::first();

        $this->assertEquals('Hello there!', $user->sayHi());
    }
}
```

surahmans:
I'm sorry, @daftspunk. Why using `PluginManager::instance()->bootAll();` only work for adding dynamic method and it doesn't work for extending relation like this,

```php
            $model->belongsToMany['wishlists'] = [
                'Indovisual\Gallery\Models\Item',
                'table' => 'indovisual_gallery_wishlists'
            ];
```

I know we can add this code to make that work,

```php
    public function setUp()
    {
        parent::setUp();

        User::extend(function($model) {
            $model->belongsToMany['wishlists'] = [
                'Indovisual\Gallery\Models\Item',
                'table' => 'indovisual_gallery_wishlists'
            ];
        });
    }
```

But, Is there another way to make it work like "boot up" our plugins?

Thanks.

daftspunk:
[...]

PR #3156, opened 2017-10-07
Added PluginManager::unregisterAll()

harmenjanssen:
### Preface
This PR follows up on issue #2592. 
To recap: in the testing environment, if you're testing plugins working together, you need to do 

```php
PluginManager::instance()->registerAll(true);
PluginManager::instance()->bootAll(true);
```

to make sure everything is setup in the same way your frontend would be using them.

### One last issue
We noticed however, that the `app` instance bound to `$this->app` in a plugin gets corrupted, sort of.

For instance, usually `$app['translator']` is set, and this is true for the first test. Every subsequent test however, throws an exception from deep within Laravel's ServiceProvider stating `ReflectionException: Class translator does not exist`. 

Adding the following to my plugin's `boot()` method shows `true` for the first test, and `false` for every subsequent test:

```php
public function boot()
{
    dump(isset($this->app['translator']));
}
```

### Solution
This lead me to think something got destroyed in `tearDown` but not reset in `setUp` again. Eventually I ended up in `PluginTestCase::runPluginRefreshCommand()`.   
If I removed the `if` statement below the comment

```php
        /*
         * First time seeing this plugin, load it up
         */
```

my tests would run correctly, because it would run `loadPlugin()` regardless of wether it was the first time or not.

This is why I offer an `unregisterAll()` method that can be called from within my `tearDown()` function, making sure plugins are registered and booted from scratch for every test.

This would look something like this:

```php
    public function setUp()
    {
        parent::setUp();

        $pluginManager = PluginManager::instance();
        $pluginManager->registerAll(true);
        $pluginManager->bootAll(true);
    }

    public function tearDown()
    {
        parent::tearDown();

        $pluginManager = PluginManager::instance();
        $pluginManager->unregisterAll();
    }
```

It feels right for `tearDown` to do the exact opposite of `setUp` and clean up after it.

This fixes my problem, but I can imagine I might miss some intricacies regarding the PluginManager, so let me know if I should elaborate either my implementation or the accompanying unit test.

I can create a repo to reproduce the problem if you want.

### DIFF
--- a/modules/system/classes/PluginManager.php
+++ b/modules/system/classes/PluginManager.php
@@ -163,6 +163,16 @@ class PluginManager
     }
 
     /**
+     * Unregisters all plugins: the negative of registerAll().
+     * @return void
+     */
+    public function unregisterAll()
+    {
+        $this->registered = false;
+        $this->plugins = [];
+    }
+
+    /**
      * Registers a single plugin object.
      * @param PluginBase $plugin
      * @param string $pluginId
--- a/tests/unit/system/classes/PluginManagerTest.php
+++ b/tests/unit/system/classes/PluginManagerTest.php
@@ -122,7 +122,7 @@ class PluginManagerTest extends TestCase
         $this->assertArrayHasKey('\database\tester', $result);
         $this->assertArrayHasKey('\testvendor\test', $result);
     }
-    
+
     public function testGetVendorAndPluginNames()
     {
         $manager = PluginManager::instance();
@@ -147,6 +147,16 @@ class PluginManagerTest extends TestCase
         $this->assertEquals('October Test Plugin', $pluginDetails['name']);
         $this->assertEquals('Test plugin used by unit tests.', $pluginDetails['description']);
         $this->assertEquals('Alexey Bobkov, Samuel Georges', $pluginDetails['author']);
+    }
+
+    public function testUnregisterall()
+    {
+        $manager = PluginManager::instance();
+        $result = $manager->getPlugins();
+        $this->assertCount(5, $result);
+
+        $manager->unregisterAll();
+        $this->assertEmpty($manager->getPlugins());
     }
 
 }


LukeTowers:
Do we need a negative of bootAll() as well?

harmenjanssen:
Good question. From the PluginManager's point of view it would comprise of no more than:

```php
function unbootAll()
{
    $this->booted = false;
}
```

A true `unbootAll` would have to call `unboot()` on all its plugins.  
It should be the plugin's responsibility to know how to unboot itself, I guess. 

For my problem specifically, the issue was in the way the plugins got loaded (and unloaded) between tests. I would use this simply to force the plugin through `loadPlugin()` once more.

But I'd be happy to add the above implementation of `unbootAll()`, if you want.

LukeTowers:
Is it necessary? What use case would actually need an unbootAll()?

harmenjanssen:
The only case I can think of is, again, between unit tests, to completely break down an application and return to a fresh state for the next test.

But maybe, in the spirit of [YAGNI](http://www.agilenutshell.com/yagni), we should just leave it at this and create it whenever the actual need pops up.

LukeTowers:
I would agree with that, this looks good to me.

harmenjanssen:
Hmm, I seem to have missed something, maybe?   
I tried using this by requiring `dev-develop` of `october/october`. 

But there's two versions of this file:

```
vendor/october/october/modules/system/classes/PluginManager.php <- this one contains my new function
modules/system/classes/PluginManager.php <- this one does not contain my function
```

I understand the last folder comes from this repo: https://github.com/octoberrain/system  
But what's the significance of that (read-only) repo? I can't really find where it's coming from (specifically: how it ends up in the root folder `/modules`).

Is there a way to get the new version of the `PluginManager` in there?  
I take it I don't _also_ have to offer my Pull Request to the System repo, right, since it explicitly states it's _readonly_?

LukeTowers:
`october/october` is not a package to be required. It is a base installation of October, it's not a dependency. Take a look at the composer.json for this repo. Could you paste your composer.json? @harmenjanssen

harmenjanssen:
Certainly. We're currently using October (incorrectly, I guess?) like this:

```js
{
    "require": {
        "php": ">=7.0",
        "october/october": "dev-develop",
        "grrr-amsterdam/garp-functional": "^1.0",
        "sentry/sentry": "^1.7",
        "doctrine/annotations": "1.4.*",
        "doctrine/cache": "1.6.*",
        "doctrine/collections": "1.4.*",
        "doctrine/instantiator": "1.0.*",
        "wikimedia/composer-merge-plugin": "dev-master"
    },
    "require-dev": {
        "fzaninotto/faker": "~1.4",
        "phpunit/phpunit": "~5.7",
        "phpunit/phpunit-selenium": "~1.2"
    },
    "autoload-dev": {
        "classmap": [
            "tests/TestCase.php",
            "tests/UiTestCase.php",
            "tests/PluginTestCase.php"
        ]
    },
    "scripts": {
        "test-plugins": [
            "find plugins -type d -name 'tests' -not -path '**/vendor/*' | xargs -n1 ./vendor/bin/phpunit"
        ],
        "post-create-project-cmd": [
            "php artisan key:generate"
        ],
        "post-update-cmd": [
            "php artisan october:util set build"
        ]
    },
    "config": {
        "preferred-install": "dist"
    },
    "minimum-stability": "dev",
    "prefer-stable": true,
    "extra": {
        "merge-plugin": {
            "include": [
                "plugins/*/*/composer.json"
            ],
            "recurse": true,
            "replace": false,
            "merge-dev": false
        }
    }
}
```

harmenjanssen:
Still: how will my changes end up in `october/system`?   
Do you guys split code from `october/october` when releasing a new build, or how does it work? 

In any case I have to wait for it to happen, right, there's no branch from `october/system` where I can check out upcoming changes?

LukeTowers:
Incorrectly would be a correct assessment.

The way that the octobercms/october repository is structured is that it is a full copy of a valid October installation, so that you can setup an October project with `composer create:project october/october` and be good to go. 

This full copy includes in its `composer.json` four core dependencies that together form the meat and potatoes of October so to speak:
```json
"october/rain": "~1.0",
"october/system": "~1.0",
"october/backend": "~1.0",
"october/cms": "~1.0",
```
`october/rain` points to https://github.com/octobercms/library, and is installed into the vendor directory.

`october/system` points to https://github.com/octoberrain/system, which is a readonly clone of the `modules/system` directory from the main repo

`october/backend` points to https://github.com/octoberrain/backend, which is a readonly clone of the `modules/backend` directory from the main repo

`october/cms` points to https://github.com/octoberrain/cms, which is a readonly clone of the `modules/cms` directory from the main repo

If you want to load the develop branch of October via Composer, then you should use the following in your `composer.json` (which should be a mirror of the main repo's `composer.json`)

```json
"october/rain": "~1.0",
"october/system": "dev-develop",
"october/backend": "dev-develop",
"october/cms": "dev-develop",
```
(october/rain needs to be ~1.0 because dev-develop doesn't satisfy the ~1.0 dependency of the october modules. We could fix this by specifying `replaces: { october/rain: ~1.0 }` in the octobercms/library composer.json but haven't yet because we haven't run into an issue where we needed the develop version of the library at that point in time. 

As for what triggers the updating of the octoberrain repos, I'm not 100% if that's automatically setup or if @daftspunk needs to do it manually.

harmenjanssen:
Interesting! Thanks for explaining that.   
To be honest, we started out with the repo from `composer create-project ...` and then cleverly removed the dependencies and replaced them with the `october/october` one. 🙂   
I'll make sure to return the original four dependencies.

_But_! I hope @daftspunk can shine some light on the update-process of the four subrepos.  
I mean, from the looks of it I would think my Pull Request should've been made directly to the `october/system` repo, right?

In fact: won't the `modules` directory in `october/october` be overwritten everytime when doing `composer install`?    
We're actually ignoring the `modules` directory in `.gitignore` in our project, shouldn't `october/october` be doing the same, since `composer install` will populate it?

LukeTowers:
I ignore the modules directory as well, but we can't add them to the .gitignore of the main repository since doing so would mean that we couldn't make any changes to the code as the code is tracked within those module directories allowing us to have a single place on github where people can submit PRs for improvements and changes to.

harmenjanssen:
Right, got it, that makes sense. 👍 

@daftspunk Hopefully you can elaborate a little on the update-process of the module repositories.

harmenjanssen:
Is there still nobody who can tell me anything about the release process of the child repos?   
Honestly right now, with no insight into the process, it's a pretty frustrating ordeal. In any other software project I would be able to checkout the develop or beta branch or whatever and take advantage of edge updates.

Colleagues of mine have also made PRs, already merged into `develop`, but unusable by outside code.  
I'm thinking of using my own October fork instead of the official repo and tweak it in order to ditch the 4 child repos, just so I can use the latest updates.

Again, _some_ update, _any_ update would be much appreciated.  
I'm willing to help out with automating the process or whatever would speed this up but right now I'm stuck with no way of progressing.

LukeTowers:
@harmenjanssen the splitting process is done by the build scripts (https://github.com/octobercms/library/tree/master/build) that are run once a day. However, something seems to be wrong with them as the readonly branches haven't been updated in a while. I'll be talking to @daftspunk about this today hopefully.

harmenjanssen:
Great, thanks. I'm looking forward to an update!

LukeTowers:
@harmenjanssen We've found the issue, @daftspunk has the build script set to run daily, however it hasn't been running correctly since Build 422. He just manually re-ran it and changed the cron scheduling for it so it should all be working now. Test out some composer updates and see if that helps you please.

harmenjanssen:
Yes, that's it! 
We're up to speed again.

Thanks for following up on this 👍

LukeTowers:
Glad to hear it! Not a problem, I was also wondering why my stuff wasn't pulling the latest changes either.

PR #3159, opened 2017-10-09
Call getSaveValue() of custom widgets in repeater

harmenjanssen:
This makes sure custom widgets embedded in a repeater get their `getSaveValue()` method calls with the given value.

--------------

I've been trying to create a unit test for this in the last two hours but I can't figure out how to create a bloody repeater programmatically, and have it use a custom widget. 

This is more or less my attempt:

```php
    public function testGetSaveValue()
    {
        $formField = new FormField('foo', 'bar');
        $formField->value = 'foobar';
        $repeater = new Repeater(
            new Controller,
            $formField,
            [
                'groups' => [
                    'group1' => [
                        'name' => 'group1',
                        'fields' => [
                            'title' => ['type' => 'text'],
                            'triple' => ['type' => 'tripler']
                        ],
                    ]
                ]
            ]
        );
        $value = $repeater->getSaveValue(
            [
                'group1' => ['title' => 'bananas', 'triple' => 33]
            ]
        );
        $this->assertSame(
            [
                ['title' => 'bananas', 'triple' => 99, '_group' => null]
            ],
            $value
        );
    }
```

(`tripler` in the example would be a custom widget that triples whatever integer its given)

Can you point me in the right direction for this? 
It's not loading the custom widget, nor complaining about a missing widget. There are no tests for other widgets that I can steal from.

### DIFF
--- a/modules/backend/formwidgets/Repeater.php
+++ b/modules/backend/formwidgets/Repeater.php
@@ -171,7 +171,44 @@ class Repeater extends FormWidgetBase
             }
         }
 
+        $formWidgets = array_map(
+            function (\Backend\Widgets\Form $form): array {
+                return $form->getFormWidgets();
+            },
+            $this->formWidgets
+        );
+
+        foreach ($value as $key => $group) {
+            $value[$key] = $this->getWidgetValuesForGroup(
+                $group,
+                array_get($formWidgets, $key, [])
+            );
+        }
+
         return array_values($value);
+    }
+
+    /**
+     * If a group contains a custom form widget, pass values through its getSaveValue() method.
+     * @param array $group
+     * @param array $allWidgets
+     * @return array The modified values
+     */
+    protected function getWidgetValuesForGroup(array $group, array $allWidgets): array
+    {
+        $keys = array_keys($group);
+        return array_reduce(
+            $keys,
+            function($out, $key) use ($allWidgets, $group) {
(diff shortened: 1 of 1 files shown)


LukeTowers:
Could you please give some background on why this is necessary? What is the detailed replication instructions for this bug?

harmenjanssen:
Sorry, totally, here goes:

### Expected behavior
As per this page: https://octobercms.com/docs/backend/widgets#form-widget-save-data
A backend form will call `getSaveValue()` on custom form widget with the submitted value.

### Actual behavior
However, when a custom widget is used within a repeater field, the widget's `getSaveValue()` is never called. 

This PR will gather all widgets used in the submitted data and modify the values by passing them through the respective `getSaveValue()` methods.

LukeTowers:
Thanks, are you able to submit a PR to the test plugin demonstrating the issue?

harmenjanssen:
Sure, just let me know where to find it 🙂

LukeTowers:
Ah, right. Here you go: https://github.com/daftspunk/oc-test-plugin

LukeTowers:
@harmenjanssen can you submit a PR to the https://github.com/octoberrain/oc-test-plugin repo to demonstrate the issue?

LukeTowers:
Implemented in https://github.com/octobercms/october/pull/4160

PR #3163, opened 2017-10-11
Fixes #3119

### LINKED ISSUE #3119: Cannot add readOnly attribute to dropdown or switch fields types
##### Expected behavior
When a model is in update mode I'd like to have de posibilities to disable some fields to be modifiables.

##### Actual behavior
The fields ignore the code in the filterFields function provided at Model level. The fields marked in the image below should be readOnly as specified in my code.

image

##### Reproduce steps
Try to add readOnly attribute to a model when the scope is update like the code below:

    public function filterFields($fields, $context = null)
    {
        if($this->type == Type::ORGANIZATION_UNIT || $this->type == Type::BANK)
        {
            $fields->is_inbound->readOnly = true;
            $fields->is_outbound->readOnly = true;
            $fields->currency->readOnly = true;

            $fields->is_inbound->value = false;
            $fields->is_outbound->value = false;
        }

        if($context == 'update')
        {
            $fields->type->readOnly = true;
            $fields->currency->readOnly = true;
            $fields->number->readOnly = true;
        }
    }

##### October build
(423)

romaldyminaya:
Added readOnly support to:
- RecordFinder
- Switch widget
- Relation widget including dropdown 

Fixes #3119

>This way I resolved the readOnly problem for some formWidgets in my project and hope it can help other people as well.

### DIFF
--- a/modules/backend/formwidgets/recordfinder/partials/_container.htm
+++ b/modules/backend/formwidgets/recordfinder/partials/_container.htm
@@ -2,6 +2,24 @@
 
     <span class="form-control"><?= e(trans('backend::lang.form.preview_no_record_message')) ?></span>
 
+<?php elseif ($field->readOnly): ?>
+    <span class="form-control" disabled="disabled">
+        <?php if ($value): ?>
+            <span class="primary"><?= e($nameValue) ?: 'Undefined' ?></span>
+            <?php if ($descriptionValue): ?>
+                <span class="secondary"> - <?= e($descriptionValue) ?></span>
+            <?php endif ?>
+        <?php else: ?>
+            <span class="form-control"><?= e(trans('backend::lang.form.preview_no_record_message')) ?></span>
+        <?php endif ?>
+    </span>
+
+    <input
+        type="hidden"
+        name="<?= $field->getName() ?>"
+        id="<?= $field->getId() ?>"
+        value="<?= e($value) ?>"
+        />
 <?php else: ?>
 
     <div class="recordfinder-widget" id="<?= $this->getId('container') ?>">
--- a/modules/backend/formwidgets/recordfinder/partials/_container.htm
+++ b/modules/backend/formwidgets/recordfinder/partials/_container.htm
@@ -1,26 +1,7 @@
 <?php if ($this->previewMode && !$value): ?>
 
-    <span class="form-control"><?= e(trans('backend::lang.form.preview_no_record_message')) ?></span>
+    <span class="form-control" disabled="disabled"><?= e(trans('backend::lang.form.preview_no_record_message')) ?></span>
 
-<?php elseif ($field->readOnly): ?>
-
-    <span class="form-control" disabled="disabled">
-        <?php if ($value): ?>
-            <span class="primary"><?= e($nameValue) ?: 'Undefined' ?></span>
-            <?php if ($descriptionValue): ?>
-                <span class="secondary"> - <?= e($descriptionValue) ?></span>
-            <?php endif ?>
-        <?php else: ?>
-            <span class="form-control"><?= e(trans('backend::lang.form.preview_no_record_message')) ?></span>
-        <?php endif ?>
-    </span>
-
-    <input
-        type="hidden"
-        name="<?= $field->getName() ?>"
-        id="<?= $field->getId() ?>"
-        value="<?= e($value) ?>">
-        
 <?php else: ?>
 
     <div class="recordfinder-widget" id="<?= $this->getId('container') ?>">
--- a/modules/backend/formwidgets/recordfinder/partials/_container.htm
+++ b/modules/backend/formwidgets/recordfinder/partials/_container.htm
@@ -3,6 +3,7 @@
     <span class="form-control"><?= e(trans('backend::lang.form.preview_no_record_message')) ?></span>
 
 <?php elseif ($field->readOnly): ?>
+
     <span class="form-control" disabled="disabled">
         <?php if ($value): ?>
             <span class="primary"><?= e($nameValue) ?: 'Undefined' ?></span>
@@ -18,8 +19,8 @@
         type="hidden"
         name="<?= $field->getName() ?>"
         id="<?= $field->getId() ?>"
-        value="<?= e($value) ?>"
-        />
+        value="<?= e($value) ?>">
+        
 <?php else: ?>
 
     <div class="recordfinder-widget" id="<?= $this->getId('container') ?>">
--- a/modules/backend/formwidgets/recordfinder/partials/_recordfinder.htm
+++ b/modules/backend/formwidgets/recordfinder/partials/_recordfinder.htm
@@ -4,7 +4,9 @@
     data-control="recordfinder"
     data-refresh-handler="<?= $this->getEventHandler('onRefresh') ?>"
     data-data-locker="#<?= $field->getId() ?>">
-    <span class="form-control">
+    <span class="form-control"
+        <?= $this->previewMode ? 'disabled="disabled"' : '' ?>>
+        
         <?php if ($value): ?>
             <span class="primary"><?= e($nameValue) ?: 'Undefined' ?></span>
             <?php if ($descriptionValue): ?>
--- a/modules/backend/formwidgets/recordfinder/partials/_recordfinder.htm
+++ b/modules/backend/formwidgets/recordfinder/partials/_recordfinder.htm
@@ -1,10 +1,16 @@
+<?php
+    $previewMode = false;
+    if ($this->previewMode || $field->readOnly) {
+        $previewMode = true;
+    }
+?>
 <div
     id="<?= $this->getId() ?>"
     class="field-recordfinder loading-indicator-container size-input-text"
     data-control="recordfinder"
     data-refresh-handler="<?= $this->getEventHandler('onRefresh') ?>"
     data-data-locker="#<?= $field->getId() ?>">
-    <span class="form-control" <?= $this->previewMode ? 'disabled="disabled"' : '' ?>>
+    <span class="form-control" <?= $previewMode ? 'disabled="disabled"' : '' ?>>
         <?php if ($value): ?>
             <span class="primary"><?= e($nameValue) ?: 'Undefined' ?></span>
             <?php if ($descriptionValue): ?>
@@ -15,7 +21,7 @@
         <?php endif ?>
     </span>
 
-    <?php if (!$this->previewMode): ?>
+    <?php if (!$previewMode): ?>
         <?php if ($value): ?>
             <button
                 type="button"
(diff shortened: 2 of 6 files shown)


modules/backend/widgets/form/partials/_field_switch.htm
```diff
@@ -2,7 +2,7 @@
 <div class="field-switch">
     <label><?= e(trans($field->label)) ?></label>
     <?php if ($field->comment): ?>
-        <p class="help-block"><?= $field->commentHtml ? trans($field->comment) : e(trans($field->comment)) ?></p>
+        <p class="help-block"><?= e(trans($field->comment)) ?></p>
```
LukeTowers: Why did you change this?

modules/backend/widgets/form/partials/_field_switch.htm
```diff
@@ -12,15 +12,15 @@
     value="0"
     <?= $this->previewMode ? 'disabled="disabled"' : '' ?>>
 
-<label class="custom-switch" <?= $this->previewMode ? 'onclick="return false"' : '' ?>>
+<label class="custom-switch" <?= ($this->previewMode || $field->readOnly) ? 'onclick="return false"' : '' ?>>
```
LukeTowers: If this is the only change that you need to make to the switch widget, then a better solution is to force readOnly = true to set previewMode = true

romaldyminaya: That's a good approach, I didn't want to touch any of the current attributes in order to prevent propagation of undesired behaviours later.

LukeTowers: See https://github.com/octobercms/october/commit/5f91c45f7997571290b3cbc1307a69b642d6c6e6#diff-24b825f292c5e6c4f2ce1bf5d6204b68 for an example of how @daftspunk implemented `disabled` support on the FileUpload form widget. It has a precedent so that should be fine.

modules/backend/widgets/form/partials/_field_switch.htm
romaldyminaya: Honestly I don't remembered to change that line of code, maybe that's because I have another project where I have the changes and I copy the entire file, but the one that's in the official octobercms develop branch is some or few commits ahead.

That line of code is not supossed to be changed :(

LukeTowers: Just change it back then :)

romaldyminaya:
@LukeTowers ,

I made the changes you've specified, should I close this pull request and create a new one ?

modules/backend/widgets/form/partials/_field_dropdown.htm
```diff
@@ -6,6 +6,15 @@
 <!-- Dropdown -->
 <?php if ($this->previewMode): ?>
     <div class="form-control"><?= (isset($fieldOptions[$field->value])) ? e(trans($fieldOptions[$field->value])) : '' ?></div>
+
+<!-- ReadOnly alternative -->
+<?php elseif ($field->readOnly): ?>
+    <div class="form-control" disabled="disabled"><?= (isset($fieldOptions[$field->value])) ? e(trans($fieldOptions[$field->value])) : '' ?></div>
```
LukeTowers: This would be DRYer if you used `if ($this->previewMode || $field->readOnly)` in the first conditional and then simply put a `<?= $field->readOnly ? 'disabled="disabled"' : ''; ?>` in the `<div class="form-control"` of the preview mode conditional.

LukeTowers: Then also put the 
```
<?php if ($field->readOnly): ?>
    <input
         type="hidden"
         name="<?= $field->getName() ?>"
         value="<?= $field->value ?>">
<?php endif; ?>
```

romaldyminaya: Sounds good, I'll push those changes 👍

romaldyminaya: The code would look like this with the modifications:

image

modules/backend/formwidgets/recordfinder/partials/_container.htm
```diff
@@ -2,10 +2,29 @@
 
     <span class="form-control"><?= e(trans('backend::lang.form.preview_no_record_message')) ?></span>
 
+<?php elseif ($field->readOnly): ?>
```
LukeTowers: Why can't this one be the same as the switch field? A much cleaner fix would be to simply add `|| $this->formField->readOnly` to https://github.com/octobercms/october/blob/master/modules/backend/formwidgets/RecordFinder.php#L179

romaldyminaya: I'll test if that option passes the model validation.

romaldyminaya: @LukeTowers it works but it didn't add the correct disabled color to the form field:
image

LukeTowers: Add the disabled attribute to the preview mode markup then

romaldyminaya: I've added the disabled attrute but it seems that the css its been overrided by another class as shown in the image below:

image

>I don't want to toch the css files so I'll see if there is another class that I could use.

romaldyminaya: The background-color of the disabled field should be **!important** I guess, what do you think ?

LukeTowers: How did it work with your code then?

romaldyminaya: It work with my previous code because the form-control wasn't inside a div with the .field-recordfinder class. As shown in the picture below:

image

In the picture above the **<div class"form-control">** is inside a form group directly.

>Note this result is based on the first modification I made.

LukeTowers: Try changing the disabled CSS to !important then: https://github.com/octobercms/october/blob/develop/modules/system/assets/ui/less/form.base.less#L143. Build the assets locally with `php artisan october:util compile less` but don't commit the built CSS file, only your changes to the form.base.less file.

romaldyminaya: It worked now! 👍  :)

image

I'll prepare the commit

modules/backend/formwidgets/recordfinder/partials/_recordfinder.htm
```diff
@@ -4,7 +4,9 @@
     data-control="recordfinder"
     data-refresh-handler="<?= $this->getEventHandler('onRefresh') ?>"
     data-data-locker="#<?= $field->getId() ?>">
-    <span class="form-control">
+    <span class="form-control"
+        <?= $this->previewMode ? 'disabled="disabled"' : '' ?>>
```
LukeTowers: One final minor style point, could you please make this and https://github.com/octobercms/october/pull/3163/files#diff-235ff395d0cbbf6067a5b0b339ce153fR9 inline with the class attribute as well instead of on new lines?

romaldyminaya: Of course :)

modules/backend/widgets/form/partials/_field_dropdown.htm
```diff
@@ -4,8 +4,18 @@
     $emptyOption = $field->getConfig('emptyOption', $field->placeholder);
 ?>
 <!-- Dropdown -->
-<?php if ($this->previewMode): ?>
-    <div class="form-control"><?= (isset($fieldOptions[$field->value])) ? e(trans($fieldOptions[$field->value])) : '' ?></div>
+<?php if ($this->previewMode || $field->readOnly): ?>
+    <div class="form-control"
+        <?= $field->readOnly ? 'disabled="disabled"' : ''; ?>
```
LukeTowers: This one too for the style fix please, thanks!

modules/backend/widgets/form/partials/_field_switch.htm
```diff
@@ -1,3 +1,8 @@
+<?php
+    if($field->readOnly){
+        $this->previewMode = true;
```
LukeTowers: I just realized, this will cause an unintentional bug when setting readOnly on a single switch field, it will cause the entire form after the switch field to be in previewMode. To fix it, use this code at the beginning of the file to set a local flag instead of relying on the Form object's global `previewMode` flag:
```php
<?php
    $previewMode = false;
    if ($this->previewMode || $field->readOnly) { 
        $previewMode = true;
    }
?>
```
Then replace all references to the global flag `$this->previewMode` in this partial to just be the local flag `$previewMode`

daftspunk:
Shouldn't `readOnly` simply be an alias for `disabled`?

LukeTowers:
@daftspunk, no, they have different effects: http://octobercms.com/docs/backend/forms#form-field-options. 

|`readOnly` | `disabled` | `previewMode`
| ----------- | ---------- | ---------------- |
| Disables field from being modified, but still includes the field with the saved data | Disables field from being modified and removes the field from the saved data | Applies the affect of disabled to all the fields in the form.

daftspunk:
Thanks for clarifying. I can see where readOnly might be useful, perhaps updating via JavaScript too. 

Once accepted, please don't merge this until we go stable with Build 420+.

LukeTowers:
Alright, sounds good. We might have to just spend some time going through every core FormWidget and FormField that currently exists in core to ensure that we're implementing those options correctly when it comes time to merge this.

romaldyminaya:
>Alright, sounds good. We might have to just spend some time going through every core FormWidget and FormField that currently exists in core to ensure that we're implementing those options correctly when it comes time to merge this.

@LukeTowers  Do you want me to review the readOnly behaviour in another widget ? I modifyed only these 3 because those are the ones that I noticed that don't implement the readOnly behaviour correctly.

LukeTowers:
@romaldyminaya no, that's fine. @daftspunk and I will need to go through every form widget and form field in October to ensure for our own selves that each is implementing it correctly. If you wanted to help by testing each and every one of them and letting us know which ones need to be changed to properly support those config options then that would be fantastic. Thanks for your work on this so far!

romaldyminaya:
Good @LukeTowers  so Today I've pushed the last changes with your considerations I guess we've finished at this point.

LukeTowers:
Awesome, thanks again @romaldyminaya! As @daftspunk requested, we'll wait until Build 42X is made stable before merging this in.

romaldyminaya:
Hi @LukeTowers ,

I was double checking my proyect's forms and realized that there are two more form widgets that don't implement the readOnly behaviour as well. Here is the resume:

- balloon-selector
-> disabled behaviour: no
-> readOnly behaviour: no
- checkbox
-> disabled behaviour: yes
-> readOnly behaviour: no

If you allow me I can fix those as well too.

modules/backend/formwidgets/recordfinder/partials/_recordfinder.htm
```diff
@@ -1,10 +1,16 @@
+<?php
```
romaldyminaya: @LukeTowers using my application I realized that I just forgot to apply the same snippet code you've suggested for the switch widget, to the recordfinder. Now the recordfinder and switch, both has the same behaviour.

PR #3164, opened 2017-10-11
Add getters for relation widgets

alxy:
These are available for the ``FormController`` as well, so this is just consistent and usefull in some cases.

### DIFF
--- a/modules/backend/behaviors/RelationController.php
+++ b/modules/backend/behaviors/RelationController.php
@@ -1594,4 +1594,22 @@ class RelationController extends ControllerBehavior
 
         return $this->makeConfig($config);
     }
+
+    /**
+     * Returns the manage widget used by this behavior.
+     *
+     * @return \Backend\Classes\WidgetBase
+     */
+    public function relationGetManageWidget() {
+        return $this->manageWidget;
+    }
+
+    /**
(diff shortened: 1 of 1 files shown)


LukeTowers:
Could you provide an example of these being used in an actual use case and also provide documentation for these methods?

alxy:
It is very hard to clearly desribe the actual use case (I obviously have one, otherwise I would not post this PR),a s it is rather complex indeed. But sometimes you just need access to these widgets from your controller, for example to get the fields array or some form data. I know that there are quite a few events that can be used as well, but for some reason ``formGetWidget`` does exist as well. I don't think that this need to be in the docs (again, I think ``formGetWidget`` isn't documented either), as it is really an advanced feature here and it is enough to have the API docs in this case.

LukeTowers:
I'm not opposed to this change, but out of curiosity why doesn't the events approach to accessing those widgets work for you? I would assume that would be more reliable as the widgets would always be available in the events but are not always going to be available at any point during the cycle when someone could call these methods.

alxy:
@LukeTowers I can describe this in detail on IRC if you like :)

LukeTowers:
How about on Slack @alxy? :)

PR #3170, opened 2017-10-13
Relocate Media Library

LukeTowers:
This PR relocates the Media Library and its related components into more logical locations. 

The `MediaLibrary` class, `MediaLibrary` test, `MediaLibraryItem` class, and the `|media` twig filter are being moved to the `System` module.

The `Cms\Controllers\Media` controller, `medialibrary` widget, and `mediafinder` formwidget are being moved to the `Backend` module.

Remaining TODO: Relocate all media translation keys to their appropriate modules for all languages. (@daftspunk said he'd take care of that).

> **NOTE:** This should be a fully backwards compatible change as any moved items had a stub class extending the new location remain in the old location that will also log a depreciation notice when used.


datune:
Which build # will this be made available? There are some [minor] UX improvements to the media manager pending on my part, which is why I'm asking ;-) Thanks for your efforts guys!

LukeTowers:
It should be in Build 426 right now @datune

PR #3177, opened 2017-10-17
Finnish language

mediaclinic:
This is an alpha release of Finnish language for October CMS


LukeTowers:
Please remove all untranslated keys (the ones that are still in English). There's no need to replicate them because missing keys will use the default backend language instead.

LukeTowers:
@mediaclinic closing in favour of #3237

PR #3181, opened 2017-10-18
Fix exception Class 'System\Twig\MediaLibrary' not found in build 426

ChVuagniaux:
Fix typo issue added on commit 08f9cd8


LukeTowers:
Great catch, thanks!

PR #3188, opened 2017-10-21
Don't block backend connection when a plugin migration trigger exception

ChVuagniaux:
Without this commit if a plugin migration fail nobody can connect to the backend


LukeTowers:
Good solution! I knew this bug existed, but since you can access the backend anyways by just going to a backend url after signing in and getting this exception I never bothered to do anything with it.

Issue #2383, opened 2016-09-21
Empty list of partials/layouts/content when /templates directory exists


vojtasvoboda:
##### Expected behavior

At backend -> CMS there should be list of content/pages/partials/layout files.
##### Actual behavior

When directory `/templates` exists at root of my project (for example with raw templates for coding) at backend -> CMS I have empty lists of partials/layouts/contents/pages. Because of [realpath function](https://github.com/octobercms/october/blob/master/modules/system/traits/ViewMaker.php#L56) which returns root path instead of false.
##### Reproduce steps

Create `/templates` directory at root. At backend show list of pages/partials/contents/layouts.
##### October build

363

daftspunk:
I was able to replicate this. Damn, we will need to think of a more robust solution for detecting the difference between

- `templates` (not real path)
- `/home/spunky/www/modules/cms/controllers/index/templates.htm` (real path)

LukeTowers:
@vojtasvoboda is this still an issue in the current build of October?

vojtasvoboda:
@LukeTowers yes it is still a problem when `/templates` folder exists, pages tree is empty:

![snimek obrazovky 2017-10-19 v 13 39 04](https://user-images.githubusercontent.com/374917/31769212-fd2bb4d4-b4d2-11e7-9424-f2e86467bf2f.png)

Build 425.

LukeTowers:
@vojtasvoboda could you dig into this a bit more and make a PR to fix it? We don't have any time to work on it ourselves right now so if it's going to be fixed it'll have to be done by the community. 

I appreciate the report and follow up though!

vojtasvoboda:
@LukeTowers please check #3190 PR. It fixes the problem described in this issue and if I correctly understood that `makePartial()` method can also take an absolute path, it should work fine.

LukeTowers:
Awesome, thanks @vojtasvoboda! Could you see if anyone could provide a second test of the fix and verify it on the PR? I don't have time right now to test it myself so if we could get a second person to verify that it works, I'll merge it. Otherwise it'll have to wait until I can find the time to review it myself.

Thanks again for reporting and providing a fix for this!

vojtasvoboda:
@LukeTowers I have put it on Slack, so we will see :-)

CptMeatball:
Just tested this and works fine!

LukeTowers:
[...]

PR #3190, opened 2017-10-22
Better realpath recognize

vojtasvoboda:
Related to #2383


CptMeatball:
I can confirm this works on both `build 419` and `build 426`.

LukeTowers:
Thanks @CptMeatball!

Issue #3114, opened 2017-09-15
Problem Richeditor & Window media insert image


thanglegeo:
![oct](https://user-images.githubusercontent.com/31978914/30466335-526edfb6-9a07-11e7-9ada-fbd1ca6604a4.jpg)

##### October build
Build 423

LukeTowers:
This was fixed in https://github.com/octobercms/october/pull/3113. Please apply those changes locally and let me know if it works.

thanglegeo:
I have edited 4 files, but it doesnt work

richeditor.css
richeditor.less
build-min.js
richeditor.js

Thanks

LukeTowers:
@thanglegeo could you please update to the latest build and empty your browser cache (or try in incognito) and let me know if it works now?

LukeTowers:
In theory fixed by #3113 or #3194

PR #3194, opened 2017-10-23
Fix richeditor popup z-index inside modal

### LINKED ISSUE #3098: Editor: Table formatter popup is hiding behind another popup

Fl0Cri:
Fixes #3098 without breaking #3111 

The 9995 z-index comes from the compiled css, but as it isn't !important, it was overridden by the inline z-index computed by the editor.

Screenshot with these changes applied:
Normal editor:
normal

Same inside a modal (fixed):
inside-modal

Editor showing a page snippet (still working):
snippet

One situation that doesn't work even with this fix is showing a snippet dialog when the richeditor is inside a modal (using the "Richeditor snippets" plugin) but it is not officially supported so not realy in the scope of these issues.
To support this specific case, it would require to raise the snippet dialog z-index, but there may be side-effects I am not able to test right now.


LukeTowers:
@Fl0Cri does this also fix #3114?

Fl0Cri:
#3114 should be already fixed by #3111.
I see no reason why the richeditor could come in front of the modal (its z-index is now 1).

Upgrading to build 424+ should fix the issue.

TimFoerster:
I checked #3098. it works perfectly!

Issue #2861, opened 2017-05-08
Trigger condition from checkbox list field type


neilcarpenter1:
I can't find any other issues relating to this, so forgive me if this has been raised before.
I was wondering if it's possible to compare a trigger condition to selected values on a checkboxlist field type?

```
roles:
    label: Roles
    type: checkboxlist
    options:
      teacher: Teacher
      governer: Governer

class:
      label: Class
      type: dropdown
      options:
        one: Class one
        two: Class two
      trigger:
        action: show
        field: roles
        condition: value[teacher]
```

LukeTowers:
It should be in theory. Are you saying that it isn't possible in practice?

neilcarpenter1:
It's not working for me no

daftspunk:
This is a complex case. Please submit a Pull Request to the [test plugin for investigation](https://github.com/daftspunk/oc-test-plugin) that demonstrates the issue. Don't forget to include step by step instructions how how to replicate it *using the test plugin*.

LukeTowers:
@neilcarpenter1 @jimcottrell could either of you make a PR to fix this bug?

jimcottrell:
I'll take a look.

daftspunk:
Thanks all who contributed to the research here.

This is solved with the following:

```
roles:
    label: Roles
    type: checkboxlist
    options:
      teacher: Teacher
      governer: Governer

class:
      label: Class
      type: dropdown
      options:
        one: Class one
        two: Class two
      trigger:
        action: show
        field: roles[]
        condition: value[teacher]
```

Note the diff:

```
        field: roles[]
```

Available in Build 435+.

PR #3195, opened 2017-10-23
Fix checkboxlist triggers for #2861

jimcottrell:
Reference: #2861 

@LukeTowers This works, but I'm open for feedback on any improvements.

The FormField class has of course to this point not had a reference to the parent form, but I didn't see any way to fix this bug without being able to grab the trigger field's object, since its field type impacts the selector the JS trigger API needs to use to watch for changes.

I don't know how often someone would set up a trigger on a field without that field being set up through a form, but I left in logic to still have a generic selector if the field has no parent form. That logic would of course still fail if the trigger field is a checkboxlist.

### DIFF
--- a/modules/backend/classes/FormField.php
+++ b/modules/backend/classes/FormField.php
@@ -22,6 +22,11 @@ class FormField
      * @var string Form field name.
      */
     public $fieldName;
+    
+    /**
+     * @var Form The form this field belongs to
+     */
+    public $form;
 
     /**
      * @var string If the field element names should be contained in an array. Eg:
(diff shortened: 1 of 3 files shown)


daftspunk:
Thank you

This isn't really something specific to the checkboxlist, more to do with how the `trigger` form field option is mutilating the field array name `[]` when converting to the trigger JS interface. 

```
        page_404:
            tab: Pages
            commentAbove: This field is enabled when the 404 page is selected
            trigger:
                field: pages[]
                condition: value[404]
                action: enable
```

See the commit here, hopefully it can resolve the issue elegantly https://github.com/octobercms/october/commit/2164c0761623ac0dcd8e4c216a27e0ca17f2d23b

Issue #2329, opened 2016-08-24
Set default value for list filter widget


mplodowski:
I don't currently see a way to set default value for filter widget.

For example I have `switch` type filter and I want to set default value to 1. This will fire my first condition specified in configuration

Maybe we can add `default` field in filter configuration that will set default value for filter like so:

```
approved:
    label: Approved
    type: switch
    conditions:
        - is_approved <> true
        - is_approved = true
    default: 1
```

daftspunk:
Makes sense for switch + checkbox types. Not so sure about the date range and group types.

LukeTowers:
The support demand for October is increasing, while the available time the founders have to give remains the same. These issues will be handled in time, however there are ways you can help:

- Submit a pull request that introduces this feature and wait for it to be reviewed
- [Make a financial contribution to the project](http://octobercms.com/fundraising) so we can allocate more time towards making improvements like these that benefit everyone

Closing as it has been over a month since any activity on this occurred.

arifnyet:
@LukeTowers  @daftspunk  i have a pr for this... default value for switch and checbox... usage as follows 


`'ShowTrashed' => [
                    'label' => 'Hide Trashed',
                    'type' => 'switch',
                    'value' => '2',
                    'conditions' => ['deleted_at IS NOT NULL', 'deleted_at IS NULL']
                ],` 


`'ShowTrashed' => [
                    'label' => 'Hide Trashed',
                    'type' => 'checkbox',
                    'value' => '1',
                    'conditions' => ['deleted_at IS NOT NULL', 'deleted_at IS NULL']
                ],`

switch 0: off 1: inter 2: on
checkbox 0: uncheck 1: checked

PR #3198, opened 2017-10-25
Arifnyet switch/checkbox   default value

arifnyet:
Pr for Set default value for list filter widget #2329

### DIFF
--- a/modules/backend/widgets/Filter.php
+++ b/modules/backend/widgets/Filter.php
@@ -566,8 +566,8 @@ class Filter extends WidgetBase
         /*
          * Set scope value
          */
-        if($scopeType=='checkbox' || $scopeType=='switch'){
-            $scope->value = isset($config['value']) ? $config['value'] : $this->getScopeValue($scope);
+        if($scopeType==='checkbox' || $scopeType==='switch'){
+            $scope->value = isset($config['default']) ? $config['default'] : $this->getScopeValue($scope);
         } else{
             $scope->value = $this->getScopeValue($scope);
         }
--- a/modules/backend/widgets/Filter.php
+++ b/modules/backend/widgets/Filter.php
@@ -566,7 +566,11 @@ class Filter extends WidgetBase
         /*
          * Set scope value
          */
-        $scope->value = $this->getScopeValue($scope);
+        if($scopeType=='checkbox' || $scopeType=='switch'){
+            $scope->value = isset($config['value']) ? $config['value'] : $this->getScopeValue($scope);
+        } else{
+            $scope->value = $this->getScopeValue($scope);
+        }
 
         return $scope;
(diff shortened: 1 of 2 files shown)


modules/backend/widgets/Filter.php
```diff
@@ -566,7 +566,11 @@ protected function makeFilterScope($name, $config)
         /*
          * Set scope value
          */
-        $scope->value = $this->getScopeValue($scope);
+        if($scopeType=='checkbox' || $scopeType=='switch'){
```
LukeTowers: Please use strict type comparisons (`===`), not loose type comparisons.

modules/backend/widgets/Filter.php
```diff
@@ -566,7 +566,11 @@ protected function makeFilterScope($name, $config)
...
+            $scope->value = isset($config['value']) ? $config['value'] : $this->getScopeValue($scope);
```
LukeTowers: The config option should be `'default'`, not `'value'`. Also, does this prevent the user from overriding the default? It seems like if a default is defined then the value retrieved from `$this->getScopeValue($scope)` would not be used.

modules/backend/widgets/filter/partials/_scope_checkbox.htm
```diff
@@ -2,6 +2,6 @@
 <div
     class="filter-scope checkbox custom-checkbox"
     data-scope-name="<?= $scope->scopeName ?>">
-    <input type="checkbox" id="<?= $scope->getId() ?>" <?= $scope->value ? 'checked' : '' ?> />
+    <input type="checkbox" id="<?= $scope->getId() ?>" <?= $scope->value==1 ? 'checked' : '' ?> />
```
LukeTowers: Again, strict type comparison please. Also, why is this check necessary?

modules/backend/widgets/Filter.php
arifnyet: @LukeTowers  done

modules/backend/widgets/Filter.php
arifnyet: @LukeTowers  done

modules/backend/widgets/filter/partials/_scope_checkbox.htm
arifnyet: @LukeTowers  it is not necessary.. done changing

arifnyet:
@LukeTowers  updated the pr

LukeTowers:
Did this address the issue with the default option overriding the user selected one all of the time?

arifnyet:
@LukeTowers sorry but i think this method is redundant with session caches.... im working on new method i close this pr for now... i submit a new one when it is ready

PR #3199, opened 2017-10-25
Check admin permissions before displaying global media buttons in Rich Editor

Harti:
**Fixes / Related:** 
https://github.com/rainlab/blog-plugin/issues/325

**Current behavior:**
If they're defined as toolbar buttons globally from the Editor Settings, an admin user will see media buttons regardless of their permissions. Using a media button to upload media without the permission to manage media will result in an exception.

**Fixed behavior:**
If they're defined as toolbar buttons globally from the Editor Settings, an admin user will only see media buttons if they have permission to do so.

### DIFF
--- a/modules/backend/formwidgets/RichEditor.php
+++ b/modules/backend/formwidgets/RichEditor.php
@@ -1,7 +1,7 @@
 <?php namespace Backend\FormWidgets;
 
 use App;
-use Backend\Facades\BackendAuth;
+use BackendAuth;
 use File;
 use Event;
 use Lang;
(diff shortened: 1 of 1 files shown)


modules/backend/formwidgets/RichEditor.php
```diff
@@ -86,6 +87,16 @@ public function prepareVars()
         $this->vars['toolbarButtons'] = $this->evalToolbarButtons();
 
         $this->vars['globalToolbarButtons'] = EditorSetting::getConfigured('html_toolbar_buttons');
+
+        if (!BackendAuth::getUser()->hasPermission(['media.manage_media'])) {
```
LukeTowers: Avoid using hasPermission, use hasAccess instead as it factors in whether the user is a super user or not. Also, you don't need to pass the check in an array, just a string is fine.

modules/backend/formwidgets/RichEditor.php
```diff
@@ -1,6 +1,7 @@
 <?php namespace Backend\FormWidgets;
 
 use App;
+use Backend\Facades\BackendAuth;
```
LukeTowers: No need to use the fully qualified path the BackendAuth facade, just `use BackendAuth` is fine.

LukeTowers:
Thanks, we'll merge this after we push build 420+ to stable (next development cycle).

LukeTowers:
Closing in favour of https://github.com/octobercms/october/pull/4059

PR #3201, opened 2017-10-26
Arifnyet switch/checkbox   default value #2329

arifnyet:
@LukeTowers 
this pr should be compatible.. 
for checkbox and switch type
it will only use the default value set by the user if the user session is not yet set... ie after login and viewing the list for the first time... the default state/value for the switch/checkbox is according to the inputted default value...
however after the user make changes to it... ie uncheck it or change it state... it will remain the same until the user change it back or clear his/her session/cookies.. 
i hope this is the intended behavior 
usage as follows

```php
'ShowTrashed' => [
    'label' => 'Hide Trashed',
    'type' => 'switch',
    'default' => '2',
    'conditions' => ['deleted_at IS NOT NULL', 'deleted_at IS NULL']
],
``` 
```php
'ShowTrashed' => [
    'label' => 'Hide Trashed',
    'type' => 'checkbox',
    'default' => '1',
    'conditions' => ['deleted_at IS NOT NULL', 'deleted_at IS NULL']
],
```
switch 0: off 1: inter 2: on
checkbox 0: uncheck 1: checked

### DIFF
--- a/modules/backend/widgets/Filter.php
+++ b/modules/backend/widgets/Filter.php
@@ -566,8 +566,8 @@ class Filter extends WidgetBase
         /*
          * Set scope value
          */
-        if($scopeType=='checkbox' || $scopeType=='switch'){
-            $scope->value = isset($config['value']) ? $config['value'] : $this->getScopeValue($scope);
+        if($scopeType==='checkbox' || $scopeType==='switch'){
+            $scope->value = isset($config['default']) ? $config['default'] : $this->getScopeValue($scope);
         } else{
             $scope->value = $this->getScopeValue($scope);
         }
--- a/modules/backend/widgets/Filter.php
+++ b/modules/backend/widgets/Filter.php
@@ -564,12 +564,11 @@ class Filter extends WidgetBase
         $scope->displayAs($scopeType, $config);
 
         /*
-         * Set scope value
+         * Set scope value if session/cookies is not set then retrieve default value for switch/checkbox
          */
-        if($scopeType==='checkbox' || $scopeType==='switch'){
-            $scope->value = isset($config['default']) ? $config['default'] : $this->getScopeValue($scope);
-        } else{
-            $scope->value = $this->getScopeValue($scope);
+        $scope->value = $this->getScopeValue($scope);
+        if(!$scope->value && isset($config['default']) && ($scopeType==="checkbox" || $scopeType==="switch")){
+            $scope->value = $config['default'];
         }
 
         return $scope;
--- a/modules/backend/widgets/Filter.php
+++ b/modules/backend/widgets/Filter.php
@@ -567,7 +567,7 @@ class Filter extends WidgetBase
          * Set scope value if session/cookies is not set then retrieve default value for switch/checkbox
          */
         $scope->value = $this->getScopeValue($scope);
-        if(!$scope->value && isset($config['default']) && ($scopeType==="checkbox" || $scopeType==="switch")){
+        if(!isset($scope->value) && isset($config['default']) && ($scopeType==="checkbox" || $scopeType==="switch")){
             $scope->value = $config['default'];
         }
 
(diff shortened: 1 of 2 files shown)


LukeTowers:
What if the value set by the user is a falsey value? I.e. 0?

arifnyet:
@LukeTowers for the values as long as it is within range ie 0/1 for checkbox 0-2 for switch.... other than that it will throw error.... should i restrict the value as well?

LukeTowers:
I meant moreso that if the user has it set to 0, (i.e. disabled), then your comparison with the getScopeValue will cause the default value to be used instead, effectively making it so that the user can't disable the filter at all.

arifnyet:
@LukeTowers hmmmm im at lost here... the default value only works for the first time the list is visited... for the second time onward it will take from session.... 

lets say for checkbox.... 

if the user set the default to 0... the user visit the list page the checkbox will render as unchecked... when the user checks it... and then move to another page... after a while he moves back to the list... the checkbox will render as checked according to its previous state....

if the user did not set the default the same will happen as above.... 

if the user set the default value as 1... the user visit the list page the checkbox will render as checked... when the user unchecks it... and then move to another page... after a while he moves back to the list... the checkbox will render as unchecked according to its previous state....

my comparison with getScopeValue is only executed if the session/cookies is yet set....

hope that clarifies the behavior

LukeTowers:
Have you tested that? Because your comparison would return that the session has not been set if `getScopeValue()` returns 0 as an integer (a falsey value). I haven't tested it myself, but that's the question I have from reading your code.

arifnyet:
@LukeTowers already did a thorough test.... scope->value can only receive value(int) within range even without my code.... so basicly the value stored is only 0/1 for checkbox and 0-2 for switch however i changed the code a bit so that only if the $scope->value did not have a value that the code will execute...

LukeTowers:
That check won't change the theory either, as even if `getScopeValue()` returns null, `$scope->value` is still going to get set.

arifnyet:
@LukeTowers if getScopeValue() returns null.. the $scope-value is going to take default value... is that not the intended behaviour?

LukeTowers:
Yes, but that isn't what `isset()` checks. `isset()` returns true even if the value is set to null.

arifnyet:
@LukeTowers i tried on http://phpfiddle.org/ many times but `scope=null;$isset($scope)` return false and can you take a look here http://php.net/manual/en/function.isset.php

LukeTowers:
Huh, looks like you were correct. Odd, I wonder why I thought that `isset()` returned true on null value variables.

arifnyet:
@LukeTowers so the fix is okay right?

LukeTowers:
I was just looking through the existing code and default values are already supported: https://github.com/octobercms/october/blob/master/modules/backend/widgets/Filter.php#L755. We just need to pass the default value to the getScopeValue method. 

Essentially all that is actually required is `$scope->value = $this->getScopeValue($scope, @$config['default']);`

arifnyet:
@LukeTowers oh i see.....

LukeTowers:
@arifnyet if you want to make that change in your PR and then test that it works correctly I'll merge it.

arifnyet:
@LukeTowers i make it but i need to ask you which is better 
this 

`$default = isset($config['default']) ? $config['default'] : null; `
`$scope->value = $this->getScopeValue($scope, $default);`

or

`$scope->value = $this->getScopeValue($scope, @$config['default']);`

can you advise?

arifnyet:
@LukeTowers and this applies to all type of scopes...

LukeTowers:
Awesome, thanks for your work on this @arifnyet!

LukeTowers:
@arifnyet could you please make a pull request to https://github.com/octobercms/docs documenting this new option?

arifnyet:
@LukeTowers Ok sure... im writing it..

PR #3205, opened 2017-10-26
Fixes #2888. Set max url length to 8000

### LINKED ISSUE #2888: Url database field too small for request logs.
##### Expected behavior
Saving all url requests to request log

##### Actual behavior
```exception 'PDOException' with message 'SQLSTATE[22001]: [Microsoft][ODBC Driver 11 for SQL Server][SQL Server]String or binary data would be truncated.' in ....\vendor\laravel\framework\src\Illuminate\Database\Connection.php:362
Stack trace:
#0 ....\vendor\laravel\framework\src\Illuminate\Database\Connection.php(362): PDOStatement->execute(Array)
#1 ....\vendor\laravel\framework\src\Illuminate\Database\Connection.php(614): Illuminate\Database\Connection->Illuminate\Database\{closure}(Object(Illuminate\Database\SqlServerConnection), 'insert into [sy...', Array)
#2 ....\vendor\laravel\framework\src\Illuminate\Database\Connection.php(578): Illuminate\Database\Connection->runQueryCallback('insert into [sy...', Array, Object(Closure))
#3 ....\vendor\laravel\framework\src\Illuminate\Database\Connection.php(363): Illuminate\Database\Connection->run('insert into [sy...', Array, Object(Closure))
#4 ....\vendor\laravel\framework\src\Illuminate\Database\Connection.php(320): Illuminate\Database\Connection->statement('insert into [sy...', Array)
#5 ....\vendor\laravel\framework\src\Illuminate\Database\Query\Processors\SqlServerProcessor.php(18): Illuminate\Database\Connection->insert('insert into [sy...', Array)
#6 ....\vendor\laravel\framework\src\Illuminate\Database\Query\Builder.php(1732): Illuminate\Database\Query\Processors\SqlServerProcessor->processInsertGetId(Object(October\Rain\Database\QueryBuilder), 'insert into [sy...', Array, 'id')
#7 [internal function]: Illuminate\Database\Query\Builder->insertGetId(Array, 'id')
#8 ....\vendor\laravel\framework\src\Illuminate\Database\Eloquent\Builder.php(936): call_user_func_array(Array, Array)
#9 ....\vendor\october\rain\src\Database\Builder.php(131): Illuminate\Database\Eloquent\Builder->__call('insertGetId', Array)
#10 ....\vendor\laravel\framework\src\Illuminate\Database\Eloquent\Model.php(1608): October\Rain\Database\Builder->__call('insertGetId', Array)
#11 ....\vendor\laravel\framework\src\Illuminate\Database\Eloquent\Model.php(1608): October\Rain\Database\Builder->insertGetId(Array, 'id')
#12 ....\vendor\laravel\framework\src\Illuminate\Database\Eloquent\Model.php(1578): Illuminate\Database\Eloquent\Model->insertAndSetId(Object(October\Rain\Database\Builder), Array)
#13 ....\vendor\laravel\framework\src\Illuminate\Database\Eloquent\Model.php(1485): Illuminate\Database\Eloquent\Model->performInsert(Object(October\Rain\Database\Builder), Array)
#14 ....\vendor\october\rain\src\Database\Model.php(1069): Illuminate\Database\Eloquent\Model->save(Array)
#15 ....\vendor\october\rain\src\Database\Model.php(1115): October\Rain\Database\Model->saveInternal(NULL, Array)
#16 ....\modules\system\models\RequestLog.php(51): October\Rain\Database\Model->save()
#17 ....\modules\cms\classes\Controller.php(197): System\Models\RequestLog::add()
#18 ....\modules\cms\classes\CmsController.php(47): Cms\Classes\Controller->run('reizen/canada/k...')
```
Urls for all intents and purposes can be up to 2048 characters long. The default url length for the log is 254 in sql server.
It should be 2048. This causes errors when urls are too long with funny warnings and such.

##### Reproduce steps
use sql server as database
make page with longer than 254 url
access it via the natural way.
Error entry will appear

##### Fix by:
https://github.com/octobercms/october/blob/master/modules/system/database/migrations/2014_10_01_000012_Db_System_Request_Logs.php#L14

Needs to be modified to support 2048 long urls. string(2048) for sql server. but I don't know about other databases.

https://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers#417184

NemoD503:
It seems to be a better to set  length to `8000`.
There is no strict spec for max url length, but there is [recommendation](https://tools.ietf.org/html/rfc7230#section-3.1.1) about length.

> Various ad hoc limitations on request-line length are found in
   practice.  It is RECOMMENDED that all HTTP senders and recipients
   support, at a minimum, request-line lengths of 8000 octets.

So url in utf-8 encoding can have from 2000 up to 8000 chars, because  [RFC3629](https://tools.ietf.org/html/rfc3629) says:

> In UTF-8, characters from the U+0000..U+10FFFF range (the UTF-16 accessible range) are encoded using sequences of 1 to 4 octets.

PS:
However, in any case, we can not guarantee that there will not be an attempt to write a longer string than we expect, so I have a question, do I still need to validate the data before saving? I can add this feature to this PR.

### DIFF
--- a/modules/system/database/migrations/2014_10_01_000012_Db_System_Request_Logs.php
+++ b/modules/system/database/migrations/2014_10_01_000012_Db_System_Request_Logs.php
@@ -3,8 +3,6 @@
 use October\Rain\Database\Schema\Blueprint;
 use October\Rain\Database\Updates\Migration;
 
-
-
 class DbSystemRequestLogs extends Migration
 {
     const MAX_URL_LENGTH = 8000; //https://tools.ietf.org/html/rfc7230#section-3.1.1
--- a/modules/system/database/migrations/2014_10_01_000012_Db_System_Request_Logs.php
+++ b/modules/system/database/migrations/2014_10_01_000012_Db_System_Request_Logs.php
@@ -3,6 +3,8 @@
 use October\Rain\Database\Schema\Blueprint;
 use October\Rain\Database\Updates\Migration;
 
+const MAX_URL_LENGTH = 8000; //https://tools.ietf.org/html/rfc7230#section-3.1.1
+
 class DbSystemRequestLogs extends Migration
 {
     public function up()
@@ -11,7 +13,7 @@ class DbSystemRequestLogs extends Migration
             $table->engine = 'InnoDB';
             $table->increments('id');
             $table->integer('status_code')->nullable();
-            $table->string('url')->nullable();
+            $table->string('url', MAX_URL_LENGTH)->nullable();
             $table->text('referer')->nullable();
             $table->integer('count')->default(0);
             $table->timestamps();
--- a/modules/system/database/migrations/2014_10_01_000012_Db_System_Request_Logs.php
+++ b/modules/system/database/migrations/2014_10_01_000012_Db_System_Request_Logs.php
@@ -3,17 +3,19 @@
 use October\Rain\Database\Schema\Blueprint;
 use October\Rain\Database\Updates\Migration;
 
-const MAX_URL_LENGTH = 8000; //https://tools.ietf.org/html/rfc7230#section-3.1.1
+
 
 class DbSystemRequestLogs extends Migration
 {
+    const MAX_URL_LENGTH = 8000; //https://tools.ietf.org/html/rfc7230#section-3.1.1
+
     public function up()
     {
         Schema::create('system_request_logs', function (Blueprint $table) {
             $table->engine = 'InnoDB';
             $table->increments('id');
             $table->integer('status_code')->nullable();
-            $table->string('url', MAX_URL_LENGTH)->nullable();
+            $table->string('url', static::MAX_URL_LENGTH)->nullable();
             $table->text('referer')->nullable();
             $table->integer('count')->default(0);
             $table->timestamps();
--- a/modules/system/database/migrations/2014_10_01_000012_Db_System_Request_Logs.php
+++ b/modules/system/database/migrations/2014_10_01_000012_Db_System_Request_Logs.php
@@ -5,15 +5,13 @@ use October\Rain\Database\Updates\Migration;
 
 class DbSystemRequestLogs extends Migration
 {
-    const MAX_URL_LENGTH = 8000; //https://tools.ietf.org/html/rfc7230#section-3.1.1
-
     public function up()
     {
         Schema::create('system_request_logs', function (Blueprint $table) {
             $table->engine = 'InnoDB';
             $table->increments('id');
             $table->integer('status_code')->nullable();
-            $table->string('url', static::MAX_URL_LENGTH)->nullable();
+            $table->string('url')->nullable();
             $table->text('referer')->nullable();
             $table->integer('count')->default(0);
             $table->timestamps();
--- a/modules/system/database/migrations/2017_10_30_000024_Db_System_Request_Logs_Update.php
+++ b/modules/system/database/migrations/2017_10_30_000024_Db_System_Request_Logs_Update.php
@@ -1,0 +1,21 @@
+<?php
+
+use October\Rain\Database\Schema\Blueprint;
+use October\Rain\Database\Updates\Migration;
+
+class DbSystemRequestLogsUpdate extends Migration
+{
+    const MAX_URL_LENGTH = 8000; //https://tools.ietf.org/html/rfc7230#section-3.1.1
+
+    public function up()
+    {
+        Schema::table('system_request_logs', function (Blueprint $table) {
+            $table->string('url', static::MAX_URL_LENGTH)->change();
+        });
+    }
+
+    public function down()
+    {
+        //this migration should not be downgraded.
+    }
+}
--- a/modules/system/models/RequestLog.php
+++ b/modules/system/models/RequestLog.php
@@ -42,7 +42,7 @@ class RequestLog extends Model
         }
 
         $record = static::firstOrNew([
-            'url' => substr(Request::fullUrl(), 0, 255),
+            'url' => mb_substr(Request::fullUrl(), 0, 8000),
             'status_code' => $statusCode,
         ]);
 


LukeTowers:
What is the real world application of having a URL that is 8000 characters long? Just because that's the recommended RFC implementation doesn't necessarily mean that is what we should be doing for a real world application. While I can see the use case for supporting URLs longer than 255 characters, I'm not quite sure that I would agree we need to support URLs with a length of 8000 characters. Also, won't this cause each row in the table to now take up 7,745 more bytes than before?

munxar:
@LukeTowers 

> What is the real world application of having a URL that is 8000 characters long?

If you store state in the query string (e.g. a shop configurator) it can easily get longer than 255 chars.

> Also, won't this cause each row in the table to now take up 7,745 more bytes than before?

No, it should not take more bytes than before, because it uses the varchar type (except you write more than 255 chars).

LukeTowers:
Thanks! Could you move the const declaration into the class itself? We don't need it to be global, just make it a class constant instead.

munxar:
Besides that, shouldn't it be a new migration that alters the column?

LukeTowers:
Your choice. Making a new migration will implement the fix for all users, changing the existing one will only affect new users.

As a note, we'll have to wait until the next dev cycle (after we release build 420+ as stable, hopefully soon) to merge this.

munxar:
imho existing migrations should never be touched once released.

LukeTowers:
@munxar ah, just realized that @NemoD503 was the original author. @NemoD503 please make the suggested changes.

NemoD503:
> Also, won't this cause each row in the table to now take up 7,745 more bytes than before?

It depends on database realisation, but @munxar is right -  for mysql and postgres it will be `varchar`, for sqlite, it will be `text` in any case.

> Could you move the const declaration into the class itself?

Of course! This was an accidental mistake by inattention, shame on me..

> Besides that, shouldn't it be a new migration that alters the column?
> imho existing migrations should never be touched once released.

Yes, for existing, running application existing migrations should never be touched once released, but october is not running application, it's a product. So it seems, that it's possible to change installation process for this product. Also this file already [has been changed](https://github.com/octobercms/october/commits/master/modules/system/database/migrations/2014_10_01_000012_Db_System_Request_Logs.php) several times.
At firs, if we create new migration with `alter`, we touch this table for **all running projects,** I think, this table can have many rows, so `alter` can run for a long time, so we can break something for all projects during october update. And secondary, bug, that will be fixed with this PR is very occasional, and  is not typical for mysql and sqlite, because mysql cuts off extra characters in the string, sqlite has `text` type without size.
So, imho **in this case** it's safer to change existing migration and fix this bug only for newly installations.

PS. If it was necessary to fix this bug for existing applications, it would be right to add new column with new size, with default null values. After that separate cron job copy old values from old column to new column. This job can run for a long time, but it's not a problem. And anyway new installation must have only one column with correct size.

munxar:
>  ... but october is not running application, it's a product.

wat? 

> Also this file already has been changed several times.

mistakes from the past should be no argument to not improve.

If we change existing migrations, the following scenario will happen:
two installations of october with the **same version**, one was installed new, the other upgraded to this version.
both have **different behaviour**, one can log urls < 255 the other with < 8000.

How should we track bugs, or reproduce them, if every installation with the **same version** behaves differently?

LukeTowers:
@munxar if you could get a reference on alter varchar length not touching existing data, then I would agree with your request for a new migration file. I know there have been migrations that have been changed in the past to add an index by default, but I wasn't overly concerned with that particular change because if anyone reported it going slow (because they had an old copy that was updated) I'd just tell them to add an index.

munxar:
@LukeTowers Yes, the reference is: I tried it :-).
10 000 000 Records in the logs.

```php
    // very fast < 1 sec
    public function up()
    {
        Schema::table('system_request_logs', function($table) {
            $table->string('url', 8000)->change();
        });
    }
    // took 2 minutes on my macbook
    public function down()
    {
        Schema::table('system_request_logs', function($table) {
            $table->string('url', 255)->change();
        });
    }
```

LukeTowers:
So alter increasing the limit will not take long, but alter decreasing it will?

munxar:
@LukeTowers exactly.

LukeTowers:
@munxar any suggestions for if someone runs down?

NemoD503:
Yes, you are right, that alter for incraising varchar will not touch data.
And, yes it's necessary to create new migration, because there is substring limit in the code, so when old users get new version, substr limit will be 8000, but database field will be 250 - it will very bad situation. But, i think it is not neccessary to decraising this migration, because, in this case data can be touched and some data can be destroyed, or some databases can throw error maybe.
So, i will make new migration on monday.

munxar:
@LukeTowers yes, run from console (no timeout) and wait ... :-)
usually october:down is not done on a production system, so not a problem at all.

NemoD503:
> Yes, you are right, that alter for increasing `varchar` will not touch data.

But, unfortunately this is not true.. On mysql `alter` always copy whole data.. ([one](https://stackoverflow.com/a/12774774/4097586)) ([docs](https://dev.mysql.com/doc/refman/5.7/en/alter-table.html#alter-table-performance))
It's possible to run with `algorithm=inplace`, but only with mysql 5.7 and only if table was created in mysql 5.7.. and it seems, that laravel `ORM` do not support this feature.. So in our case whole data will be copied on change.. So there are some risks when updating..

I've made new migration, but now I think, is this issue really critical, or maybe  it's possible to live with small field..

NemoD503:
For clarity, when I said "touch data", I mean, that data will be copied and migration time will depend on the number of rows in the table. Any way any data will not be corrupted or lost.

LukeTowers:
@NemoD503 @munxar what's the status on this? Is this still something that's still wanted?

NemoD503:
I think it is no longer necessary.

Issue #3215, opened 2017-10-31
backend auth bug


tim0991:
After sign out, i view the url /backend/backend/userroles/update/1, i get below error 

`Call to a member function isSuperUser() on null modules/backend/controllers/userroles.php line 56`
##### October build
use develop branch
commit 4a6e0e1e0e2c3facebc17e0db38c5b4d4cb05bd0

LukeTowers:
How is this a bug? The backend isn't going to work unless you are signed in

munxar:
how about something like that?
```php
        /*
         * Only super users can access
         */
        $this->bindEvent('page.beforeDisplay', function() {
            if($this->user == null) {
                 // not logged in
                 return Response::make(View::make('backend::not_authorized'), 401);    
            }

            if (!$this->user->isSuperUser()) {
                  return Response::make(View::make('backend::access_denied'), 403);    
            }            
        });
```

LukeTowers:
I would agree with that @munxar, but it should already be handled in here: https://github.com/octobercms/october/blob/develop/modules/backend/classes/Controller.php#L193-L218

LukeTowers:
Ah, I see. This would be a limited case, I think we should move https://github.com/octobercms/october/blob/develop/modules/backend/classes/Controller.php#L186-L191 below https://github.com/octobercms/october/blob/develop/modules/backend/classes/Controller.php#L193-L218

LukeTowers:
Fixed by https://github.com/octobercms/october/commit/c7a3354dfd5fcad9a435dae0f15db4ace80de4ef. Thanks for the work on this @tim0991 @munxar!

PR #3217, opened 2017-11-01
Bind auth check into page.beforeDisplay event 

tim0991:
in \Backend\Controllers\Users::__construct (#3215)

### DIFF
--- a/modules/backend/controllers/Users.php
+++ b/modules/backend/controllers/Users.php
@@ -50,12 +50,14 @@ class Users extends Controller
     public function __construct()
     {
         $this->user = BackendAuth::getUser();
-        if (!$this->user->isSuperUser()) {
-            // Prevent non-superusers from even seeing the is_superuser filter
-            $this->listConfig = $this->makeConfig($this->listConfig);
-            $this->listConfig->filter = $this->makeConfig($this->listConfig->filter);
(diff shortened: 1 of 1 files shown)


LukeTowers:
Does this actually work? The __construct method runs before page.beforeDisplay and this code needs to be run before the list behavior is initialized.

tim0991:
Yes, it works in my project
 After sign out, i view backend/backend/users/update/1 and backend/backend/userroles/update/1, they are get the same errors. 

If you don't do that,  in `Backend\Controllers\Users::__construct` will call `BackendAuth::getUser()` before sign, but  `BackendAuth::getUser()` return null, so it will be occur an error 

>Call to a member function isSuperUser() on null in modules/backend/controllers/users.php line 53

munxar:
@tim0991 
I can see multiple things that go wrong:
1. call parent::_construct BEFORE you do anything else, because it already has the BackendAuth::getUser() implemented.
2. I don't think a constructor should execute this kind of complex logic, imho it should only initialize variables.
just listen to the 'backend.page.beforeDisplay' event, and implement you logic there.

LukeTowers:
@tim0991 but does this code actually run and still accomplish it's original goal? i.e. removing the is_superuser list filter for backend users that aren't superusers?

tim0991:
Ok, maybe we need another way to fix this

PR #3220, opened 2017-11-03
Optimize theme recognition

vojtasvoboda:
**Problem**: When we have only one theme, CMS ask backend for the active theme, which makes one SQL query.

**Goal**: Without that query, we could have a webpage without any SQL queries :-)

**Cases**:

- a user has one theme which is same as the theme in `cms.activeTheme` config - no need to check DB
- a user has one theme, but it is not the same as `cms.activeTheme` config (maybe downloaded another theme and deleted the previous one) - ok, let's check DB
- a user has more themes - ok, let's check DB

**Before**:

snimek obrazovky 2017-11-03 v 12 45 51

**After**:

snimek obrazovky 2017-11-03 v 12 46 03

### DIFF
--- a/modules/cms/classes/Theme.php
+++ b/modules/cms/classes/Theme.php
@@ -148,6 +148,7 @@ class Theme
     public static function getActiveThemeCode()
     {
         $activeTheme = Config::get('cms.activeTheme');
+        $themes = static::all();
         $havingMoreThemes = count($themes) > 1;
         $themeHasChanged = !empty($themes[0]) && $themes[0]->dirName !== $activeTheme;
         $checkDatabase = $havingMoreThemes || $themeHasChanged;
--- a/modules/cms/classes/Theme.php
+++ b/modules/cms/classes/Theme.php
@@ -148,8 +148,11 @@ class Theme
     public static function getActiveThemeCode()
     {
         $activeTheme = Config::get('cms.activeTheme');
+        $havingMoreThemes = count($themes) > 1;
+        $themeHasChanged = !empty($themes[0]) && $themes[0]->dirName !== $activeTheme;
+        $checkDatabase = $havingMoreThemes || $themeHasChanged;
 
-        if (App::hasDatabase()) {
+        if ($checkDatabase && App::hasDatabase()) {
             try {
                 try {
                     $dbResult = Cache::remember(self::ACTIVE_KEY, 1440, function () {


LukeTowers:
@vojtasvoboda `$themes` is undefined. Please fix that

vojtasvoboda:
@LukeTowers wow, it was a really stupid mistake, sorry

Samuell1:
This looks like cool change but i think iterating on every request in folders can be more slow then one query and i think it will be good to put this function fully in Cache to prevent calls on every request. @vojtasvoboda

LukeTowers:
@Samuell1 can you do some benchmarking to check the performance of this? Normally filesystem calls are pretty cheap, they can work out to be cheaper than an external cache service (such as redis) in some cases.

Samuell1:
I tested it with

**4 themes:**

obrázok

**only 1 theme:**

obrázok


**Actual code without change:**

obrázok

LukeTowers:
@Samuell1 those are kind of hard to read at a glance, can you summarize with just the response time for each method as text or in a table?

Samuell1:
@LukeTowers updated my comment with table

vojtasvoboda:
Each test should be passed thousand times and then calculate average time, because as you can see, first test with 4 themes and second test are completely different :-/

I like the "1 theme" test with updated code :-)

Samuell1:
@vojtasvoboda Yeah, but still iam for last config option that will use only one theme that is configured in cms.activeTheme.

LukeTowers:
I'm not clear on what's going on here anymore, can we restate the problem and the proposed solutions and the performance impact of each solution (averaged over 1000 iterations)? I'm also not even sure the DB query is an issue as the results are cached on first use.

Samuell1:
@LukeTowers Mostly is about when there is only 1 theme in themes folder, there isnt need to access db or anything more just return `cms.activeTheme` to speed up process.

LukeTowers:
Just waiting on the performance numbers from @Samuell1 as discussed in Slack

Samuell1:
@LukeTowers i think this can get merged, i didnt do any performence checks buts should be better then every time calling DB.

LukeTowers:
@Samuell1 any change that is a change for performance sake must have actual performance numbers to back it up, that's just our policy. Just because something intuitively makes sense doesn't mean that's always how it plays out it in the real world, so we must always verify and know exactly what the performance benefit is 😄

Samuell1:
@LukeTowers Mostly benefit is about that call to db.

modules/cms/classes/Theme.php
```diff
@@ -148,8 +148,12 @@ public function isActiveTheme()
     public static function getActiveThemeCode()
     {
         $activeTheme = Config::get('cms.activeTheme');
+        $themes = static::all();
```
bennothommo: @vojtasvoboda I believe the `$themeHasChanged` check is redundant. If we have multiple available themes, we have no choice but to confirm the active theme with the database.

Could you change this line to the following and remove line 152-154:
```suggestion
        $hasMultipleThemes = count(static::all()) > 1;
```

modules/cms/classes/Theme.php
```diff
@@ -148,8 +148,12 @@ public function isActiveTheme()
...
+        $havingMoreThemes = count($themes) > 1;
+        $themeHasChanged = !empty($themes[0]) && $themes[0]->dirName !== $activeTheme;
+        $checkDatabase = $havingMoreThemes || $themeHasChanged;
 
-        if (App::hasDatabase()) {
+        if ($checkDatabase && App::hasDatabase()) {
```
bennothommo: @vojtasvoboda Following the above change, we just need to check `$hasMultipleThemes` at this point:
```suggestion
        if ($hasMultipleThemes && App::hasDatabase()) {
```

modules/cms/classes/Theme.php
bennothommo: @vojtasvoboda Please ignore this for now. I'm going to make a different suggestion in the PR.

modules/cms/classes/Theme.php
bennothommo: @vojtasvoboda Please ignore this for now. I'm going to make a different suggestion in the PR.

bennothommo:
@vojtasvoboda @Samuell1 @LukeTowers Just as a suggestion, is there any reason why we don't just ignore the DB altogether for the theme code, and just use `config/cms.php` as the single source of truth, overwriting the value directly in the config if the person chooses a different theme via the Backend (also taking into account which environment the site is running or if the site is using a `.env` file). There's only one active theme at a time, so I can't imagine it would be a complex change to make.

LukeTowers:
@bennothommo right now that's because that would break existing sites.

bennothommo:
@LukeTowers In theory, we could create a migration to move it out of the DB perhaps. I'm just trying to work out the necessity of the active theme being in the DB.

LukeTowers:
@bennothommo true, although I'm not sure how we'd handle it if someone has set an env() call on that config value. I'm not sure why it's stored in the DB, is there anything in the history?

bennothommo:
@LukeTowers yeah that's the only scenario that I have yet to think of a graceful solution for. Given that the DB "overwrites" the config when using the theme selector, perhaps this change should too, by overriding any calls for environment values and just setting the config to the selected theme as a string.

As for when the change was included, I think it was around the time @daftspunk and @alexeybobkov implemented theme selection - this was the commit I found where it was first introduced. https://github.com/octobercms/october/commit/c324c809693396446930ee3e2f010bea8b1b31d5

LukeTowers:
@bennothommo another problem with a migration would be deployments that don't persist the configuration files. Most of my October sites are setup with CD that replaces all of the files every deployment, so in that case the migration would run, the config file would be updated, and then the config file would be overwritten on the next deployment breaking everything.

Samuell1:
I think best will be to make another config value that forces to use only config theme name like we was talking @LukeTowers and it will not make any breaking changes.

SebastiaanKloos:
Currently if a database connection could not be established, the site shows an error. It would be really nice to use the config file so the website still works

SebastiaanKloos:
> @SebastiaanKloos what error gets displayed? October should work fine without a database configured, the frontend anyways.

There is a connection configurated. But when a DB connection is unavailable, it shows an error with the query where OctoberCMS tries to get the active theme

bennothommo:
@LukeTowers I have tested this before with my original review (the one I cancelled before making my suggestion). It works fine - there's not a real big speed improvement, but it does shave off an SQL query if there's only one theme in a website. I believe it's fine to merge as is now, and we can come back to my suggestion at a later time.

Samuell1:
@bennothommo I think there should be added config value that allows to skip that check and automatically passes value from config if developer sets it with default to false.

bennothommo:
@Samuell1 TBH, I don't really think it's worth the effort. It is a micro-optimization - we're talking about milliseconds difference between making a single query to the DB or not. Once it's queried, it's cached. In the normal run of things, this has zero effect on the load time.

PR #3223, opened 2017-11-08
fix compile assets path error

tim0991:


modules/backend/assets/js/october.js
```diff
@@ -15,7 +15,7 @@
 =require ../vendor/sweet-alert/sweet-alert.js
 =require ../vendor/jcrop/js/jquery.Jcrop.js
 =require ../../../system/assets/vendor/prettify/prettify.js
-=require ../../../cms/widgets/mediamanager/assets/js/mediamanager-global.js
+=require ../../backend/widgets/mediamanager/assets/js/mediamanager-global.js
```
LukeTowers: Good catch! However, wouldn't this only require `../widgets`?, not `../../backend/widgets`? @tim0991

LukeTowers: Because it's already in the backend?

tim0991: It's my fault,  i will push a new pr 😂

LukeTowers: @tim0991 no need to push a new PR, you could have just added a commit to the branch that you made these changes on

Issue #2869, opened 2017-05-16
Flash message will be shown twice


kealsera:
##### Expected behavior
Flashmessage shows each message once.

##### Actual behavior
Userinteraction will lead to a Flashmessage. After Pagerefresh this message is shown once. If the user refreshes the page again (or visits another one) the same message will be shown once more. It will not show up a third time.

Here's a video showing the bug in action using the rainlab user login:
https://youtu.be/nk8jGKkGNiE

##### Reproduce steps
I have multiple occasions where this happens, even with the UserLogin Plugin.
Thise code will show 2 messages,regardless what happens

```
public function onBannerSave()
    {
        if(Input::file('banner') != null)
        {
            try
            {
                //[..not relevant..]
                Flash::success('Banner updated successfully!');
            }
            catch(Exception $e)
            {
                Flash::error($e->getMessage());
            }
            finally
            {
                return Redirect::refresh();
            }
        }
        else
        {
            Flash::warning('No Banner provided!');
        }
    }
```

This code will show only one message 

```
public function onBannerSave()
    {

            try
            {
                //[..not relevant..]
                Flash::success('Banner updated successfully!');
            }
            catch(Exception $e)
            {
                Flash::error($e->getMessage());
            }
            finally
            {
                return Redirect::refresh();
            }
    }
```

This is my errorhtml:

```
<div class="fixed-top container" style="z-index:9999999;top:5%">
    <div class="row">

        {% flash %}
        <div class="errorMSG col-12 alert alert-{%if type == 'error'%}danger wow shake{%else%}{{type}}{%endif%}">
            <button type="button" class="close" data-dismiss="alert">x</button>
            <strong>{{type|capitalize}}</strong> {{ message }}
        </div>
        {% endflash %}
    </div>

</div>


  jQuery(document).ready(function () {
        if (jQuery(".errorMSG") == null)
            return;
        jQuery(".errorMSG").alert();
        jQuery(".errorMSG").fadeTo(3000, 500).slideUp(500, function () {
            jQuery(".errorMSG").slideUp(500);
        });

    });
```

##### October build
419

daftspunk:
Perhaps you could prepare a simple theme that demonstrates the issue.

SebastiaanKloos:
I had this problem too. I had the framework js two times in my theme. Perhaps you have the same.

kealsera:
@SebastiaanKloos you mean the framework tag?

That's unfortunetly not the case:
[...]

PR #3229, opened 2017-11-09
Purge flash messages after using {% flash %} tag

jcardell:
Fixes the problem that causes flash messages to display twice.

I'm purging the session data in the destructor, since the function Flash::getMessages()
 is not doing it by itself. That caused the flash messages to display again after a page refresh.

If this problem can be resolved in another way, let me know and i'll make the necessary changes.

Fixes issue #2869

### DIFF
--- a/modules/cms/twig/FlashNode.php
+++ b/modules/cms/twig/FlashNode.php
@@ -1,5 +1,6 @@
 <?php namespace Cms\Twig;
 
+use Flash;
 use Twig_Node;
 use Twig_Compiler;
 use Twig_Node_Expression;
@@ -15,6 +16,11 @@ class FlashNode extends Twig_Node
     public function __construct($name, Twig_Node $body, $lineno, $tag = 'flash')
     {
         parent::__construct(['body' => $body], ['name' => $name], $lineno, $tag);
+    }
+    
+    public function __destruct()
+    {
+        Flash::purge();
     }
 
     /**


frank-beta:
should the purge happen in Twig? Is it not task of the framework to purge flash after sending response to the request? what if i dont use {% flash %} to handle my flash messages? I diged into the code but couldn't found yet where purge happen, because it happen after showing the flash twice

attached sample theme not using {% flash %} that still shows flasht twice, even with this fix

[foo.zip](https://github.com/octobercms/october/files/1463193/foo.zip)

LukeTowers:
@frank-beta @jcardell It should be the responsibility of the framework to handle the forgetting as well, but perhaps this will end up being the best solution, I'm not sure if the framework exposes a getMessagesAndForget() method of some sort to provide the same functionality at a PHP code level rather than just a Twig level.

jcardell:
@LukeTowers Exactly, right now the framework doesn't expose any getMessagesAndForget method, that's why I made the changes at Twig level, buy if you think it's worth adding the getMessagesAndForget in the framework (October\Rain\Flash\FlashBag), let me know and I'll make the changes.

frank-beta:
I don't know if it is a good approach, adding Flash::purge() to boot method in Cms\ServiceProvider works, I don't know how but it works!

[ServiceProvider.php](https://github.com/octobercms/october/blob/master/modules/cms/ServiceProvider.php)

[ServiceProvider.zip](https://github.com/octobercms/october/files/1476989/ServiceProvider.zip)

Rike-cz:
Not work for me in this case: https://octobercms.com/docs/ajax/update-partials#pulling-updates

LukeTowers:

Rike-cz:
For me works good only `$compiler->write('Flash::purge();'.PHP_EOL)` added to `Cms\Twig\FlashNode::compile()`. Solution may be something like `{% flash purge %}`, where _purge_ is optional, but I do not know how to reach it.

PR #3232, opened 2017-11-10
fix issue #3064 froala selection text works

tim0991:
i checked https://github.com/octobercms/october/issues/3064 and https://github.com/octobercms/october/pull/3086, and use one half day to fix this.
i think i figure it out

kapture 2017-11-10 at 15 32 36

after i click pageLink search btn, i found there is a selection in editor
 so i think editor save the selection before and it restore now( but i don't know why it restore after click and where it be saved),while I click any place, the selection lost focus 

i realized that it must be saved after popup show, before any operations

kapture 2017-11-10 at 15 54 25

after click  insertLinkButton, i found the selection is disappeared, and i click insertButton the selection appear there with a new link, may be froala call `editor.selection.restore()` in insertButton ?

### DIFF
--- a/modules/backend/formwidgets/richeditor/assets/js/plugins/pagelinks.js
+++ b/modules/backend/formwidgets/richeditor/assets/js/plugins/pagelinks.js
@@ -57,6 +57,17 @@ $.FroalaEditor.DEFAULTS.key = 'HHMDUGENKACTMXQL==';
         }
 
         function insertLink() {
+            editor.selection.restore();
+            range = editor.selection.ranges(0);
+            editor.events.on('link.beforeInsert', function(){
+                if (range) {
+                    var sel = editor.selection.get();
+                    editor.selection.clear();
+                    sel.addRange(range);
+                    range = null;
+                }
+            });
+            
             richeditorPageLinksPlugin = this
 
             editor.$el.popup({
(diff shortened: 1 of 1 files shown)


jan-vince:
Hi @tim0991 great job! 

*Does this fix works also for images picked from Media manager? They are placed on top of editor and there is probably the same cause of it.*

tim0991:
@jan-vince NO!  

this fix don't make image picker works

tim0991:
I try to use a new method to solve this problem
i get `range obejct` before popup show, the `range object` can restore the area of user selected before opening the `insert link popup`
image
 it looks ok in text.

but when i click `choose link`
image
 the image will lose focus, and i can't get `range object` in `insertLink method`. 
because the cursor focus in the url lable
image


 i have no idea about this
how can i get `selected img dom` before i open the named `insert link` button in image popup
image

any way i hope this fix can give you some diffrent ideas

LukeTowers:
What's the main difference between this and #3086?

tim0991:
First I did not find a event named `shown.oc.popup.pageLinks`, and my focus is on how to save and restore selection

I think we need to solve the problem of selection loss in two ways, one is how to save and restore selection, the other is why the editor's focus is shifted to popup

LukeTowers:
@tim0991 the semicolons need to be removed

LukeTowers:

Issue #2753, opened 2017-03-14
Framework and Framework Extra files aren't minimized


lrs:
##### Expected behavior
/modules/system/assets/css/framework.extras.css
/modules/system/assets/js/framework.js
/modules/system/assets/js/framework/extras.js

Should be minified/uglified to help with loading speeds.

I can do this myself of course however there's a chance it would be overwritten with a build update.

##### Actual behavior
Above files aren't minifified/uglified

##### Reproduce steps
Install latest build.

##### October build
396

SebastiaanKloos:
Isn't there a minified parameter in the config files?

LukeTowers:
It would depend how you loaded it. If you used the asset combiner:
```twig
<script src="{{ ['@framework', '@framework.extras'] | theme }}"></script>
```
and then set your asset minification config settings, then yes, it should minify it for you. If you're referencing it directly then it won't.

lrs:
Thanks @LukeTowers, @SebastiaanKloos I was using what I thought was the standard way:
 ```
{% framework extras %}
```

Using this and setting `'enableAssetMinify' => true` in config/cms.php combines and minifies!

Closing ticket.

lrs:
Oops, now the AJAX framework is broken for me. Reverting back to `{% framework extras %}` resolves it.

daftspunk:
We could potentially minify it using `october:util compile assets` and toggle between the two based on `cms.enableAssetMinify` for the framework extras tag. That sounds like a great idea.

SebastiaanKloos:
+1 @daftspunk

lrs:
Thanks for keeping this issue open. Have you had any thoughts on how to progress with this?

maartenmachiels:
+1 would be nice!

You85:
+1

maartenmachiels:
Beste,

We zijn op vakantie tot 11 september.
Alvast bedankt voor het begrip en tot dan.

Mvg,

Maarten Machiels

LukeTowers:
The support demand for October is increasing, while the available time the founders have to give remains the same. These issues will be handled in time, however there are ways you can help:

- Submit a pull request that introduces this feature and wait for it to be reviewed
- [Make a financial contribution to the project](http://octobercms.com/fundraising) so we can allocate more time towards making improvements like these that benefit everyone

LukeTowers:
[...]

PR #3240, opened 2017-11-14
Select from framework asset options for #2753

jimcottrell:
Thought I'd take a crack at this based on @daftspunk's comment and at least provide a starting point.

I did come across a minor issue in the asset combiner where array_except is incorrectly used to remove an array value (array_except operates on keys).


LukeTowers:
Related #2753

LukeTowers:
@Jamesking56 did you check the Travis logs? It's because `forgetInstance()` didn't exist in the build of October that travis was utilizing but the tests were from the build where it does exist.

petehalverson:
@LukeTowers @daftspunk this one seems like low hanging fruit.

Let me know what we can do to help get this merged in.

LukeTowers:
@petehalverson could you get @jimcottrell to rebuild the assets with the latest dev branch and push that as a commit?

modules/cms/twig/FrameworkNode.php
```diff
@@ -28,17 +29,34 @@ public function compile(Twig_Compiler $compiler)
...
             ->addDebugInfo($this)
-            ->write("echo '<script src=\"'. Request::getBasePath()
-                .'/modules/system/assets/js/framework.js\"></script>'.PHP_EOL;" . PHP_EOL)
-        ;
+            ->write("\$_minify = ".CombineAssets::class."::instance()->useMinify;" . PHP_EOL);
 
         if ($includeExtras) {
             $compiler
-                ->write("echo '<script src=\"'. Request::getBasePath()
+                ->write("if (\$_minify) {" . PHP_EOL)
```
LukeTowers: Why can't the ternary approach be used here too?

modules/cms/twig/FrameworkNode.php
```diff
@@ -28,17 +29,34 @@ public function compile(Twig_Compiler $compiler)
 
         $compiler
             ->addDebugInfo($this)
-            ->write("echo '<script src=\"'. Request::getBasePath()
-                .'/modules/system/assets/js/framework.js\"></script>'.PHP_EOL;" . PHP_EOL)
-        ;
+            ->write("\$_minify = ".CombineAssets::class."::instance()->useMinify;" . PHP_EOL);
```
LukeTowers: Doesn't this break the usage of just `{% framework %}`?

LukeTowers: Never mind, saw the extra else at the end

modules/cms/twig/FrameworkNode.php
jimcottrell: It's too much for a ternary because the way I set it up, if minifying is enabled, then it loads a combined version of the minified framework + extras JS. Otherwise it loads separate files.

LukeTowers: Okay, thanks for the response! Can you confirm that it works well with both `{% framework %}` and `{% framework extras %}`?

petehalverson: Sure. I'll post some examples.

jimcottrell: I could check again if you'd like, but I do remember verifying both tags with minifying and without.

petehalverson: [twig filter example](https://f20bb1310757.octodock.com/)

petehalverson:
Thanks @jimcottrell and @LukeTowers!

PR #3246, opened 2017-11-17
[Backend] In preview mode Date with a "Null" value return the current date

pvullioud:
**Actual behavior**
Date field with null value return the current date.

**After the PR**
Date field with null value is empty

### DIFF
--- a/modules/backend/helpers/Backend.php
+++ b/modules/backend/helpers/Backend.php
@@ -120,7 +120,7 @@ class Backend
         ], $options));
         
         if(!$dateTime) {
-        return;
+            return '';
         }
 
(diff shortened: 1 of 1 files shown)


LukeTowers:
This could be a backwards-incompatible change. Does this only happen in a preview context or does it happen in update contexts as well? Create contexts?

pvullioud:
Hi @LukeTowers 
It's only happens in preview mode.
"Backend::dateTime" usage : https://github.com/octobercms/october/search?utf8=%E2%9C%93&q=%22Backend%3A%3AdateTime%22&type=

pvullioud:
Frendly ping on this. I have to re apply my patch at every update.

PR #3252, opened 2017-11-20
Add /plugins/october/demo/vendor to .gitignore

fernandoris:
To avoid conflicts when working with the repository as a submodule


LukeTowers:
@fernandoris not only does this not do what you say it does, this was made to the wrong branch.

Besides those reasons, this will not be accepted as 
A. plugins/october/demo/vendor does not exist
B. This doesn't tell us what the conflict is and why we should support it not existing.

PR #3255, opened 2017-11-21
Fix ajax validation for array syntax

bedrop-de:
allow array validation syntax if you have some fields like field.0.name etc. 

In my Project i use jQuery.repeater plugin to clone some form fields. The fields has structure like 
field[0][name], field[1][name]. After sending response to the server with ajax the validation Messages looks like field.0.name , filed.1.name etc. The validation messages are not displayed because framework.js look only for array syntax.
Model rules looks like this: 
'cats.*.content_cat_name' =>  'sometimes|required|max:255',


bedrop-de:
Hi Luke,

i fixed the previous PR. 

https://regex101.com/r/hnuDzy/3

Issue #3259, opened 2017-11-22
Outdated validation translation messages


ametad:
##### Expected behavior
Validation message in human language.

##### Actual behavior
Some error messages are not available and therefore the identifier key is shown. Per example:
Validor: file
Upon an invalid file (upload fail) the message shown: "validation.uploaded"

##### Reproduce steps
Set up an controller that expects form post that is validated with:

```
        $validator = Validator::make([
            'cv' => request('cv'),
        ], [
            'cv' => 'file',
        ]);

        if ($validator->fails()) {
            throw new ValidationException($validator);
        }
```
Then upload a file that is greater then set as 'upload_max_file' in php ini config.

The validation fails and is returning a translated message "validation.uploaded". 

The problem is that an outdated translation file of validations is used in OC:
october/modules/system/lang/en/validation.php

Laravel 5.5 has a newer version:
https://github.com/laravel/laravel/blob/593963979a6b6a27966aef9723b8ff08fb480b87/resources/lang/en/validation.php#L88

##### October build
428

LukeTowers:
Feel free to make a Pull Request adding the required information.

ametad:
Merged!

PR #3261, opened 2017-11-23
Updating validation messages

ametad:
Updating validation messages to the latest Laravel 5.5 version.

Related issue: #3259


LukeTowers:

modules/system/lang/nl/validation.php
```diff
@@ -90,10 +112,42 @@
     |
     | The following language lines are used to swap attribute place-holders
     | with something more reader friendly such as E-Mail Address instead
-    | of 'email'. This simply helps us make messages a little cleaner.
+    | of "email". This simply helps us make messages a little cleaner.
     |
     */
 
-    'attributes' => [],
+    'attributes'           => [
```
LukeTowers: I'm not sure we want these included in here, they should be up to the plugin authors to add. Ideally we'd like to implement a validation improvement that displays translated field labels instead of the raw attribute names in the FormController. Thoughts?

ametad:
Next time to develop branch, I got it! I am used to do that, but this time it must have slipped my mind...

ametad:
I will look at your changes request later, out of time right now.

modules/system/lang/nl/validation.php
ametad: In short: it is work already done that can be seen as fall back measures. It can and should be implemented in the Plugin's itself or in the app itself. But seems a waste to throw it away.
On the other hand, it could lead to wrong assumptions about how to code for OC...
What is the best thing to do?

ametad:
Hi Luke, have you seen my response to your requested changes? If you want me to just remove the attribute names, that will be alright with me. But I am wondering about best practice in this case too, that's why I answered upon your call to let hear some thoughts :-)

LukeTowers:
@ametad I would prefer to remove the attribute names, October needs to be kept simple and while that is potentially helpful I feel like it's more clutter than necessary for what it would achieve. Thoughts?

modules/system/lang/nl/validation.php
ametad: I just removed the extras to keep the core validation messages simple.

PR #3262, opened 2017-11-23
Trailing & in requestData if data is empty

SeriousKen:
This fixes a small issue where the form is submitted an extra unnamed null field is present in the request due to a trailing & symbol in `requestData` when `data` is empty.


LukeTowers:
Awesome, thanks @SeriousKen!

PR #3273, opened 2017-11-29
Fix #3109

### LINKED ISSUE #3109: Component onRender
Hi guys!

I'm calling the component into the template, and want that component's partial 'bar' will be displayed with the property 'prop'. 
{% component 'Foo::bar' prop='queue' %}
```

class Foo extends ComponentBase {
    public function onRender() {
        $this->page['propProperty'] = $this->property('prop');
    }
}
```

_../components/foo/bar.htm_

`Property value is {{propProperty}}`

##### Expected behavior
Output: 
Property value is queue

##### Actual behavior
Property value is 

When I'm calling {% component 'Foo' prop='queue' %} and use default.html partial, everything is ok.

https://github.com/octobercms/october/blob/master/modules/cms/classes/Controller.php#L992-L1010

The bud into Cms\Classes\Controller::renderComponent, the $name variable is 'Foo::bar'.
I can create PR with fix it. 
But I'm not sure, fix it into `renderComponent` method or `findComponentByName`

Samorai:
PR for fix #3109

### DIFF
--- a/modules/cms/classes/Controller.php
+++ b/modules/cms/classes/Controller.php
@@ -1261,6 +1261,16 @@ class Controller
             return $partialComponent;
         }
 
+        if (strpos($name, '::')) {
+            list($name) = explode('::', $name);
+            if (isset($this->page->components[$name])) {
+                return $this->page->components[$name];
+            }
+            if (isset($this->layout->components[$name])) {
+                return $this->layout->components[$name];
+            }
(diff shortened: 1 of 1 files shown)


LukeTowers:
Closing in favour of https://github.com/octobercms/october/issues/3109#issuecomment-348620591

Issue #3275, opened 2017-11-29
MediaManager thumbnail generator stores images only locally 


TimFoerster:
##### Expected behavior
We have a remote MediaManager storage configured.

When uploading an image over the MediaManager and click on it, a thumbnail will be generated.
The thumbnail should be stored in the `/storage/app/media/public` folder on the remote storage. 

##### Actual behavior
The thumbnail is stored on local disk under `/storage/temp/public/[..]/[..]/...`

##### Reproduce steps
1. Setup your remote storage. 
2. Set configuration `conf/cms.app`
```
        'media' => [
            'disk'   => 'my_s3',
            'folder' => 'media',
            'path'   => '/storage/app/media',
            'thumbFolder' => 'public'
        ],
```
3. Upload image and open it -> thumb will be generated

##### October build
428+

TimFoerster:
During reading the configuration file I assume that the thumbnails are stored on the remote storage under `/public` folder. 
So first of all the configuration structure should be changed to something that makes clear, where the thumbs are stored.
```
    /*
    |--------------------------------------------------------------------------
    | Resource storage
    |--------------------------------------------------------------------------
    | [...]
   */
    'storage' => [
        'uploads' => [
            'disk'   => 'local',
            'folder' => 'uploads',
            'path'   => '/storage/app/uploads',
        ],
        'media' => [
            'disk'   => 's3',
            'folder' => 'media',
            'path'   => '/storage/app/media',
            'thumbs' => [
               'disk'   => 'local',
               'folder' => '/storage/temp/public',
             ]
        ],
    ],
```

if you want to store the thumbs somewhere else then change `thumbs` to

```
            'thumbs' => [
               'disk'   => 's3',
               'folder' => '/myThumbs',
             ]
``` 
so all generated thumbs should be stored on the s3 storage under `/myThumbs'

TimFoerster:
another thing that comes into my mind is, why is there always a `folder` and a `path` attribut. What are the differences between them? Has `folder` any effect on the system?

LukeTowers:
@TimFoerster I'd accept a PR to add support for your suggested configuration of a thumbs config array in the media config array.

TimFoerster:
@LukeTowers nope, i just picked up from a slack discussion.

Samuell1:
Hey, can this be reopen? This issue is with everything that uses *temp* folder not only MediaManager for example Export uses temp folder too.

w20k:
Done @Samuell1

[...]

PR #3276, opened 2017-11-30
Store preview thumbnail to configured storage

Floriskoch:
This stores the thumbnail image in the Media Manager to the configured storage disk (local, S3, etc..). This is useful when working with a multiple server setup as locally stored files on one server won't be accessible on the other.

Open issues:
- The thumbnails will be visible in the media manager. Would be nice to store them in an invisible location.
- ???

Feedback welcome!

### DIFF
--- a/modules/backend/widgets/MediaManager.php
+++ b/modules/backend/widgets/MediaManager.php
@@ -962,9 +962,14 @@ class MediaManager extends WidgetBase
             }
 
             /*
-             * Resize the thumbnail and save to the thumbnails directory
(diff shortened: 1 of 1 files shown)


TimFoerster:
did you also checked https://github.com/octobercms/october/issues/3275 ?

Floriskoch:
@TimFoerster No I did not. Looks like a duplicate then? If so, we can close this PR.

TimFoerster:
@Floriskoch well its just a reported issue + discussion.

PR #3277, opened 2017-11-30
support min items on repeater

panakour:
This feature prevent clients to remove by mistake repeated item which consequently breaks a layout and the design of a website.

### DIFF
--- a/modules/backend/formwidgets/Repeater.php
+++ b/modules/backend/formwidgets/Repeater.php
@@ -174,6 +174,10 @@ class Repeater extends FormWidgetBase
      */
     protected function processSaveValue($value)
     {
+        if ($this->indexCount < $this->minItems) {
+            throw new ApplicationException(Lang::get('backend::lang.repeater.min_items_error', ['number' => $this->minItems]));
+        }
+
         if (!is_array($value) || !$value) {
             return $value;
         }


modules/backend/formwidgets/Repeater.php
```diff
@@ -366,4 +374,13 @@ public function getGroupTitle($groupCode)
     {
         return array_get($this->groupDefinitions, $groupCode.'.name');
     }
+
+    protected function processForMinItems()
+    {
+        if($this->minItems === null) return;
+
+        if ($this->indexCount <= $this->minItems) {
```
LukeTowers: Why is this a `<=`? instead of just a `<` sign?

modules/backend/lang/en/lang.php
```diff
@@ -585,4 +585,7 @@
         'image_size' => 'Image size:',
         'selected_size' => 'Selected:'
     ],
+    'repeater' => [
+        'min_items_error' => 'You cannot delete any item. You should modify the existing.'
```
LukeTowers: The error should be `At least :number items are required`

LukeTowers:
As another note, your layouts and design of the website should be built to be robust enough for them to have no items, because there is no way to force the user to have any data for your repeater at all in the first place.

Even if you want to go down this route, it would probably be easier and better if your enforced the item requirements in your model when you are saving the data instead of directly on the Repeater form widget. This PR will only be accepted if it can be proven to be a robust and simple change to existing logic that will be easy to maintain going forwards.

modules/backend/formwidgets/Repeater.php
panakour: The minimum requirement items will be created from the developer. 

`<=?` instead of just a `<` becasue indexCount begins from 1.

panakour:
@LukeTowers if you working with `rainlab.pages-plugin` you don't have any model to work with on custom fields..

LukeTowers:
@panakour my above comment still needs to be addressed: 
>This only affects users that want to remove existing items, it doesn't do anything to enforce that they create the repeater with at least $minItems amount of items.

panakour:
@LukeTowers what about this https://github.com/octobercms/october/pull/3277/commits/368457bb43526f74cf7058bc129fc3d9108acb82

LukeTowers:
@panakour I still don't know if that is robust enough, I'll have to review it further, specifically  the client side experience. Ping me on slack if I forget about it

LukeTowers:
Implemented by https://github.com/octobercms/october/commit/cbed54a2634f5daae08a76d62d42945185534664#diff-a8e657fa9654328055121a4e2cc4311b

PR #3280, opened 2017-11-30
Trigger queue worker restart on settings model save

jimcottrell:
Most October installations likely don’t use queues, so it might only matter to a minority of users, but I think it makes sense for settings changes to trigger queue worker restarts. That way daemon queue workers automatically pick up any new settings.

It’s a simple change that I can easily implement for my own projects, but it seems like something that might belong in October core so that it just works out of the box if, say, mail is being queued for later sending and the mail settings have been changed in the backend since the worker daemon started.

So, I’ve changed the SettingsModel behavior to call the `queue:restart` console command after saving. This works, but on the one hand I don’t like that saving a model triggers a separate app boot to run a command just to update a cached timestamp. On the other hand, it’s not really _that_ bad, and settings models are generally not going to be updated in bulk, just one at a time from the settings controller. I don’t see where Laravel exposes that functionality anywhere else, so it’s either that or update the cache directly.

Thoughts?

### DIFF
--- a/modules/system/behaviors/SettingsModel.php
+++ b/modules/system/behaviors/SettingsModel.php
@@ -1,6 +1,7 @@
 <?php namespace System\Behaviors;
 
 use App;
+use Artisan;
 use Cache;
 use System\Classes\ModelBehavior;
 use ApplicationException;
@@ -196,12 +197,14 @@ class SettingsModel extends ModelBehavior
     }
 
     /**
-     * After the model is saved, clear the cached query entry.
+     * After the model is saved, clear the cached query entry
+     * and restart queue workers so they have the latest settings
      * @return void
      */
     public function afterModelSave()
     {
         Cache::forget($this->getCacheKey());
+        Artisan::call('queue:restart');
     }
 
     /**


LukeTowers:
I'm not opposed to it in theory but I think we'd need @daftspunk's input on this. Feel free to shoot him messages on Slack if he doesn't check out this issue relatively soon.

petehalverson:
@daftspunk?

daftspunk:
LGTM, so long as this isn't using `shell_exec` to run the command or anything like that. AFAIK `Artisan::call()` is processed in the same execution cycle.

From what I gather it looks to be simply setting a cache store `illuminate:queue:restart` to a fresh timestamp. Since the logic is embedded in the console command we really have no other choice but to call it this way.

jimcottrell:
Yeah, there's no `shell_exec` or anything. It is just a cached timestamp, which, as a note to anyone else who wants to use queues any may come across this, does mean that any workers need to be sharing a cache with the web server for this to have any effect.

daftspunk:
I hit this issue in another project and resolved it by continuously restarting the daemon worker. It's good to know. Is it documented at all?

jimcottrell:
You mean you don't think a note in a random pull request is good enough?

Yeah, I'll write up a section for the docs.

jan-vince:
Hi @jimcottrell (and @daftspunk),

after this change I get errors while saving settings that use SettingsModel.

The problem is that most hosting companies (at least in Czech Republic) has ````putenv()```` disabled for security reasons (on shared hostings).

*I have opened an issue #3339 about this.*

Symfony Console package use ````putenv()```` so everyone on hostings like this is affected.

*For me personally this is about 20+ websites with October version 431 :( All of them are quite small and my clients don't want to put money in more expensive hosting plans or own managed servers.*

Is it really necessary to call ````queue:restart```` with every SettingModel save?

And can we add a global settings flag (in ````/config/cms.php``` probably) to be able to control this so people affected by disabled ````putenv()```` can prevent ````Artisan::call()```` like:

````
    public function afterModelSave()
    {
        Cache::forget($this->getCacheKey());
        
        if(  Config::get('cms.settingsSaveRestartQueue') ) {
            Artisan::call('queue:restart');
        }

    }

````

jan-vince:
Hi @jimcottrell ,

I have suggested a workaround for those who are affected by disabled ````putenv()```` function with ````try...catch```` block (#3339).

But I am not able to solve a problem when ````Artisan::call```` is executed with every SettingsModel save - which is probably not very good.

I will send a PR with ````try...catch```` but if you have better idea I would like to hear about it.

PR #3281, opened 2017-12-01
Finnish language file updates

mediaclinic:
It would be great, if the translation files could include untranslated lines also, as it makes translation process much easier. I translate them, when I have some spare time, so a bit slow but better than nothing IMHO.


mediaclinic:
Actually, seems that Finnish already translated!! Which is great! [https://github.com/octobercms/october/pull/3237](url)

Samorai:
@mediaclinic I think something went wrong :) 
Please read [Contributing Guidelines](https://github.com/octobercms/october/blob/master/CONTRIBUTING.md)

LukeTowers:
@mediaclinic try resubmitting this.

PR #3302, opened 2017-12-15
prevent from duplicate the backend url

panakour:
I did this on master branch because I have already open a pull request on develop branch #3277

### DIFF
--- a/modules/cms/controllers/Index.php
+++ b/modules/cms/controllers/Index.php
@@ -163,6 +163,10 @@ class Index extends Controller
 
         $fields = ['markup', 'code', 'fileName', 'content'];
 
+        if(Config::get('cms.backendUri') === substr($templateData['settings']['url'], 1)) {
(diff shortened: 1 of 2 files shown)


LukeTowers:

LukeTowers:
What issue exactly is this PR solving?

panakour:
If you create a page from the CMS wit the same url that the backoffice use, the system let you create it without any problem. The problem about this is that the page that has the same URL as the backend does not work and open the backend instead

LukeTowers:
This isn't a robust enough solution to merge. This would only prevent assigning a page a URL that is exactly the Backend URL. A more robust solution is to get the full list of all registered routes and prevent assigning any of them, along with a more detailed error message letting people know why exactly they can't use a specific route.

panakour:
@LukeTowers ok, I'll try to improve this.

panakour:
@LukeTowers Now it's better approach for this problem. I have opened a new pull request for it.

PR #3306, opened 2017-12-18
Prevent duplicate urls of the pages

panakour:
This block the user to create the same url like the back office has. referenced #3302


LukeTowers:
This still isn't a good enough solution. The previous comment from #3302 still applies:

>This would only prevent assigning a page a URL that is exactly the Backend URL. A more robust solution is to get the full list of all registered routes and prevent assigning any of them, along with a more detailed error message letting people know why exactly they can't use a specific route.

LukeTowers:

PR #3308, opened 2017-12-19
add default ports on database config

panakour:


LukeTowers:

LukeTowers:
Don't submit duplicate PRs, make any required changes in your original PR instead in the future. (#3307).

Please explain why this change is necessary.

panakour:
1. With one of my environment when I had forgot to add the port in the database driver, the site does not work.
2. Since php7 introduce scalar types, some third party plugins throws type error because get string instead of integer
3. I think it's a good practice to have the default values in the config, even the [laravel configuration has the default ports](https://github.com/laravel/laravel/blob/master/config/database.php#L45).

LukeTowers:
Perfect, thanks for the explanation! One thing to consider with this is that Laravel's default configuration provides the default ports as strings, not integers. Is that something that we should match or is this approach of providing them as integers a better one?

panakour:
I have tested with both (string/integer from config) and they work. I prefer integer becasue it's actually a number not string.

PR #3324, opened 2017-12-27
fix form controller redirect with empty context option

tim0991:
When i delete a data in update page, it will not going to redirect, because  i did not specify delete context

so this commit is set `defualtRedirect` to defualt context  if there is no matched context

### DIFF
--- a/modules/backend/behaviors/FormController.php
+++ b/modules/backend/behaviors/FormController.php
@@ -489,7 +489,7 @@ class FormController extends ControllerBehavior
         // source for the default redirect being default[redirect]
         $redirects['default'] = $this->getConfig('defaultRedirect', '');
 
-        if (!isset($redirects[$context])) {
+        if (empty($redirects[$context])) {
             return $redirects['default'];
         }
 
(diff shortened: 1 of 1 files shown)


LukeTowers:
@tim0991 please test if this code works instead (changing the `!isset()` to `empty()` instead)
```php
// Get the redirect for the provided context
$redirects = [$context => $this->getConfig("{$redirectContext}[{$redirectSource}]", '')];

// Assign the default redirect afterwards to prevent the
// source for the default redirect being default[redirect]
$redirects['default'] = $this->getConfig('defaultRedirect', '');
  		  
if (empty($redirects[$context])) {
    return $redirects['default'];
}		
  		  
return $redirects[$context];
```

tim0991:
yes, you are right

LukeTowers:
@tim0991 Please test this code. You still haven't added `$redirects = [$context => $this->getConfig("{$redirectContext}[{$redirectSource}]", '')];` back in, the current PR just replaces that with `$redirects = [];`.

If you could also add some unit tests for this functionality that would be greatly appreciated.

tim0991:
I'm sorry for my carelessness, and i do not know how to write unit test until now.
But i will learn it, i hope the next time when i push PR  can provide the unit test

LukeTowers:
So are you planning on adding unit tests to this PR or are you talking about doing it in the future?

tim0991:
Maybe in the future

LukeTowers:
Okay, so is this PR done then? Ready to be merged? @tim0991

tim0991:
Yes, it is a small change

PR #3326, opened 2017-12-27
error in count() structure fix

viamage:
I want to rant whoever wrote this so hard, but I will keep calm and will just say: if you do not write tests, at least **do not use notepad.exe to write your code**! Any proper IDE would give you red colors for this. 

I understand that it was giving warn in PHP instead of exception, but it's very very stupid and should be merged to master ASAP. Congrats tho, you made our whole team laugh hard. 

Regards,
James

### DIFF
--- a/modules/cms/classes/CmsCompoundObject.php
+++ b/modules/cms/classes/CmsCompoundObject.php
@@ -301,7 +301,7 @@ class CmsCompoundObject extends CmsObject
         else {
             foreach ($this->settings['components'] as $name => $settings) {
                 $nameParts = explode(' ', $name);
-                if (count($nameParts > 1)) {
+                if (count($nameParts) > 1) {
                     $name = trim($nameParts[0]);
                 }
 
--- a/tests/unit/cms/classes/CmsCompoundObjectTest.php
+++ b/tests/unit/cms/classes/CmsCompoundObjectTest.php
@@ -29,9 +29,10 @@ class CmsCompoundObjectTest extends TestCase
     public function setUp()
     {
         parent::setUp();
-
         Model::clearBootedModels();
         Model::flushEventListeners();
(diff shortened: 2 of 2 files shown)


LukeTowers:
This has existed for a very long time: https://github.com/octobercms/october/commit/61cf1ad710a8ec5f3c526e45b9eb92c8bcccc25e#diff-e28bed2e202025ac6daf188a9c3e420bR335.

While I know by reading it that it doesn't make sense, could you please explain what exactly is breaking because of this code and how this change will fix it? As much detail as possible please.

viamage:
PHP 7.2.0 on ArchLinux, reproduced on fresh PHP 7.2.0 on Debian:

Using getComponentProperties() method on Page object (in my code I'm looking for page with component which has property set to proper value)

```
$properties = $page->getComponentProperties($componentName);
```

crashed on me with:

error

so basic reasons to accept this:

- looks like it doesn't trigger warn anymore but error.
- it's obvious syntax error and it should be fixed because it's stupid and there are better methods to make people laugh.
- it will probably break a lot of stuff when people move to higher PHP version like I did. 
- there is not a single reason not to accept this and I'm not sure why are you slowing this obvious fix down by demanding explanations. No, "This obvious stupid error is there for so long that we should keep it" is not a reason.

I hope that it's enough.

Best Regards and Happy New Year
James

LukeTowers:
@viamage calm down, I'm not your enemy.

If you could add tests for this method so that there won't be future breaks, then I'll accept this. The reason I say that is to encourage more tests being added to October. I don't disagree with you at all, I just wanted to consider all the information possible. In this specific case, this is an opportunity to expand October's test coverage.

viamage:
Extended the PR, hope these will be enough. In my opinion there are some architecture issues here but they are outside the scope of this issue. 

tests

LukeTowers:
Thanks @viamage! I appreciate your time and energy to fix this!

Issue #3321, opened 2017-12-22
install october with CLI issue


buuug7:
I installed october with ` composer create-project --prefer-dist october/october
 chicken-soup`, a exception have accured. failured both in  win10 and osx(10.13) with vagrant homestead.
#### win10(version 1709 build 16299.125)
+ virtualbox 5.1.26
+ Vagrant 2.0.0
+ homstead 5.0.1

#### osx(10.13.2)
+ virtualbox 5.2.2
+ Vagrant 2.0.1
+ homstead 5.0.1

#### win10
![image](https://user-images.githubusercontent.com/12119289/34288217-dcc61eba-e726-11e7-9125-d256aadf7f10.png)
#### osx
<img width="890" alt="35b6ef78-cce1-4fd8-9511-359ee758f599" src="https://user-images.githubusercontent.com/12119289/34288438-3d945f3a-e728-11e7-9770-a16e2fb1a2fb.png">

buuug7:
i know this exception is no matter for use, this exception is not accured before 431 installed with CLI.

LukeTowers:
@buuug7 could you check to see if it's due to this change: https://github.com/octobercms/october/commit/fbf482a3f44eb28076d6df5ee46df30befaa0789

Alainy:
#3288 did cause this. I ran `git clone https://github.com/octobercms/october.git;cd october;git checkout develop` then changed `'default' => 'development'` back to `'default' => 'production'` and then ran `composer create-project`. But now the same bug occurs as in #3287. It seems that that `php artisan key:generate` expects an input from the user to continue the command and the default value is **no**. Since the user can't give any input the command is cancelled.

buuug7:
a normal laravel project need generate the application key for quick start (eg. database is optional config), but the october application need more configurations (not just only generate the application key) unless you run `php artisan october:install` or other ways of installation.  

i think `php artisan key:generate` in the `composer.json` is superfluous ,so remove the `php artisan key:generate` from the `composer.json` is good  a solution. 

```

    "scripts": {
        "post-create-project-cmd": [
           // remove this line
            "php artisan key:generate"
        ],
        "post-update-cmd": [
            "php artisan october:util set build"
        ]
    },
```

LukeTowers:
@buuug7 @Alainy how about if we switch the default environment back to production but make the post-install command `php artisan key:generate --force` instead?

buuug7:
[...]

PR #3328, opened 2017-12-29
Update composer.json

buuug7:
this PR try to solve #3321 issue, since the environment switched to development by this commit https://github.com/octobercms/october/commit/fbf482a3f44eb28076d6df5ee46df30befaa0789 , a error will accured when install october wil CLI.    

it always run the php artisan key:generate command after composer post create project, but in production environment this process is skipped.   

it's time to remove the useless code `php artisan key:generate` from `composer.json` , as it never used.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -47,9 +47,6 @@
         ]
     },
     "scripts": {
-        "post-create-project-cmd": [
-            "php artisan key:generate"
-        ],
         "post-update-cmd": [
             "php artisan october:util set build"
         ]


LukeTowers:
I'd rather see a PR that creates October's own version of key:generate that could write to config/app.php if .env isn't present instead. Could you put that together?

LukeTowers:
Or perhaps a PR to Laravel to do the above idea. @buuug7 @Jamesking56

buuug7:
@LukeTowers @Jamesking56 , i think things should ketp simple. There is no need to have october own version of `key:generate`. october already have a console command `october:env` and do more than laravel's  `key:generate`.

LukeTowers:
Rethinking it, a better solution would be to include the following:
```json
"post-create-project-cmd": [
    "php artisan october:env",
    "php artisan key:generate"
],
```

buuug7:
@LukeTowers you are awesome

LukeTowers:
@buuug7 make that change in this PR and I'll merge it

buuug7:
@LukeTowers i have already delete this fork from my github account, i give a new PR for this. this should be closed.

daftspunk:
Should be fixed in Build 433+

PR #3329, opened 2017-12-29
Fix call to a member function when BackendUser is null

PopcornPHP:
Hi. If I am not an authorized user in the backend and try to access the page
/backend/backend/users/update
an error will appear:

Call to a member function isSuperUser() on null
/modules/backend/controllers/Users.php line 53


Elisseii:
+

LukeTowers:
@PopcornPHP @Elisseii this has been fixed by https://github.com/octobercms/october/commit/69e56b0b5b237aaaabfc25f88fc8265c7549f243

PR #3331, opened 2017-12-31
Fix 'illegal string offset' warning

alxy:
In case a new administrator is created by a non-superuser and no permission is set directly on creation, ``permissions`` is not initialized correctly. Thus, when the non-superuser tries to acceess ``$this->model->permissions`` it is not populated with an empty array as expected.

I have actually no clue why it does work for superusers however, as they should certainly experience the same issue, but this is not the case.


LukeTowers:
Because super users bypass that method altogether: https://github.com/octobercms/october/blob/master/modules/backend/formwidgets/PermissionEditor.php#L67

Issue #3332, opened 2018-01-01
`storage.uploads.disk` in config/cms.php is not effective.


pikanji:
##### Expected behavior
When `storage.uploads.disk` in config/cms.php is set to `s3` (assuming that `s3` is properly configured in config/filesystem.php), the uploaded file from FileUpload widget in backend is uploaded to S3.

##### Actual behavior
The uploaded file is stored in the disk specified by `default` in config/filesystem.php, no matter what is selected in `storage.uploads.disk` in config/cms.php.

##### Reproduce steps

1. Assuming there is a working model and corresponding database table exist, add an attachMany relation to the model as described [here](https://octobercms.com/docs/database/attachments#file-attachments).
2. Add a FileUpload widget to the models backend form for the attacheMany relation added above.
3. Set `storage.uploads.disk` in config/cms.php to `s3` (assuming that `s3` is properly configured in config/filesystem.php, and destination S3 bucket is ready to be uploaded files).
4. Set `default` in config/filesystem.php to `local`.
5. Upload file from the model's create/update page in background. It will be uploaded to local storage.
6. If `default` in config/filesystem.php is changed to `s3`, the uploaded file will be stored on S3, but I guess it is not the expected behavior.

##### October build
431

LukeTowers:
Potentially related: #2917

LukeTowers:
Fixed by https://github.com/octobercms/october/commit/54ab6f080a60526fd9685518c1456bc1b7781187

PR #3333, opened 2018-01-01
Fix for #3332: storage.uploads.disk in config/cms.php not effective

pikanji:
This PR fixes issue #3332.

FileUpload widget uploads file to the disk specified by `default` in config/filesystem.php instead of `storage.uploads.disk` in config/cms.php, if we use `System\Models\File` following the instruction in [here](https://octobercms.com/docs/database/attachments#file-attachments).
Although we can still create another class extending `System\Models\File` or `October\Rain\Database\Attach\File` and use it as the model for attachOne/Many relation, `System\Models\File` seems to be the one that responsible to look at `storage.uploads.disk` in config/cms.php, because the existing methods are using `storage.uploads.*`.

### DIFF
--- a/modules/system/models/File.php
+++ b/modules/system/models/File.php
@@ -60,8 +60,8 @@ class File extends FileBase
     }
 
     /**
-     * Returns true if the storage engine is local.
-     * @return bool True if storage.uploads.disk in config/cms.php is "local".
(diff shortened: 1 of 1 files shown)


LukeTowers:
Good catch! Thanks for the fix @pikanji!

PR #3338, opened 2018-01-04
Added ability to list records with colors

flakerimi:
Might not be elegant solution but it does the job. Open for better solution.

screen shot 2018-01-04 at 16 17 04

### DIFF
--- a/modules/backend/widgets/Lists.php
+++ b/modules/backend/widgets/Lists.php
@@ -1164,7 +1164,17 @@ class Lists extends WidgetBase
 
         return Backend::dateTime($dateTime, $options);
     }
+    /**
+     * Process as background color, to be seen at list
+     */
+    protected function evalColorPickerTypeValue($record, $column, $value)
+    {
+        if (is_array($value) && count($value) == count($value, COUNT_RECURSIVE)) {
+            $value = implode(', ', $value);
+        }
 
+        return  '<span style="background:'.$value.'; padding:10px">'.$value.'<span>';
+    }
     /**
      * Validates a column type as a date
      */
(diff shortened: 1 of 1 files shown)


LukeTowers:
A better solution exists, it's called `type: partial` :) This would have to be more complex because you would need to evaluate whether to use a dark or a light text colour for the overlaid text based on the selected colour.

flakerimi:
I saw dark/light problem, but on hover you can see both, maybe just a <span style="width: 50px; height: 50px; background: $value></span> $value, so you have that before text. 

I believe its better to have type: colorpicker rather than partial in this case since it it already as type.

LukeTowers:
escape the value with `e($value)` and I'd accept it

modules/backend/widgets/Lists.php
```diff
@@ -1164,7 +1164,17 @@ protected function evalTimetenseTypeValue($record, $column, $value)
 
         return Backend::dateTime($dateTime, $options);
     }
+    /**
+     * Process as background color, to be seen at list
+     */
+    protected function evalColorPickerTypeValue($record, $column, $value)
+    {
+        if (is_array($value) && count($value) == count($value, COUNT_RECURSIVE)) {
+            $value = implode(', ', $value);
```
LukeTowers: What is the purpose of this line?

flakerimi:
Done.

LukeTowers:
Thanks @flakerimi! Could you add this to the documentation? https://github.com/octobercms/docs

flakerimi:
Its already there, https://github.com/octobercms/docs/blob/master/backend-forms.md#color-picker
No need for further explanation since is assumed behaviour. If you use it in fields, you can use it in columns too.

LukeTowers:
That's not true @flakerimi: http://octobercms.com/docs/backend/lists#column-types

Issue #3339, opened 2018-01-05
ErrorException: putenv() has been disabled for security reasons


jan-vince:
##### Expected behavior
Backed forms should be saved without any errors.

##### Actual behavior
After update to version 431, some backend forms throw error:

````
ErrorException: putenv() has been disabled for security reasons in .../vendor/symfony/console/Application.php:110
````

This is related to PHP configuration of my host that do not allow ````putenv()````.

I understand that I can change hosting company but disabled ````putenv()```` is probably more common situation and many more OctoberCMS users can be affected by this.

As this is related to Symfony package we probably cannot do much about it here but maybe someone has an experience with this and can help.

*Maybe ability to bypass Symfony/console on such hosts?*

##### October build
431

LukeTowers:
@jan-vince it would be helpful if you could provide the full stack trace of the issue.

jan-vince:
Sure, here is one of them (from backend Settings - while saving Log settings).

*When I comment all ````putenv()```` calls in ````.../symfony/console/Application.php````, everything is fine and I can save all forms.*

*I don't know yet how (and why) the symfony console package is used while saving backend forms.*

````
ErrorException: putenv() has been disabled for security reasons in /data/web/virtuals/166926/virtual/www/subdom/rezidencebilylev/vendor/symfony/console/Application.php:110
Stack trace:
#0 [internal function]: Illuminate\Foundation\Bootstrap\HandleExceptions->handleError(2, 'putenv() has be...', '/data/web/virtu...', 110, Array)
#1 /data/web/virtuals/166926/virtual/www/subdom/rezidencebilylev/vendor/symfony/console/Application.php(110): putenv('LINES=50')
#2 /data/web/virtuals/166926/virtual/www/subdom/rezidencebilylev/vendor/laravel/framework/src/Illuminate/Console/Application.php(88): Symfony\Component\Console\Application->run(Object(Symfony\Component\Console\Input\ArrayInput), Object(Symfony\Component\Console\Output\BufferedOutput))
#3 /data/web/virtuals/166926/virtual/www/subdom/rezidencebilylev/vendor/laravel/framework/src/Illuminate/Console/Application.php(177): Illuminate\Console\Application->run(Object(Symfony\Component\Console\Input\ArrayInput), Object(Symfony\Component\Console\Output\BufferedOutput))
[...]

PR #3342, opened 2018-01-08
Catch Artisan::call exception

jan-vince:
This change will allow settings to be saved even when Exception is
thrown in Artisan::call (eg. when putenv() function is disabled).

Is related to #3339 and #3280.


jan-vince:
What is fixed is shown on this video:

v431-putenv-disabled

PR #3343, opened 2018-01-08
Fixing loosely assignments + hard-coded class strings, >= PHP7.0 refactoring

arthurkushman:
Additionally:
- argument types and return types for >=7.0-7.1
- const modifiers for >=7.1

PS thought OctovberCMS built with support for php >= 7.0, but it not seems so in the code, proposing to proceed refactoring at least for php >= 7.0 with arguments typification and returning types.
Which will make code faster (at least at compile-time in OpCodes) and a lot safer.

PPS probably I'm missing something and u have already been started work on this, then I can fetch the branch and make pull request again. Thx

### DIFF
--- a/modules/cms/classes/CmsObject.php
+++ b/modules/cms/classes/CmsObject.php
@@ -9,7 +9,7 @@ use Cms\Contracts\CmsObject as CmsObjectContract;
 use ApplicationException;
 use ValidationException;
 use Exception;
-use Cms\Classes\Theme;
+
 /**
  * This is a base class for all CMS objects - content files, pages, partials and layouts.
  * The class implements basic operations with file-based templates.
--- a/modules/cms/classes/CmsObject.php
+++ b/modules/cms/classes/CmsObject.php
@@ -9,7 +9,7 @@ use Cms\Contracts\CmsObject as CmsObjectContract;
 use ApplicationException;
 use ValidationException;
 use Exception;
-
+use Cms\Classes\Theme;
 /**
  * This is a base class for all CMS objects - content files, pages, partials and layouts.
  * The class implements basic operations with file-based templates.
--- a/modules/cms/classes/MediaLibraryItem.php
+++ b/modules/cms/classes/MediaLibraryItem.php
@@ -13,7 +13,7 @@ class MediaLibraryItem extends SystemMediaLibraryItem
 {
     public function __construct()
     {
-        traceLog('Class Cms\Classes\MediaLibraryItem has been deprecated, use System\Classes\MediaLibraryItem instead.');
+        traceLog('Class ' . __CLASS__ . ' has been deprecated, use ' . SystemMediaLibraryItem::class . ' instead.');
         parent::__construct(...func_get_args());
     }
 }
--- a/modules/cms/classes/MediaLibraryItem.php
+++ b/modules/cms/classes/MediaLibraryItem.php
@@ -13,7 +13,7 @@ class MediaLibraryItem extends SystemMediaLibraryItem
 {
     public function __construct()
     {
-        traceLog('Class ' . __CLASS__ . ' has been deprecated, use ' . SystemMediaLibraryItem::class . ' instead.');
+        traceLog('Class Cms\Classes\MediaLibraryItem has been deprecated, use ' . SystemMediaLibraryItem::class . ' instead.');
         parent::__construct(...func_get_args());
     }
 }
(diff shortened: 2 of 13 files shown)


modules/cms/classes/Page.php
```diff
@@ -62,25 +62,25 @@ public function __construct(array $attributes = [])
...
+    protected function parseSettings(): void
     {
     }
 
     /**
      * Returns name of a PHP class to us a parent for the PHP class created for the object's PHP section.
      * @return mixed Returns the class name or null.
      */
-    public function getCodeClassParent()
+    public function getCodeClassParent(): ?string
```
LukeTowers: Nullable return types are not supported in PHP7.0

modules/cms/classes/Page.php
```diff
@@ -62,25 +62,25 @@ public function __construct(array $attributes = [])
...
+        return PageCode::class;
     }
 
     /**
      * Returns a list of layouts available in the theme.
      * This method is used by the form widget.
      * @return array Returns an array of strings.
      */
-    public function getLayoutOptions()
+    public function getLayoutOptions(): array
```
LukeTowers: The formatting we will go for declaring return types will be `methodName() : type`, not `methodName(): type`

modules/cms/classes/Page.php
```diff
@@ -46,7 +46,7 @@ class Page extends CmsCompoundObject
      */
     public $rules = [
         'title' => 'required',
-        'url'   => ['required', 'regex:/^\/[a-z0-9\/\:_\-\*\[\]\+\?\|\.\^\\\$]*$/i']
+        'url' => ['required', 'regex:/^\/[a-z0-9\/\:_\-\*\[\]\+\?\|\.\^\\\$]*$/i']
```
LukeTowers: Why has this spacing been adjusted?

arthurkushman:
@LukeTowers thx for guiding me - will proceed to do the best I can, also noticed that the phpunit version is outdated a little, I'm using 6.5 and there are no methods like `setExpectedException` coz they were removed a few minor versions ago, suppose it will be problematic for further contributions from other developers. Probably there should be a refactoring on those tests either in future.

Now all checks have passed - if there any issues u want to fix or reorganize - glad to work on it.

LukeTowers:
@arthurkushman there's a bunch of changed whitespace that is unnecessary, the preferred whitespace for array elements in October is to align them:
```php
[
    'foo'    => 'bar',
    'foobar' => 'baz',
]
```

modules/cms/classes/Page.php
arthurkushman: It is auto-format in PhpStorm. If the OctoberCMS team has xml code-format configuration file for PhpStorm - can easily apply it. Or make a settings correction, just in case - have one of the latest PhpStorm version 2017.3 and it was set by default.

LukeTowers: There's https://github.com/octobercms/october/blob/master/.editorconfig, but otherwise none of the core team uses PHPStorm so you'd have to change the settings yourself.

arthurkushman: Ok, thank you.

modules/cms/classes/CmsObject.php
```diff
@@ -9,7 +9,7 @@
 use ApplicationException;
 use ValidationException;
 use Exception;
-
+use Cms\Classes\Theme;
```
LukeTowers: Why is this necessary? The Theme class exists within this namespace.

arthurkushman: Definitely no reason - removed.

modules/cms/classes/MediaLibraryItem.php
```diff
@@ -13,7 +13,7 @@ class MediaLibraryItem extends SystemMediaLibraryItem
 {
     public function __construct()
     {
-        traceLog('Class Cms\Classes\MediaLibraryItem has been deprecated, use System\Classes\MediaLibraryItem instead.');
+        traceLog('Class Cms\Classes\MediaLibraryItem has been deprecated, use ' . SystemMediaLibraryItem::class . ' instead.');
```
LukeTowers: Are you going to make changes to the other traceLog messages using this pattern as well?

arthurkushman: Sure, it is needed for consistency. I'll push changes in a while.

LukeTowers:
@arthurkushman is this all done now?

arthurkushman:
Yes, this is done.

PR #3347, opened 2018-01-10
Fix operations to uploaded files such as deletion

### LINKED ISSUE #3346: `storage.uploads.disk` in config/cms.php is still not effective for some part
My fix for #3332 was not enough. Although the problem I reported there was fixed, I found other functions, such as deletion of uploaded files, are still not working due to the same problem (`System\Models\File` is partially using `default` in config/filesystems.php).
To fix this issue, I suggest to make its base class `October\Rain\Database\Attach\File` a bit more scalable instead of adding more overriding methods to `System\Models\File`.

I have issued PRs for the fix.
* octobercms/library#294
* #3347 

##### Expected/Actual behavior & Reproduce steps
Please see #3332, and deletion of the uploaded file is still not working.

##### October build
431

pikanji:
This fixes #3346. This requires octobercms/library#294.

The modification to apply `storage.uploads.disk` is generalized and moved to `October\Rain\Database\Attach\File` by octobercms/library#294. Most of the modification are done in there.

`System\Models\File` now only overrides getStorageConfigKey() method to provide the key `storage.uploads.disk` to specify the configuration.

### DIFF
--- a/modules/system/models/File.php
+++ b/modules/system/models/File.php
@@ -2,8 +2,6 @@
 
 use Url;
 use Config;
-use File as FileHelper;
-use Storage;
 use October\Rain\Database\Attach\File as FileBase;
 
 /**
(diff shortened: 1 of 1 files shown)


LukeTowers:
Will require changes request in https://github.com/octobercms/library/pull/294 as well

LukeTowers:
This should theoretically be fixed by https://github.com/octobercms/library/commit/48fed9b7e4c4c22ccc6b3c2e5b1273da3e2c38dd & https://github.com/octobercms/october/commit/aea4857ebaaa132499033014309bd5a6032dd66d

pikanji:
Confirmed.

PR #3349, opened 2018-01-10
create october:key command

buuug7:
create october:key command for this issue #3321 , this #3328  PR should be closed.   

if the the key like `  'key' => env('APP_KEY', 'some key')`, this will not work.   

only work with  `'key' => 'some key'`  situation.  

But I still think the best solution is remove ` php artisan key:generate` from `post-create-project-cmd`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -48,7 +48,7 @@
     },
     "scripts": {
         "post-create-project-cmd": [
-            "php artisan key:generate"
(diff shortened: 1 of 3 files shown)


LukeTowers:
Closing in favour of a change to composer.json

buuug7:
(-_-, ok

PR #3353, opened 2018-01-11
update composer.json for solve #3321 issue

buuug7:
This PR try to solve the #3321 issue. the #3328 PR should be closed.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -48,6 +48,7 @@
     },
     "scripts": {
         "post-create-project-cmd": [
+            "php artisan october:env",
             "php artisan key:generate"
         ],
         "post-update-cmd": [


LukeTowers:
For future reference you don't need to close and open a new PR to make changes, just make the changes to the same branch as the PR is from and they'll get automatically included. For this PR could you revert the changes made in https://github.com/octobercms/october/commit/fbf482a3f44eb28076d6df5ee46df30befaa0789 and change the default environment back to production?

buuug7:
@LukeTowers , No, the environment is still `development`, why revert it back to `production`?

LukeTowers:
@buuug7 Never mind, I forgot that the change was the first part necessary to get it working with this change as well.

LukeTowers:
@buuug7 does this change now finally solve the issues with creating projects from composer?

buuug7:
@LukeTowers yes, it sovled. but rethink it ,it would be more nice if remove all from `post-create-project-cmd`.  leave it empty . is it ? 
```
 "post-create-project-cmd": [ ],
```

itsam:
@LukeTowers I thing @buuug7 is right. You should leave "post-create-project-cmd": [ ]
The "php artisan october:env" causes General error: 1 no such table: system_settings appears when `php artisan october:install` is executed (obviously since there's the .env present). Also, .env selects mysql by default (which makes it cumbersome to change to sqlite)

LukeTowers:
@itsam then perhaps we should change it to no longer be run by composer but instead be triggered at the end of the `october:install` command. 

Speaking for only myself I don't use the `october:install` command nor the `october:update` command, but it is important to support those. I still want both of these commands to be run on new project creation given that setting the application key is very important for security and having an environment file is also a good thing. Perhaps we should get @daftspunk to weigh in on this?

buuug7:
@LukeTowers @itsam  this really cause the error for use `php artisan october:install` command. to solve this problem, we have two ways: 
+ leave the `post-create-project-cmd` empty in the composer.json file.
+ improvement the `October\Rain\Config\ConfigWriter` support `'key' => env('Key', 'Value')` type

default the ConfigWrite is support this type
```
 * The following value types are supported for writing:
 * - strings
 * - integers
 * - booleans
 * - nulls
 * - single-dimension arrays
```

LukeTowers:
@buuug7 I would say that adding proper support for `env()` values to October\Rain\Config\ConfigWriter would be the best solution here.

buuug7:
@LukeTowers  October\Rain\Config\ConfigWriter  have issue for write some key with `'key' => env('Key', 'Value')`, I tried it today but it still didn't work out.
```
    /*
     * Match configured with env function
     * match as 'someKey' => env('someKey', 'someValue')
     * */
    protected function buildEnvExpression($targetKey, $arrayItems = array())
    {
        $expression = array();

        // Opening expression for array items ($1)
        $expression[] = $this->buildArrayOpeningExpression($arrayItems);

        // The target key opening ($2)
        $expression[] = '([\'|"]' . $targetKey . '[\'|"]\s*=>\s*env\s*\(\s*[\'|"]\S+[\'|"]\s*\,\s*)';

        // The target value to be replaced ($3)
        $expression[] = '(([\'|"]\S+[\'|"])|(\d+)|[Nn][Uu][Ll][Ll])';
        return '/' . implode('', $expression) . '/';
    }

```

daftspunk:
Rolled this back, causing a plethora of knock on issues. Using the `october:env` approach should be opt-in. It's not a good idea to have some environments using basic config and some using env config by default, makes troubleshooting a nightmare.

PR #3361, opened 2018-01-17
Replace URL parameters dynamically without needing a list in advance

### LINKED ISSUE #3358: ListController display can render slowly depending on model's $appends usage.
##### Expected behavior
The `ListController` behavior will put together tables of model data with minimal unnecessary data access.

##### Actual behavior
If a list defines a record URL, then rendering each row calls the model's `toArray` function so that any model attribute can be used as a URL parameter. The `toArray` call adds all attributes defined in the model's `$appends` property. If the model appends any attributes that result in a database query, then those queries will run for each record, generally to no real purpose, since there's likely no corresponding parameter in the URL (I've personally not yet had a reason to use anything more than `:id`).

Many models will of course not fit this pattern, but any model that does has backend lists render more slowly than normal and particularly with larger page sizes can become noticeable to the end user.

##### Reproduce steps
- Create a model that uses `$appends` to include extra attributes, and define those attribute getters to query the database (or even just add a call to `sleep`).
- Set up a backend controller using the standard `ListController` behavior to create a list display for the model.
- Add a few records to the model table.
- Make use of a tool that will display queries and execution time (debugbar or equivalent)
- Loading the list index page will result in any appended attribute query (or `sleep`) running for each row.

##### October build
431

There are a couple of possible workarounds, but I think the ideal solution would be to parse URLs for the needed parameter names rather than grabbing all attribute values just in case they're used. I don't see a specific function to do that, though `October\Rain\Router\Helper` could be modified to handle that without too much effort.

As a side note, it's a little strange that setting the list's `recordUrl` results in `toArray` being used to include all attribute values, but `recordOnClick` just makes use of `getAttributes`. So one fix could be to just use `getAttributes` both places, though someone somewhere might depend on the current behavior.

I'd be happy to put together a pull request for this, but I wanted to check if there's something I'm overlooking, and I also want to make sure there's the "will" to add a fix for it, so I don't just add a pull request to the backlog.

jimcottrell:
Fixes #3358. Depends on octobercms/library#297.

`Backend\Widgets\Lists::getRecordUrl` called `toArray` on models, allowing access to all attributes and relationships in the serialization, but requiring possibly a large amount of data to be loaded. This is specifically a replacement for that behavior to access only attributes for which there are parameters in the URL, but I figured as long as I was making changes, I may as well update the `FormController` redirect and `Lists::getRecordOnClick` to allow access to any model attributes, including those accessible via get mutators.

### DIFF
--- a/modules/backend/behaviors/FormController.php
+++ b/modules/backend/behaviors/FormController.php
@@ -464,7 +464,7 @@ class FormController extends ControllerBehavior
         }
 
         if ($model && $redirectUrl) {
-            $redirectUrl = RouterHelper::parseValues($model, array_keys($model->getAttributes()), $redirectUrl);
+            $redirectUrl = RouterHelper::replaceParameters($model, $redirectUrl);
         }
 
         return ($redirectUrl) ? Backend::redirect($redirectUrl) : null;
--- a/modules/backend/widgets/Lists.php
+++ b/modules/backend/widgets/Lists.php
@@ -562,12 +562,7 @@ class Lists extends WidgetBase
             return null;
         }
 
-        $data = $record->toArray();
-        $data += [$record->getKeyName() => $record->getKey()];
-
-        $columns = array_keys($data);
-
-        $url = RouterHelper::parseValues($data, $columns, $this->recordUrl);
+        $url = RouterHelper::replaceParameters($record, $this->recordUrl);
         return Backend::url($url);
     }
 
@@ -582,8 +577,7 @@ class Lists extends WidgetBase
             return null;
         }
 
-        $columns = array_keys($record->getAttributes());
-        $recordOnClick = RouterHelper::parseValues($record, $columns, $this->recordOnClick);
+        $recordOnClick = RouterHelper::replaceParameters($record, $this->recordOnClick);
         return Html::attributes(['onclick' => $recordOnClick]);
     }
 


petehalverson:
@LukeTowers @daftspunk, we'd see a significant backend performance boost from this PR in our e-commerce plugin. Do you need to see an example in the OC test plugin?

LukeTowers:
@petehalverson sure, that'd be fantastic! Thanks for following up, issues get lost if there's no checkins, I appreciate the follow up

w20k:
@LukeTowers I will! 

@petehalverson could you still submit a PR to OC-Test with examples? Which will simplify the verification process 😉

LukeTowers:
@jimcottrell if @petehalverson is busy right now, would you be willing to submit examples to the [test plugin](https://github.com/octoberrain/oc-test-plugin)?

w20k:
@LukeTowers, I've started testing this feature. Looks promising! 😄

w20k:
@LukeTowers, I've tested this feature with some simple examples that I could find (within Test-Plugin). 

Works really great. 
Approximatley 10 times faster (see samples below), thought I'd really like to have more examples with model redicrect (tested only with 'save and close') or 'hardcore parametrized' URLs, if anyone could suggest plugin with many URL params, would be awesome!. 

Performance analysis:

## URL with one parameter.
**URL example**: `$('#RecordFinder-formPhone-phone').recordFinder('updateRecord', this, **':id'**) -> $('#RecordFinder-formPhone-phone').recordFinder('updateRecord', this, **'4'**)`

**Without PR**:
Time:  0.0094 Seconds
‌Time:  0.0158 Seconds
‌Time:  0.0123 Seconds
‌Time:  0.0129 Seconds

**With PR**:
Time:  0.0013 Seconds
Time:  0.0019 Seconds
‌Time:  0.0013 Seconds
‌Time:  0.0024 Seconds

PR #3363, opened 2018-01-17
Develop fix type juggling, strings to ::class accross all project, flow structs

arthurkushman:
- fixed type juggling 
- fixed hard coded strings to ::class
- flow structures ex:
```php
if (condition) {
  return false;
} else {
  return true;
}
// to 
if (condition) {
  return false;
}
return true;
```


LukeTowers:
Lots of stuff to review :)

If you're looking for other ways to support the project, inline documentation for more of the available events would be absolutely fantastic! Here is an example of what would be needed: https://github.com/octobercms/october/blob/master/modules/cms/classes/CmsObject.php#L150-L179. Here is how that documentation gets rendered in our API docs: http://octobercms.com/docs/api/cms/object/listintheme; and here is a spreadsheet of most of the available events within October: https://docs.google.com/spreadsheets/d/17vv4zn_ezp9iFBgYhYuYsmUG4MXW9uuIw7aOIrVcHiM/edit#gid=0

modules/cms/classes/MediaLibrary.php
```diff
@@ -17,7 +17,7 @@ class MediaLibrary extends SystemMediaLibrary
      */
     protected function init()
     {
-        traceLog('Class Cms\Classes\MediaLibrary has been deprecated, use System\Classes\MediaLibrary instead.');
+        traceLog('Class Cms\Classes\MediaLibrary has been deprecated, use ' . SystemMediaLibrary::class . ' instead.');
```
36864: For consistency, would it not be best to also replace the first class name?
`traceLog('Class ' . static::class . ' has been deprecated, use ' . SystemMediaLibrary::class . ' instead.')`

modules/cms/classes/Page.php
```diff
@@ -70,7 +70,7 @@ protected function parseSettings()
      * Returns name of a PHP class to us a parent for the PHP class created for the object's PHP section.
      * @return mixed Returns the class name or null.
      */
-    public function getCodeClassParent()
+    public function getCodeClassParent() : string
     {
         return '\Cms\Classes\PageCode';
```
36864: Looks like you missed this one

modules/cms/classes/MediaLibrary.php
LukeTowers: @36864, no, because if someone has extended Cms\Classes\MediaLibrary, then their message will be "Class Your\Custom\Class has been deprecated", which is not the case.

modules/cms/classes/Controller.php
```diff
@@ -148,18 +148,16 @@ public function run($url = '/')
             $url = Request::path();
         }
 
-        if (!strlen($url)) {
+        if ('' === $url) {
```
LukeTowers: Who invited yoda? 😄 In all seriousness though, wouldn't an `if (empty())` be better here?

modules/cms/helpers/File.php
```diff
@@ -30,7 +30,7 @@ public static function validateName($fileName)
     public static function validateExtension($fileName, $allowedExtensions, $allowEmpty = true)
     {
         $extension = strtolower(pathinfo($fileName, PATHINFO_EXTENSION));
-        if (!strlen($extension)) {
+        if ('' === $extension) {
```
LukeTowers: yoda vs `empty` again

tests/unit/cms/classes/CodeParserTest.php
```diff
@@ -1,5 +1,6 @@
 <?php
 
+use Cms\Classes\LayoutCode;
```
LukeTowers: I'd prefer the `use` statements were in order

modules/cms/classes/MediaLibrary.php
36864: `\Cms\Classes\MediaLibrary::class` then?

modules/cms/classes/Controller.php
arthurkushman: =) funny, I would rather switch sides (totally agree with stick to common style), then make this statements with empty as `empty()` checks for many (`null`, `false` etc) values, which is expensive, only in those cases where we expecting `string` as input value, like in this example.

tests/unit/cms/classes/CodeParserTest.php
arthurkushman: Didn't understand that - in lexicographical order?

modules/cms/classes/Page.php
arthurkushman: Sure, thx - will fix that.

modules/cms/classes/Controller.php
LukeTowers: Are you sure it's that expensive? Run some benchmarks on it and then get back to me with the results. `empty()` is a language construct, it's pretty fast. I personally prefer using it because it's easier to read for developers and it encompasses more potential states for the variable to be in that all ultimately mean the same thing, that there is no useful data in the variable to work with.

modules/cms/classes/MediaLibrary.php
36864: Isn't this whole PR about pedantry? I just find it odd that you'd use `::class` for one thing and not another in the same line. And yes, `self::class` would be perfect here.

arthurkushman: I can fix that to `self::class` and yes I'm pedantic )

modules/cms/classes/Controller.php
arthurkushman: Definitely, the decision is yours - tests:
```php
<?php
// empty
$start = microtime(true);
$str = '';
for ($i = 0;$i < 1e6;++$i) {
    if (empty($str)) {

    }
}
$t = microtime(true) - $start;
echo 'empty time:' . $t . PHP_EOL;
// equation
$start = microtime(true);
for ($i = 0;$i < 1e6;++$i) {
    if ($str === '') {

    }
}
$t = microtime(true) - $start;
echo 'string time:' . $t . PHP_EOL;
```  

Results:
1...
```
empty time:0.14257001876831
string time:0.1290819644928
```
2...
```
empty time:0.14147520065308
string time:0.12645101547241
```
3...
```
empty time:0.1415810585022
string time:0.12579417228699
```

LukeTowers: Looks like approximately a hundredth of a second's difference over 1 million iterations in favour of `"" ===` over `empty()`. Personally in my mind that's not really enough to make `empty()` less desirable. Do you have any further input on it though? (Also, `$str` is defined in the empty test and reused in the `=== ''` test, so that might make a very minor difference).

arthurkushman: Ok, I'll turn all the "" === $str into empty()

LukeTowers:
@arthurkushman what's the status of this now?

arthurkushman:
I presume all things we wanted to fix are done, all the classes and operators fixed either.

PopcornPHP:
Why was the use of this class removed?

https://github.com/octobercms/october/commit/c78b3ef78619a442aee045ff84d4fb62aee378ec#diff-8e0645f0b59fa1fe39d9774e50d10d4cL6

LukeTowers:
@PopcornPHP fixed it.

PR #3370, opened 2018-01-23
Twig environment "strict_variables"

matteotrubini:
https://twig.symfony.com/doc/2.x/api.html#environment-options

`strict_variables` will ensure a cleaner approach in themes' development.

### DIFF
--- a/modules/cms/classes/Controller.php
+++ b/modules/cms/classes/Controller.php
@@ -488,6 +488,7 @@ class Controller
 
         $useCache = !Config::get('cms.twigNoCache');
         $isDebugMode = Config::get('app.debug', false);
+        $strictVariables = (Config::get('cms.enableTwigStrictVariables', null) !== null) ?: $isDebugMode;
         $forceBytecode = Config::get('cms.forceBytecodeInvalidation', false);
 
         $options = [
--- a/modules/cms/classes/Controller.php
+++ b/modules/cms/classes/Controller.php
@@ -488,7 +488,8 @@ class Controller
 
         $useCache = !Config::get('cms.twigNoCache');
         $isDebugMode = Config::get('app.debug', false);
-        if (($strictVariables = Config::get('cms.enableTwigStrictVariables', null)) === null) {
+        $strictVariables = Config::get('cms.enableTwigStrictVariables', false);
+        if ($strictVariables === null) {
             $strictVariables = $isDebugMode;
         }
         $forceBytecode = Config::get('cms.forceBytecodeInvalidation', false);
--- a/modules/cms/classes/Controller.php
+++ b/modules/cms/classes/Controller.php
@@ -494,7 +494,7 @@ class Controller
         $options = [
             'auto_reload' => true,
             'debug' => $isDebugMode,
-            'strict_variables' => $isDebugMode,
+            'strict_variables' => $strictVariables,
         ];
 
         if ($useCache) {
--- a/modules/cms/classes/Controller.php
+++ b/modules/cms/classes/Controller.php
@@ -488,7 +488,9 @@ class Controller
 
         $useCache = !Config::get('cms.twigNoCache');
         $isDebugMode = Config::get('app.debug', false);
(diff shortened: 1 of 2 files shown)


matteotrubini:
I changed my mind: `strict_variables` couldn't inherit `app.debug` values as too many views was coded relying on the fact that currently Twig is silently ignoring invalid variables. Too many components would become unusable.

But I still thing that `strict_variables` should be configurable.

What do you think about add a `cms.twigStrictVariables` config (default to false)?

LukeTowers:
@matteotrubini named `cms.enableTwigStrictVariables` to match the current options would be accepted. I agree that the too much pre-existing code relies on the current behaviour; especially when you consider the number of production sites that never switched debug mode off.

matteotrubini:
I agree 👍

modules/cms/classes/Controller.php
```diff
@@ -488,11 +488,13 @@ protected function initTwigEnvironment()
 
         $useCache = !Config::get('cms.twigNoCache');
         $isDebugMode = Config::get('app.debug', false);
+        $strictVariables = (Config::get('cms.enableTwigStrictVariables', null) !== null) ?: $isDebugMode;
         $forceBytecode = Config::get('cms.forceBytecodeInvalidation', false);
 
         $options = [
             'auto_reload' => true,
             'debug' => $isDebugMode,
+            'strict_variables' => $isDebugMode,
```
LukeTowers: This needs to be `$strictVariables`, please test each config value and ensure that they're working as intended (true = enabled, false = disabled, null = value of app.debug)

matteotrubini: Ouch, a bad copy&paste from my testing installation :sweat:
I made all config combination test with `dd()` just after var assignment.

modules/cms/classes/Controller.php
```diff
@@ -488,11 +488,13 @@ protected function initTwigEnvironment()
 
         $useCache = !Config::get('cms.twigNoCache');
         $isDebugMode = Config::get('app.debug', false);
+        $strictVariables = (Config::get('cms.enableTwigStrictVariables', null) !== null) ?: $isDebugMode;
```
LukeTowers: Wouldn't this set $strictVariables = true when Config::get returns false?

matteotrubini: It's equivalent to:
```
if (Config::get('cms.enableTwigStrictVariables', null) !== null) {
    $strictVariables = Config::get('cms.enableTwigStrictVariables', null);
} else {
    $strictVariables = $isDebugMode;
}
```
or
```
$strictVariables = Config::get('cms.enableTwigStrictVariables', null);
if ($strictVariables !== null) {
    $strictVariables = $isDebugMode;
}
```
REF: https://en.wikipedia.org/wiki/Elvis_operator

matteotrubini: Today I need more coffee ... :coffee:

LukeTowers:
@matteotrubini just simplify this, no need to make it fancy:
```php
// We have to default to false because updating october does not overwrite config files so existing
// installs will not have this config key present, so the default must be the same as the default would
// be if the config key existed.
$strictVariables = Config::get('cms.enableTwigStrictVariables', false);
$strictVariables = is_null($strictVariables) ? $isDebugMode : $strictVariables;
```

modules/cms/classes/Controller.php
```diff
@@ -488,11 +488,16 @@ protected function initTwigEnvironment()
 
         $useCache = !Config::get('cms.twigNoCache');
         $isDebugMode = Config::get('app.debug', false);
+        $strictVariables = Config::get('cms.enableTwigStrictVariables', false);
+        if ($strictVariables === null) {
```
LukeTowers: use `is_null()` please. In the first place, the performance difference between them is ridiculously slim, in the second place PHP7 actually reverses the performance difference. PHP5.6 `===` is ~100 nano seconds faster, PHP7 `is_null()` is ~5 nano seconds faster. We use `is_null` elsewhere in the code so just use it here too please.

PR #3373, opened 2018-01-24
Use config variable for data cache key.

whsol:
If we deploy octobercms on two servers (in the same path on both), and setup one cache storage for it (for example redis).
System will store information about cached "partials" in redis, and servers will rewrite this information because they are stored under the same key. System validate cache by compare "mtime", between "mtime" in redis and filesystem "mtime".
Often we have "mtime" from server1 in redis which compare with file mtime on server2, so system run rebuild cache very often.

And when we have big traffic we able to get error:
prod.ERROR: Symfony\Component\Debug\Exception\FatalThrowableError: Class 'Cms5a667baecf93e170794683_4ce1b5d369a0cfcf4b16d519cea731c1Class' not found in /var/www/html/modules/cms/classes/CodeParser.php:180


LukeTowers:
Please add a default comment to the config/cache.php file to document the addition of this feature.

viamage:
@whsol could you get back at it, any help needed? per Luke comment above, also some travis checks failing? would be awesome to see this fixed and merged.

LukeTowers:
@viamage if @whsol doesn't want to do that feel free to resubmit.

LukeTowers:
Awesome, thanks very much @whsol!

PR #3374, opened 2018-01-25
Preparing for Plugin Management UX

Teranode:
This will fix the refererence to https://github.com/octobercms/october/issues/2714 also see in _disable_form.htm about the js used to alter the checkboxes.

Issue #2774, opened 2017-03-23
i can't enable the disabled plugins


tim0991:
##### Expected behavior
I hope the plugins can be enable in normal

##### Reproduce steps
1. i disable plugins in backend
2. i can't found a 'enable' opiton when i want to enable plugins

![image](https://cloud.githubusercontent.com/assets/26053904/24235793/cc636ebc-0fd9-11e7-8e87-c1578a19a8bc.png)

![image](https://cloud.githubusercontent.com/assets/26053904/24235817/e4a163c6-0fd9-11e7-825f-fc52298393ed.png)
3. i try the 'Freeze updates' option,  it show me  below, it means 'Freeze updates' is 'enable'?
![image](https://cloud.githubusercontent.com/assets/26053904/24235850/129d3fa2-0fda-11e7-8425-11ee9f398a05.png)

##### October build
v1.0.401

LukeTowers:
Duplicate of #2714. See https://github.com/octobercms/october/issues/2714#issuecomment-287591391 for the solution, basically just leave those boxes unticked and click apply / save / confirm (whatever that popup says for a positive action button).

I do agree that this UX could be much improved by populating the checkbox with the current status of the plugin so that it was more obvious what exactly you were doing to a plugin when using that popup.

daftspunk:
Great suggestion, someone PR this?

Issue #2714, opened 2017-02-27
Plugin Freeze/Unfreeze


pjrola:
##### Expected behavior
Ability to "unfreeze" a plugin being currently frozen from updates. 

##### Actual behavior
Update hop up displays currently frozen plugin with no clear way to unfreeze. A checkbox displays however it is un checked. This checkbox should be checked. When unchecking it should update the system_plugin_versions table. Currently you have to manually navigate into the table to change the value from 1 to 0 in the is_frozen column.

##### Reproduce steps
Navigate to Settings > Updates & Plugins click manage plugins, choose a plugin and click Enable-Disable. Notice freeze updates checkbox appears. Click Freeze updates note there doesn't seem to be a way to un freeze after a plugin has previously been frozen.

##### October build
1.0.1

daftspunk:
I can see how this might be confusing UX, but when you leave the checkbox unticked and click Apply, it will unfreeze them.

LukeTowers:
@daftspunk Could we check the checkbox in that dialog when it has been frozen so that it is more obvious? Currently it's confusing (even to me) when I'm using that dialog and I'm uncertain if my changes have taken effect, or how to reverse them.

daftspunk:
If you have selected 1 frozen and 1 unfrozen, is the checkbox checked or unchecked?

LukeTowers:
Right now, it's unchecked regardless of how many plugins are selected. I think we'd have to just implement the current status indicator if one plugin is selected and improve the user interface (maybe a better explanation of what will happen) for when more than one plugin is selected.

daftspunk:
> we'd have to just implement the current status indicator if one plugin is selected

I'm on board with this!

Teranode:
@daftspunk would this be a good way to implement it? 
![](https://image.prntscr.com/image/oVFCQU55RqiCbAgdkAVA5A.png)

each plugin can be toggled rather than doing a select/deselect all.

LukeTowers:
@Teranode a switch is better, but I think it could probably be made even simpler. Would you be able to do it in a table layout?

| Plugin | Freeze Updates | Disable Plugin |
| --- | --- | --- |
| RainLab.User | Off | Off |
| Uxms.Restful | Off | On |

Teranode:
@LukeTowers ![](https://image.prntscr.com/image/caSkxRi1Sn2g6Cl5NyM7YA.png)

would this work, also changed the words to a yes/no since its more a question where you want to do an action

[...]

PR #3379, opened 2018-02-02
Plugin Management UX Improvements

LukeTowers:
With the assistance of @Teranode (#3374), here are the improvements to the Plugin Management UX as required by #2714 & #2774. 

Old Plugin Management Listing Page:
screen shot 2018-02-01 at 8 40 35 pm

New Plugin Management Listing Page:
screen shot 2018-02-01 at 8 41 04 pm

Old Plugin Bulk Management Actions:
screen shot 2018-02-01 at 8 40 43 pm

New Plugin Bulk Management Actions:
screen shot 2018-02-01 at 8 41 19 pm


munxar:
Fantastic job! I really like it.
One thing I would interpret differently are the icons in the dropdown.
I would switch the disable plugin (pause icon) with the disable updates (disabled icon) *and*
the enable plugin (play icon) with the enable updates (check icon).
Why?
The update process is like something that runs forward as time goes by. By stopping it we actually pausing it, if enabled it will run the updates again (at least on a meta level).
On the other hand the check and disabled icon would more naturally fit to my understanding of enabled, disabled.

For the reset plugins: The bomb indicates something blows up, which is (should) not be the case. As I understand we just force to download the plugins source code again? No migration, so no data loss. Imho it should by a sync icon or something.

Again just my first impression and totally biased, do with this input whatever you like.
Again, good job!

LukeTowers:
@munxar good point on the icons, something to consider. In regards to the bomb icon, it does destroy your data because it's effectively running `down` and then `up` on the plugin's migrations, which is the same as removing the tables completely and recreating them, destroying any data added by the user.

LukeTowers:
Updated actions:

screen shot 2018-02-02 at 12 23 33 am

munxar:
I'd vote only for superusers. Good point!

LukeTowers:
Merged this into develop; I changed it so that only superusers can reset plugin data; and only when the site is in debug mode.

alvaro-canepa:
It would be great to have a button to apply new updates on plugins in development stage. The same as logout/login from backend to apply new updates, without to reload entire version branchs.

Eoler:
>It would be great to have a button to apply new updates on plugins in development stage.

The Rainlab.Builder (Versions) already does it for plugin you're working on.

alvaro-canepa:
@Eoler but you need to install it only for one feature. Not all developers use it. I prefer to build plugins using command line scaffolding.

LukeTowers:
@alvaro-canepa pull requests always welcome :)

SiyuQian:
@LukeTowers Good Work! I have one question about this UX improvement. Why not have a toggle button on the /backend/system/updates to allows the users disable and enable plugins without entering the mange plugin page(backend/system/updates/manage)

LukeTowers:
@SiyuQian honestly because of the effort that it would require. I would have had to redirect the manage action as well as implement all of that work on the normal page, so I went with just getting the improvements out there to start and then probably at a later date we'll go back and move it to the plugin listing page.

Teranode:
I too initally thought about doing it from the updates page, but I was on a supposedly made time crunch from work. Honestly i like where it is, while I like lazy/efficent end goals, i personally prefer to make people make that one extra step :V (as a developer)

GinoPane:
When is it expected to go live?

LukeTowers:
In a few weeks @GinoPane; basically as soon as @daftspunk releases a new build.

modules/system/controllers/updates/_list_manage_toolbar.htm
```diff
@@ -1,52 +1,94 @@
...
-            data-request-confirm="<?= e(trans('system::lang.plugins.refresh_confirm')) ?>"
-            data-trigger-action="enable"
-            data-trigger=".control-list input[type=checkbox]"
-            data-trigger-condition="checked"
-            data-stripe-load-indicator>
-            <?= e(trans('system::lang.plugins.refresh')) ?>
-        </button>
-        -->
+
+        <ul class="dropdown-menu" data-dropdown-title="<?= e(trans('rainlab.user::lang.users.bulk_actions')) ?>">
```
adrenth: This translation message identifier is invalid.

LukeTowers: @adrenth could you make a PR to fix that?

Issue #3393, opened 2018-02-12
storm.css : .loading-indicator shouldn't have background property set.


Alex360hd:
Hello,

It's a very, very small thing, but the .loading-indicator css rules in storm.css add have a background:#f9f9f9; property.

I think that the correct behavior for that div is to have the same background than the element where it's displayed. So, you shouldn't write any background rules to that div, or eventually background: transparent;

Best regards,

Alex

LukeTowers:
Submit a PR to the LESS file in question and include a before and after screenshot with both regular usage and usage where the background is causing a problem please.

Alex360hd:
PR created.

First PR, be indulgent ^^

LukeTowers:
Will close when PR merged

PR #3394, opened 2018-02-12
Update loader.less

Alex360hd:
This PR reference #3393 

Here is before / after screenshot.

loading-indicator-screen

### DIFF
--- a/modules/system/assets/ui/less/loader.less
+++ b/modules/system/assets/ui/less/loader.less
@@ -17,7 +17,6 @@
     color: @color-loading-indicator-text;
     font-size: 14px;
     font-weight: 500;
-    background: @body-bg;
     text-align: left;
     z-index: @zindex-loader;
 


LukeTowers:
And the before after screenshot for where it's used in core?

Alex360hd:
Here is screenshot when used with default backend list page (one generated by the builder plugin).

backend-loader-list

Another example when you click on "refresh" button on the event log backend setting page.

backend-loader-log

As my PR remove the background css rule of the backend loader, it will always be displayed with the background color of the parent element, which I suppose is a better behavior than inherit the body background color.

LukeTowers:
Excellent, thanks!

LukeTowers:

Alex360hd:
Ok no problem, I didn't know sorry.

LukeTowers:
@Alex360hd it turns out this wasn't the best solution as it's caused a bunch of other issues so it will be reverted: https://github.com/octobercms/october/pull/3840.

It looks like your case is custom, at least it's not a widget configuration / design combination that I recognize. In this case you should write your own CSS specifically targeting your specific implementation and only include it on the pages where it is needed.

Let me know if you have any questions!

Alex360hd:
No problem to me, it's better to have an unwanted grey box (that you could configure in your plugin if it's really unwanted) than a buggy text overlay on most backend ajax buttons.

PR #3421, opened 2018-02-28
Fixing stale cache issue

interworks-morr:
Occasionally with certain IIS setups, the file will contain an out-of-date class name.  This patch will verify the class name from the file before returning, and clear the cache if invalid.

### DIFF
--- a/modules/cms/classes/CodeParser.php
+++ b/modules/cms/classes/CodeParser.php
@@ -192,8 +192,10 @@ class CodeParser
 
         if (is_file($path)) {
             if ($className = $this->extractClassFromFile($path)) {
-                $data['className'] = $className;
-                return $data;
+                if (class_exists($className)) {
+                    $data['className'] = $className;
+                    return $data;
+                }
             }
 
             @unlink($path);


LukeTowers:
Can you think of a way that this can be unit tested?

interworks-morr:
I'm not sure of a way to unit test it because I couldn't reliably recreate the issue.  To troubleshoot, I had to wait for it to occur naturally and add debug statements while the site was down.  I'm still unsure of what caused the class name cache to expire but leave the files behind.

It would be a bit janky, but one way to test it would be to manually corrupt the class name in one of the files and then refresh the page to ensure that it cleared and restored cache as expected.

LukeTowers:
Are you able to run that test manually to verify that this change fixes the issue? @interworks-morr

interworks-morr:
I just manually corrupted my `storage/cms/cache/b7/81/default.htm.php` file.  The existing version threw an exception.  With this patch applied, it appears to have caught the issue and rebuilt the corrupted cache.

LukeTowers:
Sweet, thanks for the contribution @interworks-morr!

PR #3432, opened 2018-03-06
Cache page number in List widget

tobias-kuendig:
I am proposing a minor optimization to the way the list widget works when displaying multiple pages. This PR is one solution to this problem. I am open to discuss any other methods to solve this problem.

Let's say I am on page 5 of 10 in my list view and I clilck on a record. If I make the desired modifications and click on `Save` or `Cancel` I am redirected back to the list but I am now on page 1 and not on page 5.

My solution is to save the last visited page number into a session variable (like it is done with the search query and visible columns) and load it when the List widget creates the pagination.

### DIFF
--- a/modules/backend/widgets/Lists.php
+++ b/modules/backend/widgets/Lists.php
@@ -263,6 +263,9 @@ class Lists extends WidgetBase
 
         if ($this->showPagination) {
             $this->vars['pageCurrent'] = $this->records->currentPage();
+            // Store the currently visited page number in the session so the same
+            // data can be displayed when the user returns to this list.
+            $this->putSession('lastVisitedPage', $this->vars['pageCurrent']);
             if ($this->showPageNumbers) {
                 $this->vars['recordTotal'] = $this->records->total();
                 $this->vars['pageLast'] = $this->records->lastPage();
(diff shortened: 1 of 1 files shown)


modules/backend/widgets/Lists.php
```diff
@@ -540,8 +543,13 @@ protected function getRecords()
             $records = $model->getNested();
         }
         elseif ($this->showPagination) {
-            $method = $this->showPageNumbers ? 'paginate' : 'simplePaginate';
-            $records = $model->{$method}($this->recordsPerPage, $this->currentPageNumber);
+            $method            = $this->showPageNumbers ? 'paginate' : 'simplePaginate';
+            $currentPageNumber = $this->currentPageNumber;
+            // Restore the last visited page from the session if available.
+            if ( ! $currentPageNumber && $currentPageNumber = $this->getSession('lastVisitedPage')) {
+                $this->putSession('lastVisitedPage', false);
```
LukeTowers: why store an explicit false to the `lastVisitedPage` session var?

tobias-kuendig: Yes, good question. This was from my first implementation. I wanted to restore the page number only the first time I return to the list but I just noticed that all the other settings are saved over multiple page views as well so maybe it would be more consistent if we just store the number until the session ends.

What do you think?

```php
            $currentPageNumber = $this->currentPageNumber;
            if ( ! $currentPageNumber) {
                // Restore the last visited page from the session if available.
                $currentPageNumber = $this->getSession('lastVisitedPage');
            }
```

tobias-kuendig:
I just noticed, the page should not be cached if any filtering is active.

modules/backend/widgets/Lists.php
```diff
@@ -540,8 +543,13 @@ protected function getRecords()
             $records = $model->getNested();
         }
         elseif ($this->showPagination) {
-            $method = $this->showPageNumbers ? 'paginate' : 'simplePaginate';
-            $records = $model->{$method}($this->recordsPerPage, $this->currentPageNumber);
+            $method            = $this->showPageNumbers ? 'paginate' : 'simplePaginate';
+            $currentPageNumber = $this->currentPageNumber;
+            if ( ! $currentPageNumber && empty($this->searchTerm)) {
```
LukeTowers: get rid of the spaces around the bang `!` please :)

LukeTowers:
@tobias-kuendig could you find someone on slack or something willing to test this as a double check?

edhardie:
@tobias-kuendig I've just seen this in Slack and given it a quick test. Can confirm I'm returned to correct page after editing a record. Page was not cached when I had a search term applied.

pvullioud:
nice addition thanks. work well on preview and edit

PR #3438, opened 2018-03-09
Silence errors coming from OpCache

danhrrn:
I noticed an error being thrown by OpCache:

```
Zend OPcache API is restricted by "restrict_api" configuration directive
```

This PR temporarily silences it until a solution can be found.

### DIFF
--- a/modules/cms/classes/CodeParser.php
+++ b/modules/cms/classes/CodeParser.php
@@ -335,8 +335,8 @@ class CodeParser
          * Compile cached file into bytecode cache
          */
         if (Config::get('cms.forceBytecodeInvalidation', false)) {
-            if (function_exists('opcache_invalidate')) {
-                @opcache_invalidate($path, true);
+            if (function_exists('opcache_invalidate') && opcache_invalidate($path, true)) {
+                opcache_invalidate($path, true);
             }
             elseif (function_exists('apc_compile_file')) {
                 apc_compile_file($path);
--- a/modules/cms/classes/CodeParser.php
+++ b/modules/cms/classes/CodeParser.php
@@ -336,7 +336,7 @@ class CodeParser
          */
         if (Config::get('cms.forceBytecodeInvalidation', false)) {
             if (function_exists('opcache_invalidate')) {
-                opcache_invalidate($path, true);
+                @opcache_invalidate($path, true);
             }
             elseif (function_exists('apc_compile_file')) {
                 apc_compile_file($path);


LukeTowers:
The same could be said of any error; why should we silence it? Wouldn't it be a better idea to address it properly instead?

danhrrn:
@LukeTowers it seems to be that it depends on your environment and backend authentication whether this error shows up or not.
I am unable to find the real root cause of this issue. It only appears when the user visits the frontend site whilst logged in to the backend. This is also the second SP server I have encountered this on - please check your own. I doesn't show up on Homestead or the like.

danhrrn:
I have looked into it more - it is being thrown because `opcache.restrict_api`'s value is 1.

From [php.net](http://php.net/manual/en/opcache.configuration.php#ini.opcache.restrict-api), `opcache.restrict_api`...
> Allows calling OPcache API functions only from PHP scripts which path is started from specified string. The default "" means no restriction.

What would be a suitable permanent fix for this?

danhrrn:
@LukeTowers I have discussed this issue with a ServerPilot support technician.
The conditional here is correctly checking if the function exists, but it should probably also be checking the value of `opcache.restrict_api` before calling `opcache_invalidate()`.
If changing the code isn’t an option, the `php.ini` file can be changed to turn this feature on.
Being a SP user, it is probably in your interest to get this changed.

LukeTowers:
I haven't run into this issue on my sites personally, even those I use ServerPilot for. What are you doing to trigger it? Additionally, perhaps we should add a check similar to https://github.com/websharks/comet-cache-pro/pull/295/files#diff-200db4a444463a4f331e7412898f3addR357 instead so that we can detect if it will throw an error instead of just suppressing it.

danhrrn:
@LukeTowers I literally deployed a brand new DO server, installed SP and added an October app. This error gets thrown.

I will work on a check that can be run. I am thinking...
- During installation, there is an extra check that takes place. If this option is turned off, they can choose to suppress the issue if they want. Sound okay?

LukeTowers:
@danhrrn there already is an option for this: https://github.com/octobercms/october/blob/master/config/cms.php#L348

danhrrn:
Eugh that doesn't fix the issue either.

xariketi:
Hi! I also have the same issue and i also use Serverpilot. I notice that i get this error only on php 7.2. If i change php to 7.1 i don't get error message.

danhrrn:
@xariketi you can look into changing this value to blank in your php.ini file, that works

LukeTowers:
@xariketi see https://github.com/octobercms/october/pull/3438#issuecomment-372037977 as for a suggestion for what we can do to fix this.

PR #3440, opened 2018-03-10
Fix incorrect usage of the count() function

mxschmitt:
Fixes the error: `ErrorException: count(): Parameter must be an array or an object that implements Countable in /var/www/superkorb/modules/system/controllers/requestlogs/_referer_field.htm:1`
In PHP 7.2 there was a change, which won't allow anymore to give null into the function. So we need to check it before, if the object is an array which is countable.

//  Edit: This error occurs, when you gonna view the request log and check for the referrers in a PHP 7.2 installation.

### DIFF
--- a/modules/system/controllers/requestlogs/_referer_field.htm
+++ b/modules/system/controllers/requestlogs/_referer_field.htm
@@ -1,4 +1,4 @@
-<?php if (count($formModel->referer) > 0): ?>
+<?php if (is_array($formModel->referer)?count($formModel->referer):0 > 0): ?>
     <div class="form-control control-simplelist with-icons">
         <ul>
             <?php foreach ((array) $formModel->referer as $referer): ?>


LukeTowers:
This seems like it could be a rather large breaking change across all uses of count(), depending on how it was utilized. Would you be willing to grep the code base for all uses and implement appropriate fixes?

mxschmitt:
@LukeTowers sorry, I'm not good in using grep and know the codebase for october not very well.

LukeTowers:
@mxschmitt in your local OctoberCMS instance you can run the following command to get a list of files and line numbers where `count` is used. Run this command while in both `/vendor/october` and `/modules`:
`grep -Erno "count\(" .`

LukeTowers:
@mxschmitt is this still an issue?

LukeTowers:
Implemented in https://github.com/octobercms/october/commit/4abd74febc916a4e38fe81ff7d3e8e7993259b53

zukinaru:
Thanks a lot

Issue #3382, opened 2018-02-05
Add support for more than one nested folder in the CMS file browser


SiyuQian:
##### Expected behavior
Display the whole folder directory tree in the admin panel

##### Actual behavior
The backend view:
![image](https://user-images.githubusercontent.com/24470452/35805071-4dfc1e5a-0adf-11e8-97c2-55fdde8a4411.png)

My folder structure:
![image](https://user-images.githubusercontent.com/24470452/35805112-77ef8fe4-0adf-11e8-85ff-4736a21bd9f1.png)

##### Discussion
I am not sure did anyone found this issue yet. 
Maybe you did this on purpose?
I found this could make me hard to manage my twig templates, I know that you can name the file like xxx-xxx-xxx.htm but I feel this way is quite ugly. 
 
##### October build
431

munxar:
afaik, you can just have one level of subdirs. so pillar folder is ok, slider is one level to deep.
this is documented somewhere but i couldn't find it right now.

LukeTowers:
@SiyuQian @munxar http://octobercms.com/docs/cms/themes#subdirectories. Personally I'm not opposed if someone wanted to make a PR to add support for infinite subdirectories.

SiyuQian:
Hey Luke,
Thank you for the documentation link. BTW, Luke can you help me to confirm that October Team is agreed on merge the 'add support for infinite subdirectories' to the project. (You know, I just wanted my codes more meaningful) If it is OK, I might do this when I finished all my paid projects.
Cheers,
Siyu Qian

LukeTowers:
@SiyuQian I personally am not opposed to it; and as long as the implementation and user experience can retain simplicity at the core of it all then I don't think any of the other core members will mind either. Be prepared for the potential of lots of review comments on your PR though :)

munxar:
@LukeTowers @SiyuQian I have good news: By just setting this attribute to null, i could use any depth from code.
https://github.com/octobercms/library/blob/4d1418ec5a86c86d4761b8cea00e68bcd8ce523e/src/Halcyon/Model.php#L88
Basically every file based model (static pages, email templates, ...) uses this, the change should be in every subclass. But the backend didn't reflect this change (work needed here).
Now the hard part: UX.
many, many, many CMSs failed with a horrible unusable file tree representations and I could imagine *why* the core team decided to not go deeper than 2 levels.
On the other hand, there are good UX examples for a deep nested folder structure:
- material design (google drive)
[...]

PR #3441, opened 2018-03-11
Add support for more than one nested folder in the CMS file browser #3382

SiyuQian:
#3382  **Add support for more than one nested folder in the CMS file browser**
This is a quick implementation of supporting max two levels of subdirectories for ONLY partials.
TODO: Investigate a good UI/UX example to implement infinite levels of subdirectories without making backend as a mess

To October Team,

I suggest that we can support max two levels of subdirectories only in cms type: partial.
I have another solution that we implement the Partials' UI as the Assets' UI does, so which supports infinite nested folder.
Please advise me what to do next. 
Also, could you please explain that how the $maxNesting works, I have overridden the variable in modules/cms/classes/Partial.php but I still not be able to extract the items with two or more nested folder and files in the Template file without changing the vendor/october/rain/src/Halcyon/Datasource/FileDatasource.php which I think is not robust enough.

Cheers,
Siyu Qian

### DIFF
--- a/modules/cms/classes/Partial.php
+++ b/modules/cms/classes/Partial.php
@@ -9,6 +9,12 @@
 class Partial extends CmsCompoundObject
 {
     /**
+     * From david: I dont think this variable $maxNesting is still working
+     * TODO: Make sure that we use this variable to control the max nesting level which allows us to have different max nesting possiablities in the different cms type(pages, partials ...)
+     */
+    protected $maxNesting = null;
+
+    /**
      * @var string The container name associated with the model, eg: pages.
      */
     protected $dirName = 'partials';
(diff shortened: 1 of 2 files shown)


LukeTowers:
@kenng we need to figure out the best UX for this change and then also probably change Halcyon\Datasource to support it

LukeTowers:
@SiyuQian @kenng is there still interest in getting this added?

SiyuQian:
Hey Luke,
For sure, did you guys get any ideas about the UX/UI part?

Regards,
Siyu Qian

LukeTowers:
@SiyuQian can we use the existing UX for the asset folders UI?

LukeTowers:
@SiyuQian that's fine. The Halcyon implementation is currently undergoing some major work right now to support the database as a datastore so that would need to be completed first anyways.

PR #3446, opened 2018-03-12
Added trigger change event

scorewinner:
Enable an MLMediaFinder implementation from @munxar without copy pasting the whole js code.


munxar:
@LukeTowers note, this is my co-worker helping me on https://github.com/rainlab/translate-plugin/pull/325. so this PR would be needed for a better implementation.

PR #3452, opened 2018-03-15
Update ImportModel.php

mcharytoniuk:
This PR ca be a backwards incompatible change in some cases, but substantially improves memory usage. Instead of loading all array upfront while importing it uses `fetch` (https://csv.thephpleague.com/8.0/reading/ ) callback feature that applies `processImportRow` function to each row one by one instead of loading the entire array into one variable. I experience some issues while trying to import 250k+ records set and this is the way for me to reduce the memory footprint drastically. 

This change will break the user code if the implementation relied on the fact that `processImportData` returned an array - now it's an `iterator` instance (it can be mitigated by adding `iterator_to_array` call in the userspace implementation). Typical usage should stay the same, as `foreach ($results as $result) {...}` still works, but the memory usage is much lower and it helps a lot with large datasets.

### DIFF
--- a/modules/backend/models/ImportModel.php
+++ b/modules/backend/models/ImportModel.php
@@ -141,13 +141,9 @@ abstract class ImportModel extends Model
             ));
         }
 
-        $result = [];
-        $contents = $reader->fetch();
-        foreach ($contents as $row) {
-            $result[] = $this->processImportRow($row, $matches);
-        }
-
-        return $result;
+        return $reader->fetch(function ($row) use ($matches) {
+            return $this->processImportRow($row, $matches);
+        });
     }
 
     /**


LukeTowers:
I'm not sure I understand the difference between this and your last PR: https://github.com/octobercms/october/pull/3450

mcharytoniuk:
It pushes this optimization further but in the contrary to the last PR, it introduces backwards incompatible change. In the last PR I removed the assignment to the `$contents` variable (now it holds the `iterator` instead of an array). In this PR I am removing the need of storing the complete results array in `$result` variable.

LukeTowers:

mcharytoniuk:
@LukeTowers  it optimized memory usage by pushing import results one by one instead of putting them all in an array upfront. It's more like a generator approach. I had issues with memory usage while importing large datasets and it was the fix.

LukeTowers:
@mcharytoniuk unfortunately we can't introduce a backward's incompatible change like this. Potentially this could be implement through the use of a config property on the importexport configuration though, any thoughts on that?

mcharytoniuk:
@LukeTowers it does not change the API. It's backward compatible, I made it like that because I needed compatibility with plugins and the rest of OctoberCMS, but maybe I can't see something.

LukeTowers:
> contrary to the last PR, it introduces backwards incompatible change. 

What did you mean by this then?

mcharytoniuk:
Ah, sorry, it was so long ago I contradicted myself, you are right. Return type of a function is an iterator instead of array to not load each result to the memory at the same time.

LukeTowers:
So what would you propose adding to October then? Knowing that we can't break backwards compatibility in that way.

PR #3453, opened 2018-03-15
Placeholder for TagList

GinoPane:
I've noticed that placeholder option is missing for TagList, so I've added it within this PR.

```
                    ...
                    'type'  => 'taglist',
                    'placeholder' => 'Add items...',
                ]
```

taglist_placeholder

Also small adjustments were made to a default style of select items.

Before:
taglist_before

After:
taglist_after

Alignments were fixed, removal cross is slightly bigger, text color was changed.

### DIFF
--- a/modules/backend/formwidgets/taglist/partials/_taglist.htm
+++ b/modules/backend/formwidgets/taglist/partials/_taglist.htm
@@ -7,8 +7,8 @@
     id="<?= $field->getId() ?>"
     name="<?= $field->getName() ?>[]"
     class="form-control custom-select <?= !count($fieldOptions) ? 'select-no-dropdown' : '' ?> select-hide-selected"
-    data-token-separators="<?= $customSeparators ?? ''; ?>"
-    data-placeholder="<?= !empty($placeholder) ? e(trans($placeholder)) : ''; ?>"
+    <?php if (!empty($customSeparators)): ?>data-token-separators="<?= $customSeparators ?>"<?php endif ?>
+    <?php if (!empty($placeholder)): ?>data-placeholder="<?= e(trans($placeholder)) ?>"<?php endif ?>
     multiple
     <?= $field->getAttributes() ?>>
     <?php foreach ($availableOptions as $key => $option): ?>
--- a/modules/backend/formwidgets/taglist/partials/_taglist.htm
+++ b/modules/backend/formwidgets/taglist/partials/_taglist.htm
@@ -7,8 +7,9 @@
     id="<?= $field->getId() ?>"
     name="<?= $field->getName() ?>[]"
     class="form-control custom-select <?= !count($fieldOptions) ? 'select-no-dropdown' : '' ?> select-hide-selected"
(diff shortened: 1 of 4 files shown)


amdad:
x at left is better because when you want delete multiple items you don't need to move cursor. You just click in one place as next tag in row fall into place. If x is at right each label have different width and you must move right/left to target x button. So it is just small UX trick.

GinoPane:
@amdad I understand your point but still I have to argue with it.

1. It is common and expected to find removal icon on the right for **left-to-right** languages. We read at first and then decide what to do with what we just have read. We can almost treat this like stable metaphor for such icons. Just look at this search [results ](https://goo.gl/NzBQJT):

image

By placing it somewhere else we'll break this metaphor and we'll add some cognitive overloads for users because they will have to deal with uncommon input.

And still, it will be completely fine in this case to place "X" icon on the left for **right-to-left** layout.

2.  The described beneficial situation seems artificial, it's not so often when you need to delete many consecutive tags at once. In reality you'll delete one and then read the next one while deciding if you need to delete it too. I don't see much benefit here actually.

It would be nice to find any researches (if there are some) about this.

amdad:
I fully agree from reading perspective x at the end is better. I was aware this when presenting UX view. Reason why someone place it before text by design. Better reading vs better clicking. No clear winner.  I'm not opposing here for anything. Just add one cent to discussion.

amdad:
One way to resolve dilemma between read and click (delete), is to make whole button clickable not just x. Then it is even easier to delete unwanted items, and x position doesn't matter for usability (multi removing).

munxar:
@amdad please don't. That is even more a non standard way, plus it makes selecting Text imposible.
On touch interfaces you acidentally click an item (e.g for scrolling) and instead remive it? Imho no good UX.
I'm full on @ginopane side.

daftspunk:
We should look for examples in other areas. The CMS tabs have the "X" on the right hand side, so they should remain on the right hand side. This positioning is the result of a debate held long ago, on this same topic.

- The "X" should be on the right hand side

GinoPane:
Hey @daftspunk! What's the status of this one?

GinoPane:
Hey @daftspunk! What is the status here?

modules/backend/formwidgets/taglist/partials/_taglist.htm
```diff
@@ -7,7 +7,8 @@
     id="<?= $field->getId() ?>"
     name="<?= $field->getName() ?>[]"
     class="form-control custom-select <?= !count($fieldOptions) ? 'select-no-dropdown' : '' ?> select-hide-selected"
-    <?php if ($customSeparators): ?>data-token-separators="<?= $customSeparators ?>"<?php endif ?>
+    data-token-separators="<?= $customSeparators ?? ''; ?>"
```
LukeTowers: Please leave this line the way that it was, doing it this way leaves an empty attribute on the generated HTML that I would prefer to avoid. @GinoPane

GinoPane: @LukeTowers ok, no problem, done and applied the same for placeholder attribute

PR #3455, opened 2018-03-16
Reflect Updated Laravel 5.5+ default changes

Teranode:


config/session.php
```diff
@@ -16,7 +16,7 @@
     |
     */
 
-    'driver' => 'file',
+    'driver' => env('SESSION_DRIVER', 'file'),
```
LukeTowers: env() changes aren't going to be merged because October doesn't use the .env file by default, users need to run `php artisan october:env` to get it.

Teranode:
For now hold off on approving this, seems its redundunt until php 7.3

ayumihamsaki:
@Teranode @LukeTowers Can I recommend you set the default to `strict` it makes no sense to have `null` which is off and `lax` that doesn't fully protect the websites and is not recommend to be set.

(When PHP 7.3 comes out)

> With this flag set the browser will provide some protection against things like CSRF and XSSI, but not full protection, it should only be used as a defense in depth measure. For cross-site requests that use an unsafe HTTP method, like POST, the browser will not attach the cookie. That is, if Facebook set the SameSite=Lax flag on their cookies, and this page issued a POST request to facebook.com, then the browser would not attach your cookie for facebook.com to the request.

Taken from Scott's website found here: https://scotthelme.co.uk/tough-cookies/

LukeTowers:
@ayumihamsaki see #3644 instead

LukeTowers:
Closed because #3644 was merged instead.

LukeTowers:
Correction: #3688

Issue #3445, opened 2018-03-12
Folder issues with S3 media uploads


shortformblog:
##### Expected behavior
I'm using the Drivers Plugin to manage my images through Amazon S3. I would like to use the main directory of my bucket to manage those images, which would help as I'm trying to use an existing base of images I have managed via imgix.

##### Actual behavior
However, two things happen that prevent me from doing so:

1. Selecting the root of the bucket fails to add the initial slash after the domain/bucket name, so the link shows up broken (e.g. http://mybucket.comimage.jpg instead of http://mybucket.com/image.jpg). However, the image does appear to upload. Setting a non-root folder solves the issue, but I would like to use the root folder for this.

2. I cannot open up some of my existing folders in the Media interface, so as a result, I cannot access my cache of images which is currently organized in a /year/month/ format. However, a folder I created in October works fine. I believe this could possibly be a permissions issue in my bucket but I checked AWS and no permissions were set on the folders.

I did not find a prior Github Issue about these problems, but they seem to have been mentioned on the October forums in the past: https://octobercms.com/forum/post/amazon-s3-andor-cdn-image-linking-issue

##### Reproduce steps
Set Amazon S3 for media uploads, put the "folder" in cms.php at the root of the bucket. Putting in nothing throws out an error.

```
 'media' => [
  'disk'   => 's3',
  'folder' => '/',
  'path'   => 'https://mybucket.com/',
],
```

##### October build
(October build number)

Build 434

rajakhoury:
@shortformblog  I reproduced the steps you mentioned. 
1) yes the Media item url is missing a slash between the bucket name and file path 
2) I created folders manually and through the backend Media Library but there's is not problem in accessing those folders/files from the backend just the urls are missing the slash..

shortformblog:
@rajakhoury Yeah, I'm admittedly not 100 percent sure that folder issue isn't on my end—though when I checked there didn't seem to be anything on the folder preventing its use. But that said, it might be helpful, as a possible addition to that, for October's media folder to offer more feedback to the end user as to why a folder won't open or a file won't function. Currently it does nothing and it would be useful for troubleshooting to state if there's a problem.

I'm glad to have confirmation on the first issue being a problem, though, as that was the thing that clearly seemed like a bug to me.

rajakhoury:
I am sharing the steps I did and let's see if others can confirm. 

### Default Test ###
1. Using default config

        'media' => [
            'disk'   => 's3',
            'folder' => 'media',
[...]

PR #3461, opened 2018-03-20
Fix S3 Media Library Item Url

rajakhoury:
Fix for #3445 the Public Url of S3 Media items in the backend when the storage folder is set as a bucket Root folder. 
The File path was missing a slash between the bucket name & file path.


LukeTowers:
Thanks for looking into this @rajakhoury!

rajakhoury:
@LukeTowers Thanks to you for always helping out.

PR #3463, opened 2018-03-21
Belarusian Map for JS Slugify

GinoPane:
Added a missing character from Belarusian alphabet and also fixed mapping for one Russian character.


GinoPane:
client.php was committed with no content changes, reason for commit - CRLF/LF differences detected by IDE.

Issue #282, opened 2014-06-12
[Bug] Blog - error loading image


bestxp:
1. press create new post
2. put `![0](image)`
3. choose image and start upload
4. error

```
Error uploading file "day_of_russia.jpg". SQLSTATE[42S22]: Column not found: 1054 Unknown column 'public' in 'field list' (SQL: insert into `system_files` (`public`, `file_name`, `file_size`, `content_type`, `disk_name`, `updated_at`, `created_at`) values (1, day_of_russia.jpg, 91986, image/jpeg, 539946373bdb2608636191.jpg, 2014-06-12 06:18:31, 2014-06-12 06:18:31))
```

PR #3465, opened 2018-03-21
Set a trigger condition to be a field from the parent form for a field in a repeater

MartynasJanu:
Repeater fields can now be triggered by fields from the parent form. Just add '@' before field name. Example:

```
    type:
        label: Type
        span: auto
        options:
            numbers: Numbers
            texts: Texts
        type: dropdown
    type_data:
        label: Type data
        span: auto
        type: repeater
        form:
            fields:
                number:
                    label: Number
                    span: auto
                    type: number
                    trigger:
                        action: show
                        field: '@type'
                        condition: 'value[number]'
                text:
                    label: Text
                    span: auto
                    type: text
                    trigger:
                        action: show
                        field: '@type'
                        condition: 'value[text]'
```

### DIFF
--- a/modules/backend/classes/FormField.php
+++ b/modules/backend/classes/FormField.php
@@ -19,6 +19,11 @@ class FormField
     const NO_SAVE_DATA = -1;
 
     /**
+     * @var string A special character in yaml config files to indicate a field higher in hierarchy
+     */
+    const HIERARCHY_UP = '^';
+
+    /**
      * @var string Form field name.
      */
     public $fieldName;
@@ -461,6 +466,12 @@ class FormField
         $triggerField = array_get($this->trigger, 'field');
         $triggerCondition = array_get($this->trigger, 'condition');
 
+        // Check for parent field reference for the trigger condition
+        $triggerFieldParentLevel = $this->getNumberOfPrecedingSymbols($triggerField, self::HIERARCHY_UP);
+        if ($triggerFieldParentLevel > 0) {
+            $triggerField = substr($triggerField, $triggerFieldParentLevel);
+        }
+
         // Apply these to container
         if (in_array($triggerAction, ['hide', 'show']) && $position != 'container') {
             return $attributes;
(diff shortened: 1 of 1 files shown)


modules/backend/classes/FormField.php
```diff
@@ -461,6 +461,12 @@ protected function filterTriggerAttributes($attributes, $position = 'field')
         $triggerField = array_get($this->trigger, 'field');
         $triggerCondition = array_get($this->trigger, 'condition');
 
+        $triggerFieldGlobal = false;
```
LukeTowers: Why do you set this flag up here separate from where you actually use it instead of right by where it gets used?

MartynasJanu: Mainly because I prefer setting variables at the beginning of the function but also because I want to get rid of the '@' in the ```$triggerField``` as soon as possible, since later there may be code added which may fail if '@' is present. Not sure if it makes much of a difference though in terms of readability: all ```$trigger*``` variables in this method are set at the beginning but used only later.

LukeTowers: Fair enough. Did you go through any other flags to indicate it's a global field other than `@`? As a side thought, how difficult do you think it would be to target fields that are a specified level above the current field? I.e. repeater in a repeater, wants to target a field on the parent repeater? Would that be feasible?

MartynasJanu: Good idea. Just implemented and tests setting a number of '@' symbols to indicate how high in the hierarchy to go. For example: ```@type``` would indicate immediate parent, ```@@type``` would indicate parent's parent, etc. All works fine, so I could upgrade, however, do You have any better ideas for the syntax? I'm rather new to October, so not much on general conventions and syntax "habbits" here. If multiple '@' symbols are fine, then I'll amend

LukeTowers: I don't necessarily mind the `@` symbol, but I ask because it's used by component partials to indicate to the parser to use an overridable partial path: http://octobercms.com/docs/plugin/components#component-partials

Other symbols used include `~` and `$` in some file path parsing areas where they mean application path and plugin directory path respectively.

All that to say I'm not against it, just wondering if there's a better option out there. Perhaps a caret? `^`?

MartynasJanu: I like the caret ```^```. It kinda points "up" and also has a similar meaning in git syntax which is normally quite familiar to developers. Ok, thanks for the tip, I will amend the PR

MartynasJanu:
Updated for direct parent level referencing and switched to using caret ```^``` symbol

LukeTowers:
@MartynasJanu awesome! Could you update the documentation [here](http://octobercms.com/docs/backend/forms#field-trigger-events) to provide instructions on the caret symbol and how it works?

MartynasJanu:
@LukeTowers done [#282](https://github.com/octobercms/docs/pull/282)

PR #3477, opened 2018-03-28
Added a config flag for request cache. See octobercms/library pull 308.

Josniii:
This commit adds a config flag in `config/cache.php` to enable/disable the request cache being added in [this commit](https://github.com/octobercms/library/pull/308) to the core library, as discussed there with @LukeTowers 

Please let me know if you think something should be worded differently!

### DIFF
--- a/config/cache.php
+++ b/config/cache.php
@@ -78,4 +78,17 @@ return [
 
     'prefix' => 'october',
 
+    /*
+    |--------------------------------------------------------------------------
+    | Disable Request Cache
+    |--------------------------------------------------------------------------
+    |
+    | The request cache remembers cache retrievals from your cache store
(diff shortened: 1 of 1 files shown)


config/cache.php
```diff
@@ -78,4 +78,17 @@
 
     'prefix' => 'october',
 
+    /*
+    |--------------------------------------------------------------------------
+    | Disable Request Cache
+    |--------------------------------------------------------------------------
+    |
+    | The request cache remembers cache retrievals from your cache store
```
LukeTowers: Change `remembers` to `stores`, `your` to `the`, and replace the last line with `Set to true to disable this in memory request cache` please. Thanks!

Josniii: Done 👍

PR #3479, opened 2018-03-30
Fix mail.beforeAddContent listener to use the $plain argument as well

hlev:
`System\Serviceprovider` adds a listener to `mail.beforeAddContent` but was not passing the `$plain` argument provided by the dispatched event.
See https://github.com/octobercms/library/blob/9d3c73e1d4e6cc562dea6b930bcb4e63d2b64d3b/src/Mail/Mailer.php#L335

This resulted in that when plain-text only emails are attempted with `[ 'text' => 'template.code' ]`, `template.code` became `null` in the handler and could not be found. As a consequence `lluminate/View/FileViewFinder` trying to find it on the filesystem where it did not exist, because mail templates/views/layouts may be stored in the database alone.

`InvalidArgumentException: View [template.code] not found.`

### DIFF
--- a/modules/system/ServiceProvider.php
+++ b/modules/system/ServiceProvider.php
@@ -330,7 +330,8 @@ class ServiceProvider extends ModuleServiceProvider
          */
         Event::listen('mailer.beforeAddContent', function ($mailer, $message, $view, $data, $raw, $plain) {
             $method = $raw === null ? 'addContentToMailer' : 'addRawContentToMailer';
-            return !MailManager::instance()->$method($message, $raw ?: $view ?: $plain, $data, $view === null);
+            $plainOnly = $view === null; // When "plain-text only" email is sent, $view is null, this sets the flag appropriately
+            return !MailManager::instance()->$method($message, $raw ?: $view ?: $plain, $data, $plainOnly);
         });
     }
 
--- a/modules/system/ServiceProvider.php
+++ b/modules/system/ServiceProvider.php
@@ -330,7 +330,7 @@ class ServiceProvider extends ModuleServiceProvider
          */
         Event::listen('mailer.beforeAddContent', function ($mailer, $message, $view, $data, $raw, $plain) {
             $method = $raw === null ? 'addContentToMailer' : 'addRawContentToMailer';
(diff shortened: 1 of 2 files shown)


hlev:
I was trigger happy. This only seems to fix the part of the issue that causes the exception. The email is still sent in both formats. Templates cannot be saved in the backend as plain-text only, so at least now the template is found, but the email is still sent both in HTML and plain-text.

modules/system/classes/MailManager.php
```diff
@@ -98,7 +98,7 @@ public function addContentToMailer($message, $code, $data)
      * Internal method used to share logic between `addRawContentToMailer` and `addContentToMailer`
      * @return void
      */
-    protected function addContentToMailerInternal($message, $template, $data)
+    protected function addContentToMailerInternal($message, $template, $data, $plainOnly)
```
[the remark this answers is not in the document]
hlev: Thanks @LukeTowers, fixed that and updated a handful of docblocks.

hlev:
@LukeTowers https://github.com/octobercms/october/pull/3479/commits/293845e7b3086dbcedc9385cd0df286fc54c7f9c fixes the issues and addresses review comments as well. Still good to go.

modules/system/ServiceProvider.php
```diff
@@ -328,9 +328,9 @@ protected function registerMailer()
         /*
          * Override standard Mailer content with template
          */
-        Event::listen('mailer.beforeAddContent', function ($mailer, $message, $view, $data, $raw) {
+        Event::listen('mailer.beforeAddContent', function ($mailer, $message, $view, $data, $raw, $plain) {
             $method = $raw === null ? 'addContentToMailer' : 'addRawContentToMailer';
-            return !MailManager::instance()->$method($message, $raw ?: $view, $data);
+            return !MailManager::instance()->$method($message, $raw ?: $view ?: $plain, $data, $view === null);
```
LukeTowers: Can you add a comment in here to explain `$view === null`'s purpose?

hlev: Sure @LukeTowers, pushed a new commit.

PR #3491, opened 2018-04-09
Add key generation command

juliangums:
It would be very useful if you could add the `october:key` command I've created to have an easy way of re-generating the key in the config file. Laravels `key:generate` is not useful as it writes it to `.env` file.
Command is tested and can just be merged.

### DIFF
--- a/modules/system/ServiceProvider.php
+++ b/modules/system/ServiceProvider.php
@@ -235,6 +235,7 @@ class ServiceProvider extends ModuleServiceProvider
         $this->registerConsoleCommand('october.fresh', 'System\Console\OctoberFresh');
         $this->registerConsoleCommand('october.env', 'System\Console\OctoberEnv');
         $this->registerConsoleCommand('october.install', 'System\Console\OctoberInstall');
+        $this->registerConsoleCommand('october.key', 'System\Console\OctoberKey');
 
         $this->registerConsoleCommand('plugin.install', 'System\Console\PluginInstall');
         $this->registerConsoleCommand('plugin.remove', 'System\Console\PluginRemove');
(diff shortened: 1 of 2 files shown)


LukeTowers:
This is unnecessary as `key:generate` in October already does this. As a side note, you really should be using a `.env` file in the first place.

LukeTowers:
See https://github.com/octobercms/library/blob/develop/src/Foundation/Console/KeyGenerateCommand.php

juliangums:
yeah, when there's an `.env` file present it'll work. no worries

LukeTowers:
Just so you're aware, the `key:generate` command in October does not touch the .env file; only the config/app.php file directly. This is a decision by @daftspunk. If you need to update the key in the .env file you have to run `key:generate` with the `--show` flag and copy the generated key to the .env file directly.

juliangums:
Yeah, I want to change the config value in `config/app.php` so the logic actually works perfectly. However, it still fails, if there is no `.env` file present.
```
file_get_contents(/home/vagrant/Code/gta/.env): failed to open stream: No such file or directory
```
There's no apparent reason for me why it should fail if you're not writing to it anyway.
But if this is expected behaviour, that's good. If not, I am happy to look into why it's throwing the error and make a new pull request with the solution. Let me know

juliangums:
Got the latest release of October and I also just compared both files. Both the same. It rather seems to fail on the Laravel command class it is extending in this line `https://github.com/laravel/framework/blob/5.5/src/Illuminate/Foundation/Console/KeyGenerateCommand.php#L96`

juliangums:
@LukeTowers thanks! Wasn't that file, but when I comparend the Laravel class it's extending I figured out my `laravel/framework` was out of date. Mondays...

LukeTowers:
Specifically the one on the develop branch :)

PR #3494, opened 2018-04-10
Add humans.txt file to october:mirror command

haleksandre:
Add `humans.txt` to be mirrored by the `october:mirror` command.


LukeTowers:
Protip: you can also add your own custom paths here: http://octobercms.com/docs/api/system/console/mirror/extendpaths

PR #3522, opened 2018-04-21
Add alignment and size options in loader indicator

wellyfrs:
Add option to horizontally align the loader indicator in the center
and option to change the size (only small size available for now)
through the JavaScript API. Also, add comments to explain the use of
the new options.

Reference used: https://octobercms.com/docs/ui/loader

### DIFF
--- a/modules/system/assets/ui/js/loader.base.js
+++ b/modules/system/assets/ui/js/loader.base.js
@@ -12,6 +12,8 @@
  * JavaScript API:
  *
  * $('#buttons').loadIndicator({ text: 'Saving...', opaque: true }) - display the indicator in a solid (opaque) state
+ * $('#buttons').loadIndicator({ centered: true }) - display the indicator aligned in the center horizontally
+ * $('#buttons').loadIndicator({ size: small }) - display the indicator in small size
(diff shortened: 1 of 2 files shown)


LukeTowers:
Where are the styles for this? Also could you add screenshots demoing this?

wellyfrs:
The styles are in [modules/system/assets/ui/less/loader.less](https://github.com/octobercms/october/blob/master/modules/system/assets/ui/less/loader.less) and you can see them in the cited reference: https://octobercms.com/docs/ui/loader. What I did was just make styles available through JavaScript API options, like this:

`$('#buttons').loadIndicator({ text: 'Saving...', opaque: true, size: 'small', centered: true })`

I also made it possible to add these options as data attributes: `data-load-indicator-centered` and `data-load-indicator-size="small"`. For example:

```
<div class="loading-indicator-container">
    <button class="btn btn-primary"
            data-request="onSave"
            data-load-indicator="Saving..."
            data-load-indicator-opaque
            data-load-indicator-centered
            data-load-indicator-size="small">
        Save
    </button>
</div>
```

LukeTowers:
Ah, I see @wellyfrs, I should have clicked the link in the first place :) Thanks!

Issue #3067, opened 2017-08-28
Nested repeater groups do not include group indexes.


CptMeatball:
**simplified repeater_groups.yaml**

    html:
      fields:
          title:
              type: text
    repeater:
      fields:
        repeater:
            type: repeater
            groups: $/path/to/group/elements-group.yaml

##### Expected behavior
When nesting a repeater that includes itself by a group definition, the child repeater loads and renders the correct repeater fieldss.

Data stored In DB:

    [{
		"title": "Test1",
		"_group": "title"
	},
	{
		"repeater": {
		       {
				"title": "Test2",
		                "_group": "title"

			}
		},
		"_group": "repeater"
	}]

##### Actual behavior
When using a repeater that includes itself with the same group,. It does not store the group index, and therefore does not know how the data should be rendered.

stored data in DB:

    [{
		"title": "Test1",
		"_group": "title"
	},
	{
		"repeater": {
			"3": {
				"title": "Test2"
			}
		},
		"_group": "repeater"
	}]

##### Reproduce steps
0. Install October and create plugin that uses a repeater with a repeater group.
1. In said group, define a group item that is a repeater that loads the same group. (includes itself).
2. Enter data in the create/update view of the controller + save
3. Reload page and see that the repeater does see that there are items; but does not know which data and how to render.

##### October build
419

LukeTowers:
This is a complex case. 

In order to aid in our understanding of the problem, please submit [a Pull Request to the test plugin](https://github.com/daftspunk/oc-test-plugin) for investigation that demonstrates the issue. Don't forget to include step by step instructions how how to replicate it _using the test plugin_.

Narkoleptika:
@LukeTowers 

I can confirm this bug. I created a pull request to the oc-test-plugin with all relevant information: daftspunk/oc-test-plugin#50

==

In an unrelated note:

@daftspunk 
I owe you an apology for being a dick the last time I was on this repo: https://github.com/octobercms/october/issues/2722#issuecomment-294069118

If it's any consolation, I copied the wrong quote for the point I was trying to make. Which made me look like even more a a D-bag and brought myself sufficient shame for seppuku.

I'm sorry

LukeTowers:
@Narkoleptika Could you dig deeper into the issue and see if you can find a fix?

Narkoleptika:
@LukeTowers 

Here ya go! #3523

LukeTowers:
[...]

PR #3523, opened 2018-04-22
Fix for missing _group field when saving data from a nested group repeater

Narkoleptika:
**Repro and details (screenshots, expected, actual, etc)**: daftspunk/oc-test-plugin#50
**Closes**: #3067

This fixes a bug where the `_group` field isn't saved with nested group repeater forms. This causes problems for subsequent renders of the backend form with the nested group repeater.


LukeTowers:
Thanks @Narkoleptika, excellent job providing the bug report and reproduction instructions! Looking forwards to any future contributions from you :)

PR #3524, opened 2018-04-23
Add searchOnEnter option to make the Search widget fire on enter key

manogi:
Adding an option to change from the "search on every keystroke" functionality to one where the search is only fired when hitting the enter key. This makes sense when having larger datasets or CMS users who are slow typers.

### DIFF
--- a/modules/backend/widgets/Search.php
+++ b/modules/backend/widgets/Search.php
@@ -1,4 +1,6 @@
-<?php namespace Backend\Widgets;
+<?php
+
+namespace Backend\Widgets;
 
 use Lang;
 use Backend\Classes\WidgetBase;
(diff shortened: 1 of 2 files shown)


modules/backend/widgets/Search.php
```diff
@@ -1,4 +1,6 @@
-<?php namespace Backend\Widgets;
+<?php
```
LukeTowers: Leave this formatting the way it was please, that's the standard across all of October's files.

LukeTowers:
Carried over from https://github.com/octobercms/october/pull/2703#discussion_r183301699

LukeTowers:
Please add this to the documentation as well, then I can merge this

manogi:
Added it to the docs: octobercms/docs#290

PR #3533, opened 2018-04-29
Add the option to resize images on upload

adsa95:
This pull request adds the option to automatically resize and/or change the quality of images that are uploaded to the media library.

While processing the image is stored in the server temp directory. This is because the resizer requires a local path to the file it's working with.

If the user has not chosen to resize or change the quality of uploaded images, it's done in the same way it was before.

I think this is an important feature to be able to save space in a way that is transparent to CMS users. An alternative would be to add some events so that a plugin can be used for resizing.


config/cms.php
```diff
@@ -252,6 +252,15 @@
     | folder - a folder prefix for storing all generated files inside.
     | path   - the public path relative to the application base URL,
     |          or you can specify a full URL path.
+    |
+    | For the 'media' resource you can also specify:
+    |
+    | imageMaxWidth     - Resize all media manager image uploads to be
+    |                     within this with (in pixels, 0 = ignore)
```
LukeTowers: `within this with` typo, should be `within this width`

LukeTowers:
@adsa95 does the https://octobercms.com/docs/api/media/file/upload event work to implement this feature in a plugin?

LukeTowers:
Let's go with that approach instead for now then. Thanks for your work on this though, we may revisit it at some point in the future!

PR #3537, opened 2018-05-02
Overwrite SameName Image Upload Using MediaManager

detrojaashish:
Right Now In OctoberCMS Images we upload using MediaManager, the same name Images are replaced every time when upload the image. To resolve that that issue I have modified the MediaManager.php file. Using that I have resolved my issue But the one other problem arise with this when composer runs file is reverted to original version. Above I have done a change to resolve that issue If it is right than I request you to please update your file to resolve it permanently.


LukeTowers:
This submission as is wouldn't be acceptable, we'd need a more robust solution that asks the user if they want to override the file or add it when uploading files with the same name. @detrojaashish are you interested in implementing that?

LukeTowers:

PR #3540, opened 2018-05-04
[Discussion] Left align number inputs in backend forms

tobias-kuendig:
Currently `number` is the only backend form field type that is right aligned.

I see the reason why number inputs might be right aliged, but in my opinion this makes backend forms look inconsistent. It also makes it harder to see what fields contain data, especially on bigger screens where the form fields can become very wide.

grafik

I suggest to left align number inputs to get more consistency and better UX in backend forms.

grafik

### DIFF
--- a/modules/system/assets/ui/less/form.less
+++ b/modules/system/assets/ui/less/form.less
@@ -162,10 +162,6 @@
         padding-bottom: 5px;
     }
 
-    &.number-field {
-        > .form-control { text-align: right; }
-    }
-
     &.radio-align {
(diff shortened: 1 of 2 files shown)


pvullioud:
It make sense in some cases but not in other.
OC allow you to add custom CSS for the backend on the settings page. That is the easiest way to choose the best alignment for your project

Image
source: https://ux.stackexchange.com/questions/81865/why-arent-html5-typed-number-inputs-right-aligned-and-should-they-be

PR #3549, opened 2018-05-15
Add extensibility to auth restore view (like signin view)

wellyfrs:
In the backend login view, a view event is fired which is useful for adding some content to this page, for example. This pull request adds the same event type to the password recovery view.


wellyfrs:
The backend login view: [https://github.com/octobercms/october/blob/master/modules/backend/controllers/auth/signin.htm](https://github.com/octobercms/october/blob/master/modules/backend/controllers/auth/signin.htm)

PR #3555, opened 2018-05-21
Singularize placeholders for automatically generated titles

GinoPane:
Titles for manage-list and manage-pivot modals are automatically generated. In most cases model names are plural whilst titles expect singular nouns. Currently we end up with such titles by default:

october_add_a_new_posts

This PR is to fix such titles.

It would be also nice to title-case all titles, but title_case function does not behave correctly (for example, it title-cased articles for me).

### DIFF
--- a/modules/backend/behaviors/relationcontroller/partials/_manage_list.htm
+++ b/modules/backend/behaviors/relationcontroller/partials/_manage_list.htm
@@ -2,9 +2,11 @@
     <?= Form::open() ?>
         <div class="modal-header">
             <button type="button" class="close" data-dismiss="popup">&times;</button>
-            <h4 class="modal-title"><?= e(trans($relationManageTitle, [
-                'name' => trans($relationLabel)
-            ])) ?></h4>
+            <h4 class="modal-title">
+                <?= e(trans($relationManageTitle, [
+                    'name' => trans(str_singular($relationLabel))
+                ])) ?>
+            </h4>
         </div>
 
         <div class="list-flush">
--- a/modules/backend/behaviors/relationcontroller/partials/_manage_pivot.htm
+++ b/modules/backend/behaviors/relationcontroller/partials/_manage_pivot.htm
@@ -2,7 +2,11 @@
     <?= Form::open() ?>
         <div class="modal-header">
             <button type="button" class="close" data-dismiss="popup">&times;</button>
-            <h4 class="modal-title"><?= e(trans($relationManageTitle, ['name'=>trans($relationLabel)])) ?></h4>
+            <h4 class="modal-title">
+                <?= e(trans($relationManageTitle, [
+                    'name' => trans(str_singular($relationLabel))
+                ])) ?>
+            </h4>
         </div>
         <?php if (!$relationSearchWidget): ?>
             <div class="modal-body">


LukeTowers:
Doesn't this make it so that even if a developer manually specified the plural version they wouldn't be able to have it anymore?

GinoPane:
It only effects on the model name. If the user specifies a custom title it won't be affected.

LukeTowers:
@w20k could you test this?

w20k:
Hi @LukeTowers doesn’t do what was intended, because right now what ‘RelationController’ does it sets either label or fieldName -> `$this->vars['relationLabel'] = $this->config->label ?: $this->field;`. So, anyone could specify their singular object name the way they want inside the relation_config.yaml -> field:label.

LukeTowers:
I'm not sure I understand that @w20k, could you provide an example of what you're talking about?

w20k:
@LukeTowers  to make it simple you cannot override popup title with your own plural label, even if you want to (for some reason).

Example:
In my relation config, I have a Comments field with label: 'My Comments'.
It will be converted from: `Add a new My Comments` to `Add a new My Comment`

LukeTowers:
@w20k can you post an example of the config where that would happen? I need to see it in context to understand what you're talking about.

w20k:
Sure-sure, @LukeTowers 

Example without change (Test-Plugin::Posts):

**Set label to '1 Comment'**

_Config_
screenshot_2018-12-03_at_11_00_40

_Posts with comments popup_
screenshot_2018-12-03_at_11_11_00


**Set label to '1 Comments'**

_Config_
screenshot_2018-12-03_at_11_02_26

_Posts with comments popup_
screenshot_2018-12-03_at_11_05_21


Example with PR applied (Test-Plugin::Posts):

**Set label to '1 Comments'**

_You don't have an option to override. While without this fix you could make it singular if needed. Not for popup only, though. I don't see a point._
screenshot_2018-12-03_at_11_13_29

LukeTowers:
What ideas do you have to still provide users the ability to override this behaviour?

w20k:
To use `str_singular` only when label ain't set. 
For me, as I understand OctoberCMS logic, might be wrong, the label is a custom user-specific field, which should have the power to override 'field' name.

LukeTowers:
That idea looks good to me. @GinoPane could you implement @w20k's suggestion?

GinoPane:
Sorry guys, I was on vacation.
I'll take a look. Thanks!

GinoPane:
Ok, I've finally checked this and I don't think I really agree..

Let me explain. 

If you don't set label for relation you'd get an exception:

image

That's why we'll always have a label and we'll never rich the singularization part of
`$this->vars['relationLabel'] = $this->config->label ?: str_singular($this->field);`
and the change won't make any sense.

And I really think if you set plural label it should get singularized, that was the intention of the PR. And the source of the issue actually this:
`'add_a_new' => 'Add a new :name',`. 

So we either need to **singularize** everything which comes as label to the title (just to match the grammar) **or** provide another PR which will detect if label is plural or singular and provide a relevant localisation template for both, but it seems more difficult to implement.

w20k:
That's not correct...it worked for sure, i'll double check.

LukeTowers:
@w20k any updates on this?

w20k:
@LukeTowers I think we can close this one. 
I don't agree on the solution provided by @GinoPane where we'd `singularize everything which comes as a label to the title`, that's definitely a wrong solution here. We need another solution (PR) which will detect if a label is plural or singular.

LukeTowers:
I don't really care if it singularizes everything, my only concern is making sure that the developer can override it

w20k:
With current proposal won't be possible.

GinoPane:
I'm ok with the ability to override and singularize if we don't have the label provided, the problem is that I got exception when I tried the solution provided by @w20k (see the comment https://github.com/octobercms/october/pull/3555#issuecomment-452393822). If you could provide more clear suggestion maybe, I'd try to apply it.

w20k:
@GinoPane it's not related to my fix, but the fact it's required to have `label` when using relation.
`RelationController: protected $requiredRelationProperties = ['label'];`

Samuell1:
I think better will be to allow user to change modal, button names because you have soo many language and you cant solve it for every language with one function.

Samuell1:
Looks like for buttons there is already PR: https://github.com/octobercms/october/pull/4444

DanHarrin:
@Samuell1 @w20k @LukeTowers @GinoPane so would a fix like #4444 be applicable for this case as well? I can PR it if needed.

Samuell1:
@DanHarrin yes, that will be best option to allow developer customize titly fully.

w20k:
Will be closing this PR in favor of #4444.

PR #3558, opened 2018-05-22
Added filter option to list relations

### LINKED ISSUE #2587: No option to use filter in relation manager
##### Expected behavior
Option to use filter in relation manager.

##### Actual behavior
No option to use a filter in the relation manager.

##### Reproduce steps
-

##### October build
Latest build

SaifurRahmanMohsin:
List-type relations can now have filters just as normal lists can.
The filter events for extensibility has not been added to this.
This PR also resolves #2587

### DIFF
--- a/modules/backend/behaviors/RelationController.php
+++ b/modules/backend/behaviors/RelationController.php
@@ -62,6 +62,11 @@ class RelationController extends ControllerBehavior
     protected $viewWidget;
 
     /**
+     * @var Backend\Classes\WidgetBase Reference to the filter widget object.
+     */
+    protected $filterWidget;
+
+    /**
      * @var Backend\Classes\WidgetBase Reference to the widget used for relation management.
      */
     protected $manageWidget;
@@ -256,6 +261,7 @@ class RelationController extends ControllerBehavior
         $this->vars['relationToolbarButtons'] = $this->toolbarButtons;
         $this->vars['relationViewMode'] = $this->viewMode;
         $this->vars['relationViewWidget'] = $this->viewWidget;
+        $this->vars['relationFilterWidget'] = $this->filterWidget;
         $this->vars['relationViewModel'] = $this->viewModel;
         $this->vars['relationPivotWidget'] = $this->pivotWidget;
         $this->vars['relationSessionKey'] = $this->relationGetSessionKey();
@@ -620,6 +626,7 @@ class RelationController extends ControllerBehavior
             $config->recordsPerPage = $this->getConfig('view[recordsPerPage]');
             $config->showCheckboxes = $this->getConfig('view[showCheckboxes]', !$this->readOnly);
             $config->recordUrl = $this->getConfig('view[recordUrl]', null);
+            $config->filter = $this->getConfig('view[filter]', null);
 
             $defaultOnClick = sprintf(
                 "$.oc.relationBehavior.clickViewListRecord(':%s', '%s', '%s')",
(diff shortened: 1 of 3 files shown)


modules/backend/behaviors/RelationController.php
```diff
@@ -713,6 +720,26 @@ protected function makeViewWidget()
                     }
                 }
             }
+
+            /*
+             * Prepare the filter widget (optional)
+             */
+            if ($this->getConfig('view[filter]')) {
```
LukeTowers: Could you switch this to `if ($config->filter)` since that is actually the variable getting used later?

LukeTowers:
@SaifurRahmanMohsin are you able to add the filter events for extensibility to this using the same approach as the events for the list and form widgets? Also, in order to merge this it will need to be added to the documentation (https://github.com/octobercms/docs). 

Looks pretty good so far though, your contribution is much appreciated!

modules/backend/behaviors/RelationController.php
SaifurRahmanMohsin: Good catch. Yes, give me a minute and I'll update the PR.

SaifurRahmanMohsin:
I haven't tried adding the filter extensibility yet since I would need to test it with some data. Whereas the code I committed here has been tested with the oc-crm-plugin (I used the contacts relation to filter based on company name) and it works fine. I'll make a new PR should I add extensibility or you're welcome to add it as well.

LukeTowers:
@SaifurRahmanMohsin I would prefer to include the extensibility as part of this. You could use the Test plugin to test and submit a PR to the test plugin with your tests of the extensibility. Test plugin is available at https://github.com/octoberrain/oc-test-plugin

LukeTowers:
@SaifurRahmanMohsin okay. If you can create a PR to demonstrate the relation filters working in the test plugin then I'll be able to merge this.

Farrow:
I was about to look at introducing this functionality into the relation manager myself as I need to be able to filter based on an additional relation, but I found your pull request.

Unfortunately I can't get it to work for me, though perhaps I'm doing something wrong. When I click on the filter I get an error popup saying "A widget with class name 'relationSubsectorsViewListFilter' has not been bound to the controller on line 513 of [...]/modules/backend/Classes/Controller.php"

Have I missed some config setting somewhere maybe?

SaifurRahmanMohsin:
@Farrow I had the same issue until I changed `data-request="listFilter::onFilterUpdate"` to `data-request="<?= $this->getEventHandler('onFilterUpdate') ?>"` so now the filter works perfectly well with text-type fields which is what I mainly needed for my project.

I didn't check the other filter types until now and I see the problem lies with the fact that with the other filter types the handler is called in javascript, specifically [here](https://github.com/octobercms/october/blob/master/modules/system/assets/ui/js/filter.js#L345-L346) as `$form.request(this.options.updateHandler`. The problem is that this $form object does not have access to the underlying controller and therefore the widgets which are already bound are not visible and that's why that error is thrown. Frankly, I'm not certain how to overcome the issue so I'm fiddling around with the code to find a solution. If you can figure it out, feel free to comment out the fixes or make a PR.

LukeTowers:
@SaifurRahmanMohsin are you still interested in this? @w20k is this something you could help with?

SaifurRahmanMohsin:
Yeah, of course, I am. But I'm kind of stuck with work for the past several weeks and haven't been able to do any other open source work at the moment. Help would be appreciated.

w20k:
@SaifurRahmanMohsin, If you could describe in details how to reproduce it or have an example? Would appreciate it!

LukeTowers:
Will be replaced by https://github.com/octobercms/october/pull/4241

PR #3567, opened 2018-05-30
on('keyup') changed to on('input'). resolves #1270

### LINKED ISSUE #1270: October uses onkeyup instead of oninput
October CMS uses `onkeyup` event to trigger functions related to user input, like slug generation. There's better way of handling it. `oninput` event gives immediate effect and therefore provides much smoother experience. The only drawback is that IE8 doesn't support it, so if you want to comfort this browser, a fallback must be used.

viamage:
I was browsing issues and found quite easily solvable #1270, which remains open for years, is confirmed but doesn't have its PR (@danhrrn who was writing fixes seems to be gone from github). I've replaced on('keyup' with on('input' within the source as discussed in issue thread.

All events working fine or better:
http://uploads.keios.eu/video/october-oninput.html

Not quite sure if anything more is expected from PR as long as testing goes.

---

Below @tschallacka explanation, as requested by Luke Towers
 
> From my personal experience on a different project, oninput is best to use because it catches everything. Also when something is pasted in it without using the keyboard.
> It also prevents race conditions where the "order of events" is going bottoms up.
> for example, if you wish to catch when someone holds a key down and do things, you need oninput because onkeyup will only trigger when the key is released
> 
> See https://jsfiddle.net/70nk1b5x/
> Just keep a key pressed in the fields. The oninput updates on every entry.

```
$('#keyup').on('keyup',function() {
    $('#output_keyup').text(this.value.length);
});
$('#oninput').on('input',function() {
    $('#output_oninput').text(this.value.length);
});
<input type="text" id="keyup">

   onkeyup characters typed <span id="output_keyup"></span><BR/>
<input type="text" id="oninput">

oninput characters typed <span id="output_oninput"></span>
```


LukeTowers:
Thanks @viamage! Good job on getting it done :)

PR #3575, opened 2018-06-01
Cms module refactor

SeriousKen:
This is a refactor of the CMS module to move the Twig environment creation into the container avoiding the repetition of creation in the `CmsCompoundObject` and the `Controller`. It removes the controller dependency on the `DebugExtension` which didn't seem necessary and it removes it from the `Cms\Twig\Extension` constructor, which gets it instead from the Twig context.

It shouldn't change any behaviour but this does fix an issue with the Pages plugin where the lack of `DebugExtension` caused [an issue](https://github.com/rainlab/pages-plugin/issues/337)


modules/cms/twig/ComponentNode.php
```diff
@@ -34,7 +34,7 @@ public function compile(Twig_Compiler $compiler)
         }
 
         $compiler
-            ->write("echo \$this->env->getExtension('Cms\Twig\Extension')->componentFunction(")
+            ->write("echo \$this->env->getExtension('Cms\Twig\Extension')->componentFunction(\$context,")
```
LukeTowers: What's the initial backslash on $context for?

SeriousKen: Escapes the $ in the double-quoted string, otherwise PHP tries to do variable expansion.

LukeTowers: I gotcha. Wouldn't it be simpler to reverse the order of the single and double quotes then? Or am I missing something else?

SeriousKen: I just added the $context variable to what was there already. If you swap them wouldn't you then need to double escape the backslashes anyway?

LukeTowers: Fair point, that would be the case.

LukeTowers:
@w20k could you review this?

modules/cms/ServiceProvider.php
```diff
@@ -55,6 +63,49 @@ public function boot()
...
+            return new TwigLoader;
+        });
+
+        App::singleton('cms.twig.environment', function ($app) {
+            $loader = $app->make('cms.twig.loader');
+
+            $useCache = !Config::get('cms.twigNoCache');
+            $isDebugMode = Config::get('app.debug', false);
+            $strictVariables = Config::get('cms.enableTwigStrictVariables', false);
+            $strictVariables = is_null($strictVariables) ? $isDebugMode : $strictVariables;
```
w20k: Was there a reason to switch from the short version (`$strictVariables = $strictVariables ?? $isDebugMode;`) to is_null verification?

LukeTowers: @SeriousKen ^

SeriousKen: @w20k to be honest I'm not sure why, but looking at that code `$strictVariables` is never `null`; it's either `true` or `false`, so neither line of code works as I imagine it was intended. Surely it should be `$strictVariables = $strictVariables ?: $isDebugMode`?

w20k: @SeriousKen, could you change that part and also, if you don't mind add few Unit tests?

bennothommo:
Closing - this PR appears to be abandoned.

Issue #3571, opened 2018-05-31
Minor Bug: Saving... Appears over buttons etc.


ayumihamsaki:
I am sure this is not correct.

Everytime I click on the save button in the cms a grey text says "Saving..." 

See below screenshot:

![untitled](https://user-images.githubusercontent.com/17784082/40791335-2aece182-64ef-11e8-90cd-7f1877ca2218.png)

It's like an overkill and I suggest removing it. Having the spinning icon is more than enough! 

So I suggest removing the word "Saving..." in grey that is very hard to read and just looks not very nice. Or at least don't have it overlapping the button. (Either remove it or place it under the buttons)

I see this in both Latest Google Canary and Firefox nightly.

Would it not be better to have it like this:

![untitled](https://user-images.githubusercontent.com/17784082/40794692-d3458e76-64f7-11e8-926e-2856854a8c3b.png)

~October seems to use very old CSS and no flexbox or css grids etc.

LukeTowers:
@ayumihamsaki I've seen this myself too recently, this is due to a change that was made in the CSS for the loading indicator that removed it's background colour because it didn't make sense in a different context. We just need to re-add the background colour, ideally only targeting the button context. The idea behind having it appear over the buttons with a solid background colour is to prevent the user from clicking on any of the other buttons while an action is occurring.

LukeTowers:
Found it: https://github.com/octobercms/october/pull/3394. @ayumihamsaki could you investigate a styling change that would only target the button group loading indicator to apply the `body-bg` background colour?

ayumihamsaki:
@LukeTowers Thanks for the extra info, now makes sense why I have seen this for last few months. Will have a think on some styling, also I think OctoberCMS has very poor CSS, for example it would most likely fail WCAG 2.0 (AA) spec. The light grey background behind a sightly darker grey icon surely must fail and is hard to view! Therefore most likely there would need to be quite a lot of work done to Octobercms CSS to correct many errors.

I feel this fix is not straight forward as October uses many things: saving, updating, creating, deleting etc. all in many places. Will have a think on it, but don't expect a quick solution from my side on this one lol.

LukeTowers:
@ayumihamsaki thanks for looking into this! Is there anyway we can retain the old behaviour of putting the loading indicator directly overtop of the buttons with a solid background colour?

Also, I appreciate the sense of urgency but it's not really that urgent :)
Low: Should be taken care of eventually, whenever someone gets around to caring about it
[...]

PR #3576, opened 2018-06-01
Fix Issue with Minor Bug: Saving... Appears over buttons etc.

ayumihamsaki:
See Issue: [3571](https://github.com/octobercms/october/issues/3571)


modules/system/assets/ui/less/loader.less
```diff
@@ -12,6 +12,10 @@
 @color-stripe-loader:                        @brand-accent;
 @loader-image-path:                          @image-path;
 
+.form-buttons .loading-indicator {
```
LukeTowers: Minor quibble, but please put more specific rules lower in the file. So in this case put this below the `.loading-indicator` rule group

ayumihamsaki: @LukeTowers done :+1:

LukeTowers:
@ayumihamsaki awesome, thanks! This'll take effect the next time I compile the assets and commit them with `php artisan october:util compile assets`

PR #3577, opened 2018-06-02
Soft components support

### LINKED ISSUE #1431: Add support for conditional components using `@component`
Now you can only specify what plugins your theme depends on. It doesn't make sense for theme authors to bake in support for every popular plugin if it pulls in all those dependencies as well. However, a theme might have support for it but not require it. This would need a mechanism for themes to check if a plugin is activated and then include code.

This would also be a huge benefit to the theme marketplace since the users can search for themes that supports RainLab.Blog, for example.

viamage:
This is a proposal to resolve #1431, making it a PR per LukeTowers request.

### DIFF
--- a/modules/cms/classes/Controller.php
+++ b/modules/cms/classes/Controller.php
@@ -1237,31 +1237,55 @@ class Controller
      * @param array  $properties  Component properties
      * @param bool   $addToLayout Add to layout, instead of page
      * @return ComponentBase Component object
+     * @throws CmsException
+     * @throws SystemException
      */
     public function addComponent($name, $alias, $properties, $addToLayout = false)
     {
         $manager = ComponentManager::instance();
 
-        if ($addToLayout) {
-            if (!$componentObj = $manager->makeComponent($name, $this->layoutObj, $properties)) {
-                throw new CmsException(Lang::get('cms::lang.component.not_found', ['name'=>$name]));
+        if($isSoftComponent = $this->isSoftComponent($name)){
+            $name = $this->parseComponentLabel($name);
+            $alias = $this->parseComponentLabel($alias);
+        }
+        try {
+            if ($addToLayout) {
+                if (!$componentObj = $manager->makeComponent($name, $this->layoutObj, $properties)) {
+                    throw new CmsException(Lang::get('cms::lang.component.not_found', ['name' => $name]));
+                }
+
+                $componentObj->alias = $alias;
+                $this->vars[$alias] = $this->layout->components[$alias] = $componentObj;
+            } else {
+                if (!$componentObj = $manager->makeComponent($name, $this->pageObj, $properties)) {
+                    throw new CmsException(Lang::get('cms::lang.component.not_found', ['name' => $name]));
+                }
+
+                $componentObj->alias = $alias;
+                $this->vars[$alias] = $this->page->components[$alias] = $componentObj;
             }
 
-            $componentObj->alias = $alias;
-            $this->vars[$alias] = $this->layout->components[$alias] = $componentObj;
-        }
-        else {
-            if (!$componentObj = $manager->makeComponent($name, $this->pageObj, $properties)) {
-                throw new CmsException(Lang::get('cms::lang.component.not_found', ['name'=>$name]));
+            $this->setComponentPropertiesFromParams($componentObj);
+            $componentObj->init();
+
+            return $componentObj;
+            /**
+             * We catch two exceptions below, as missing plugin can throw both and catch \Exception would be too global.
+             * We also debug-log it, to not be TOO silent about this.
+             */
+        } catch(CmsException $e){
+            \Log::debug('Missing component! '. $e->getMessage() . ' '. $e->getTraceAsString());
+            if(!$isSoftComponent){
+                throw $e;
             }
-
-            $componentObj->alias = $alias;
-            $this->vars[$alias] = $this->page->components[$alias] = $componentObj;
+        } catch(SystemException $e){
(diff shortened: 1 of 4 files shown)


viamage:
So first thing to discuss here would be if try-catch approach is best here.

Second would be logging - I'm quite sure we should inform website admin that there is something wrong in theme (and lacking component that is called in theme is something wrong in theme, right?) - but if we log it or not depends on what this enhancement use cases are. I personally would keep it, but I understand reasons to possibly remove it.

LukeTowers:
Soft components don't need to be logged when they're missing, they're explicitly defined as optional so them not being present isn't a problem. Could you make a PR to the docs to document this feature and remove the logging?

Samuell1:
@viamage Hey, any news on this?

modules/cms/classes/Controller.php
```diff
@@ -1237,31 +1237,55 @@ public function param($name, $default = null)
...
         $manager = ComponentManager::instance();
 
-        if ($addToLayout) {
-            if (!$componentObj = $manager->makeComponent($name, $this->layoutObj, $properties)) {
-                throw new CmsException(Lang::get('cms::lang.component.not_found', ['name'=>$name]));
-            }
-
-            $componentObj->alias = $alias;
-            $this->vars[$alias] = $this->layout->components[$alias] = $componentObj;
+        if($isSoftComponent = $this->isSoftComponent($name)){
```
LukeTowers: space after if please `if (`

modules/cms/classes/Controller.php
```diff
@@ -1237,31 +1237,55 @@ public function param($name, $default = null)
...
+            $this->setComponentPropertiesFromParams($componentObj);
+            $componentObj->init();
+
+            return $componentObj;
+            /**
+             * We catch two exceptions below, as missing plugin can throw both and catch \Exception would be too global.
+             * We also debug-log it, to not be TOO silent about this.
+             */
+        } catch(CmsException $e){
+            \Log::debug('Missing component! '. $e->getMessage() . ' '. $e->getTraceAsString());
```
LukeTowers: remove log please

modules/cms/classes/Controller.php
```diff
@@ -1237,31 +1237,55 @@ public function param($name, $default = null)
...
-            $this->vars[$alias] = $this->page->components[$alias] = $componentObj;
+            $this->setComponentPropertiesFromParams($componentObj);
+            $componentObj->init();
+
+            return $componentObj;
+            /**
+             * We catch two exceptions below, as missing plugin can throw both and catch \Exception would be too global.
+             * We also debug-log it, to not be TOO silent about this.
+             */
+        } catch(CmsException $e){
```
LukeTowers: space after `catch` please (`catch (`)

modules/cms/classes/Controller.php
```diff
@@ -1237,31 +1237,55 @@ public function param($name, $default = null)
...
+            \Log::debug('Missing component! '. $e->getMessage() . ' '. $e->getTraceAsString());
+            if(!$isSoftComponent){
```
LukeTowers: space after if please `if (`

modules/cms/classes/Controller.php
```diff
@@ -1237,31 +1237,55 @@ public function param($name, $default = null)
...
+                throw $e;
+            }
+        } catch(SystemException $e){
```
LukeTowers: space after catch please (`catch (`)

modules/cms/classes/Controller.php
```diff
@@ -1237,31 +1237,55 @@ public function param($name, $default = null)
...
+            \Log::debug('Missing component! '. $e->getMessage() . ' '. $e->getTraceAsString());
```
LukeTowers: remove log please

modules/cms/classes/Controller.php
```diff
@@ -1237,31 +1237,55 @@ public function param($name, $default = null)
...
+            if(!$isSoftComponent){
```
LukeTowers: space after if please `if (`

modules/cms/classes/Controller.php
```diff
@@ -1411,4 +1435,25 @@ protected function verifyCsrfToken()
             $token
         );
     }
+
+    /**
+     * Removes '@' from soft component name
+     * @param string $label
+     * @return string
+     */
+    private function parseComponentLabel($label){
```
LukeTowers: protected instead of private please

modules/cms/classes/Controller.php
```diff
@@ -1411,4 +1435,25 @@ protected function verifyCsrfToken()
...
+        if($this->isSoftComponent($label)){
```
LukeTowers: space after if

modules/cms/classes/Controller.php
```diff
@@ -1411,4 +1435,25 @@ protected function verifyCsrfToken()
...
+            return str_replace('@','',$label);
```
LukeTowers: space after arguments (`'@', '', $label`)

modules/cms/classes/Controller.php
```diff
@@ -1411,4 +1435,25 @@ protected function verifyCsrfToken()
...
+        }
+        return $label;
+    }
+
+    /**
+     * Checks if component name has @.
+     * @param string $label
+     * @return bool
+     */
+    private function isSoftComponent($label){
```
LukeTowers: protected instead of private please

DanHarrin:
Want these changes completed as a new PR?

w20k:
@DanHarrin would be nice! 👍 You could add credits to @viamage.

w20k:
Closing this PR in favor of #4539

Issue #3532, opened 2018-04-28
Update PhpUnit version in composer.json


Ognj3n:
The latest stable version of PhpUnit is 7.*. 

However installation source code requires 5.7 PhpUnit version, which can make debugging of PhpUnit tests pretty messy for first timers in October CMS. 

Will be glad to see this updated.

LukeTowers:
@Ognj3n feel free to make a PR to octobercms/october and octobercms/library and any RainLab plugins using unit testing to upgrade the version.

viamage:
Latest PhpUnit is 7.x, but php-selenium requires php-unit <7, so 6.5 I guess. 

I wonder if we should wait for https://github.com/giorgiosironi/phpunit-selenium/issues/422 or upgrade to 6.x for now?

w20k:
Was upgraded by @bennothommo to 6.5 version (https://github.com/octobercms/october/blob/master/composer.json#L46)

bennothommo:
Upgraded to PHPUnit 8 as part of https://github.com/octobercms/october/pull/4893.

Issue #3360, opened 2018-01-16
Update phpunit dependency from ~5.7 to ~6.0 to match Laravel 5.5


jofrysutanto:
##### Expected behavior
Writing tests in OctoberCMS is expected to support Laravel TestCase helpers.

##### Actual behavior
Some of Laravel TestCase helper traits such as `assertDatabaseHas()` breaks due to missing expected PHPUnit class dependency:

```
Fatal error: Class 'PHPUnit\Framework\Constraint\Constraint' not found in /acme-project/vendor/laravel/framework/src/Illuminate/Foundation/Testing/Constraints/SoftDeletedInDatabase.php on line 8
```

##### Reproduce steps
Clean installation of October build 420+ (Laravel 5.5)
Create new unit tests that utilises Laravel `InteractsWithDatabase` test helper trait.

##### October build
420+

Updating composer phpunit dependency to ~6.0 seem to fix the issue, which matches Laravel 5.5 dependency: https://github.com/laravel/laravel/blob/v5.5.28/composer.json#L17

LukeTowers:
OctoberCMS uses PHP Unit 6.5 as of https://github.com/octobercms/october/commit/6fb6e5f71ca74d8cad62b4e5816f9ef30192a954#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780

PR #3051, opened 2017-08-22
Replace old selenium functional testing for the new Laravel Dusk

arzola:
Hi OctoBros! 👊

I have updated the functional tests to meet the new Laravel dusk features and be able to test the CMS features and also provide to the users a powerful and easy way to test their plugins.

I added a `dusk` folder inside the config directory to be able to config the default service provider used by dusk and be able to easily integrate in any continuous integration platform.

I think this is a huge advantage to avoid the selenium nightmare, and provide to the users a super easy and natural functional testing.

The user only have to setup the enviroment to `APP_ENV=dusk` and done

image

PR #3580, opened 2018-06-04
MediaLibrary test updates

viamage:
Updates for depreciated methods and empty tests. This is phpunit 7.0.1 runned on php7.2 before this patch:

https://gist.github.com/viamage/2ac3e80412bcbf15d0c2fdd2d2c41869

And this is after:

https://gist.github.com/viamage/3b27e4b31075f9969ff309ebbfaeaa6e

setExpectedException is depraciated since phpunit 5.x and can be safely replaced, the other is just additional assertion that removes the test from risky list and adds additional coverage for checking if result is indeed string (should be done with typehint, but if no typehint let there be test at least). 

Two remaining "risky" tests are from October/Rain library, not spammy so i'll leave them be for now.

### DIFF
--- a/tests/unit/system/classes/MediaLibraryTest.php
+++ b/tests/unit/system/classes/MediaLibraryTest.php
@@ -41,8 +41,9 @@ class MediaLibraryTest extends TestCase // @codingStandardsIgnoreLine
      */
     public function testInvalidPathsOnValidatePath($path)
     {
-        $this->setExpectedException('ApplicationException');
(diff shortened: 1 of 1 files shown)


LukeTowers:
Awesome, thanks @viamage! Perhaps you can look at https://github.com/octobercms/october/issues/3532, https://github.com/octobercms/october/issues/3360, and https://github.com/octobercms/october/pull/3051?

PR #3583, opened 2018-06-06
Handle unsortable columns before passing them to the DB #3517

### LINKED ISSUE #3517: Cannot prevent backend SQL errors appear on screen
##### Expected behavior

SQL (or any other detailed) errors should be logged to an error file but should not appear on screen, since security scan tools treats them as information leak. Or at least we should be able to enable/disable this feature. 

##### Actual behavior

SQL error pops up on screen (this happens during security scans with modified requests)

SQLSTATE[HY000]: General error: 1366 Incorrect integer value: 'test' for column 'published' at row 1 (SQL: update `table` set `published` = test, `updated_at` = 2018-04-19 14:09:45 where `id` = 2)

image

This happens even if `APP_DEBUG` is `false` 

##### Reproduce steps

Modify an update request using Burp/Intruder or similar tool, change or remove some posted values to an invalid values.

##### October build

v1.0.434

semiherdogan:
Fixes #3517

### DIFF
--- a/modules/backend/widgets/Lists.php
+++ b/modules/backend/widgets/Lists.php
@@ -17,6 +17,8 @@ use Backend\Classes\WidgetBase;
 use October\Rain\Database\Model;
 use ApplicationException;
 use DateTime;
+use Config;
+use Flash;
 
 /**
  * List Widget
@@ -1299,6 +1301,12 @@ class Lists extends WidgetBase
     public function onSort()
     {
         if ($column = post('sortColumn')) {
+
+            $debug = Config::get('app.debug', false);
+            if (!$debug && !$this->isSortable($column)) {
+                Flash::error('Column not sortable.');
+                return;
+            }
 
             /*
              * Toggle the sort direction and set the sorting column


LukeTowers:
The appropriate way to handle this is to throw an exception, not do the error displaying at this level.

endo64:
@LukeTowers Without this change, the error is already displaying in a SweetAlert box with all the SQL details and perceived as a security breach by security tools like Burp. Please see my comment here: https://github.com/octobercms/october/issues/3517#issuecomment-383132876

LukeTowers:
@semiherdogan You're missing my point. Flashing an error message to the user directly is a bad practice because you can't actually handle that error correctly in other cases. If we implement this solution as is, then we won't be able to catch this error occurring in error logs, and we won't be able to handle this as a part of a more robust error message display solution (catching exceptions and displaying sanitized messages when debug mode is disabled). 

A much more maintainable solution would be this:
```php
public function onSort()
{
     if ($column = post('sortColumn')) {
          if (!$this->isSortable($column)) {
              throw new ApplicationException("The $column column is not marked sortable. Define sortable: true on $column if it is able to be sorted.");
          }

          // rest of method
     }
}
```

This allows us to catch this exception further up the stack in order to have a more flexible approach to handling the exception; while also having the default behavior of displaying a flashed error message with the contents of the exception message.

LukeTowers:

endo64:
@LukeTowers In your example security tools will complain about a possible XSS because if `$column` is changed to `<script>alert(1);<script>` it will be shown in the response.

Even the output is sanitized and correctly escaped before displaying, security tools will still complain because they'll think they put a malicious content in request and it appeared on the response.

Again, if we return `ApplicationException`, HTTP response code will be 500 and security tools will complain about it for same reason as above.

How about silently ignoring this error and continue without sorting? Because this error should never happen unless the request is altered.

semiherdogan:
@LukeTowers since we trying to validate the column whether is sortable or not,  how about throwing ValidationException instead of ApplicationException, or before `Flash::error()` adding `\Log::error("The $column column is not marked sortable.")`.

LukeTowers:
@endo64 I agree, sanitization is required in this case, my example was just that, an example. And @semiherdogan your argument about the `ValidationException` has merit too, so it's worth discussing the best option. 

I will say that this (or perhaps a very similar) error can and does happen without the request being altered, if you have a column that cannot be sorted by in the DB (i.e. a column that pulls from a model attribute) but you haven't specified that it can't be with `sortable: false` then it can trigger the SQL Exception.

> Even the output is sanitized and correctly escaped before displaying, security tools will still complain because they'll think they put a malicious content in request and it appeared on the response.

That sounds like a problem with your security tools, not with October or error handling in general.

In regards to the 500 response code, perhaps we should go with @semiherdogan's suggestion of a validationexception with a 400 response code.

I prefer going the exception route rather than the flash error message route because this is not something that occurs from a user not fulfilling the requirements for the request (i.e. not selecting any records and then clicking delete selected), this is something that will result from a malformed request or a improperly structured application.

LukeTowers:

Issue #3554, opened 2018-05-20
Bug with Components Close icon gets hidden in FireFox on touch screen


ayumihamsaki:
I have been using October CMS for years but only just clicked on this bug!

It happens when using Firefox and not in Google Canary so I am guessing it is a pseudo CSS error.

## The Bug

When you hover over the component icons the close icon in the top right will disappear in Firefox, seems to work ok in Chome.

~Maybe problem in Opera, as well as FireFox and Opera share this kind of issue...

(Also I should note I am using the latest version of OctoberCMS (v435) and Browsers)

## Show the problem in the wild

Below shows a screenshot of the problem in Firefox Nightly:

![1](https://user-images.githubusercontent.com/17784082/40280510-9b209d46-5c4c-11e8-9fff-b9f6696f4076.png)

Below shows a screenshot of it working fine in Google Canary:

![2](https://user-images.githubusercontent.com/17784082/40280511-9f2bd27a-5c4c-11e8-8228-23ad4913f92a.png)

## Issue

I had a quick check but I can't seem to find the cause of the conflict? The reason the close icon is hidden is that when you hover the component box it is not making a blue background and so the close button is the same color as the background on hover making it hidden.

LukeTowers:
@ayumihamsaki are you able to determine the difference between Firefox and Chrome that's causing Firefox to not pickup the hover style for the components?

ayumihamsaki:
@LukeTowers Hi I think I have found the problem, but I honestly don't fully understand the CSS code used in October CMS.

The issue is in the following file found here:

https://github.com/octobercms/october/blob/master/modules/cms/assets/css/october.components.css

The hover code part is this:

```
.draggable-component-item:hover,
.component-list .components div.layout-cell:hover,
div.control-componentlist div.components div.layout-cell:hover {
  background: #3498db;
}
```

But it is not working because it seems to be overwritten by this piece of code:

```
.touch div.control-componentlist div.components div.layout-cell > div:hover,
.touch div.control-componentlist div.components div.layout-cell > div:active,
.touch div.control-componentlist div.components div.layout-cell > div:active:focus {
  background: #ffffff;
  color: #475354;
}
```

Logically to me I don't understand why you have a CSS hover and background color white for that section?

The simple solution that works is to do this in the CSS:

```
/*
.touch div.control-componentlist div.components div.layout-cell > div:hover,
.touch div.control-componentlist div.components div.layout-cell > div:active,
[...]

PR #3585, opened 2018-06-07
Fix for css issue regarding component close button

Teranode:
References https://github.com/octobercms/october/issues/3554


LukeTowers:
@Teranode this change needs to be done in the source LESS files: https://github.com/octobercms/october/blob/master/modules/cms/assets/less/october.components.less#L245

LukeTowers:
@Teranode thanks! @ayumihamsaki could you verify this resolves the issue?

PR #3604, opened 2018-06-22
Fix Security issue in media manager

Teranode:
Currently there is a way to manipulate folder names using the request headers for the medafinder folder path. Since `MediaLibrary::validatePath()` does not validate folder names but rather only validates the existence of a folder as a parent directory or subdirectory. Using that manipulation, if the folder does not exist as a parent directory or subdirectory, the folder will be created using the path query.

We can utilize `$this->validateFileName()` to verify if the name of the folder path name being used for the folder name is valid during the folder creation function and the name applying function for the path query.

### DIFF
--- a/modules/backend/widgets/MediaManager.php
+++ b/modules/backend/widgets/MediaManager.php
@@ -119,7 +119,7 @@ class MediaManager extends WidgetBase
     public function onGoToFolder()
     {
         $path = Input::get('path');
-
+        
         if (Input::get('clearCache')) {
             MediaLibrary::instance()->resetCache();
         }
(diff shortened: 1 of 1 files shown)


LukeTowers:
Awesome @Teranode! Could you please add some tests for invalid vs valid paths?

LukeTowers:
@Teranode no worries, basically you would add to https://github.com/octobercms/october/blob/master/tests/unit/system/classes/MediaLibraryTest.php to test a valid path to make sure it passes validation and test an invalid path to make sure it fails validation. If you don't want to give it a try let me know and I will take care of it for you

daftspunk:
All good, I'll get it. Thanks @Teranode

Teranode:
Ill let u handle that :|

Issue #3471, opened 2018-03-26
All caches are flushed on backend login


mrbohnke:
All caches are flushed when you login to backend, right in the moment when you hit the submit button on the login form. Is this a feature? is it necessary? IDK if this is expected or not, we noticed it trying to discover why logins to the backend took sooo long, and thought maybe it was something in our plugin, but after disabling *all* plugins the behavior persists.

##### October build
434

mrbohnke:
Nobody? nothing? 💔

mrbohnke:
I'm just surprised about nobody else being afflicted by this problem since it is very evident, frustrating and we can't manage to understand how or why it is happening. I totally understand that time is scarce and the issues many, didn't want to look pushy.
Thank you very much for your time, this project is awesome and we will totally contribute.

LukeTowers:
@mrbohnke it all depends on what your cache situation is in relation to other people encountering the same issue.

If you're using a non-default option, you've already limited the number of people that might potentially encounter the issue to the number of people using that caching system and then it's further limited by people also operating on the same scale as you. In request caching of the cache system (pretty meta right?) was just implemented by someone who used an external cache server (Elasticache) https://github.com/octobercms/library/pull/308 for their caching system. It was an issue for them but not a large number of people had such a setup so it ultimately came down to them addressing their need and providing the changes to us in the form of a PR for everyone else to benefit from as well (with the added benefit of the burden of supporting that feature being passed to us instead of them).

As for this issue in particular, https://github.com/octobercms/october/issues/2659 is directly related. More specifically, this is the line of code that is causing you grief: https://github.com/octobercms/october/blob/master/modules/system/classes/UpdateManager.php#L159. See @daftspunk's comment in #2659 for the rational / explanation

mrbohnke:
Thank you VERY much, we actually saw this issue in the very beginnings of our research on the problem and, as soon as I read october:up, thought it wasn't related. Right now, reading more carefully, I see he actually says that it also accours on login —bad researcher!
We will continue from there and will update on any news if necessary.
Again, thank you and sorry for all the trouble.

petehalverson:
@mrbohnke You're not alone, I've run into this before as well and thought it was odd.

Would it make sense to introduce a config option and allow the automatic update check on login to be disabled?

LukeTowers:
@petehalverson I'd agree with that, feel free to make a PR

LukeTowers:
[...]

Issue #2659, opened 2017-02-06
Cache is cleared on october:up


toughdeveloper:
##### Expected behaviour
Values set using the Cache service (http://octobercms.com/docs/services/cache) to remain after `october:up` has run.

##### Actual behaviour
Cached is cleared completely. 

##### Reproduce steps
Set something in the Cache. Run `october:up` (or log into admin), Cached value will now be lost. 

##### October build
396

---

To note, this has become an issue as clients logging into admin have been clearing the cache. As it seems the `october:up` command is ran upon login. The cache is being reset several times a day and is difficult to rely upon. 

I assume the `Cache::flush` method is being run within october:up to avoid issues with updates, would it make more sense to simply use the ::forget call on the October cached values?

daftspunk:
Interesting, the cache needs to be cleared as part of the "update" process, perhaps not necessarily as part of the migration process (up). We'll take a look, thanks.

toughdeveloper:
Just wondered if there was any update on this issue ? Thanks

LukeTowers:
No update just yet, sorry.

The support demand for October is increasing, while the available time the founders have to give remains the same. These issues will be handled in time, however there are ways you can help:

- Submit a pull request that introduces this feature and wait for it to be reviewed
- [Post a bug bounty](https://www.bountysource.com/teams/october) for this specific issue so that more people can become involved with creating a fix
- [Make a financial contribution to the project](http://octobercms.com/fundraising) so we can allocate more time towards making improvements like these that benefit everyone

daftspunk:
We need to leave this in because there is a pretty decent flaw in Laravel's file based cache at the moment that is being counteracted by this issue's current functionality. In short the cache doesn't clean up after itself and it ends up being huge. We'll need to address the framework level issue first as it is blocker to this one.

LukeTowers:
Closing as it has been over a month since any activity on this occurred.

LukeTowers:
Also, has Laravel 5.5 resolved the issue with the file based cache @daftspunk? If so we could change this behaviour.

PR #3613, opened 2018-06-27
Only clear cache when updates are applied

FranciscoMateo:
By default, when a user make login process the UpdateManager::update() method is lauched and the cache always is erased. With this way I make me sure that caché is remove it when some plugin is updated.

Beside I added 'clearCacheOnUpdateManeger' option on config/cache.php to allow us enable or not the erase cache on UpdateManager::update() method.

### DIFF
--- a/config/cache.php
+++ b/config/cache.php
@@ -2,109 +2,109 @@
 
 return [
 
-	/*
-	|--------------------------------------------------------------------------
-	| Default Cache Store
-	|--------------------------------------------------------------------------
-	|
-	| This option controls the default cache connection that gets used while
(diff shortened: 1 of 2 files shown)


LukeTowers:
I don't see the need for a config variable for this if we're detecting whether or not updates have actually been applied before clearing the cache.

Thoughts?

LukeTowers:
@FranciscoMateo I don't understand what you mean by 
> Probably some pages have some big database and needs handle this kind of situation when their plugins are updated.

mrbohnke:
Shall this help with #2659 and #3471

LukeTowers:

Issue #3608, opened 2018-06-25
assets/js/framework.js forces the use of the 'unsafe-eval' Content-Security-Policy directive.


MeLlamoPablo:
##### Expected behavior
The OctoberCMS backend works with the following [Content-Security-Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP):

```
default-src 'none'; script-src 'self'; connect-src 'self'; img-src 'self'; style-src 'self';
```

##### Actual behavior
The OctoberCMS backend crashes because the code violates the previous Content-Security-Policy. Here is the error:

```
Uncaught Error: Error parsing the data-request-update attribute value. EvalError: Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "script-src 'self' 'unsafe-inline' data: connect.facebook.net sdk.accountkit.com www.google-analytics.com".
    at paramToObj (framework.js?v1:459)
    at n.fn.init.$.fn.request (framework.js?v1:426)
    at Function.$.request (framework.js?v1:437)
    at HTMLDocument.<anonymous> (backend:238)
    at j (jquery.min.js?v1:2)
    at Object.fireWith [as resolveWith] (jquery.min.js?v1:2)
    at Function.ready (jquery.min.js?v1:2)
    at HTMLDocument.I (jquery.min.js?v1:2)
```

The error is produced at exactly [this line in assets/js/framework.js](https://github.com/octobercms/october/blob/5ee8ee09170fa52cd6bb5dbb2416345aeea186e5/modules/system/assets/js/framework.js#L459).

This error doesn't occur when including 'unsafe-eval' in the CSP header, like so:

```
default-src 'none'; script-src 'self' 'unsafe-eval'; connect-src 'self'; img-src 'self'; style-src 'self';
```

This error occurs because `eval` is used to parse arbitrary JSON data from the client, I presume. I don't understand the reasoning for using `eval > stringify > parse`, as I believe that simply `parse` should be enough, and that would fix this issue.

##### Reproduce steps

1. Create a fresh OctoberCMS install.
2. Put it behind an NGINX reverse proxy.
3. Add the following directive to the NGINX configuration:
    ```
    add_header Content-Security-Policy  "default-src 'none'; script-src 'self'; connect-src 'self'; img-src 'self'; style-src 'self';" always;
    ```
4. Navigate to `https://my-site.test/backend`, and log in

##### October build
v1.0.435

[...]

PR #3622, opened 2018-07-02
Proof of concept: replace eval with JSON5.parse

MeLlamoPablo:
This PR replaces most usages of `eval` with `JSON5.parse`. This is only a proof of concept and shouldn't be merged. Ideally, the JSON5 library should be bundled with October and not pulled from `unpkg`.

I suggest we keep the discussion in #3608 for simplicity.


LukeTowers:

Issue #3619, opened 2018-06-29
Add config option cms.restrictBaseDir for relaxing file path restrictions


lthurston:
Per a discussion with @LukeTowers and @daftspunk in Slack, I'm going to add an undocumented config option (for use in dev envs only!) that allows realpath() as alternative path validation.

https://octobercms.slack.com/team/U712K9ZQ9

I'm referencing commit fb2aa1730cc08e8563908b54b70e942bd2988f5e and its removal of realpath() as the places where this change will be made.

I should be able to work on this the first week of July.

LukeTowers:
@lthurston on second thought you can make it documented, have it set to true by default with a strong warning against setting it to false.

lthurston:
@LukeTowers @daftspunk Can you put some eyes on this to confirm I'm doing this correctly? If it looks good, I'll PR it.

https://github.com/lthurston/october/commit/71c165bf34725defc591d251a54dcb79f542be9d

lthurston:
Err, this isn't quite right. I'll circle back tomorrow.

LukeTowers:
@lthurston let me know when it's ready to review. Feel free to submit it as a WIP PR while you work on it

PR #3626, opened 2018-07-03
Add config flag for allowing insecure paths (#3619)

lthurston:
In the testing that I was doing with a plugin that was symlinked, not all of these conditions needed to be modified in order for the plugin to "work" as it should. Which is another way of saying, I only verified the functionality of the one condition (the one in ViewMaker::makeFileContents()). Nevertheless, I assume there are situations similar to mine where these functions might receive paths that contain symlinks.

### DIFF
--- a/config/cms.php
+++ b/config/cms.php
@@ -362,4 +362,23 @@ return [
     
     'enableTwigStrictVariables' => false,
 
+    /*
+    |--------------------------------------------------------------------------
+    | Base Directory Restriction
+    |--------------------------------------------------------------------------
+    |
+    | Restricts loading backend template and config files to within the base
+    | directory of the application.
+    |
+    | WARNING: This should always be enabled for security reasons. However, in
+    | some cases you may need to disable this; for instance when developing
+    | plugins that are stored elsewhere in the filesystem for organizational
+    | reasons and then symlinked into the application plugins/ directory.
+    |
+    | NEVER have this disabled in production.
+    |
+    */
+
+    'restrictBaseDir' => true,
+
 ];
--- a/modules/system/traits/ConfigMaker.php
+++ b/modules/system/traits/ConfigMaker.php
@@ -7,6 +7,7 @@ use Event;
 use SystemException;
 use Backend\Classes\Controller;
 use stdClass;
+use Config;
 
 /**
  * Config Maker Trait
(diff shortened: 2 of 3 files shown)


modules/system/traits/ConfigMaker.php
```diff
@@ -1,5 +1,6 @@
 <?php namespace System\Traits;
 
+use Illuminate\Support\Facades\Config;
```
LukeTowers: why not just `use Config;`?

modules/system/traits/ViewMaker.php
```diff
@@ -4,6 +4,7 @@
 use Lang;
 use Event;
 use Block;
+use October\Rain\Support\Facades\Config;
```
LukeTowers: why not just `use Config;`?

modules/system/traits/ViewMaker.php
```diff
@@ -221,7 +224,10 @@ public function getViewPath($fileName, $viewPath = null)
      */
     public function makeFileContents($filePath, $extraParams = [])
     {
-        if (!strlen($filePath) || !File::isFile($filePath) || !File::isLocalPath($filePath)) {
+       if (!strlen($filePath) ||
```
LukeTowers: incorrect indentation

modules/system/traits/ConfigMaker.php
lthurston: Oops, there's something screwy with my IntelliJ. I'll fix it.

modules/system/traits/ViewMaker.php
lthurston: Ditto.

modules/system/traits/ViewMaker.php
lthurston: Ok, I'll fix it.

lthurston:
@LukeTowers I made the requested changes.

LukeTowers:

lthurston:
Thanks you!

config/cms.php
```diff
@@ -362,4 +362,27 @@
...
+    | from the plugins/ directory.
+    |
+    | To disable this base directory restriction, the safest option is like
+    | this:
+    |
+    | 'restrictBaseDir' => env('APP_ENV') !== 'dev'
+    |
+    | Ensure that your production environments are properly configured to set
+    | APP_ENV to "production"
+    |
```
LukeTowers: Please change to 
```
Restricts loading backend template and config files to within the base directory of the application.
WARNING: This should always be enabled for security reasons. However, in some cases you may
need to disable this; for instance when developing plugins that are stored elsewhere in the filesystem
for organizational reasons and then symlinked into the application plugins/ directory. 
NEVER have this disabled in production.
```

PR #3629, opened 2018-07-07
Fill component properties properly

GinoPane:
Notes:
- Fix required for rainlab/blog-plugin#394;
- CmsCompoundObject::getComponent() returns components with parameters supplied, and therefore paramName(), propertyName() becomes meaningful

### DIFF
--- a/modules/cms/classes/CmsCompoundObject.php
+++ b/modules/cms/classes/CmsCompoundObject.php
@@ -211,7 +211,8 @@ class CmsCompoundObject extends CmsObject
      * This method is used only in the back-end and for internal system needs when
      * the standard way to access components is not an option.
      * @param string $componentName Specifies the component name.
-     * @return \Cms\Classes\ComponentBase Returns the component instance or null.
+     *
+     * @return ComponentBase|null Returns the component instance or null.
      */
     public function getComponent($componentName)
     {
@@ -219,11 +220,22 @@ class CmsCompoundObject extends CmsObject
             return null;
         }
 
-        return ComponentManager::instance()->makeComponent(
+        $component = ComponentManager::instance()->makeComponent(
             $componentName,
             null,
             $this->settings['components'][$componentSection]
         );
+
+        if (!is_null($component)) {
+            /** @var Controller $controller */
+            $controller = Controller::getController();
+
+            if (!is_null($controller)) {
+                $controller->setComponentPropertiesFromParams($component);
+            }
+        }
+
+        return $component;
     }
 
     /**
--- a/modules/cms/classes/CmsCompoundObject.php
+++ b/modules/cms/classes/CmsCompoundObject.php
@@ -211,8 +211,7 @@ class CmsCompoundObject extends CmsObject
      * This method is used only in the back-end and for internal system needs when
      * the standard way to access components is not an option.
      * @param string $componentName Specifies the component name.
-     *
-     * @return ComponentBase|null Returns the component instance or null.
+     * @return \Cms\Classes\ComponentBase Returns the component instance or null.
      */
     public function getComponent($componentName)
     {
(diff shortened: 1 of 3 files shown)


GinoPane:
Hi @LukeTowers ! How do you feel about the proposed changes?

LukeTowers:
@GinoPane to be honest I'm not wild about it, I'll have to do some more digging into how this works before I decide whether we should be instantiating components returned from this method with the current controller's values.

LukeTowers:
@GinoPane tell you what, just have this PR be the change from protected to public and then populate the params on the components from the blog plugin instead

GinoPane:
@LukeTowers let me get this clear, you suggest to return component from `getComponent` as it was before and then do this on the plugin level:

```
if (!is_null($component)) {
            /** @var Controller $controller */
            $controller = Controller::getController();

            if (!is_null($controller)) {
               $controller->setComponentPropertiesFromParams($component);
            }
 }
```

It would work of course, but my concern against it is that all other plugins that would like to implement the same `setUrl` thing for their components would need to implement the same component initialisation (or reuse Rainlab/Blog's method).

What about leaving the initialisation in the core but also making it optional by using an optional method parameter defaulting to `false`? It will make the change backward-compatible and won't break ay existing functionality if we worry about it.

LukeTowers:
@GinoPane yes, but I don't want `Controller::getController()` being called anywhere, the component already has access to the controller in the `$this->controller` property.

LukeTowers:
@GinoPane ping

GinoPane:
Hi @LukeTowers! 
To make the latest changes more or less reusable I've put the required method into `ComponentBase`, so it will be available to all components.

But to be honest I think that logically the method better belongs to `ComponentManager` maybe.

LukeTowers:
@ginopane why did you move it into `ComponentBase`? What was wrong with where it was?

GinoPane:
As I have said, to make it shareable. So every plugin/component could use the same without reimplementing the same logic.

We could populate params on a plugin level only as you suggested, but the question arises where to put the method. We can't put it into specific component, because then we'll have to reimplement it for other components which also need it. Two options left - a trait with this method and an abstract class as parent for components based on ComponentBase (so it would be, for example, ComponentBase -> Abstract class with populating method -> Posts component).

If you are happy with one of alternatives, I can reimplement this.

LukeTowers:
@GinoPane I don't really see a need for it to be shareable right now, having the bulk of the logic only live in the blog plugin is fine with me for now. If more plugins need this (and start using this functionality) then we can discuss centralizing it appropriately but for now the only change I want done to the octobercms/october repo is to make the one method public instead of private so that the logic within the blog plugin will work correctly to initialize the component with the URL parameters.

LukeTowers:
@GinoPane I noticed :) Will need to run it by @daftspunk first since it's changing a method from protected to public, we'll be having a meeting next week and I'll bring it up then. Thanks again for your work on this so far!

LukeTowers:
@daftspunk was supposed to get back to me on this, hasn't yet.

GinoPane:
Hi @LukeTowers! Any chance for this to get finally reviewed? 😄

LukeTowers:
@GinoPane try bugging @daftspunk about it on slack

GinoPane:
Hi @daftspunk! What is the status of this one? I tried to reach you in slack with no luck.

daftspunk:
I've had a look at https://github.com/rainlab/blog-plugin/pull/394 and https://github.com/rainlab/blog-plugin/pull/392 

There is too much gymnastics going on with the code but also no clear use case on why a custom slug is needed. The slug represents the attribute on the model, so it's OK for it to be hard coded in this way. I'm not really sure why this should be fixed, without a use case it's difficult to know.

GinoPane:
Actually, the rainlab/blog-plugin#392 represents the issue. If you set parameter name to some not expected value, the logic which relies on it won't work.

A basic example with default behaviour (post list is generated by blogPosts component)

image

image

Links are properly generated.

Let's change the slug param name:
image

Voila, blogPosts component cannot generate proper links anymore:
image

There are no any constraints on parameter name values, so it is easy to break them.

GinoPane:
@daftspunk , @LukeTowers I'm not quite sure if anything related was changed somewhere in the code base  but it turned out that this one is no longer needed, the issue could be fixed on plugin level.

I'll make an update for corresponding rainlab/blog PR soon.

PR #3630, opened 2018-07-08
Add option for Richeditor to upload files as attachmany relation

joakimbo:
This will add the option for users to define a relation for files uploaded through the richeditor.

Implemented:
- Users can upload files as relation through the richeditor
- Users can delete uploaded relational files through the browse popup
- Users can select(insert) attached files through the browse popup

Known issues:
- Width of row in popup makes table too wide

### DIFF
--- a/modules/backend/formwidgets/RichEditor.php
+++ b/modules/backend/formwidgets/RichEditor.php
@@ -44,15 +44,14 @@ class RichEditor extends FormWidgetBase
 
     /**
      * @var array Contains the options used for file upload
-     *
      * Supported options:
-     * - mode: mediaManager, relationUpload
-     * - relationName: (name of model relation to use for upload, must be of type attachMany and
-     * the model must inherit from October\Rain\Database\Attach\File)
+     * - mode: [relation | medialibrary]
+     * - relation: Name of relation (relation also needs to be an attachmany relationship where the
+     * attached model inherits from October\Rain\Database\Attach\File)
      */
     public $uploadOptions = [
-        'mode' => 'mediaManager',
-        'relationName' => '',
+        'mode' => 'medialibrary',
+        'relation' => '',
     ];
 
     //
@@ -129,17 +128,17 @@ class RichEditor extends FormWidgetBase
     protected function evalUploadOptions()
     {
         if($this->uploadOptions['mode']
-           && $this->uploadOptions['mode'] == 'relationUpload'
-           && $this->uploadOptions['relationName']
-           && !empty(trim($this->uploadOptions['relationName']))) {
-            if($this->model->getRelationType($this->uploadOptions['relationName']) == 'attachMany'
-               &&  $this->model->{$this->uploadOptions['relationName']}()->getRelated()
+           && $this->uploadOptions['mode'] == 'relation'
+           && $this->uploadOptions['relation']
+           && !empty(trim($this->uploadOptions['relation']))) {
+            if($this->model->getRelationType($this->uploadOptions['relation']) == 'attachMany'
+               &&  $this->model->{$this->uploadOptions['relation']}()->getRelated()
                instanceof \October\Rain\Database\Attach\File) {
-                return 'relationUpload';
+                return 'relation';
             }
         }
 
-        return 'mediaManager';
+        return 'medialibrary';
     }
 
     protected function checkUploadPostback()
--- a/modules/backend/formwidgets/RichEditor.php
+++ b/modules/backend/formwidgets/RichEditor.php
@@ -45,7 +45,7 @@ class RichEditor extends FormWidgetBase
     /**
      * @var array Contains the options used for file upload
      * Supported options:
-     * - mode: [relation | medialibrary]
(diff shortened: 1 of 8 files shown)


joakimbo:
@LukeTowers I'm unsure if it would work in the case of the richeditor but I saw that the Froala options could be [edited after init](https://github.com/froala/wysiwyg-editor/issues/778#issuecomment-142609365) would it be better to add a new js with something the following
```
+function ($) { "use strict";
    $(document).render(function() {
        if ($.FroalaEditor) {
            $('.field-richeditor textarea').on('froalaEditor.initialized', function (e, editor) {
                editor.opts.someParam = someValue;

            });
        }
    })
}(window.jQuery);
```
rather than needing to rebuild the build.min.js?

LukeTowers:
@joakimbo @daftspunk would need to run the build process to rebuild build.min.js, which would probably work the best.

joakimbo:
@LukeTowers What do you think about moving this:
```
// richeditor.js
// File upload
froalaOptions.imageUploadURL = froalaOptions.fileUploadURL = window.location
froalaOptions.imageUploadParam = froalaOptions.fileUploadParam = 'file_data'
froalaOptions.imageUploadParams = froalaOptions.fileUploadParams = { X_OCTOBER_MEDIA_MANAGER_QUICK_UPLOAD: 1 }
```
into the mediamanager.js plugin (editing the froala .opts directly), say in its _init func or a private func which then gets called from the init?
The "upload as relation" could then be moved into a plugin which would be included in the build.min and we simply check which mode is activated in richeditor.js and turn off the other plugin?

joakimbo:
I can see that the imageManager froala plugin is omitted from the build.js, any specific reason for this? It would be beneficial if the relationUploader could make use of this for managing images. instead of writing a new manager.

LukeTowers:
@joakimbo could the options be the following:
```
mode: [relation | medialibrary] Method of storing uploaded images
relation: Name of relation (use relation instead of relationName since that's what's use elsewhere in October already, relation also needs to be an attachmany relationship)
```
Also it would be nice if you wrote the code leaving it open enough for the addition of other parameters in the future (perhaps `mediaPath` and `mediaDisk` as well?) Those wouldn't need to be implemented right away though.

joakimbo:
@LukeTowers Sure, I'll see about changing the options part later, a bit curious what you would want to achieve with mediaPath and mediaDisk though, the media manager is handled through config files and my thought is that the relation would handle the disk and path for relations.

LukeTowers:
@joakimbo say you wanted to store and manage all the files uploaded through a specific richeditor field through it's own mediamanager disk and/or folder

joakimbo:
@LukeTowers That would require some changes to the MediaManager widget and MediaLibrary class right? Since the MediaLibrary acts as singleton, we would need to either temporarily override the disk/folder or we need to be able to create several medialibraries. If this is done we could however use the mediamanager to handle relational files aswell?

LukeTowers:
@joakimbo basically the way to do it right now is to dynamically set the config values for the media library before instantiating it, although to properly implement such a feature in the future we may need to make changes to the MediaLibrary class to make it easier. 

And yes, the path options would only be available for the mediamanager, the fileupload is it's own set of functionality.

joakimbo:
@LukeTowers The basic functionality should now be available, I'm sure there are places where some optimization could be good but a more direct problem is that the content (table data) in the popup modal overflows outside of the modal, where would be the best place to put css to remedy this problem?

LukeTowers:
@joakimbo let me know when it's ready to review. Also, screenshots of the new functionality would be greatly appreciated! :)

joakimbo:
@LukeTowers I think it's ready for a first review now.
New functionality is basically just a record list of attachMany files.
image
This opens in the same way the mediaManager does by clicking the browse (folder icon) in the image or file insert of the richeditor.

LukeTowers:
@joakimbo would it make sense to instantiate a FileUpload form widget directly attached to the RichEditor widget instead to handle that part of the upload so that the UX is more consistent?

joakimbo:
@LukeTowers Hmm, I do agree that it would make sense, my starting point was simply to try and mimic how the mediaManager handled the uploads for the richeditor. The only thing to figure out would be how to "properly" instantiate and attach it but also to hook into insert and delete to be able to pull information necessary about the files for textarea.

LukeTowers:
Awesome @joakimbo, I appreciate all your work on this!

joakimbo:
@LukeTowers So I've gotten the widget to work however if we want to capture selections in the widget we would need to create a new fileupload.php (could inherit from original) and its respective js and partials right?

LukeTowers:
@joakimbo not necessarily, if you can bind / initialize it as a widget attached to the richeditor widget instance it should in theory work correctly without having to do too much custom work.

joakimbo:
@LukeTowers Hi, sorry not updating this. I have been busy for some time now but if this has yet to be solved by some other means then I am still interested. I would however need to re-acquaint myself with the code and most likely wont have time until closer to the holidays.

Issue #3639, opened 2018-07-15
Fileupload in relationcontroller form does not respect disabled option in filterFields method


joakimbo:
##### Expected behavior
Defining a fileupload in fields.yaml for a model and setting ```disabled: true``` in either the fields.yaml or in the models filterFields function makes the widget activate previewMode when used in a relationcontroller form.

##### Actual behavior
Fileupload widget only sets previewMode to true if ```disabled: true``` is set in fields.yaml, setting disabled to true in filterFields does not propagate down to fileupload widget if managed through relationcontroller.

##### Reproduce steps
1. Create a model which manages a relation through a relationcontroller
2. The related model defines a fileupload widget in its fields.yaml
3. Using filterFields function of related model to set the fileupload field to disabled = true has no effect on widget when managed through the relationcontroller form.
4. Setting the fileupload ```disabled: true``` in fields.yaml propagates to the relationcontroller manage form and no upload can be done

##### October build
437

Edit: Did some further quick testing of a fileupload directly in a model (not in a relationcontroller form) and the behavior of setting the disabled value in filterFields does not seem to be supported there either.

Edit 2: One way to achieve this is to listen to the ```backend.form.extendFieldsBefore``` event in the Plugins boot function, I'm unsure if my original approach "should" be supported.

joakimbo:
Related https://github.com/octobercms/october/issues/3444

LukeTowers:
@joakimbo does filterFields give you access to the FileUpload formwidget to be able to set previewMode manually?

joakimbo:
@LukeTowers The ```$fields``` param only contains objects of Backend\Classes\FormField which doesn't seem to have any reference to possible widgets.

joakimbo:
@LukeTowers Could we simply re-run the widgets init func after the filterFields?
https://github.com/octobercms/october/blob/fb2aa1730cc08e8563908b54b70e942bd2988f5e/modules/backend/widgets/Form.php#L1026
i.e
```
/*
 * Allow the model to filter fields.
 */
protected function applyFiltersFromModel()
{
    /*
     * Standard usage
     */
    if (method_exists($this->model, 'filterFields')) {
        $this->model->filterFields((object) $this->allFields, $this->getContext());
        foreach ($this->formWidgets as $widget) {
            $widget->init();
        }
    }
    /*
     * Advanced usage
     */
    if (method_exists($this->model, 'fireEvent')) {
        $this->model->fireEvent('model.form.filterFields', [$this]);
[...]

PR #3645, opened 2018-07-18
Fixes propagation of filterFields to formWidgets

joakimbo:
Ref. to issue: https://github.com/octobercms/october/issues/3639

Edit: Some widgets check their field values before their own config, wasn't sure if this was intentional (feels like it should be config first, then field value)


LukeTowers:
I'll be honest, I don't like this approach. It adds a lot of what seems to be boilerplate code and the naming doesn't really reflect what's actually going on in most use cases. Discussion can continue in #3639.

joakimbo:
Fair enough.

PR #3651, opened 2018-07-24
PHP 7.2

scottbedard:
The current version of PHP is 7.2, this commit just adds that version to our test config.
http://php.net/supported-versions.php

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ language: php
 php:
   - 7.0
   - 7.1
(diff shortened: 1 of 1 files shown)


LukeTowers:
@scottbedard Re #3650 you don't have to close and open new PRs to change the target branch, just click edit next to the title and you can change the target branch there.

scottbedard:
@LukeTowers Oh nice, I didn't realize we could do that

PR #3653, opened 2018-07-27
Add file check for image upload form widgets

gergo85:
Expected behavior  
if the uploaded image is not exist, show the error message.

Actual behavior  
if the uploaded image is not exist, show nothing.

Reproduce steps  
1. Upload an image via media finder or single image upload form widget.
2. Delete or rename the file.

October build  
All


gergo85:
Before:
before

After:
after

daftspunk:
This has been implemented in 65936513313d5cbd2bdcd7893043c926aecd7108 and will be available in Build 449+

Thanks!

acasar:
This doesn't work as expected. I added a comment to the commit: https://github.com/octobercms/october/commit/65936513313d5cbd2bdcd7893043c926aecd7108

I'm also providing a more detailed explanation on how to reproduce the issue here:

1. We load a form with an empty mediafinder input

image


2. We select an image -> this is where the bug happens:

image


3. After we save and reload the page, the issue is gone:

image

PR #3658, opened 2018-07-30
Add custom asset list style

gergo85:
Before:
before

After:
after


gergo85:
Reproduce steps

1. Change the secondary color of back-end area.
2. Go to the CMS page.
3. Click on any asset from list.

PR #3468, opened 2018-03-23
OpCache Restrict

danhrrn:
https://github.com/octobercms/october/pull/3438

PR #3664, opened 2018-07-31
Extend opcache check if it is enabled

vosco88:
When trying to deploy October on shared hosting (www.websupport.sk) I received following error when accessing the frontend pages - ErrorException:> Zend OPcache API is restricted by "restrict_api" configuration directive.
Backend was working normally. 
I propose the following change to check not only if the opcache_invalidate exists but also if opcache is enabled.
Same check is actually present in twig vendor files - but so far it is working without having the check there.

### DIFF
--- a/modules/cms/classes/CodeParser.php
+++ b/modules/cms/classes/CodeParser.php
@@ -338,7 +338,7 @@ class CodeParser
          * Compile cached file into bytecode cache
          */
         if (Config::get('cms.forceBytecodeInvalidation', false)) {
-            if (function_exists('opcache_invalidate')) {
+            if (function_exists('opcache_invalidate') && ini_get('opcache.enable')) {
                 opcache_invalidate($path, true);
             }
             elseif (function_exists('apc_compile_file')) {


LukeTowers:
@danharrin is this related to the issue you had brought up a few months ago?

DanHarrin:
@LukeTowers It certainly is the same issue I was facing.

In my case, I changed the value of `restrict_api` to an empty string in my server PHP config file. October seems to reply on some of the functionality that is restricted by this feature, so it's best not to touch anything in the codebase and instead include a check on installation.

LukeTowers:
@DanHarrin not sure what you mean by that. Also, could you link your issue / PR where we discussed it earlier?

DanHarrin:
@LukeTowers I have found my original PRs but since my old account was compromised and since deleted, I appear as a ghost. I am happy to recreate the PR under this account if you want to close it.

https://github.com/octobercms/october/pull/3438 (our discussion)
https://github.com/octobercms/october/pull/3468 (my suggested fix)

LukeTowers:
@DanHarrin it seems like this is a slightly related but technically different issue; i.e. both this and #3468 should be implemented.

LukeTowers:
@DanHarrin because this checks to see if the opcache is enabled while yours checks to see if the opcache is restricted to a certain directory.

LukeTowers:
@vosco88 is this still an issue?

vosco88:
@LukeTowers Hi Luke, yes it is, after updating the sites to 443 I still needed to add the check to make it work.

LukeTowers:
@vosco88 thanks for confirming. I'll merge it in and then @DanHarrin can submit a separate PR if he is still interested in his issue.

PR #3687, opened 2018-08-11
Add customization of froala lineBreaker tags

vanmil:
This PR adds functionality in the backend to customise the froala lineBreaker tags (https://www.froala.com/wysiwyg-editor/examples/line-breaker) in the same way as one can customize the 'allowed empty tags', ''allowed tags', 'do not wrap tags' and 'remove tags'.

screen shot 2018-08-11 at 18 05 12

### DIFF
--- a/modules/backend/models/editorsetting/fields.yaml
+++ b/modules/backend/models/editorsetting/fields.yaml
@@ -96,6 +96,13 @@ tabs:
             type: textarea
             span: auto
 
+        html_line_breaker_tags:
+            label: backend::lang.editor.line_breaker_tags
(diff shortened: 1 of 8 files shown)


modules/backend/models/editorsetting/fields.yaml
```diff
@@ -96,6 +96,13 @@ tabs:
             type: textarea
             span: auto
 
+        html_line_breaker_tags:
+            label: backend::lang.editor.line_breaker_tags
```
LukeTowers: Missing the lang file changes for these keys

Issue #1795, opened 2016-02-19
Bug: Disabled plugins showing up in main nav


oadslug:
When logging in as superuser disabled plugins are showing up in the main nav (top nav);  it is also possible to view, edit, create, save entries in disabled plugins. They only disappear again when entering Settings/Manage Plugins page or Builder page.

While I haven't tested this I assume that disabled plugins will also show up for any user that has permission to access them. 

This could potentially introduce a security issues.

daftspunk:
Is this a bug report? If so can you provide replication instructions, which plugin does this occur on?

I disable my plugins and they do not appear in the top nav.

oadslug:
It is happening every time for me on this project, in it's current state, but I would have no way of knowing how to duplicate this issue. Possibly a conflict with the Builder plugin. Regardless, this doesn't seem like something that should be possible under any circumstances. I'll try and provide more details if I can. Or try and replicate the issue. 

Current installed plugins are:

RainLab.Blog	1.2.2 - disabled
RainLab.User	1.2.1
RainLab.Location	1.0.3
RainLab.UserPlus	1.0.3
RainLab.Builder	1.0.6
Bedard.Debugbar	1.0.3
October.Test	1.0.3 - disabled
Terrain.Backend	1.0.1
ShahiemSeymor.Roles	1.0.10
Autumn.Pages	1.0.11 - disabled

Terrain.Site	1.0.3 (project plugin)
Terrain.Products 1.0.3 (project plugin)

Hessel91:
Maybe you can use some of my information, I'm having an sort of same issue.
I created an smaller version of the company plugin. everything works fine.
However the plugin only shows up in the main nav when superuser is on. If superuser is of, and all acces is granted it will not appear in the main nav

daftspunk:
We need clear replication instructions to be able to assist. See the contribution guidelines for more information.

uxms:
This bug is duplicated for us, too. Tried both local installation (debug true) and on live site (debug false).

Replication steps:
1- Disable any plugin from manage plugins page
2- Log out from backend (if you not logged-out, plugin seems disabled, neither setting page nor menu page reachable)
3- re-login backend. menu pages shown and settings pages is reachable.
4- Click Updates on settings
5- Plugin disabled from menu and settings, also is not reachable as expected.


Hope this detailed a bit more..

uxms:
[...]

Issue #3208, opened 2017-10-27
Registering plugin schedules in test environment can cause DB error


jimcottrell:
##### Expected behavior
Plugin test cases should be able to be completely isolated from the app environment.

##### Actual behavior
If a plugin accesses the DB in its `registerSchedule` function-- say, via a settings model to determine when or whether to schedule a task--this access happens while bootstrapping the test app's console kernel, which is before the app is configured to use in-memory SQLite by `PluginTestCase`.

In the first test that runs there will be no error, I believe because for the first app bootstrap the plugin manager isn't yet aware of the plugin being tested. But any further tests will trigger the error `PDOException: SQLSTATE[HY000] [2002] No such file or directory` when the settings model attempts to load while registering the schedule.

When I initially noticed this, the errors were actually triggering while testing a plugin other than the one registering a schedule, which was a little hard to figure out at first. So, I know that it's even possible to disrupt another plugin's tests. I assume that was just because in my case the one plugin requires the other that registered the schedule and this wouldn't happen just because the other plugin is installed.

##### Reproduce steps
- Create a plugin with a settings model
- Add a `registerSchedule` function to the plugin where you access a value on the settings model.
- Add a test case extending `PluginTestCase` with a couple of dummy tests (has to be more than one).
- Run PHPUnit on the plugin's tests

##### October build
426

I'm not sure what the best answer is. Perhaps there should be a check for the app environment that doesn't register schedules if testing?

LukeTowers:
@jimcottrell is this related to https://github.com/octobercms/october/pull/3156 at all?

jimcottrell:
I hadn't seen that issue. In my case, I don't boot or register any plugins, so it's not exactly the same situation, but it seems like it could share the root cause. If I modify `October\Rain\Support\Traits\Singleton` to let me forget a singleton instance and create a test `tearDown` that uses that for `PluginManager`, I don't see my error. It seems like state persisting between app boots in externally-managed singletons could be leading to both issues.

LukeTowers:
@jimcottrell what would you recommend as a fix for this issue then?

jimcottrell:
If `PluginTestCase` could either completely reset the `PluginManager` instance's state or had a way to force that instance to be forgotten after tests, that would be a fairly simple change that I think would fix my schedule issue and should eliminate the need to unregister all plugins after a test as in #3156. I can make a PR if you'd like. I'd probably go with forgetting the singleton on teardown.

LukeTowers:
[...]

PR #3691, opened 2018-08-15
#1795 - Bug: Disabled plugins showing up in main nav

GinoPane:
The reason of underlying #1795 issue is that `disabled.json` is being removed on backend login (when UpdateManager is called) and therefore the information is lost.
To my mind `disabled.json` is not necessary at all, but anyway I left it as it was and fixed the issue by populating of missing data from DB. DatabaseServiceProvider is not booted there yet, that's why I had to use DB facade.

### DIFF
--- a/modules/system/classes/PluginManager.php
+++ b/modules/system/classes/PluginManager.php
@@ -431,7 +431,7 @@ class PluginManager
      *
      * @return string
      */
-    public function normalizeIdentifier(string $identifier): string
+    public function normalizeIdentifier($identifier)
     {
         foreach ($this->plugins as $id => $object) {
             if (strtolower($id) == strtolower($identifier)) {
--- a/modules/system/classes/PluginManager.php
+++ b/modules/system/classes/PluginManager.php
@@ -6,9 +6,10 @@ use File;
 use Lang;
 use View;
 use Config;
+use ApplicationException;
 use RecursiveIteratorIterator;
 use RecursiveDirectoryIterator;
-use ApplicationException;
+use Illuminate\Support\Facades\DB;
 
 /**
  * Plugin manager
@@ -425,10 +426,12 @@ class PluginManager
 
     /**
      * Takes a human plugin code (acme.blog) and makes it authentic (Acme.Blog)
-     * @param  string $id
+     *
+     * @param string $identifier
+     *
      * @return string
      */
-    public function normalizeIdentifier($identifier)
+    public function normalizeIdentifier(string $identifier): string
     {
         foreach ($this->plugins as $id => $object) {
             if (strtolower($id) == strtolower($identifier)) {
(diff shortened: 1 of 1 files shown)


modules/system/classes/PluginManager.php
```diff
@@ -425,10 +426,12 @@ public function getIdentifier($namespace)
...
     /**
      * Takes a human plugin code (acme.blog) and makes it authentic (Acme.Blog)
-     * @param  string $id
+     *
+     * @param string $identifier
+     *
      * @return string
      */
-    public function normalizeIdentifier($identifier)
+    public function normalizeIdentifier(string $identifier): string
```
LukeTowers: Please leave return types out, October doesn't use them yet.

GinoPane: Done

daftspunk:
Thank-you for this!

Fixed in ab7dda260baf2b0489f6fe2753a5fabc6f910b93
Available in Build 440+

petehalverson:
This breaks `php artisan october:up` when attempting to seed a new database.

https://travis-ci.org/aspendigital/docker-octobercms/jobs/417373243#L2390-L2391

@jimcottrell, you should chime in and propose the two different solutions we discussed.

jimcottrell:
I think it's just one solution in the end, but if it works like I think it should, it should also handle the root cause of the scheduling issue addressed in #3208, which is mitigated in the testing environment but can still impact setting up a new project with a fresh database.

What I'm thinking is maybe we catch DB exceptions in `populateDisabledPluginsFromDb` from this PR. On an exception, we'd have the `PluginManager` go into kind of a "system-only" mode where it won't load any plugins, as we could assume the exception there should mean this is a fresh database. So, when the task schedule is put together, no plugins will be consulted, fixing the issue of them possibly accessing system settings to determine scheduling and causing a DB error before migrations can run to create the system settings table in the first place. `UpdateManager` would just have to be modified to trigger normal plugin loading if necessary after migrating modules.

I think that would take care of it. Thoughts?

LukeTowers:
@jimcottrell what's wrong with just a 
```php
if (Schema::hasTable('system_plugin_versions'))
```
statement to fix it?

jimcottrell:
@LukeTowers I don't really see the point of running an extra query on every page load for a situation that's resolved as soon as `october:up` is run the first time. It also wouldn't address the scheduling issue, which can occur when setting up projects with plugin code already present (say, everything checked out from a repo). That's not a huge thing, but it is an annoyance.

jimcottrell:
Oh, that's true. I was wrong about it running every time.

I'd refer you back to #3208 on the scheduling discussion.

Of course you're correct that checking for table existence deals with the immediate problem, but I think this is a good opportunity to deal with both issues since the root cause is a fresh database. It makes sense to me to go deeper and ensure that if using a database, that plugins aren't loaded until the system can be set up.

LukeTowers:
@jimcottrell could you submit a PR with your proposed changes?

PR #3693, opened 2018-08-16
Change the Status reportwidget icons

gergo85:


gergo85:
Before:
before

After:
after

PR #3696, opened 2018-08-16
Updated to log when name is changed

SeriousKen:
Enhancement that logs when a CMS object has been renamed but the content is unchanged

### DIFF
--- a/modules/cms/models/ThemeLog.php
+++ b/modules/cms/models/ThemeLog.php
@@ -73,7 +73,7 @@ class ThemeLog extends Model
         $newContent = $template->toCompiled();
         $oldContent = $template->getOriginal('content');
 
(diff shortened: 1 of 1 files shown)


LukeTowers:
Not sure what the purpose of this is exactly.

SeriousKen:
@LukeTowers when logging changes to themes renaming a page/partial doesn't get logged unless the content also changes. I think it would make sense to log when theme elements are renamed and who and when it happend.

LukeTowers:
@SeriousKen so this changes the ThemeLogging to also log whenever a page's template is changed?

SeriousKen:
@LukeTowers it changes ThemeLogging to log when the filename changes. So, for example, if I change `home.htm` to `index.htm` that is not currently logged.

Issue #3675, opened 2018-08-08
Adding attach/detach events for models


kharanenka:
##### Expected behavior
We have model [events](https://octobercms.com/docs/database/model#events).
It will be useful to add events for [Many To Many](https://octobercms.com/docs/database/relations#inserting-many-to-many-relations) relations (attach/detach methods).
This is useful if we need to clear cache or add logic after creating/removing relation.
I tried to create pivot models and use [events](https://octobercms.com/docs/database/model#events) (afterSave, afterDelete) in them, but they do not work in pivot models.

For example:
```php
$obUser->group()->attach($iGroupID);

User::extend(function($obUser) {
    $obUser->bindEvent('relation.afterAttach', function($sRelationName, $iNewGroupID, $arPivotData = []) {
        //$sRelationName == 'group'
        //$iNewGroupID == $iGroupID
    });
});
```
or
```php
$obUser->group()->attach($iGroupID);

Event::listen('relation.afterAttach', function($sRelationName, $obMainModel, $iAttachedElementID, $arPivotData = []) {
        //$sRelationName == 'group'
        //$obMainModel == $obUser
        //$iAttachedElementID == $iGroupID
})
```

osmanzeki:
As an FYI and/or reference, this package and plugin does this:

https://github.com/damianlewis/october-laravel-pivot
https://github.com/damianlewis/oc-sortable-relations-plugin

LukeTowers:
Pull requests welcome ;)

kharanenka:
How did you install these packages in your project? These packages are not published on packagist.org.

LukeTowers:
@kharanenka google composer using github repos as packages.

kharanenka:
Do you understand reasons I created this task? I do not need these functions in project. I am a developer of [Shopaholic](https://octobercms.com/plugins/search?search=shopaholic) plugins and we need to have these functions in OсtoberCMS for development of our plugins. If solution with addition of package is not right for you, why did not you say this first?
Why do we need to figure this out after creating an PR?

If you do not have time to add new features to OctoberCMS, and you are waiting for PR from developers, why do not we discuss the implementation in advance? So that later there would be no problems with her merge?

LukeTowers:
@kharanenka my apologies for the confusion, I thought that you were asking how to include it at the project level. Do you have any questions about how I want to see it implemented in the library?

kharanenka:
Yes, give us recommendations on this task so that we can do issue and you could immediately accept it.

[...]

PR #3700, opened 2018-08-16
#3675 Add damianlewis/october-laravel-pivot package to composer.json

kharanenka:
#3675 This package introduces new events for sync(), attach(), detach() or updateExistingPivot() methods on BelongsToMany relation

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,7 +38,8 @@
         "october/backend": "~1.0",
         "october/cms": "~1.0",
         "laravel/framework": "5.5.*",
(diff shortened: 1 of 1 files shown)


LukeTowers:
@kharanenka this will not be accepted as we are not going to be adding dependencies on extensions to October in the core framework of October, that just doesn't make sense. You can however submit a PR to octobercms/library that would implement the logic of that package in the core library, but we're not going to accept it as an external dependency.

PR #3701, opened 2018-08-17
Improve the System Status dashboard widget

gergo85:
Add notification, when the events log or request log are disabled:
widget

The log settings in this case:
img2

The warning icon is clickable (/system/settings/update/october/system/log_settings) and has tooltip:
hover

### DIFF
--- a/modules/system/reportwidgets/Status.php
+++ b/modules/system/reportwidgets/Status.php
@@ -4,6 +4,7 @@ use Lang;
 use Config;
 use BackendAuth;
 use System\Models\Parameter;
+use System\Models\LogSetting;
 use System\Classes\UpdateManager;
 use System\Classes\PluginManager;
 use Backend\Classes\ReportWidgetBase;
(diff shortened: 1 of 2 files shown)


gergo85:
@LukeTowers Any suggestion for this idea?

w20k:
Hi @gergo85, don't mind if add my 50 cents. From the UX perspective, it's better to use yellow color, because logging, in some cases, is not mandatory and not critical at all :) 
For the icon would use bubble with 'i' (information), but there is already a bubble icon - 'icon-exclamation' or 'icon-info', would do the trick.

gergo85:
This is the suggested version:
widgetv2

LukeTowers:
I'm fine with the triangle warning sign coloured yellow

gergo85:
I can't modify this repository :(

LukeTowers:
@gergo85 feel free to resubmit

gergo85:
@LukeTowers Alright. I created a new PR (#3866).

PR #3704, opened 2018-08-17
Add functionality to use a model method to provide the availableColors for the color-picker widget

vanmil:
This PR adds the functionality to use a custom model method to determine the availableColors (the colors used by the widget) of the color-picker form widget. It uses the same concept as the [dropdown documentation](https://github.com/octobercms/docs/blob/master/backend-forms.md#dropdown)'s fourth method to provide options.

The behaviour of the colorpicker availableColors now is:
- When omitted or left empty, it uses the default colors (stays the same)
- When providing an array of colors, it the colors from this array (stays the same)
- When providing a string function name, it looks at the model and uses that custom model function. 

Example usage:
In your plugin `plugins/example/example/models/example/fields.yaml` set availableColors to your custom method
```
icon_color:
    type: colorpicker
    availableColors: getBrandColors
```
Then in your `plugins/example/example/models/Example.php` define a the method which returns an array containing hex color codes
```
public function getBrandColors($fieldName, $value, $formData){
    return [BrandSetting::get('primary_color'), BrandSetting::get('secondary_color')];
}
```

### DIFF
--- a/modules/backend/formwidgets/ColorPicker.php
+++ b/modules/backend/formwidgets/ColorPicker.php
@@ -95,10 +95,9 @@ class ColorPicker extends FormWidgetBase
     {
         if (is_array($this->availableColors)) {
             return $this->availableColors;
-        }
-        elseif (is_string($this->availableColors) && !empty($this->availableColors)) {
+        } elseif (is_string($this->availableColors) && !empty($this->availableColors)) {
             if ($this->model->methodExists($this->availableColors)) {
-                return $this->model->{$this->availableColors}(
+                return $this->availableColors = $this->model->{$this->availableColors}(
                     $this->formField->fieldName,
                     $this->formField->value,
                     $this->formField->config
--- a/modules/backend/formwidgets/ColorPicker.php
+++ b/modules/backend/formwidgets/ColorPicker.php
@@ -1,6 +1,8 @@
 <?php namespace Backend\FormWidgets;
 
 use Backend\Classes\FormWidgetBase;
+use Lang;
+use ApplicationException;
 
 /**
  * Color picker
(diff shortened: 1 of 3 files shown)


modules/backend/lang/en/lang.php
```diff
@@ -9,6 +9,7 @@
         'invalid_type' => 'Invalid field type used :type.',
         'options_method_invalid_model' => "The attribute ':field' does not resolve to a valid model. Try specifying the options method for model class :model explicitly.",
         'options_method_not_exists' => "The model class :model must define a method :method() returning options for the ':field' form field."
+        'colors_method_not_exists' => "The model class :model must define a method :method() returning html color HEX codes for the ':field' form field."
```
LukeTowers: Please test run all code branches & relevant locales first: https://travis-ci.org/octobercms/october/jobs/417747998, you're missing a comma at the end of both of these lines

modules/backend/formwidgets/ColorPicker.php
```diff
@@ -78,10 +80,37 @@ public function prepareVars()
...
+     * @return array
+     */
+    protected function getAvailableColors()
+    {
+        if (is_array($this->availableColors)) {
+            return $this->availableColors;
+        }
+        elseif (is_string($this->availableColors) && !empty($this->availableColors)) {
+            if ($this->model->methodExists($this->availableColors)) {
+                return $this->model->{$this->availableColors}(
```
LukeTowers: Change to `return $this->availableColors = $this->model->{$this->availableColors}` to locally cache the results of that method and also put the elseif on the same line as the end bracket please.

LukeTowers:
Thanks for your  work on this @vanmil!

PR #3706, opened 2018-08-18
Check for plugin version table existence when loading disabled plugin data or registering schedules

jimcottrell:
@LukeTowers This is what I propose for #3691 and @munxar's issue brought up in #3208, which I have also run into. I decided that deferring plugin loading like I was describing in #3691 was fine but unnecessary. This is simple enough.

I don't think anyone would count on running plugin task scheduling without getting everything set up and at least running `october:up` on their database, but this would break that particular use case.


daftspunk:
We can't merge this because it fixes 2 separate issues.

1/2 fixed in ab4074cec1fb271394f6728762bba48e60284a0d

Issue #3698, opened 2018-08-16
Remove hard height constraint from previewing uploaded images that causes images to stretch


gergo85:
I upload a long width logo:
![before](https://user-images.githubusercontent.com/2959112/44216370-0ad86700-a175-11e8-8927-67aae2df6485.JPG)

If I deleted the height css style, I get right ratio:
![after](https://user-images.githubusercontent.com/2959112/44216462-46733100-a175-11e8-842e-57e8f74b3086.JPG)

Are the css dimensions (especially the height property) necessary?
/modules/backend/formwidgets/fileupload/partials/_config_form.htm (line 15)
/modules/backend/formwidgets/FileUpload.php (line 201-229)

LukeTowers:
@gergo85 feel free to submit a PR to change it to use max-height and max-width instead to constrain the proportions to something reasonable. Remember, people can upload all sorts of wacky sized images

gergo85:
I fixed it, but there is an another issue. If I upload a too small image (height < 170px), the image thumbnail generator increases the image height to 170px. So the image will become ugly.

gergo85:
Done.

PR #3718, opened 2018-08-22
Remove hard height constraint from previewing uploaded images

gergo85:
Ref #3698


LukeTowers:
@gergo85 could you submit screenshots of various image ratios before and after this change? I.e. too large, too small, just right, etc?

PR #3723, opened 2018-08-23
Increase the close button visibility

gergo85:
Before:

before

After:

after


daftspunk:
Thanks @gergo85!

Added in 6099096e5da7d213025ccdd2671db36073c10876
Available in Build 443+

PR #3724, opened 2018-08-23
Add more common aliases

gergo85:
Ref #3240


gergo85:
@LukeTowers Where can I enable the assets minification?

LukeTowers:
@gergo85 https://github.com/octobercms/october/blob/master/config/cms.php#L198

gergo85:
OK. Thank you for answer.

Issue #3721, opened 2018-08-22
CombineAsset::combineToFile and localPath parameter


Rike-cz:
https://github.com/octobercms/october/blob/ab7dda260baf2b0489f6fe2753a5fabc6f910b93/modules/system/classes/CombineAssets.php#L204

$localPath is not used in this method.

LukeTowers:
What's your point?

Teranode:
@LukeTowers I beleive he's trying to find out why it's there if it's not in the function:

```
        // Disable cache always
        $this->storagePath = null;
        list($assets, $extension) = $this->prepareAssets($assets);
        $rewritePath = File::localToPublic(dirname($destination));
        $combiner = $this->prepareCombiner($assets, $rewritePath);
        $contents = $combiner->dump();
        File::put($destination, $contents);
```

Rike-cz:
@Teranode We talked about it with @LukeTowers later on Slack and I was called to make a pull request #3727

PR #3727, opened 2018-08-24
Add assets prefix implementation within combineToFile method

Rike-cz:
#3721 
Add missing implementation of assets `$localPath` prefix within CombineAssets::combineToFile method. Skipping aliases included.

### DIFF
--- a/modules/system/classes/CombineAssets.php
+++ b/modules/system/classes/CombineAssets.php
@@ -207,7 +207,7 @@ class CombineAssets
         $this->storagePath = null;
 
         // Prefix all assets
-        if($localPath) {
+        if ($localPath) {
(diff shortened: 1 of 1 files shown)


modules/system/classes/CombineAssets.php
```diff
@@ -206,6 +206,17 @@ public function combineToFile($assets = [], $destination, $localPath = null)
         // Disable cache always
         $this->storagePath = null;
 
+        // Prefix all assets
+        if($localPath) {
```
LukeTowers: Add a space between if and opening parenthesis please (i.e. `if (`)

Rike-cz:
Thank you, fixed.

LukeTowers:
Thanks @Rike-cz!

PR #3729, opened 2018-08-28
Fix repeater issues with child widgets (including nested repeaters) and duplicate field names

jimcottrell:
I believe that #3523 doesn't go deep enough into this issue (and to be fair, @Narkoleptika said he hadn't pursued it beyond fixing the immediate problem), so this PR reverts that commit to address the root cause.

There are two problems addressed here:

1. Repeaters add group information to the save value, but they do not pass through the data processing to any child widgets (which includes nested repeaters). This is what causes the lack of the `_group` index in child repeater save data in #3523, but it also prevents any other child widget that overrides `getSaveValue` from doing its own processing. This PR fixes that by building the save value through calling `getSaveData` on each child form widget and then tacking on group data if necessary as before.
2. Repeaters add index and group information to the form based on the repeater's field name (without array prefix), causing an issue if multiple repeaters have the same field name, which can come up through nesting or other usage. This could be fixed different ways, but I've just added an MD5 hash of the full field name to the simple field name.

The original test plugin PR (octoberrain/test-plugin#50) can be used to verify that this change works. I plan on eventually submitting a functional test case for this to prevent regressions, but the functional tests need some updating first, which is being discussed elsewhere.

### DIFF
--- a/modules/backend/formwidgets/Repeater.php
+++ b/modules/backend/formwidgets/Repeater.php
@@ -11,7 +11,7 @@ use October\Rain\Html\Helper as HtmlHelper;
 class Repeater extends FormWidgetBase
 {
     const INDEX_PREFIX = '___index_';
-    const GROUP_PREFIX = '___group_';
+    const EXISTS_PREFIX = '___repeater_';
 
     //
     // Configurable properties
@@ -57,7 +57,7 @@ class Repeater extends FormWidgetBase
     protected $defaultAlias = 'repeater';
 
     /**
-     * @var string Form field name for capturing an index.
+     * @var string Form field name for capturing index and group values.
      */
     protected $indexInputName;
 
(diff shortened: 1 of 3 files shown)


Narkoleptika:
I agree with this. I believe I even mentioned something along these lines in octoberrain/test-plugin#50, but ended up going with a less is more approach to, as @jimcottrell put it, fix the immediate problem.

Assuming that this doesn't cause a regression with nested group repeaters, this is definitely the way to go.

Also, to confirm the second issue addressed here, I'm fairly certain that I've run into this at least once before. In that case (nested repeater had the same name as a top level repeater), I ended up changing the name of the nested repeater to get everything working.

LukeTowers:
@jimcottrell does this also fix https://github.com/rainlab/translate-plugin/issues/307?

jimcottrell:
@LukeTowers I don't think it does by itself, no, but it should at least bring things closer. If I get some time, maybe I'll dig into that.

alxy:
Another important question: Does this break the [MLRepeater](https://github.com/rainlab/translate-plugin/blob/master/formwidgets/MLRepeater.php)? I am tempted to believe that...

LukeTowers:
@alxy could you test this with MLRepeater?

LukeTowers:
Made redundant with https://github.com/octobercms/october/pull/4234

PR #3742, opened 2018-08-31
Adapting the request log list for mobile screens

VoroninWD:
On my phone with a diagonal of 5 inches - this list is very difficult to view

The list looks like on the left screenshot, and how does the list look like after my edits on right screenshot:

photo_2018-08-31_08-45-54

I think this is more convenient and informative, since the column 'status_code' is always 404, and does not carry a payload

### DIFF
--- a/modules/system/controllers/requestlogs/config_list.yaml
+++ b/modules/system/controllers/requestlogs/config_list.yaml
@@ -8,7 +8,7 @@ modelClass: System\Models\RequestLog
 recordUrl: system/requestlogs/preview/:id
 noRecordsMessage: backend::lang.list.no_records
 recordsPerPage: 30
-showSetup: false
+showSetup: true
 showCheckboxes: true
 defaultSort:
     column: count
--- a/modules/system/controllers/requestlogs/config_list.yaml
+++ b/modules/system/controllers/requestlogs/config_list.yaml
@@ -8,7 +8,7 @@ modelClass: System\Models\RequestLog
 recordUrl: system/requestlogs/preview/:id
 noRecordsMessage: backend::lang.list.no_records
 recordsPerPage: 30
-showSetup: true
+showSetup: false
 showCheckboxes: true
 defaultSort:
     column: count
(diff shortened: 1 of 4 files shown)


modules/system/controllers/requestlogs/config_list.yaml
```diff
@@ -8,7 +8,7 @@ modelClass: System\Models\RequestLog
 recordUrl: system/requestlogs/preview/:id
 noRecordsMessage: backend::lang.list.no_records
 recordsPerPage: 30
-showSetup: true
+showSetup: false
```
LukeTowers: Better mobile styles is not a good enough reason to prevent people from utilizing the showSetup feature to customize their backend lists.

ebashu-on-holidays:
@LukeTowers, I recorded the animation for comparison

screenshot_2018-09-05-00-42-46-877_com android chrome

modules/system/models/requestlog/columns.yaml
```diff
@@ -5,8 +5,7 @@
 columns:
     status_code:
         label: system::lang.request_log.status_code
-        searchable: yes
```
LukeTowers: Why did you remove these properties?

ebashu-on-holidays: Because status_code is always 404

LukeTowers: Just leave it as it is and set invisible to true.

LukeTowers:
The checkbox in the header in your comparison GIF looks like it is no longer vertically centred.

ebashu-on-holidays:
On mobile screen:
screenshot_2018-09-10-15-14-50-210_com android

On desktop screen:


For checkbox in the header I added special CSS style, but it did not help to solve the problem completely. Mobile browsers scale images as 1 pixel PC = 2 pixels on the smartphone. Even when I wrote `top: -1.59em;` it stuck to the bottom edge, and when `top: -1.6em;` it stuck to the top.

screenshot_2018-09-10-15-14-50-210_com android chrome

ebashu-on-holidays:
I'll ask you not to merge the branches yet, some problems have appeared on other pages

LukeTowers:

PR #3748, opened 2018-09-03
Apply custom secondary color to Balloon Selector

gergo85:
Before:
before

After:
after


LukeTowers:
@gergo85 is the default brand secondary colour the orange colour? I.e. does this change the colour of the balloon selector on installs that haven't changed their brand secondary colour from the default?

gergo85:
Indeed, the two color are different. Other element doesn't use this color.

PR #3750, opened 2018-09-03
Fixed the readonly issue

gergo85:
The readonly attribute doesn't work if the type of form field is checkbox or switch.


PubliAlex:
the readOnly attribute doesn't work with field of type "relation" too, do I have to create another issue for that ?

LukeTowers:
@PubliAlex create a PR to implement readOnly on relation formwidgets please.

PR #3760, opened 2018-09-05
Reset the administrator's role

gergo85:
If I set a role for user, there is no way to delete from role. This solution add a "reset" feature.


gergo85:
role

modules/backend/lang/en/lang.php
```diff
@@ -121,6 +121,8 @@
         'email' => 'Email',
         'role_field' => 'Role',
         'role_comment' => 'Roles define user permissions, which can be overriden on the user level, on the Permissions tab.',
+        'role_none' => 'None',
+        'role_none_comment' => 'This administrator does not belong to any rules.',
```
LukeTowers: `any rules` should be `any roles`

modules/backend/models/User.php
```diff
@@ -157,7 +157,7 @@ public function getGroupsOptions()
 
     public function getRoleOptions()
     {
-        $result = [];
```
LukeTowers: should be 
```php
$result = [
    0 => ['backend::lang.user.role_none', 'backend::lang.user.role_none_comment']
];
```
Don't access uninitialized variables, that's bad practice.

gergo85:
I modified the code.

LukeTowers:

PR #3761, opened 2018-09-05
Correcting the class for correctly displaying the bootstrap grid in the tabs

ebashu-on-holidays:
```yaml
cssClass: row
fields:

# ......... Works very well

tabs:
    # This 'cssClass' is assigned to all tabs, but for the right
    # display - must be assigned for each tab separately
    cssClass: row
    fields:

        # ......... Indenting is not correct

        meta_title:
            tab: SEO
            label: Meta title
            span: storm
            cssClass: col-xs-6

        meta_keywords:
            tab: SEO
            label: Meta keywords
            span: storm
            cssClass: col-xs-6

        meta_description:
            tab: SEO
            label: Meta description
            span: storm
            cssClass: col-xs-12
```

opera- _2018-09-05_140933_bigbear

### DIFF
--- a/modules/backend/assets/less/layout/layout.less
+++ b/modules/backend/assets/less/layout/layout.less
@@ -125,7 +125,7 @@ body {
     display: table;
     table-layout: fixed;
     height: 100%;
-    width: 100%;
+    min-width: 100%;
 
     > .layout-row {
         display: table-row;
--- a/modules/backend/classes/FormTabs.php
+++ b/modules/backend/classes/FormTabs.php
@@ -182,6 +182,10 @@ class FormTabs implements IteratorAggregate, ArrayAccess
      */
     public function getPaneCssClass($index = null, $label = null)
     {
+        if (is_string($this->paneCssClass)) {
+            return $this->paneCssClass;
+        }
+
         if ($index !== null && isset($this->paneCssClass[$index])) {
             return $this->paneCssClass[$index];
         }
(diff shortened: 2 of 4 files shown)


ebashu-on-holidays:
Returned the value of `cssClass` as it was, since this option can be used in other plugins for other purposes. Added a new option `contentCssClass`, it applies the style to the contents of all tabs

ebashu-on-holidays:
The last commit removes the problem with the width of the block in the sidebar

LukeTowers:
Isn't this already implemented by `paneCssClass`?

ebashu-on-holidays:
@LukeTowers, you suggest doing something like this?

```yaml
tabs:
    paneCssClass:
        0: row
        1: row
        2: row
        3: row
    fields:
```

Can it then be easier to add the ability to add to `paneCssClass` not only the array but also the string?

```yaml
tabs:
    paneCssClass: row
    fields:
```

LukeTowers:
@ebashu-on-holidays you could easily adapt the getPaneCssClass method to just always return the same string if `paneCssClass` was defined as a string instead of an array.

LukeTowers:
Just waiting on changes to be made to https://github.com/octobercms/docs/pull/320

ebashu-on-holidays:
@LukeTowers , you forgot to merge branches

LukeTowers:
@ebashu-on-holidays thanks

PR #3765, opened 2018-09-07
Added method getLayout() to Controller

ebashu-on-holidays:


LukeTowers:
@ebashu-on-holidays why is this required?

ebashu-on-holidays:
In my plugin, I added fields to the Layout and Page:
```php
Event::listen('backend.form.extendFields', function ($widget) {
    $this->extendCMSFields($widget, Layout::class);
    $this->extendCMSFields($widget, Page::class);
});

private function extendCMSFields($widget, $instance)
{
    if (!$widget->model instanceof $instance) {
        return;
    }

    $widget->addFields([
        'settings[some_meta]' => [
            'tab'   => 'cms::lang.editor.meta',
            'label' => 'Some meta',
        ],
    ]);
}
```

From the component, I want to access the information recorded in the layout and page settings. This method exists for the page, but this method is not available for the layout
```php
$this->layoutSettings = $this->controller->getLayout()->settings;
$this->pageSettings = $this->controller->getPage()->settings;
```

Issue #2445, opened 2016-10-24
The checkbox value is false instead of true


gergo85:
##### Description

I added a new form field with the `$form->addTabFields` method. The default value of field was _true_.
##### Expected behavior

The checkbox or switchbox is checked.
##### Actual behavior

The checkbox or switchbox is not checked.
##### Reproduce steps

```
Event::listen('backend.form.extendFields', function($form)
{
    if ($form->model instanceof Backend\Models\Preference) {
        $form->addTabFields([
            'fp_show_total' => [
                'tab'     => 'Exampe',
                'label'   => 'Something',
                'type'    => 'checkbox',
                'default' => true
            ]
        ]);
    }
}
```
##### October build

376

Patroklo:
Ok, I THINK, that's a preferences thing. I mean, if I use that code in any form that is not in the preferences section it shows the default value correctly. But it doesn't with the preferences form (/backend/backend/preferences).

But if I click on "Reset to default" button, it starts working correctly!

I haven't checked the code, but it seems that when you have stored data for preferences and add new fields, they will always be shown as empty even if you give them a default value, because you really don't have that preference with that value, you have an empty value there, so it makes a lot of sense.

daftspunk:
As @Patroklo says, it is a quirk with the preferences model. If you are introducing new fields, the values will be empty if the preferences are already established. You should use the "reset to default" to rebuild the preferences from their default values.

LukeTowers:
@daftspunk I think it may be more closely related to a quirk in the form controller/widget - i.e. default values are not applied in any context other than create (or when the model for the form doesn't exist yet) irregardless of whether there is a value for the field or not.

I was planning on looking into improving that at some point in the future so that default values could apply on dynamic fields that get loaded into a model that doesn't have a value set yet because it wasn't a part of the initial save of the model. Do you have any thoughts on that?

daftspunk:
Ok, I trust your judgement on this. I would need to see the problem demonstrated along with the fix and we are good to go.

LukeTowers:
For sure, I'll take note of this issue and mention it when I make a PR

bbrandtlv:
Hey, i'm having same issue, but in i declared in fields.yaml
```
        special_offer:
            label: mishai.sitemanager::lang.hotel_model.special_offer
            type: checkbox
            default: true
            tab: mishai.sitemanager::lang.hotel_model.special_offer

```
looks like default: true simply not working

daftspunk:
[...]

Issue #2013, opened 2016-05-16
[Bug?] Repeater fields not honouring default values


DanielHitchen:
Hi,

When using the repeater widget, if you set a default value for a form field within it, the default value is ignored whenever you add a new item.

Not sure whether this is a bug or whether this functionality just needs adding to the widget, but it would be great if the repeater honoured the default field values/options set.

daftspunk:
This is a complex case. Please submit a Pull Request to the [test plugin for investigation](https://github.com/daftspunk/oc-test-plugin) that demonstrates the issue. Don't forget to include step by step instructions how how to replicate it *using the test plugin*.

gabsource:
If I understood the problem, it happens if the _data_ passed to the form contains the _key_ whether the value is `null`, `0` or empty string.

The FormField test looks like this : 
```
if (!array_key_exists($key, $result)) {
    return $default;
}
```

If the form is populated whit a model where the _key_ exists it does not look at the default value : 

```
$model = new MyModel;
$model->myField = null; // if you remove this, the field in populated with the default value
```

DanielHitchen:
@gabsource thanks for your input.

I found where the default field value is set within the backend ```modules/backend/widgets/Form.php``` and changed lines 896-898 from:
```
$defaultValue = !$this->model->exists
    ? $field->getDefaultFromData($this->data)
    : null;
```
to
```
$defaultValue = $field->getDefaultFromData($this->data);
```
for the moment in time until a better solution can be found - but this does what I want for now.

Thanks

harmenjanssen:
I created a Javascript fallback to deal with this problem.
See this gist for details: https://gist.github.com/harmenjanssen/6895778f4b85910ba56452abc98fdc3e

Basically: add a data-attribute to the fields in the repeater containing the default value, which will be picked up by Javascript.

haralake:
Is there a way to solve this issue with repeater default values on theme settings, without hack the core files?

LukeTowers:
Submit a PR to fix it @haralake, hacking the core files is fine as long as you contribute those hacks and they're well thought out so that they stay when you next update :)

HammenWS:
@LukeTowers I have the same issue and created a PR for this #3766

bennothommo:
Fixed by https://github.com/octobercms/october/pull/4098 and https://github.com/octobercms/october/commit/a7bc442df5fb615a1cf044b046ccce0e374aff73.

PR #3766, opened 2018-09-07
Getting default values from data in formwidget

HammenWS:
When adding a new repeater field in update context, default values for the added fields won't be respected.

Some issues where this was mentioned:
https://github.com/octobercms/october/issues/2445#issuecomment-274621932
#2013 

You can replicate this in the test plugin by adding a default value for some field in the config/repeater_fields.yaml 

When you *create* a new item using this repeater field, the default value will be set.
But when you *update* an item, the default value won't be set when adding a new group.

This PR should fix this problem.

### DIFF
--- a/modules/backend/widgets/Form.php
+++ b/modules/backend/widgets/Form.php
@@ -1081,7 +1081,11 @@ class Form extends WidgetBase
      * This should be done when model exists or when explicitly configured
      */
     protected function shouldFetchDefaultValues() {
-        return !$this->model->exists || object_get($this->config, 'shouldFetchDefaultValues');
+        $shouldFetchDefaultValues = object_get($this->config, 'shouldFetchDefaultValues');
+        if ($shouldFetchDefaultValues === false) {
+            return false;
+        }
+        return !$this->model->exists || $shouldFetchDefaultValues;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


modules/backend/widgets/Form.php
```diff
@@ -1069,13 +1069,21 @@ protected function getFieldValue($field)
...
 
         return $field->getValueFromData($this->data, $defaultValue);
     }
 
+    /**
+     * Checks if default values should be taken from data.
+     * This should be done when model exists or when explicitly configured
+     */
+    protected function shouldFetchDefaultValues() {
+        return !$this->model->exists || object_get($this->config, 'shouldFetchDefaultValues');
```
LukeTowers: If we're using a config value for this it should support being explicitly set to false and having that override `!$this->model->exists`

HammenWS:
@LukeTowers 
I didn't think about such a usecase, but I understand why this is wanted. I added the changes to this PR.

HammenWS:
@LukeTowers Hello? Have you seen my changes?

LukeTowers:
@HammenWS could you post the YAML file where this option would be used?

LukeTowers:
Could you rename the `shouldFetchDefaultValues` config property (and related variables) to: `enableDefaults`? And then add documentation (and necessary supporting code to the FormBehavior) to https://octobercms.com/docs/backend/forms#configuring-form for how people can use `enableDefaults` (not defined, based on whether the model exists or not - default behavior; true, always use them; false, never use them)

LukeTowers:
Closing in favour of #4098

Issue #3746, opened 2018-09-02
Support Swoole


tamer-hassan:
Already got OctoberCMS working with Swoole after some work with https://github.com/scil/LaravelFly  running in "Map" mode.

but have only two main issues with current OC code:

- the stage at which Execution Context (back-end/front-end) setting happens is very early
See: https://github.com/octobercms/library/blob/master/src/Foundation/Bootstrap/RegisterOctober.php#L40-L50
It is not until System\ServiceProvider is registered that ExecutionContext is actually checked
See:
https://github.com/octobercms/october/blob/master/modules/system/ServiceProvider.php#L73
So why not make that check at that later stage?
This would allow us to run the bootstrapper's code only once, and only register the System ServiceProvider on every new request.

- Global vars usage is troublesome. Had a problem where all AJAX widgets were not working and traced it down to the use of global $_POST in `post()` helper function in:
https://github.com/octobercms/library/blob/master/src/Support/helpers.php#L42-L51
it was always coming back empty. however, `request()->post()` works for both php-fpm and Swoole.
also LaravelS mentions cannot use global vars like $_POST/$_GET etc: https://github.com/hhxsv5/laravel-s#important-notices

tamer-hassan:
https://github.com/tamer-hassan/october-fly

LukeTowers:
@tamer-hassan I don't really have any idea what you're asking for nor do I have the time to research a bunch of other projects to figure it out either. Try simplifying what you're saying so that I can easily grasp what it is you're trying to do / want me to do about your question.

LukeTowers:
@tamer-hassan how about you submit a PR with the required changes to support Swoole to the https://gitub.com/octobercms/library repository and we can carry on the discussion there?

tamer-hassan:
@LukeTowers PR's submitted, with changes as minimalist as possible.

1. Moving only the execution-context setting of back-end/front-end to System\ServiceProvider allows Swoole-based projects to work properly when only System\ServiceProvider is configured as a 'providers_in_request' (as in [OctoberFly example config](https://github.com/tamer-hassan/october-fly/blob/master/config/laravelfly-app-config.example.php) ), and everything else booted on worker (kept in memory). I have also tested the changes on php-fpm and it works as expected.

2. using request() helper instead of Globals $_GET / $_POST fixes issues with Swoole like inability to login, and also lets all AJAX based widgets work. Also tested on php-fpm an it all works as expected.

[...]

PR #3773, opened 2018-09-11
Swoole support: avoid header_remove()

tamer-hassan:
Header modification is not supported with Swoole. Functions like header() or header_remove() will cause an error.
See:
https://github.com/swooletw/laravel-swoole/wiki/Z1.-Notices
https://github.com/scil/LaravelFly/wiki/Coding-Requirement

Discussion here: https://github.com/octobercms/october/issues/3746


LukeTowers:
Please also leave a link to this PR in the comment about why that code exists.

daftspunk:
Fixed by c5bc804d7398522783fbd76b2764f9541747867d

Thanks!

PR #3783, opened 2018-09-13
Swoole support: avoid exit() usage

tamer-hassan:
because exit() would kill the swoole http server worker in which the request is run.

discussion: https://github.com/octobercms/october/issues/3746

### DIFF
--- a/modules/backend/behaviors/ImportExportController.php
+++ b/modules/backend/behaviors/ImportExportController.php
@@ -645,7 +645,11 @@ class ImportExportController extends ControllerBehavior
          * Output
          */
         $csv->output($options['fileName']);
-        exit;
+
+        // Do not kill Swoole http server if we're running that
+        if (php_sapi_name() != "cli") {
+            exit;
+        }
     }
 
     //
--- a/modules/backend/behaviors/ImportExportController.php
+++ b/modules/backend/behaviors/ImportExportController.php
@@ -1,5 +1,6 @@
 <?php namespace Backend\Behaviors;
 
+use App;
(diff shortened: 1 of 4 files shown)


tamer-hassan:
`exit` / `die` will kill the Swoole http server worker that the request is running in. There's just no direct replacements for these 2 language constructs (they're not normal functions).
See:
https://github.com/swooletw/laravel-swoole/wiki/Z1.-Notices
https://github.com/scil/LaravelFly/wiki/Coding-Requirement

I don't see how this would introduce bugs as the check for `php_sapi_name()` will never be `cli` unless we're running in Swoole http server. All what is touched here is File transfer!

I also think those exit's were only added as an additional safety and not absolutely required. With the changes in commit above, Swoole file upload now works without 500 server error after successful upload through media manager, because the script now finishes cleanly.

In any case, if you still see a replacement for `exit` here as a must, we can do it the Python way:

define an exception named `SystemExit', throw it instead of calling exit() and catch it in index.php with an empty handler to finish script execution cleanly.


```
// file: index.php
class SystemExit extends Exception {}
try {
   /* code code */
}
catch (SystemExit $e) { /* do nothing */ }
// end of file: index.php

///////////////////

// some deeply nested function or .php file    

if (SOME_EXIT_CONDITION)
   throw new SystemExit(); // instead of exit()
```

idea from comments on: http://php.net/manual/en/function.exit.php

daftspunk:
This is fixed by 5190c8177bbb6531836ad8fa8458111361a35608

Thanks!

Issue #1313, opened 2015-07-26
Export problems


axomat:
Hi
Its great to see the new options in Import and Export but here is some feedback.

1. I had to remove the debugbar plugin which adds html to the exported file and causes the browser to think it is an html page. Removing the plugin makes it work but the browser still thinks the csv file is an html page so I think there is a download type issue somewhere.

2. I cant find a way to get eloquent to get() an attribute appended to the model. It only works if you get all columns which may be a limitation of laravel but annoying if, like me, you make a lot of use of appended attributes. This worked in the old method.
EDIT:
I have solved this by using the following in the import/export model
return Subscriber::get()->toArray();
instead of 
return Subscriber::get($columns);
which is how the documentation suggests it should be used.

3. Exporting columns that have been created from relations is much more difficult (it was easy in the old method), in fact I dont know how to do it but I presume some laravel magic in the import/export model would do it,

So in conclusion, please could we have the old "simple" export back as an alternative to the new one? It worked really easily, the user could do a search and refine the list of records before downloading them (this does not work at all with the new method), and it was really easy to understand for the user "if you can see the data on the page, you will see it in the downloaded file". As a programmer it was really easy as well, just a line in a config file. I used it in nearly all my controllers, instead now I have to go through all my plugins and rewrite to avoid the issues above.

Please?!

axomat:
A further update on this, I have reproduced a search option by using the form facility. For anyone else trying this, I had to make the fields used in the form all fillable in the importexport model.

daftspunk:
I was going to suggest toArray(), perhaps I will add this to the documentation. The debugbar needs to detect if the response is not text/html and disable itself, this will cause problems for the Theme Export feature too. Exporting with relations is straight forward, simply make the columns visible and toArray() will take care of the rest. Eg:

```
protected $visible = ['somerelation.somefield'];
```

Or see the `addVisible()` method.

axomat:
Thanks Daftspunk fantastic response as always!

Floriskoch:
Where can I find some more documentation on this? I'm trying to export own plugin but the relation fields stay empty. I've looked here https://octobercms.com/docs/backend/import-export but it doesn't say anything about exporting relations. The above comments also don't work. I've swapped the `protected $visible` property for a `public $attributes` property but that does not work either. Would be great to get some more guidance on this subject. Thanks!

chrisvidal:
Same here, I can't seem to find any documentation on how to export simply a relation or 2 levels of relation also

[...]

PR #3787, opened 2018-09-15
Adds missing Content-Type header to CSV-export. (mentioned in #1313)

cleverer:
CSV Export got interpreted as HTML due to wrong Content-Type header. (Tested with Safari 11.1.2 on macOS 10.13.6 (17G65)).

Since the Content-Type gets set in the `BinaryFileResponse->prepare()` method (which does not get called here) the headers are missing it.

If someone can come up with a more elegant solution than hardcoding the type, please feel free to let me know. I'm relatively new to OctoberCMS.

### DIFF
--- a/tests/unit/backend/models/ExportModelTest.php
+++ b/tests/unit/backend/models/ExportModelTest.php
@@ -72,7 +72,7 @@ class ExportModelTest extends TestCase
         $response->prepare($requestMock);
 
         $this->assertTrue($response->headers->has('Content-Type'), "Response is missing the Content-Type header!");
-        $this->assertTrue($response->headers->contains('Content-Type', 'text/plain'), "Content-Type is not \"text/csv\"!");
+        $this->assertTrue($response->headers->contains('Content-Type', 'text/plain'), "Content-Type is not \"text/plain\"!");
 
         @unlink(temp_path() . '/' . $csvName);
     }
--- a/tests/unit/backend/models/ExportModelTest.php
+++ b/tests/unit/backend/models/ExportModelTest.php
@@ -8,7 +8,16 @@ class ExampleExportModel extends ExportModel
 {
     public function exportData($columns, $sessionKey = null)
     {
-        return [];
+        return [
+            [
+                'foo' => 'bar',
+                'bar' => 'foo'
+            ],
+            [
+                'foo' => 'bar2',
+                'bar' => 'foo2'
+            ],
+        ];
     }
 }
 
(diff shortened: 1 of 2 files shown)


cleverer:
Actually, `Symfony\Component\HttpFoundation\BinaryFileResponse` (which gets returned by `Response::download()`) provides the method `deleteFileAfterSend()` to achieve the desired effect.

cleverer:
I was able to write a test.

tests/unit/backend/models/ExportModelTest.php
```diff
@@ -48,4 +57,24 @@ public function testEncodeArrayValue()
...
+        $requestMock = $this
+            ->getMockBuilder('Illuminate\Http\Request')
+            ->setMethods()
+            ->getMock();
+        $response->prepare($requestMock);
+
+        $this->assertTrue($response->headers->has('Content-Type'), "Response is missing the Content-Type header!");
+        $this->assertTrue($response->headers->contains('Content-Type', 'text/plain'), "Content-Type is not \"text/plain\"!");
+
+        @unlink(temp_path() . '/' . $csvName);
```
LukeTowers: Is this required since download() should be deleting the file?

cleverer: Exactly.

I wanted to avoid unnecessary side-effects, so I tried to use the minimal amount of function calls. Since I know exactly what unlink does, I figure it would be safer.

LukeTowers: Isn't there a way to mock the send? I'd like to be able to verify that the file was automatically deleted after making the request

cleverer: I think we could just call `->send()` on the response. But I'm not that familiar with tests in October/Laravel, maybe there is a more elegant solution.

However, I'd suggest to write a separate test for this…

LukeTowers: @cleverer what's the rationale for the second test instead of just calling send()?

cleverer: Well it is a unit test, so it should test each piece of functionality separately, right? One test checking the headers being set right and one checking the file getting deleted afterwards. I think it is easyer to work with tests that just test a small and well defined piece of code.

But if the tests for October usually are more generalized, I'm happy to follow the lead and make them consistent with the others.

LukeTowers: In my mind we're testing the proper functionality of the export, which includes correct headers and cleaning up after itself, so you might need to change the name of the test method to reflect that but I think it should be one piece.

PR #3788, opened 2018-09-16
integrate flag-icon for language select

mariavilaro:
Integration of flag-icon-css library for language selection dropdown.

I didn't delete existing flag.less because I I'm not sure if it's still needed for backwards compatibility with anything. If it can be deleted, some styles should be copied to flag-icon.less (a small margin and I don't know if anything else)


modules/backend/models/Preference.php
```diff
@@ -178,45 +178,45 @@ protected function getFallbackLocale($locale)
     public function getLocaleOptions()
     {
         $localeOptions = [
-            'ar' => [Lang::get('system::lang.locale.ar'), 'flag-sa'],
```
LukeTowers: Is there anyway CSS trickery that we could do in order to retain the existing `flag-locale` class syntax instead of requiring the `flag-icon` addition?

Something like:
```css
*[class*="flag-"] {
     .flag-icon()
}
```

mariavilaro: I'm sorry I have no idea what that css you wrote do, I'm not a css expert at all (I'm a backend developer...)

w20k:
Should be something like that - "*[class^='flag-'] and *[class*=' flag-']", but new mixing and flags list should be added for a new css style (.flag-{flagname} ;)

LukeTowers:
@w20k if I merge this into a separate branch can you build on what @mariavilaro did and make it so it would be a drop in replacement for the existing setup?

modules/backend/models/Preference.php
LukeTowers: `*` targets all elements, `[]` is an attribute selector, `*=` means value is included anywhere (as opposed to `^=` starting with and `$=` ending with).

w20k:
@LukeTowers , sure can do, but most probably on a weekends ;) Still, might be some issues, if somewhere there is the same class name starting from 'flag-'. Some testing is needed!

LukeTowers:
There shouldn't be a class name starting with `flag-` but give it a shot and let me know how it goes. I'll create the new branch for you to make your PR to.

LukeTowers:
@w20k base your PR onto the wip/flagicons branch and we can merge it from there when all the development is done.

LukeTowers:
Whatever works @w20k :)

w20k:
Seems like working ;) Even squared version.

back-end_preferences___octobercms

LukeTowers:
Submit the PR then :)

PR #3794, opened 2018-09-17
Fix for unit test

tschallacka:
It always tests against an absolute url instead of the url set in the app.url configuration

When october is installed in a subdirectory and the url is different than the hardcoded test url. Of if one has set a test domain via the hosts.conf it would always test against a wrong url.

```
$ ./vendor/bin/phpunit
PHPUnit 5.7.27 by Sebastian Bergmann and contributors.

..............................................................  62 / 332 ( 18%)
....F......................................................... 124 / 332 ( 37%)
.............................................................. 186 / 332 ( 56%)
.............................................................. 248 / 332 ( 74%)
.............................................................. 310 / 332 ( 93%)
......................                                         332 / 332 (100%)

Time: 1.55 minutes, Memory: 38.00MB

There was 1 failure:

1) ControllerTest::testThemeUrl
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'http://localhost/themes/test'
+'http://localhost/exitoffice/themes/test'

```

Not sure if app.config + cms.themesPath is correct, but I would think it would have to be so that the settings that are set match the settings that are generated by the internal code.

### DIFF
--- a/tests/unit/cms/classes/ControllerTest.php
+++ b/tests/unit/cms/classes/ControllerTest.php
@@ -475,7 +475,7 @@ ESC;
     public function testComponentWithOnRender()
     {
         $theme = Theme::load('test');
-        $controller = new Controller($theme); 
+        $controller = new Controller($theme);
         $response = $controller->run('/component-custom-render')->getContent();
 
         $content = <<<ESC
--- a/tests/unit/cms/classes/ControllerTest.php
+++ b/tests/unit/cms/classes/ControllerTest.php
@@ -19,18 +19,19 @@ class ControllerTest extends TestCase
         include_once base_path() . '/tests/fixtures/plugins/october/tester/components/ContentBlock.php';
         include_once base_path() . '/tests/fixtures/plugins/october/tester/components/Comments.php';
         include_once base_path() . '/tests/fixtures/plugins/october/tester/classes/Users.php';
+        
     }
 
     public function testThemeUrl()
     {
         $theme = Theme::load('test');
         $controller = new Controller($theme);
-
+        
         $url = $controller->themeUrl();
-        $this->assertEquals(config('app.url') . 'themes/test', $url);
+        $this->assertEquals(config('app.url').substr(config('cms.themesPath').'/test', 1), $url);
 
         $url = $controller->themeUrl('foo/bar.css');
-        $this->assertEquals(config('app.url') . '/themes/test/foo/bar.css', $url);
+        $this->assertEquals(config('app.url').substr(config('cms.themesPath'), 1).'/test/foo/bar.css', $url);
 
         //
         // These tests seem to bear different results
@@ -475,7 +476,7 @@ ESC;
     public function testComponentWithOnRender()
     {
         $theme = Theme::load('test');
-        $controller = new Controller($theme);
+        $controller = new Controller($theme); 
         $response = $controller->run('/component-custom-render')->getContent();
 
         $content = <<<ESC
--- a/tests/unit/cms/classes/ControllerTest.php
+++ b/tests/unit/cms/classes/ControllerTest.php
@@ -27,10 +27,10 @@ class ControllerTest extends TestCase
         $controller = new Controller($theme);
 
         $url = $controller->themeUrl();
-        $this->assertEquals('http://localhost/themes/test', $url);
+        $this->assertEquals(config('app.url') . 'themes/test', $url);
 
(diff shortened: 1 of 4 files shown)


tests/unit/cms/classes/ControllerTest.php
```diff
@@ -25,12 +25,12 @@ public function testThemeUrl()
     {
         $theme = Theme::load('test');
         $controller = new Controller($theme);
-
+        
         $url = $controller->themeUrl();
-        $this->assertEquals('http://localhost/themes/test', $url);
+        $this->assertEquals(config('app.url').substr(config('cms.themesPath').'/test', 1), $url);
```
LukeTowers: Your URL building is off, firstly it's missing a slash at the start and secondly this doesn't properly cover every case of custom configuration. Isn't their a url helper that you can use instead? Something like `url(config('cms.themesPath'))`?

tschallacka: I'm not entirely familiar with what all the helper functions are that are available and what the correct way is to build things. This was my first effort to get this one failing unit test to pass and make it useable.

How it builds now is app.url which ends with a / at the end like http://localhost/ or http://127.0.0.1/ when having the default config which needs then the theme added which is stored in config like '/theme' as default, which then needs the '/' at the start removed which is the substr for.

i'll have a look at the possible helper functions tomorrow(now +13h)

tschallacka: I modified it with a regex. The url helper doesn't remove double slashes. 
Right now I build the url with a `/` in between that will get removed by the regex if they have a config set up without trailing and leading slash for the corresponding settings. It now passes on several configuration ways.

LukeTowers: @tschallacka could you instead look at adding that handling to the URL helper? I'd rather the built in helper works for all configurations instead of just one random unit test.

tschallacka: Ask and you shall receive. I had to add a few extra files to the project. I hope I put them in the correct places. I'm not sure if i'm supposed to put it in the modules/system or somewhere else. If it needs to be put somewhere else/registered somewhere else please let me know.

LukeTowers:
Clean and simple fix to this: use `themes_path($pathRelativeToThemesDir)`: http://octobercms.com/docs/services/helpers#method-themes-path

tschallacka:
Okay. Can try that. If that works, do you still want to keep the url helper modification that removes double  slashes?

LukeTowers:
No, that's fine. Keep it simple :)

LukeTowers:
Darn it. 

Could you add a helper to https://github.com/octobercms/library/blob/6eff938b1d4d8c794893e5ab35d4f8a817bf627b/src/Support/helpers.php `themes_url()` that would basically pass `url($configThemesDirWithoutBeginningSlash . '/' . $passedPathWithoutBeginningSlash)`? While you're at it, could you create on for plugins_url() too?

tschallacka:
I'll have a look see and add those, maybe tonight, maybe tomorrow.

LukeTowers:
Appreciate it

tschallacka:
Okay, an issue i'm running against is that the rain/library is intended for standalone inclusion in seperate laravel projects. The features you're requesting are octobercms specific, since laravel doesn't support a cms configuration option file by default.

Would it be wiser to implement the code I have in the october repo(this one) or continue to make it for the library repo?

LukeTowers:
@tschallacka the October Rain library already has other references to the OctoberCMS configuration files and there are already path helpers for plugins and themes so I don't mind adding these helpers. Besides, using just the October Rain library by itself isn't really something that we care to support. I appreciate the foresight though!

w20k:
@tschallacka are you still interested in this?

tschallacka:
@w20k yes I am

LukeTowers:
@tschallacka what's the status of this?

tschallacka:
@LukeTowers You just closed the other pull request this stuff tests. I guess you can close this one too.

https://github.com/octobercms/library/pull/350#issuecomment-485013631

LukeTowers:
@tschallacka if you still want to get the PR implemented that's fine, I can reopen the other PR, but otherwise I'll close it for now. Let me know what you want done.

tschallacka:
Currently closing would be fine, i don't have the time to give in currently.

PR #3798, opened 2018-09-18
Fix input trigger on source field types with multiple selected values

jimcottrell:
The problem is demonstrated in octoberrain/test-plugin#55. Currently, the form field trigger doesn't work on source fields with multiple selected values. This PR fixes the issue by running through all selected fields (e.g. `checkboxlist` with multiple selections) and handling an array as a result of `$.val()` on the trigger field (e.g. `taglist` or generic `<select>` using multiple attribute).


LukeTowers:
Thanks @petehalverson for the wonderful Octodock system, made this so easy to test! Thanks for the detailed test and fix @jimcottrell, much appreciated!

Issue #3264, opened 2017-11-24
Error at the CombineAssets


rizhenkov:
##### Expected behavior
When I add .less file in the plugin component, I expext that file will be compiled to .css as described in docs and added to styles.

##### Actual behavior
When I add .less file:
```php
public function onRun(){
    $this->addCss(['assets/less/main.less']);
}
```
System tryed to load file from incorrect path started from root and fails with error. 

File (really) located here (absolute path): 
`/home/site/plugins/company/pluginname/assets/less/main.less`

Path from the error (where system trying to get it).
```
file_get_contents(/plugins/company/pluginname/assets/less/main.less): failed to open stream: No such file or directory
```

##### Reproduce steps

1. Create empty plugin and add component
2. Create plugin less file at the assets/less/main.less
3. Add less file to styles from component's onRun method
```php
public function onRun(){
    $this->addCss(['assets/less/main.less']);
}
```
4. Add component to page and try to access it from public. It fails.

##### October build
428

First candidate where is may be bug:
class CombineAssets at line 456  (I think, error in the unnamed function)
```php
protected function getDeepHashFromAssets($assets)
    {
        $key = '';

        $assetFiles = array_map(function ($file) {
            return file_exists($file) ? $file : File::symbolizePath($file, null) ?: $this->localPath . $file;
        }, $assets);

```

Stack trace: 
![default](https://user-images.githubusercontent.com/987517/33216984-32a84632-d147-11e7-93de-6537c62d71db.png)

LukeTowers:
The AssetMaker trait sets `$this->assetPath` to a public path (relative to the project) by default instead of an absolute path (required for the asset combiner). Try this instead:
```php
public function onRun()
{
    $publicAssetPath = $this->assetPath;
    $this->assetPath = $this->guessViewPath('/assets');
    $this->addCss(['assets/less/main.less']);
    $this->assetPath = $publicAssetPath;
}
```

LukeTowers:
But you're right, this behaviour could be improved, perhaps to try using an absoluteAssetPath when using the combiner. We would need to maintain backwards compatibility with any changes though.

riartem:
Thank you for a quick solution. Works! 👍 
But I really think that was a bug and need for fix :)

miy4mori:
same issue.
finally,I have to use css instead of scss.

LukeTowers:
@miy4mori did you try the solution: https://github.com/octobercms/october/issues/3264#issuecomment-346857066?

[...]

PR #3802, opened 2018-09-19
Fix asset combiner

cleverer:
Opening a pull request for easier discussion of #3264.

I'm not that familiar with the inner workings of October, I think we should get an error, if we have a path like `~/themes/demo/assets/test.css`but in my tests it did work.

### DIFF
--- a/modules/system/traits/AssetMaker.php
+++ b/modules/system/traits/AssetMaker.php
@@ -112,11 +112,7 @@ trait AssetMaker
     public function addJs($name, $attributes = [])
     {
         if (is_array($name)) {
-	        $assetPath = $this->assetPath;
-	        if (!starts_with($assetPath, [base_path()])) {
-		        $assetPath = base_path($assetPath);
-	        }
-            $name = $this->combineAssets($name, base_path($assetPath));
+            $name = $this->combineAssets($name, $this->getLocalPath($this->assetPath));
         }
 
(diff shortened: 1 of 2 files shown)


modules/system/traits/AssetMaker.php
```diff
@@ -142,7 +146,11 @@ public function addJs($name, $attributes = [])
     public function addCss($name, $attributes = [])
     {
         if (is_array($name)) {
-            $name = $this->combineAssets($name);
+	        $assetPath = $this->assetPath;
+	        if (!starts_with($assetPath, [base_path()])) {
```
LukeTowers: Can you put this in a helper method, something like `getLocalPath()`?

cleverer: Sure thing!

LukeTowers:
Thanks @cleverer. Could you add some unit tests to test the following cases:
```php
$this->assetPath = $defaultPublicPath;
$this->assetPath = $pathWithSymbols // both types of symbols
$this->assetPath = $pathThatAlreadyIncludesBasePath
```

LukeTowers:
@cleverer the second approach works for me, the public path would be generated by https://github.com/octobercms/october/blob/master/modules/backend/classes/WidgetBase.php#L54

LukeTowers:
Awesome, thanks very much @cleverer!

PR #3804, opened 2018-09-19
Added a few tests to test the new helpers committed

tschallacka:
A few unit tests for the new helpers that were made as a pull request to https://github.com/octobercms/library/pull/350 

I added the unit tests here because the library itself provides no app() or config() when running an unit test on it that is needed, and here it can be tested in the Environment where it is essential that it works properly.(to be honest, I couldn't be bothered to mock app and config)

It's thought as an optional module, not sure if it should be here, but an extra unit test couldn't hurt I would think.

for full history https://github.com/octobercms/october/pull/3794


LukeTowers:

PR #3812, opened 2018-09-22
Repaired Form widget renderField method

Rike-cz:
In `Backend\Widgets\Form::renderField` method is calling of `prepareVars()` needed before using `$this->allFields` otherwise is this array always empty!

### DIFF
--- a/modules/backend/widgets/Form.php
+++ b/modules/backend/widgets/Form.php
@@ -239,6 +239,8 @@ class Form extends WidgetBase
      */
     public function renderField($field, $options = [])
     {
+        $this->prepareVars();
+        
         if (is_string($field)) {
             if (!isset($this->allFields[$field])) {
                 throw new ApplicationException(Lang::get(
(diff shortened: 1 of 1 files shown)


LukeTowers:
@Rike-cz please post a detailed explanation of the bug you are encountering and how this fixes that.

Rike-cz:
When I try to render Backend\Widgets\Form field by field, I used in my backend controller this:
```
$widget = $this->makeWidget('Backend\Widgets\Form', $field_definition);
```
Fields are defined in `$field_definition` and correctly stored to `Backend\Widgets\Form::$fields`. After that I want something like:
```
echo $widget->renderField('somefield');
```
...to render first field, but ApplicationException is thrown. I discovered that reason is empty `Backend\Widgets\Form::$allFields` array, because is filled with `Backend\Widgets\Form::defineFormFields` method. So solution is to force call `Backend\Widgets\Form::defineFormFields` before accessing `Backend\Widgets\Form::$allFields` array. Simplest way to fix it is to place `Backend\Widgets\Form::prepareVars` as a first call in `Backend\Widgets\Form::renderField` method.

I am not sure if is it comprehensible...

PR #3813, opened 2018-09-23
Flag icons fix

w20k:


modules/system/assets/ui/less/flag.less
```diff
@@ -16,31 +15,30 @@
...
-  display: inline-block;
-  width: unit((4 / 3), em);
-  line-height: 1em;
-  &:before {
-    content: "\00a0";
-  }
-  &.oc-flag-squared {
-    width: 1em;
-  }
+   background-size: contain;
```
LukeTowers: indentation four spaces please

LukeTowers: No problem, just ping me when it's fixed :)

PR #3815, opened 2018-09-24
Add filters to log pages

gergo85:


modules/backend/models/accesslog/columns.yaml
```diff
@@ -3,23 +3,34 @@
...
         type: timetense
         width: 160px
 
+    type:
+        label: backend::lang.access_log.type
+        invisible: true
+
+    ip_address:
+        label: backend::lang.access_log.ip_address
+        searchable: yes
```
LukeTowers: Can you make this one invisible as well?

gergo85: Of course :)

gergo85: I made it.

gergo85:
@LukeTowers Is it correct?

modules/backend/models/accesslog/columns.yaml
LukeTowers: On second review I didn't notice that it was already present and visible; please leave this field visible by default.

gergo85: Now It is visible by default.

gergo85:
@LukeTowers Could I change something?

PR #3816, opened 2018-09-24
Filtering the list without session

panakour:
The current behavior of october loads a list page by filtering the list depend on the session of the previous selected value. 
That's pretty good but there is cases where we want the client to see the list page for the first load with the same filter value without putting it in the session.

This PR let the developer to set a filter with setting in which prevent the filter from saving to session.

For example the bellow filter `email_verified` prevents to save it to session and load the list page for the first time with the same value
```
    email_verified:
        label: Email Verified
        type: switch
        default: 2
        session: false
        conditions:
        - is_email_verified = '0'
        - is_email_verified = '1'

```

ps: this changes keep the behavior of october the same as before without any breaking changes.

### DIFF
--- a/modules/backend/widgets/Filter.php
+++ b/modules/backend/widgets/Filter.php
@@ -869,8 +869,10 @@ class Filter extends WidgetBase
             $scope = $this->getScope($scope);
         }
 
-        $cacheKey = 'scope-'.$scope->scopeName;
-        $this->putSession($cacheKey, $value);
+        if (!isset($scope->config['session']) || $scope->config['session']) {
+            $cacheKey = 'scope-'.$scope->scopeName;
+            $this->putSession($cacheKey, $value);
+        }
 
         $scope->value = $value;
     }


LukeTowers:
@panakour If you could add a PR to the docs documenting this I'll merge it :)

panakour:
@LukeTowers any update on this?

LukeTowers:
@panakour does this cause any issues where visually the filter would be applied but in actuality it would not? I.e. applying a filter with session: false and then applying another filter?

panakour:
@LukeTowers this does not cause any issue. it works as expected

LukeTowers:
@w20k could you test this?

w20k:
@LukeTowers, tested - works like expected, I guess. Not, really sure when it would be useful. It's not persistent to any action: refresh, changing page, going back from the details and so on.

LukeTowers:
@panakour @w20k makes a good point, this would also prevent it from persisting across pages in the list. What are your thoughts on that?

panakour:
Hmmm. I didn't think about it. I will reply as soon as possible with my thoughts...

w20k:
Maybe we should think of a better field name, not 'session: true|false'? Just an idea.

LukeTowers:
@w20k with the current behavior does it keep the filter visually selected when going to the next page? That would be undesired because then the filter would be visually active but not actually active with filtering the results. I would like a better name for the property if we can think of one

w20k:
@LukeTowers, It works the same, but the key difference is: it's not persistence to any page changes.
It is selected.

Example with back button:
dec-18-2018 19-44-24

w20k:
@LukeTowers, here you go.
dec-18-2018 21-59-24

LukeTowers:
Yeah, we can't have that. We'll need to prevent filters that don't have their value stored in the session  from retaining the visually selected value on pagination

LukeTowers:
On second thought, we can't accept this in it's current state. @panakour please rethink the original problem and propose a different solution to solve it, this solution is just going to cause problems.

PR #3818, opened 2018-09-24
Flag.less 4 spaces fix

w20k:


LukeTowers:
What's the difference between this and https://github.com/octobercms/october/pull/3813?

w20k:
@LukeTowers , could you close this one. I forgot to pull branch changes that's why there is an error with this pull request. Seems like it was a bad idea to make changes from the Hospital :)

PR #3824, opened 2018-09-26
Fix (Trait 'AssertGD\GDAssertTrait' not found)

gitlog:
$ vendor/bin/phpunit

Fatal error: Trait 'AssertGD\GDAssertTrait' not found in vendor/october/rain/tests/Database/Attach/ResizerTest.php on line 7


LukeTowers:
Thanks!

Issue #3823, opened 2018-09-25
Dashboard make it 12 instead of 10


ayumihamsaki:
Most CSS Grid systems use 12 because it makes the columns equal sizes, for example:

1 column = 12 spaces
2 columns = 6 spaces
3 columns = 4 spaces
4 columns = 3 spaces

However the Dashboard at the moment is really messy due to it having 10 spaces and giving the following:

1 column = 10 spaces
2 columns = 5 spaces
3 columns = Can't be done as whole number
4 columns = Can't be done as whole number

So I'm asking is it a good idea and would it be possible to change the dashboard to have a 12 column width spacing instead of 10. Then my dashboards would become less messy and not all over the place, which is happening at the moment due to this strange sizing of 10. I am sure a lot of people have noticed this as well.

LukeTowers:
Feel free to make a PR to implement that as an option if it doesn't already exist @ayumihamsaki

daftspunk:
This is merged. Thanks @ayumihamsaki !

PR #3825, opened 2018-09-26
Use 12 column layout in dashboard instead of 10 column layout

ayumihamsaki:
I have updated the grid system from 10 spaces to 12 spaces. I have also updated the jquery.isotope library from v1.5.26 to v3.0.6 in doing so it has fixed many bugs! This PR is related to github issue: https://github.com/octobercms/october/issues/3823


LukeTowers:
@ayumihamsaki I would prefer this was simply made available as an option, given that the existing sites all use 10 columns, I don't want to be suddenly breaking dashboard layouts.

ayumihamsaki:
@LukeTowers Do you want this as a plugin? And just update the library as there are errors when you resize the screen with the jquery.isotope library v1.5.26 so that would need to be updated for everyone anyway.

LukeTowers:
@ayumihamsaki no it could be a backend preference setting item perhaps? Or a file config item? The updated library should definitely be in core though.

LukeTowers:
No, keep the library update in this PR. And we'll wait for @daftspunk's input on what's the best way to make this configurable before moving forwards on that front.

ayumihamsaki:
Ok, thanks. 

For future reference, the updated library sorts out some issues on the responsive layout of the grids when you resize the screen resolution on the dashboard page.

ayumihamsaki:
This PR has been closed and will be broken up into different separate PR's.

Update Dashboard Columns from 10 to 12, see here: https://github.com/octobercms/october/pull/3965

Update jquery.isotope library from v1.5.26 to v3.0.6, see here: x

Issue #3826, opened 2018-09-27
Custom theme data doesn't support array fields


SeriousKen:
Custom settings on the theme doesn't support array fields, for example if you add the following to the demo theme's `theme.yaml`

``` yaml
form:
    fields:
        custom[forename]:
            label: Forename
        custom[surname]:
            label: Surname
```

It says:

> Unexpected type of array when attempting to save attribute "custom", try adding it to the $jsonable property. 

I've looked into this and it seems the `beforeSave` method is being called too late. I've created a fix by moving the code to the `model.saveInternal` event instead as this is fired before the validation that triggers the error.

SeriousKen:
Found a workaround. Use ThemeData::extend to add the relevant fields to jsonable.

LukeTowers:
That is the recommended way of accomplishing that :) I'll add it to the docs as a note though.

PR #3827, opened 2018-09-27
Fix for issue #3826

SeriousKen:
Moves code that prepares the attributes to an earlier event (`model.saveInternal`). See issue https://github.com/octobercms/october/issues/3826


SeriousKen:
Found a workaround so not needed.

PR #3831, opened 2018-09-28
Fixed supplying unavailable page number from session to paginate method in Lists Widget

slowpokefarm:
This PR is a fix for a bug that occurs when list page number is retrieved from Session, but records were deleted and therefore no records are available on this page and pagination is also broken.


bennothommo:
This was fixed by https://github.com/octobercms/october/pull/4020.

PR #3832, opened 2018-09-29
Update EventLog.php

sajjad-ser:
Pleas use Model alias because i overwrite this alias in my Persian plugin for persian date support

### DIFF
--- a/modules/system/models/EventLog.php
+++ b/modules/system/models/EventLog.php
@@ -2,7 +2,7 @@
 
 use App;
 use Str;
-use October\Rain\Database\Model;
+use Model;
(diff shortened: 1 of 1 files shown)


w20k:
Why can't you use like alias (use My\Full\Classname as Another;) ?

LukeTowers:
@w20k he means that he's overridden the Model facade to add Persian date support but it doesn't apply in this case because this file is referring directly to the base October\Rain model instead of going through the Model facade. Basically he needs every reference to `Model` to use the `Model` facade instead of directly referencing the October\Rain Model class in order for his plugin to work correctly.

Issue #2631, opened 2017-01-27
Repeaters add collapse / expand feature


ChVuagniaux:
Repeater is an very good feature but when you have many on a page this can be difficult to see them all and re-order them. For improve that I suggest to add buttons for can Collapse a repeater or all repeaters.

For add this feature I have do the following PR : #2632

### Collapse / Expand all items or just one

![ezgif com-b647557930](https://cloud.githubusercontent.com/assets/12028540/22365668/18c63050-e479-11e6-8b88-cd3144ee0038.gif)

---

### Realtime "Title" generation
System get the first Text item 

![ezgif com-d5c5b839bf](https://cloud.githubusercontent.com/assets/12028540/22366923/4b462764-e47f-11e6-9fee-162bda767f5b.gif)

---

### Easily reorder many items
![ezgif com-e51381fa6f](https://cloud.githubusercontent.com/assets/12028540/22367401/997a73c0-e481-11e6-8dcb-3cb60fba0017.gif)

## Compatibility
This fork fine with OctoberCMS but I have a display problem with the [Rainlab Translation plugin](https://github.com/rainlab/translate-plugin). The Translate plugin add switch lang button in absolute position at the same position that "Collapse all" "Expand all" buttons. 

For fix that I have do a PR (rainlab/translate-plugin#214) in the Rainlab Translation plugin repository.

surahmans:
Wow, That's great!

LukeTowers:
I like this a lot! Could you add the ability to configure how the repeater titles are built though? I'd like to be able to control that even if there isn't a text field within the repeater.

daftspunk:
Real nice work here. A couple of adjustments have been made (e7609aa5e8eef1342320fc0e7793cb88748b789f)

- Expand all / Collapse all have been removed
- Ctrl/Meta + click will collapse/expand all
- Nesting support has been tested + fixed
- Button styles are consistent
- Reordering behavior and appearance is fixed
- A "titleFrom" option has been added `data-title-from` allowing a form field name to be specified, used to source the title value (suggested by @LukeTowers)

iotch:
> Expand all / Collapse all have been removed

I wonder how to collapse all items on mobile

LukeTowers:
@iotch any recommendations for the best UX to have on mobile to achieve that?

iotch:
Sure! Keep those buttons and don't rely on actions not available on mobile devices (hover, control/command modifiers, etc.) :)

[...]

PR #3842, opened 2018-10-03
Repeater: toggle/collapse all on double click/tap

iotch:
Idea/Attempt on #2631


iotch:
This is quick example, implementing the same behavior on double-tap will require more efforts. And I don't have more time now, unfortunately.

LukeTowers:
Thanks @iotch, I'll leave this open until there's time to implement double tap as well. As a note, we'd need to continue to support the modifier keys + click method as well as that is what currently exists and what people are currently used to.

LukeTowers:

PR #3844, opened 2018-10-04
Fix filter popovers in modals

tobias-kuendig:
Since popovers and modals currently share the same z-index (600) a popover that gets created for a filter inside a modal is invisible. It is hidden below the modal.

This PR adds a check to see if a filter is created from within a modal window. If yes, the popover will be added to the modal container. Otherwise the default container value `false` is used.

image

### DIFF
--- a/modules/system/assets/ui/js/filter.js
+++ b/modules/system/assets/ui/js/filter.js
@@ -223,7 +223,15 @@
         var self = this,
             scopeName = $scope.data('scope-name'),
             data = this.scopeValues[scopeName],
-            isLoaded = true
+            isLoaded = true,
+            container = false
+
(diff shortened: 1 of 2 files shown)


LukeTowers:
@tobias-kuendig could you provide before and after GIFs to demonstrate the issue? Also, the assets will need to be recompiled with the latest dev branch.

tobias-kuendig:
Sure, here is the current behaviour. Note that I'm using the dev tools to increase the z-index of the popover to show that it is actually hidden below the modal.

not-working

This is the behaviour with this PR applied.

working

I did a rebase to bring this PR up to date with the latest changes from develop and recompiled the assets.

PR #3852, opened 2018-10-10
Change the delete buttons style to danger

gergo85:


w20k:
@gergo85 Could you also provide a screenshot before/after. Thanks!

gergo85:
Before:
before

After:
after

gergo85:
The remove button of Manage plugins page is also danger style.

PR #3885, opened 2018-10-22
PR for Date Modifed

Teranode:
Sorry for the late PR but for whatever reason, my local branch wont push to my remote fork so Ima have to do things a little manually here.


LukeTowers:
@Teranode there's still content_toolbar, page_toolbar, & partial_toolbar

PR #3886, opened 2018-10-23
Update October CMS from jQuery V2 to V3.3.1 and add JQuery Migrate

ayumihamsaki:
This PR combines the work done on the previous two PR's and put's everything into one place ready to be added to October CMS for review.


modules/backend/assets/js/vendor/jquery-migrate.min.js
```diff
@@ -0,0 +1,215 @@
+/*! jQuery Migrate v3.0.1 | (c) jQuery Foundation and other contributors | jquery.org/license */
```
LukeTowers: Are you sure this is the minified version of jquery-migrate?

modules/system/assets/ui/js/drag.scroll.js
```diff
@@ -399,7 +399,7 @@
     DragScroll.prototype.dispose = function() {
         this.scrollClassContainer = null
 
-        $(document).off('ready', this.proxy(this.fixScrollClasses))
```
LukeTowers: Is this the correct replacement? Note that this is using `.off` instead of `.on`. @w20k can you comment on this?

Teranode: @LukeTowers since you're disposing `DragScroll.prototype.dispose` you'll also want to unbind the event handlers that are attached to it.

modules/backend/assets/js/vendor/jquery-migrate.min.js
Teranode: yes this is the minified version of migrate: https://code.jquery.com/jquery-migrate-3.0.1.min.js

modules/system/assets/ui/js/drag.scroll.js
w20k: @LukeTowers , I guess it was a copy-past mistake :) @ayumihamsaki could you roll back '$(document).off('ready', fn);' From the 2.x version was removed ['on(ready, fn)'](https://jquery.com/upgrade-guide/3.0/#breaking-change-on-quot-ready-quot-fn-removed). Off + event still present.

w20k:
@LukeTowers , @ayumihamsaki @Teranode - question to everyone who is here ;) Why do we still want to add  Jquery.Migrate support? To have a backward compatibility?

ayumihamsaki:
@w20k 

> question to everyone who is here ;) Why do we still want to add Jquery.Migrate support? To have a backward compatibility?

To answer your question, my reason for wanting to add it is because of two reasons, the first being backward compatibility to October's Core Files and the second reason is to add backward compatibility to developers plugins.

modules/system/assets/ui/js/drag.scroll.js
ayumihamsaki: @LukeTowers Corrected resolve issues now for you to re-check.

w20k:
@LukeTowers, give me a day or so, will re-test core plugins, my own and few others just to be 200% sure.
For me, 'fully tested' - no ;)

w20k:
Could not find any critical issue (popped up once, but couldn't reproduce it. Something with the page rendering, might be the same thing with height), expect this one:
URL: backend/system/mailbrandsettings
```
jquery.min.js?v443:2 Uncaught TypeError: Cannot read property 'scrollHeight' of undefined
    at HTMLDocument.adjustPreviewHeight (mailbrandsettings.js?v443:16)
    at HTMLDocument.dispatch (jquery.min.js?v443:2)
    at HTMLDocument.y.handle (jquery.min.js?v443:2)
    at Object.trigger (jquery.min.js?v443:2)
    at HTMLDocument.<anonymous> (jquery.min.js?v443:2)
    at Function.each (jquery.min.js?v443:2)
    at e.fn.init.each (jquery.min.js?v443:2)
    at e.fn.init.trigger (jquery.min.js?v443:2)
    at HTMLDocument.triggerRenderOnReady (framework.js?v443:529)
    at l (jquery.min.js?v443:2)
```

Also, found UI bug with pagination -> 'loading message BG issue', UX bug -> in the settings menu, under 'Administrators'. Roles/User/Groups -> there is no 'Cancel' button. Only a single option: 'save|save and close'.

Teranode:
modules\system\assets\ui\vendor\flot\jquery.flot.navigate.js is ded

so i would recommend an alternative chart script.

ayumihamsaki:
I like D3 (see here: https://d3js.org/) they have so many projects that extend it.

ayumihamsaki:
@LukeTowers  @w20k @Teranode

Hi just reading the notes here once a day. Is this ready now to submit? Also the flot issue would that be a separate issue that we could try to replace with D3 (https://d3js.org/) at a later date?

w20k:
I guess we still need to tackle scrollHeight, or find out why it was present for me.

ayumihamsaki:
@w20k I would guess it's the code order making it `null`, why not just wrap it with 

```
$(document).ready(function() {
// ..
});
```

Teranode:
I cant do anything major regarding this til this is merged.
@w20k ill take a look to see if i can reproduce

w20k:
@Teranode -> Will re-test this one and post you back with a comment.

ayumihamsaki:
@w20k Sure can push that, does that code have a syntax error?

```
$(function () {
    var previewContents = $('#hhh').html();
    previewFrame = $('#hhh').get(0);

    createPreviewContainer(previewFrame, previewContents);
}); <-- Missing Semicolon
```

Correct to add the Semicolon? Before I push it.

w20k:
@ayumihamsaki , nope not needed. As I've understood, it's a part of OctoberCMS style to leave Semicolons (like: asset/ui/js/filter.dates.js ).

Most probably comma should be there or not, if it's a global object: 
```
var previewContents = $('#<?= $this->getId('mailPreviewTemplate') ?>').html() , <-- here
previewFrame = $('#<?= $this->getId('mailPreviewContainer') ?>').get(0)
```

ayumihamsaki:
@w20k Thanks for quick reply, pushed the update

Issue #3603, opened 2018-06-22
Improve styling on popup inspector for CMS components


ayumihamsaki:
The components pop-up is not really responsive and makes reading and using the components very hard and almost impossible to read.

## Full screen example:

![1](https://user-images.githubusercontent.com/17784082/41750968-b925f236-75b6-11e8-8ab2-e2418d03dd59.png)

Everything is squeezed and impossible to read.

Not responsive has a fixed width of 400px;

## Tablet Sized screen example:

![2](https://user-images.githubusercontent.com/17784082/41751025-09d13768-75b7-11e8-81d1-9b86089b7acf.png)

Pop-Up hanging out of the screen, really suprised no one has complained on this before?

## Mobile screen example:

![3](https://user-images.githubusercontent.com/17784082/41751107-97b8bde4-75b7-11e8-8f6b-208f54ae3cd8.png)

Please can other people comfirm if this is an issue with their CMS as well. Needs fixing badly to make responsive and easy to read.

(These screenshots were done on Google Chrome V69).

Seems to be these three files:

https://github.com/octobercms/october/blob/master/modules/system/assets/ui/js/popover.js

https://github.com/octobercms/october/blob/master/modules/system/assets/ui/storm.css

https://github.com/octobercms/october/blob/master/modules/system/assets/ui/storm.less

ayumihamsaki:
The code line is:

`<div class="control-popover control-inspector in placement-bottom" style="width: 400px; left: 675px; top: 338px;">`

I'm guessing it is injected using Javascript, as says:

```
element.style {
    width: 400px;
    left: 675px;
    top: 338px;
}
```

Then in Mobile has the following style css media:

```
@media (max-width: 480px)
body.popover-open .control-popover {
    overflow: auto;
    overflow-y: scroll;
    position: fixed;
    margin: 0;
    padding: 10px;
    width: 100% !important;
    z-index: 603;
    top: 0 !important;
    right: 0 !important;
    bottom: 0 !important;
    left: 0 !important;
}
```

That's all the styling I can see at present.

ayumihamsaki:
Request to have the following upgrades would seem better:

### Mobile

![1](https://user-images.githubusercontent.com/17784082/41768822-41db9484-7605-11e8-9b72-1a31f393551c.png)

### Desktop

![2](https://user-images.githubusercontent.com/17784082/41768620-aa031556-7604-11e8-8603-5d058a572542.png)

[...]

PR #3891, opened 2018-10-25
Fix positioning for popovers

Teranode:
in response to https://github.com/octobercms/october/issues/3603


LukeTowers:
@Teranode could you provide a before and after GIF of this happening and it being fixed? @ayumihamsaki perhaps you could if @Teranode doesn't want to?

Teranode:
@LukeTowers 

before: 


after:

PR #3892, opened 2018-10-25
Add auto focus to the first text field of forms

gergo85:


LukeTowers:
@gergo85 this is something that could be done by a plugin with a JS file, there's no need to add it in the YAML files.

PR #3897, opened 2018-10-27
Fix resize event after element removal

Teranode:
This prevents an error from popping up after the popover is hidden/removed `this.$container` is no longer available after removal so we need to unbind the `resize` event before that happens. Also gave the resize event a namespace so that it wouldn't unbind "all" `$('window')` resize events.

@LukeTowers :)

### DIFF
--- a/modules/system/assets/ui/js/popover.js
+++ b/modules/system/assets/ui/js/popover.js
@@ -117,7 +117,9 @@
         this.reposition()
 
         $(window).on('resize.popoverReposition', function() {
-            self.reposition()
+            if (self.$container) {
+                self.reposition()
+            }
         })
 
         /*
--- a/modules/system/assets/ui/js/popover.js
+++ b/modules/system/assets/ui/js/popover.js
@@ -41,6 +41,7 @@
     }
 
     Popover.prototype.hidePopover = function() {
+        $(window).off('resize.popoverReposition')
         this.$container.remove();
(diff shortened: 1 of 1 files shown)


LukeTowers:
@Teranode what happens if there are multiple popovers though? How about just add a check to the resize listener that checks to see if the contain exists before attempting to trigger the reposition?

Teranode:
I believe it will just create a new resize.popopverReposition event for the new element that utilizes the showing.oc.popover event

LukeTowers:
@Teranode but `$(window).off('resize.popoverReposition')` would remove all listeners to the `resize.popoverReposition` event would it not? Wouldn't that cause issues with multiple popovers on the same page removing each other? 

Do you have any thoughts on the check being added to the resize listener to see if the container exists before triggering the reposition?

Teranode:
wouldnt know how can you put it?

modules/system/assets/ui/js/popover.js
```diff
@@ -41,6 +41,7 @@
     }
 
     Popover.prototype.hidePopover = function() {
+        $(window).off('resize.popoverReposition')
```
LukeTowers: @Teranode you'll need to remove this part as well. And you can unnamespace the listener.

LukeTowers:
Perfect, thanks @Teranode!

PR #3898, opened 2018-10-27
fix hide event firing twice

Teranode:
the setTimeout fires after the modal already hides so it ends up firing twice which causes an error cause the element's no longer there

this is the error:
```
Uncaught TypeError: Cannot read property 'get' of null
    at Popup.triggerEvent (storm-min.js?v443:3732)
    at Popup.hide (storm-min.js?v443:3741)
    at storm-min.js?v443:3730
```

Present in the Master and Development branch
@LukeTowers :)

### DIFF
--- a/modules/system/assets/ui/js/popup.js
+++ b/modules/system/assets/ui/js/popup.js
@@ -274,7 +274,6 @@
         // Wait for animations to complete
         var self = this
         setTimeout(function() { self.setBackdrop(false) }, 250)
-        setTimeout(function() { self.hide() }, 500)
     }
 
     Popup.prototype.triggerEvent = function(eventName, params) {


LukeTowers:
@Teranode what's the first trigger that hides it? I don't see two calls in the code.

Teranode:
```
    /*
     * Only use the popup loading indicator if the handlers are an exact match.
     */
    $(document)
        .on('ajaxPromise', '[data-popup-load-indicator]', function(event, context) {
            if ($(this).data('request') != context.handler) return
            $(this).closest('.control-popup').removeClass('in').popup('setLoading', true)
        })
        .on('ajaxFail', '[data-popup-load-indicator]', function(event, context) {
            if ($(this).data('request') != context.handler) return
            $(this).closest('.control-popup').addClass('in').popup('setLoading', false).popup('setShake')
        })
        .on('ajaxDone', '[data-popup-load-indicator]', function(event, context) {
            if ($(this).data('request') != context.handler) return
            $(this).closest('.control-popup').popup('hideLoading')
        })
```
that gets called after you finish with the popup and the popup hides with a confirmation (that code only applies when its ajax) so "after" hiding it calls on
```
setTimeout(function() { self.hide() }, 500)
```
again if the request was xhr

w20k:
While testing 'modernize.js' upgrade, stumble upon this issue. 

Steps to reproduce (GIF):
nov-01-2018 14-00-07

Teranode:
@w20k that's what this PR fixes

w20k:
I know @Teranode ;) Just added GIF for steps, if needed.

LukeTowers:
@w20k did you test this fix as well?

w20k:
@LukeTowers - Yep, tested. No errors! 
I hope we could have core updated, soon, this 'min-width' issue in BuilderPlugin just making my crazy! :)

LukeTowers:
@w20k next core update will probably encompass all of the JS upgrades @ayumihamsaki is taking care of.

PR #3900, opened 2018-10-29
FormWidget TagList add handler onPaginationMore

gitlog:
At the moment we are faced with the problem of a complete selection of tags instead of partial. Because of this, the search for tags takes too much.
In this pool request, we implemented automatic tag loading by parts when scrolling through the list.

This version fully compatible to previous


LukeTowers:
@gitlog could you post a GIF of before and after? Also will need documentation on how to use this feature.

LukeTowers:

PR #3904, opened 2018-11-01
Modfied error maker with extra funcitonality

tschallacka:
For my integration unit tests I had some failing due to a fatal error.
The only way to detected this is to check if the `$controller->vars['fatalError']` had been set which is not very elegant.

This would be more elegant, by allowing a simple call `$controller->hasFatalError()` because the status of some controllers with fatal errors is still 200.

### DIFF
--- a/modules/backend/traits/ErrorMaker.php
+++ b/modules/backend/traits/ErrorMaker.php
@@ -17,6 +17,22 @@ trait ErrorMaker
     protected $fatalError;
 
     /**
+     * @return boolean Wether a fatal error has been set or not.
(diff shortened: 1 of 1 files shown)


modules/backend/traits/ErrorMaker.php
```diff
@@ -16,6 +16,22 @@
      */
     protected $fatalError;
 
+    /**
+     * @return boolean Wether a fatal error has been set or not.
```
LukeTowers: Typo, `Wether` should be `Whether`

Issue #3905, opened 2018-11-02
Disabled plugin's controllers are still accessible


alfreedom:
##### Expected behavior
The controller pages of a disabled plugin should not be displayed in the backend with the url of the page

##### Actual behavior
If you enter a page of a plugin using your url, show the page, regardless of whether the plugin is disabled

##### Reproduce steps
1. Go to a backend controller page of a plugin.
2. Copy the page url.
3. Disable the plugin in the configuration panel
4. Open the copied url again, the page will be show

##### October build
None

w20k:
@LukeTowers - confirmed.
Sidebar is removed, but with absolute URL it's accessible.

Settings:
![screenshot 2018-11-02 at 14 29 08](https://user-images.githubusercontent.com/1053320/47915516-c921a780-deab-11e8-9a36-9cc3d6f8d808.png)

Plugins Page:
![screenshot 2018-11-02 at 14 29 16](https://user-images.githubusercontent.com/1053320/47915511-c4f58a00-deab-11e8-9647-923a5f750c91.png)

w20k:
@LukeTowers, I will already started debugging ;)

w20k:
@LukeTowers, found the spot, but not yet sure about how plugins $id is generated (is it as simple as ucfirst($author) . '.' . ucfirst($plugin)? ).

It looks up for a plugin controller, but doesn't check if plugin is disabled or not.
https://github.com/octobercms/october/blob/master/modules/backend/classes/BackendController.php#L107

LukeTowers:
@w20k you should be able to use this: https://github.com/octobercms/october/blob/master/modules/system/classes/PluginManager.php#L505

```php
PluginManager::instance()->isDisabled($author . '\\' . $plugin);
```

That could probably go right after the list(author, plugin) call to prevent the further code from running

w20k:
@LukeTowers tried this way while testing. Didn't work. Namespace and pluginId should start with capitalized letter .

This worked like charm, will make a PR either today, or tomorrow.
```
if (!PluginManager::instance()->isDisabled(ucfirst($author) . '.' . ucfirst($plugin))){
....
}
```

w20k:
Found one strange behaviour, looking at it right now.

LukeTowers:
@w20k it should pass the request off to be handled by the CMS 404 page, or however a 404 in the backend typically works

w20k:
Sorry for taking so looong ;) 

[...]

PR #3923, opened 2018-11-15
Fix for not showing disabled plugins

w20k:
Related to issue: #3905

### DIFF
--- a/modules/backend/classes/BackendController.php
+++ b/modules/backend/classes/BackendController.php
@@ -1,15 +1,16 @@
 <?php namespace Backend\Classes;
 
+use Str;
 use App;
-use Closure;
-use Config;
 use File;
+use View;
+use Config;
+use Response;
 use Illuminate\Routing\Controller as ControllerBase;
 use October\Rain\Router\Helper as RouterHelper;
-use Response;
-use Str;
+use Closure;
 use System\Classes\PluginManager;
-use View;
+
 
 /**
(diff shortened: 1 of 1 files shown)


modules/backend/classes/BackendController.php
```diff
@@ -1,14 +1,15 @@
...
-use File;
-use View;
+use Closure;
 use Config;
-use Response;
+use File;
 use Illuminate\Routing\Controller as ControllerBase;
 use October\Rain\Router\Helper as RouterHelper;
-use Closure;
+use Response;
```
LukeTowers: Why did you change the order of imports? @w20k

w20k: No particular reason. Already thought about that, after commit.
But for me it's just default Intellij Idea formatting, when set to auto :)

LukeTowers: Can you revert it? October's import statements aren't organized that way so I see no reason to change them willy nilly :)

w20k: Sure thing :) Sorry, had a glass of wine before pull, so let it slip somehow :D

LukeTowers: No worries, take care of it when able :) I appreciate all your help on the project so far!

w20k: If you could be still available in few min will make PR update ;)

Issue #3903, opened 2018-11-01
Datepicker time positioning different to calendar


SeriousKen:
The position of the popup for the calendar positions itself for best fit in the viewport but the time popup is _always_ below the time field even if there isn't enough room. This creates an inconsistent experience for our users inputting dates.

LukeTowers:
@SeriousKen pictures please

SeriousKen:
The calendar uses autopositioning, looking at the code in Pikaday which is used for the calendar it uses its own popover code rather than bootstrap.

![calendar](https://user-images.githubusercontent.com/29771532/48052043-c28e7b00-e19e-11e8-85ae-778cb905d08e.jpg)

The timepicker uses the bootstrap popover always aligned to the bottom of the input.

![clock](https://user-images.githubusercontent.com/29771532/48052044-c28e7b00-e19e-11e8-9671-50e18ac39267.jpg)

w20k:
Found the culprit: `placement: 'bottom',` it's by default set to be `bottom` ;)

w20k:
Found a simple solution for now, needs more testing, but guess will work. 
Cheers ;)

LukeTowers:
@SeriousKen @w20k implemented a fix, I've merged it into develop.

PR #3927, opened 2018-11-16
Add TimePicker.placement = auto (only supports: top/bottom position)

w20k:
Fix list:
- add new option for TimePIcker.placement => auto
- set default placement for timePicker widget to be auto by default
- rebuild js assets  
Example:
nov-16-2018 21-02-20

Fix for issue: #3903 

Hi, @LukeTowers had to fix plugin itself, thought it's quite outdated and abandoned :) I hope it's suitable for OctoberCMS needs.


LukeTowers:
Please use the same indentation style as the source file, I think it uses tabs

LukeTowers:
@w20k turn on invisible characters in your editor, it will quickly become apparent :)

w20k:
@LukeTowers finally got it working 😄 The old the not bad, but ugly Vim helped ))

PR #3940, opened 2018-11-26
Colorpicker: support RGBA format

toymaster:
The current implementation of the colorpicker.js converts any custom colour into HEX format loosing opacity value even if showAlpha is enabled in the form field definition.
It is an annoying issue when RGBA values are actually required. The only workaround is adding an additional form field to keep opacity value. 
  
Proposed solution with minimal syntax and semantic changes to colorpicker.js:

1.  Create Sub-function covertColor
- convert custom colour object into RGBA string (toRgbString) if color._a is < 1 
- convert custom colour object into HEX string (toHexString) if not
2. Call covertColor when hide, move, change

I tested this on my stuff and it looks ok.

More to consider: 
1
Hide and Show configuration gives the same functionality, setting value, to buttons Ok, Close(x) and clicking outside the picker. If removing Hide and Show the picker behavior will be as normally expected - set value on OK, do nothing on Close/clocking outside.
2
If accepted hexColor js variable and data-hex-color attribute used further in the original code may look semantically misleading. 
3
If accepted Form Widget docs must be updated to specify showAlpha option for the Colorpicker, not only availableColors.


modules/backend/formwidgets/colorpicker/assets/js/colorpicker.js
```diff
@@ -57,24 +57,36 @@
...
                 }
             })
         }
     }
 
+    /**
+     * Convert colour to hex or rgba depnding on alpha (opacity) value
+     */
+    ColorPicker.prototype.covertColor = function(color) {
+        if (color._a < 1 ) {
```
w20k: If a color is undefined, it will throw an error.

Could you refactor this part? Start of by checking, if color is defined then alpha logic: `color._a < 1 ? color.toRgbString() : color.toHexString()`.

toymaster: Hi Denis,
1
I think verifying color looks a bit excessive. There should be something fundamentally wrong with 3rd-party Spectrum or Colorpicker partial, _colorpicker.htm, to get color undefined, etc. 1st is beyond the scope, 2nd looks ok so far. There is no verification for color in the original code as well.
 
2
Your shorthand code is perfect. I will teas and apply it. Thank you.

modules/backend/formwidgets/colorpicker/assets/js/colorpicker.js
```diff
@@ -57,24 +57,31 @@
...
+                    self.setCustomColor(customColor)
                 }
             })
         }
     }
 
+    /**
+     * Convert colour to hex or rgba depending on alpha (opacity) value
+     */
+    ColorPicker.prototype.covertColor = function(color) {
```
LukeTowers: typo, function name should be `convert`, not `covert`, unless you're saying that the color is now a spy.

toymaster: I fixed this, no conspiracy anymore

LukeTowers:
@toymaster what's the status on this?

toymaster:
I will check it again and update soon

toymaster:
@LukeTowers  you may play with tile background settings in my test plugin
https://linode.lilydaleheights.vic.edu.au/backend/lhc/pagepiling/tiles
un: test psw: 0ctober

I have made some changes since the initial request and use the code in a production web-site (http://go.lilydaleheights.vic.edu.au) for a few months without any issues. 

1. Functional change to colorpicker.js 
Sub-function covertColor converts custom colour object into either RGBA or HEX string depending on opacity value. It is called on hide, move and change events. 

2. Semantic changes to colorpicker.js/_colorpicker.htm
Since color may have non-hexadecimal value all hex-worded variables/attributes are renamed 
data-hex-color > data-color
hexColor > color
hex > colorValue

toymaster:
3. Functional change to colorpicker.js
option clickoutFiresChange: false to make sure color value is set only when clicking ok

Issue #3944, opened 2018-11-28
[Feature Request] set Project id from command line


obuchmann:
It would help to script october deployments if there would be an artisan command to set the project id.

Working on an PR.

obuchmann:
I mean to attach a project id, like the System\Controllers\Updates::onAttachProject() method does.

The implementation could be as following:

```
$manager = \System\Model\UpdateManager::instance();
$result = $manager->requestProjectDetails($projectId);

\System\Model\Parameter::set([
   'system::project.id'    => $projectId,
   'system::project.name'  => $result['name'],
   'system::project.owner' => $result['owner'],
]);
```

LukeTowers:
@obuchmann I'm fine with that. If you could make your PR to add it to `october:util` in the form of `php artisan october:util set project=$projectIdHere` that would be accepted.

PR #3946, opened 2018-11-28
Util set project Command

obuchmann:
implements feature #3944

### DIFF
--- a/modules/system/console/OctoberUtil.php
+++ b/modules/system/console/OctoberUtil.php
@@ -331,9 +331,6 @@ class OctoberUtil extends Command
         }
     }
 
-    /**
-     * Set ProjectId from october.com
-     */
     protected function utilSetProject()
     {
         $projectId = $this->option('projectId');


LukeTowers:
I don't think this works. If it's set to required then won't it require that option even when the util command isn't set project? Also, what would the full command signature be to trigger this?

obuchmann:
I'm sure, it works. Options are always optional. `InputOption::VALUE_REQUIRED` only means that if you specify that option an value has to be set via `=`, its no flag.

Of course you can specify the `projectId` option on all other util commands too. But they are not processed. This follows the way `debug` and `force` are implemented.

The signature is `artisan october:util set project --projectId=<id>`. I also added the signature as comment.

w20k:
@LukeTowers , @obuchmann is correct on value_required.

```
InputOption::VALUE_REQUIRED
This value is required (e.g. --iterations=5 or -i5), the option itself is still optional;
```

It could be via `space` or `=`, see [docs](https://symfony.com/doc/master/components/console/console_arguments.html) for examples.

LukeTowers:
Perfect, thanks for checking into that! Please document it here: https://github.com/octobercms/october/blob/master/modules/system/console/OctoberUtil.php#L29 and remove it from the method docblock.

modules/system/console/OctoberUtil.php
```diff
@@ -328,4 +330,28 @@ protected function utilGitPull()
...
 
+    /**
+     * Set ProjectId from october.com
+     * Call via: artisan october:util set project --projectId=<id>
+     */
+    protected function utilSetProject(){
+
+        $projectId = $this->option('projectId');
+
+        if(null == $projectId){
```
LukeTowers: please use `empty($projectId)` instead, and add a space after your `if`

modules/system/console/OctoberUtil.php
```diff
@@ -328,4 +330,28 @@ protected function utilGitPull()
         }
     }
 
+    /**
+     * Set ProjectId from october.com
+     * Call via: artisan october:util set project --projectId=<id>
+     */
+    protected function utilSetProject(){
```
LukeTowers: Opening brackets go on new lines for methods and classes

modules/system/console/OctoberUtil.php
obuchmann: resolved in https://github.com/octobercms/october/pull/3946/commits/32d9e14c2e3e2a4c15a3333325a1c6f33bc34c5b

modules/system/console/OctoberUtil.php
```diff
@@ -328,4 +331,27 @@ protected function utilGitPull()
         }
     }
 
+    /**
```
LukeTowers: remove this docblock

modules/system/console/OctoberUtil.php
obuchmann: resolved in https://github.com/octobercms/october/pull/3946/commits/32d9e14c2e3e2a4c15a3333325a1c6f33bc34c5b

modules/system/console/OctoberUtil.php
obuchmann: resolved in https://github.com/octobercms/october/pull/3946/commits/0680384f9a121b4e56a14d322f1f64723de238d2

Issue #3947, opened 2018-11-29
Improve visibility of code editor buttons in the backend


ayumihamsaki:
This issue has annoyed me for years and today, I had enough and decide to code the fix.

The issue see screenshot, below:

![1a](https://user-images.githubusercontent.com/17784082/49227605-5ab80280-f3e1-11e8-9b9a-5a8a3f40e234.png)

As you can see, you can not see the buttons in default mode as they are hidden in the code (very annoying!)

The old CSS code:

```
.field-codeeditor .editor-toolbar {
    position: absolute;
    top: 10px;
    right: 25px;
    z-index: 10;
}
```

My suggestion is to re-code it to make it exactly the same as the search box in the same place, therefore see the following screenshot example now:

![2a](https://user-images.githubusercontent.com/17784082/49227750-c4d0a780-f3e1-11e8-95d6-76c1bd5235ee.png)

CSS File Location: modules/backend/formwidgets/codeeditor/assets/css/codeeditor.css?v443

I will try to create a PR in a bit - if others think this is fine.

ayumihamsaki:
Update: Actually I might just move it across to the left a bit so it is not covering the scrollbar like the search box is doing:

```
.field-codeeditor .editor-toolbar {
    border-radius: 0px 0px 5px 5px;
        border-top-left-radius: 0px;
        border-top-right-radius: 0px;
        border-bottom-right-radius: 5px;
        border-bottom-left-radius: 5px;
    right: 25px;
    background-color: #ddd;
    border: 1px solid #cbcbcb;
        border-top-color: rgb(203, 203, 203);
        border-top-style: solid;
        border-top-width: 1px;
        border-right-color: rgb(203, 203, 203);
        border-right-style: solid;
        border-right-width: 1px;
    border-top-color: rgb(203, 203, 203);
    border-top-style: solid;
    border-top-width: 1px;
    border-right-color: rgb(203, 203, 203);
    border-right-style: solid;
    border-right-width: 1px;
    border-right-color: rgb(203, 203, 203);
    border-right-style: solid;
    border-right-width: 1px;
    border-top: 0 none;
    max-width: 125px;
    overflow: hidden;
    margin: 0;
    padding: 4px;
    position: absolute;
    top: 0px;
    z-index: 99;
    white-space: normal;
}
```

Giving the results:

(With Scrollbar)

[...]

PR #3948, opened 2018-11-29
Improve visibility of code editor buttons in the backend

ayumihamsaki:
PR for https://github.com/octobercms/october/issues/3947

Have tested it in various Browsers and Desktop, Tablet and Mobile all seem fine.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -43,7 +43,8 @@
     "require-dev": {
         "fzaninotto/faker": "~1.7",
         "phpunit/phpunit": "~5.7",
-        "phpunit/phpunit-selenium": "~1.2"
+        "phpunit/phpunit-selenium": "~1.2",
+        "meyfa/phpunit-assert-gd": "1.1.0"
     },
     "autoload-dev": {
         "classmap": [
--- a/modules/backend/formwidgets/codeeditor/assets/css/codeeditor.css
+++ b/modules/backend/formwidgets/codeeditor/assets/css/codeeditor.css
@@ -19,5 +19,7 @@
 .field-codeeditor .editor-toolbar > ul > li > a:hover > i,.field-codeeditor .editor-toolbar > ul > li > a:focus > i{opacity:1;filter:alpha(opacity=100)}
 .field-codeeditor.editor-fullscreen{z-index:301;position:fixed !important;top:0;left:0;height:100%;border-width:0;-webkit-border-radius:0;-moz-border-radius:0;border-radius:0}
 .field-codeeditor.editor-fullscreen .editor-code{-webkit-border-radius:0;-moz-border-radius:0;border-radius:0}
-.field-codeeditor.editor-fullscreen .editor-toolbar{z-index:302}
 .field-codeeditor.editor-fullscreen .ace_search{z-index:303}
+.field-codeeditor.editor-fullscreen .editor-toolbar{z-index:302;border-radius:0 0 5px 5px;right:25px;background-color:#ddd;border:1px solid #cbcbcb;border-right-color:#cbcbcb;border-right-style:solid;border-right-width:1px;border-top:0 none;max-width:125px;overflow:hidden;margin:0;padding:4px;position:absolute;top:0;white-space:normal}
+.field-codeeditor.editor-fullscreen .editor-toolbar > ul > li > a{color:#666}
+.secondary-tabs .editor-toolbar > ul > li > a{color:#fff}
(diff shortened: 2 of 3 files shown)


ayumihamsaki:
Sorry forget my last comment, with regards to that screenshot - that would not work when you go into fullscreen mode.

Actually I have this PR running on a test website all day. If I want to see the first line of code, I just switch from normal mode to full screen and the code is located in a different place, due to the two different screensizes.

ayumihamsaki:
Will experiment more for solution

w20k:
One of the options would be to move 'editor-toolbar' on top of nav-bar using css (hacky solution: .editor-toolbar{top: -30px;right: 35px;} and color white for an anchor tag): 

screenshot 2018-11-30 at 01 58 04

ayumihamsaki:
@LukeTowers @w20k I was thinking the exact same thing (but more of ripping out the HTML code and moving it properly) and also adding that grey bar to the top of the full screen mode. Giving Users more features like "Search" in full screen mode.

ayumihamsaki:
Note: z-index in the LESS file is 10px + 3px so that is why I changed the CSS to 13px to match it correctly (mistake in original file).

w20k:
@ayumihamsaki That would be awesome and logical to do. You could add PR for 'ace.js' and test it while you're dealing with this task, and make one more JS lib PR for me to speed up the process a bit.

ayumihamsaki:
@w20k @LukeTowers I think Luke can just complete this issue and then I will open a fresh new PR for the ace update and update any issues there. Makes things easier completing each issue one at a time.

With regards to testing, what about testing DPI? I've been wondering how October handles high to low and vice vera resolution devices.

w20k:
@ayumihamsaki Sounds good to me!

If you could write down steps, how could we test it or what's your idea behind - 'high DPI testing' (haven't done DPI testing myself. Don't have a knowledge on topic). I could try it out, or also add ideas ;)

ayumihamsaki:
PLEASE DONT COMPLETE THIS PR YET - I AM RUNNING MORE TESTS ON IT.

ayumihamsaki:
@w20k The reason for high DPI testing is that people say a normal computer has a 1 dpi and an ipad pro has a 2 dpi etc.

You can find a list here of all the sizes: http://screensiz.es/

For testing we use DPR see here for more details on testing:

https://developers.google.com/web/tools/chrome-devtools/device-mode/emulate-mobile-viewports#optional_controls_eg_touch_media_queries_dpr

LukeTowers:
What does it look like in full screen?

ayumihamsaki:
I tested in various modes all looks fine to me. My full screen default I use for testing is 1920x1080 as I think that is the most common one.

ayumihamsaki:
I was thinking, maybe add some rules for updating things such as leaving Version Numbers and/or where the update library location is for future people. For example, I got zero idea how to go ahead and update this file vendor inside the codeeditor found here:

https://github.com/octobercms/october/blob/master/modules/backend/formwidgets/codeeditor/assets/vendor/emmet/emmet.js

ayumihamsaki:
Still working on this pr .......

ayumihamsaki:
@LukeTowers @w20k 

Ok I have done a combination of things:

1. Light background (non- main coding area) the icons are dark grey color:

pr-1

2. Main Coding area (place the icons in the grey area and also set the icons to a white color:

pr-2

3. Full screen mode (added my original idea):

pr-3

Will update the pr for more testing.

w20k:
Perfect, for me @ayumihamsaki Lets wait for @LukeTowers opinion.

LukeTowers:
@ayumihamsaki does the full screen mode block the top right if you're looking at a minified file which would have text all the way to the end?

Otherwise looking good to me!

ayumihamsaki:
> does the full screen mode block the top right if you're looking at a minified file which would have text all the way to the end?

Yes it would block on the full screen mode at the top right.

I personally think the number of times that would cause a real world issue would be very small. It would only be an issue on the first line of code and if that was a problem, then just add an extra line or move the code down when editing and move it back up when finished.

Code is not the same as language in the sense code it is standardized using left to right (ltr) and language is left to right (ltr) and right to left (rtl). So that is why I think the problem would be very small (in fullscreen).

If you think its going to cause a big issue then maybe we should create a full icon bar above the fullscreen coding window at the top of the screen?

I leave you to decide on this.

LukeTowers:
I'm fine with your comments on it. Thanks!

PR #3950, opened 2018-12-01
Revert changes made to self.hide after ajax

Teranode:
This is needed as it removes leftover control-popover backdrops in any plugin backend that utilizes ajax in a popup modal (which is any of them that has relations or nested relations), so we will need to take a look at the dashboard widgets again.

### DIFF
--- a/modules/system/assets/ui/js/popup.js
+++ b/modules/system/assets/ui/js/popup.js
@@ -274,6 +274,7 @@
         // Wait for animations to complete
         var self = this
         setTimeout(function() { self.setBackdrop(false) }, 250)
+        setTimeout(function() { self.hide() }, 500)
     }
 
     Popup.prototype.triggerEvent = function(eventName, params) {


LukeTowers:
@w20k could you look into this and the original PR (https://github.com/octobercms/october/pull/3898) and see if you can recommend a fix that would work everywhere?

w20k:
@LukeTowers could we revert this change then, merge it with the dev branch? What I've understood from the @Teranode comments it's a critical issue for backend plugins that utilize ajax in popup modal.

A simple solution would be to test hide function if an object is still visible. Needs testing.
@Teranode if you have an example on your hands, could you push it to OC-Test? To make sure I won't break anything else 😄We really need JS tests...

Teranode:
dont have a gif example on me but consider this screenshot:


Basically the backdrop and the contents of the modal disappears, but the modal itself doesnt

, and with each nested modal, every time u save or update it will leave a modal block until you press esc to remove it completely

LukeTowers:
> A simple solution would be to test hide function if an object is still visible.

I'd agree with @w20k, we can do that to fix the original dashboard widget issue. Also, @Teranode could you add an example of nested popups to the oc-test-plugin repository for future testing?

w20k:
Got the fix. It's hilariously simple `if (!this.isOpen) return`. Though need to test plugins hidden logic.

PR #3953, opened 2018-12-02
Fix max & min items is repeater when using groups

c2webstudio:
Updated repeater when using the grouped functionality after adding the item was not removing the add more prompt if max-Items where reached.

### DIFF
--- a/modules/backend/formwidgets/repeater/assets/js/repeater.js
+++ b/modules/backend/formwidgets/repeater/assets/js/repeater.js
@@ -84,7 +84,7 @@
     }
 
     Repeater.prototype.clickAddGroupButton = function(ev) {
-        var  $self = this;
+        var $self = this;
         var templateHtml = $('> [data-group-palette-template]', this.$el).html(),
             $target = $(ev.target),
             $form = this.$el.closest('form'),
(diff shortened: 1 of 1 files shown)


LukeTowers:
I think the real issue is that https://github.com/octobercms/october/blob/master/modules/backend/formwidgets/repeater/assets/js/repeater.js#L45 is not getting fired in your specific case. Could you look into why that is @c2webstudio?

c2webstudio:
Yes @LukeTowers  the https://github.com/octobercms/october/blob/master/modules/backend/formwidgets/repeater/assets/js/repeater.js#L45 is not fired as in my case i am adding a group that fires the https://github.com/octobercms/october/blob/master/modules/backend/formwidgets/repeater/assets/js/repeater.js#L47

modules/backend/formwidgets/repeater/assets/js/repeater.js
```diff
@@ -84,6 +84,7 @@
     }
 
     Repeater.prototype.clickAddGroupButton = function(ev) {
+        var  $self = this;
```
LukeTowers: One too many spaces between `var` and `$self`

LukeTowers:
@c2webstudio fair enough :) added a comment, fix that and I'll merge it

PR #3956, opened 2018-12-03
Prevent plugins that cannot be instantiated from being loaded

bennothommo:
Per @LukeTowers [suggestion here](https://github.com/rainlab/builder-plugin/pull/261#issuecomment-443528752), this functionality wraps a portion of the `loadPlugin` function with a try..catch block. If, for any reason, the plugin cannot be instantiated (for example, if a parse error occurs or reserved keywords are used in the namespace), instead of bricking the site, this will simply prevent the plugin from being loaded.

### DIFF
--- a/modules/system/classes/PluginManager.php
+++ b/modules/system/classes/PluginManager.php
@@ -11,7 +11,6 @@ use Schema;
 use RecursiveIteratorIterator;
 use RecursiveDirectoryIterator;
 use ApplicationException;
-use System\Models\EventLog;
 
 /**
  * Plugin manager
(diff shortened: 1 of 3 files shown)


LukeTowers:
This is great @bennothommo! Could we either throw an exception or log an error when an invalid plugin is detected instead of just chugging along as if nothing happened though? What are your thoughts on that?

bennothommo:
@LukeTowers Good idea. I did try having some logging originally, however, when it was using a translatable string it caused all the system translation strings to remain untranslated. I can only assume that when the plugins are loaded, the translator isn't fully operational at that point. It does work when using a static string however.

LukeTowers:
I'm fine with a static string, at least then people would be able to google the issue. And yes, that issue does happen when attempting to use the translator before everything finishes loading, I've had to wrap code that uses the translator directly in App::before() to get it working properly but I don't think that would work in this case.

bennothommo:
Error log now in place :)

PR #3961, opened 2018-12-05
Add data-request-hide to hide items during request

firemankurt:
Adds ability to hide an item like a save button group during an ajax request.  For me, I needed a way to hide the save button while my dropdown change made a request that affected other form fields with AJAX results.  There is a delay during the request and I don't want user to submit form until the request completes because it changes what form fields are shown.


firemankurt:
Example of use:
<form data-request-data="id:'14'">
    <div class="form-group dropdown-field span-right">
        <label for="employee_id">Employee</label>
        <select 
          class="form-control custom-select" 
          id="employee_id" 
          data-request="onChoice" 
         data-request-update="morepartial: '#more_fields_target'"
          data-request-loading=".wait" 
          data-request-hide=".save_btn_grp" 
          name="employee_id">
              <option value="0">-- Employee Not Assigned --</option>
              <option value="102">BLOUIN, TYRELL - 341</option>
              <option value="138">BRAY, AUSTIN - 394</option>
              <option value="78" selected="selected">BRENNAN, MARK - 316</option>
              <option value="44">BRUBAKER, JEFF - 269</option
        </select>
    </div>

    <div id="more_fields_target"></div>

    <div class="form-group wait" style="display: none;">     
      Please wait!
    </div>

    <div class="form-group save_btn_grp" style="">
      <button type="submit" data-request="onSave" data-request-flash="">Save
      </button> or
      <button type="button"  data-dismiss="modal">Cancel
      </button>
    </div>

</form>

w20k:
Hi, @firemankurt why not just add a [data-attach-loading](https://octobercms.com/docs/ajax/extras#loader-button), which will do the same thing. Will prevent users from submitting form before response?

firemankurt:
data-attach-loading does not hide the save button.  It also does not prevent the submission of the form during long request while I am trying to show the user new relevant data from their select box choice. 
When the user sees the save button, they expect that the form is ready to save and that pressing it will do something.  Slow connections are a real problem for this case.

w20k:
@firemankurt could you update the docs and I'll merge this one! Thanks ;)

LukeTowers:
@firemankurt @w20k on second thought, why don't you just use the loading container method like https://github.com/octobercms/october/blob/master/modules/backend/controllers/users/create.htm#L19 and the `data-load-indicator` attribute? 

Response required before merge approved.

firemankurt:
If you look at the original request you will notice that that has been tried.  I need to hide items, not show them during load process.

LukeTowers:
@firemankurt I don't understand how what you're asking for is different from what already exists. Could you provide screenshots?

firemankurt:
data-load-indicator attribute does not work on frontend.

LukeTowers:
Anything that works on the backend can work on the frontend, you just need to include the relevant assets. I'm not actually sure what asset is responsible for data-loading-indicator though. 

It would still be helpful to see screenshots though

firemankurt:
Here is an example page that does not work with the methods you suggest. The layout has framework extras loaded.
I want the save button to hide when the drop-down selection changes.

[ajax.htm.txt](https://github.com/octobercms/october/files/3175964/ajax.htm.txt)

You will need to remove the .txt extension from this page file.

LukeTowers:
The file that handles the loading container is https://github.com/octobercms/october/blob/fb2aa1730cc08e8563908b54b70e942bd2988f5e/modules/system/assets/ui/js/loader.base.js which is included in storm.js. You would also need to include storm.css or https://github.com/octobercms/october/blob/fb2aa1730c/modules/system/assets/ui/less/loader.less directly for it to work.

firemankurt:
Sounds pretty heavy on requirements and more complicated then just adding in the more flexible and lighter code that I am submitting in this PR.  Please reconsider forcing me to use data-load-indicator instead of my simpler solution.

Samuell1:
I think better will be to allow passing callback in `loading` and then it can be more complex and useful in other ways.

Example:
`<button data-request="onSend" data-request-loading="$('.myDiv').hide()" data-request-complete="$('.myDiv').show()">`

LukeTowers:
Actually, why can't you just use `data-request-loading` and CSS to hide your button when your loading indicator is shown?

Samuell1:
@LukeTowers i think he cant because loading works in reverse it shows element not hides.

bennothommo:
@LukeTowers @firemankurt I don't mind this, as a nice counterpart to `data-request-loading`. It doesn't require the dev to load the "extras" framework assets, or require the user to override the CSS when using the "extras" framework and `data-attach-loading`.

firemankurt:
Thanks guys for considering my request. And thanks for the new rate at which PRs are getting processed. We in the community notice your hard work and commitment to October CMS.

daftspunk:
Very much appreciate your contribution @firemankurt and feedback from everyone in this thread. Also apologies for the delay in reviewing this. 

The solution makes sense, however, I'm not sure it is common enough that we should add it to the core. This can be solved by writing a relatively small piece of JS code and including it within your relevant controllers

```
    // Warning: Code is untested!
    $(document)
        .on('ajaxPromise', '[data-request][data-request-hide]', function() {
            var $target = $(this),
                requestHide = $target.data('request-hide')

            if ($.type(requestHide) == 'string') {
                requestHide = $(requestHide)
            }

            requestHide.hide()
        })
        .on('ajaxFail ajaxDone', '[data-request][data-request-hide]', function() {
            var $target = $(this),
                requestHide = $target.data('request-hide')

            if ($.type(requestHide) == 'string') {
                requestHide = $(requestHide)
            }

            requestHide.show()
        })

```

This is how the core and plugins handle these requirements when they arise, and they are often even more targeted, not needing to go so far as to define a generic API for it

daftspunk:
As an alternative solution, we could look at implementing `data-request-loading` in the backend, would could then be used to address #4729

firemankurt:
What ended up being the final decision on this issue? Was my solution implemented or is there a different way to hide a button group during an Ajax request that updates a form field based on another field value?

LukeTowers:
@firemankurt see https://github.com/octobercms/october/pull/3961#issuecomment-570867476. Current recommendation is to include that snippet on whatever controllers you need to have this ability.

daftspunk:
Yes, turns out there were plenty of these load indicator instances in the backend, that writing a custom one for your use case is the norm

Issue #3958, opened 2018-12-03
Update the dashboard area with improved downscaling


ghost:
I'm just going to open this issue, but I don't have time to fix it right now.

All the widgets in the dashboard for many plugins and the default ones are not responsive! If you open the dashboard screen in tablet mode or open chrome browser and open dev-tools and resize the screen you will see that there is no responsive code at all to any of the widgets.

Here's an example:

![example](https://user-images.githubusercontent.com/17784082/49383661-2c079800-f711-11e8-9c2a-86621cad2379.png)

Maybe someone could have a look at it, or next year I will fix the css code there.

For now I just open the issue.

LukeTowers:
@ayumihamsaki part of the problem is that it isn't that easy for developers to make the widgets responsive with breakpoints because I don't think you can target the width of the container holding your elements, just your screen width. Any thoughts on that?

ayumihamsaki:
@LukeTowers My initial thoughts is using flexbox and rem instead of px. Needs investigation, but I think can easily be upgraded, to make it easy for developers.

LukeTowers:
@ayumihamsaki feel free to experiment with it, I'm not so sure it'll be that easy. Some other thoughts are allowing different layouts at smaller screen sizes and / or having developers specifying a min width for their widgets

w20k:
Or as an option, we could help developers build 'dashboard widgets' the smart way. Provide them with guidelines or a set of tests (or a test tool) for different resolution/options.

ayumihamsaki:
@LukeTowers @w20k I strongly, I think I could code something for a baseline, which will make all the widgets responsive and developers wont need to do any extra work.

I throw out some ideas:

- Responsive fonts
- Media Queries for screen sizes
- Flexbox
- REM instead of PX
- Overflow

> Or as an option, we could help developers build a 'dashboard widgets' the smart way. Provide them with guidelines or a set of tests (or a test tool) for different resolution/options.

I think is also a good idea.

But October CSS Code is old and still using floats and old this still. So an upgrade would be a good idea.

(Slow upgrade doing each section).

ayumihamsaki:
p.s. 

"How to empower developers to make dashboard widgets more responsive"

I wouldn't say it totally developers fault, I'm sure its the October CMS base line css code fault.

w20k:
@ayumihamsaki , @LukeTowers - what do you guys think, if we could create a separate branch or rep. to test these ideas? And to find the best possible solution?

[...]

PR #3965, opened 2018-12-06
Update Dashboard Columns from 10 to 12

ayumihamsaki2:
This PR has been re-coded to split the old PR into two separate PR's. This PR relates to github Issue: https://github.com/octobercms/october/issues/3823

### DIFF
--- a/modules/backend/widgets/ReportContainer.php
+++ b/modules/backend/widgets/ReportContainer.php
@@ -49,7 +49,7 @@ class ReportContainer extends WidgetBase
      *         sortOrder: 1
      *         configuration:
      *             title: 'Traffic overview'
-     *             ocWidgetWidth: 10
+     *             ocWidgetWidth: 12
      */
     public $defaultWidgets = [];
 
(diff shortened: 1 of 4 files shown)


ayumihamsaki2:
If anyone asks why bother to create this PR, having a 12 pattern grid instead of 10 allows us to have equal numbers and create much more tidy Dashboards, see below a 12 pattern grid example:

12_column_grid_gold

ayumihamsaki:
I will quote Luke's comment:

>No, keep the library update in this PR. And we'll wait for @daftspunk's input on what's the best way to make this configurable before moving forwards on that front.

So I thought, Luke was going to talk to daftspunk on that.

Also while I am writing this comment the issue here: https://github.com/octobercms/october/issues/3958#issuecomment-449896342

Will be creating a 12-column demo test file and expanding this pr code in to that test file.

daftspunk:
Making this configurable as either 10/12 would add too much needless complexity. It's unlikely that it will break any dashboards beyond adding a 20% margin on the right hand side.

Let's merge this because it seems like an oversight in the original design. We'll make it the primary feature of Build 449 and people can update their dashboards accordingly.

Thanks!

Issue #3867, opened 2018-10-15
List of Updated Javascript Packages in October for Future Reference


ghost:
This is a question, but why are so many packages in October so out of date and there seems to be so many of them in this state.  The more I check the more worrying it makes me feel.

I am referring to all the Javascript packages mostly, although the CSS is very old as well.

With regards to the Javascript packages most are using jQuery V1 still and alot say their dated from as far back as 2006. Surely they must be very buggy and have lots of secury holes in them?

As for the CSS it contains lots of floats and no CSS Flexbox or CSS grids.

I just wondered is there a plan to update everything? Or is it being swept under the carpet as it seems like quite a big issue.

Honestly, it really does seem every Javascript package is years out of date. This is why lots of them are using Eval and causing that issue.

LukeTowers:
@ayumihamsaki simply put, because no one has bothered to update them. Feel free to submit PRs as desired, testing thoroughly for backwards compatibility.

ayumihamsaki:
@LukeTowers Will try my best and will periodically create PR's and link them to this issue, will take time due to backwards compatibility and conflicts, if other people could also try and HELP that would be great. I think there are well over 100 out of date JS files on a quick search through the master of this github.

If you want to make a plan for this let me know otherwise will just update things slowly ...

ayumihamsaki:
These paths:

```
https://github.com/octobercms/october/tree/master/modules/backend/assets/js
https://github.com/octobercms/october/tree/master/modules/backend/assets/vendor
https://github.com/octobercms/october/tree/master/modules/cms/assets/js
https://github.com/octobercms/october/tree/master/modules/cms/assets/vendor/jsdiff
https://github.com/octobercms/october/tree/master/modules/system/assets/js
https://github.com/octobercms/october/tree/master/modules/system/assets/vendor
https://github.com/octobercms/october/tree/master/modules/system/assets/ui/js
https://github.com/octobercms/october/tree/master/modules/system/assets/ui/vendor
```

...and any other Javascript file inside a **vendor** or **assets** folder.

LukeTowers:
@w20k is this something you're interested in helping with?

ayumihamsaki:
@LukeTowers @w20k Thanks.

ayumihamsaki:
Issue: Sweet Alert last updated 2015, by Daftpunk. There is no Version Number on anything found in this folder: https://github.com/octobercms/october/tree/master/modules/backend/assets/vendor/sweet-alert

[...]

PR #3966, opened 2018-12-06
Update jquery.isotope library from v1.5.26 to v3.0.6

ayumihamsaki2:
Update jquery.isotope library from v1.5.26 to v3.0.6. I could not find any special instructions to update from V1 to V3.

IE8 and 9 support has been dropped; should not be a problem because global usage for those two are at zero now.

This update is for github issue: https://github.com/octobercms/october/issues/3958

@w20k if you got any time would you like to run some tests on this please and thank you for all your help.


ayumihamsaki2:
I need to also ref this github issue: https://github.com/octobercms/october/issues/3867

ayumihamsaki:
Added Patch for Testing found here: https://github.com/ayumihamsaki/patches/blob/master/isotope/0001-added-to-create-patch.patch

w20k:
@ayumihamsaki, cloned himself 👍 Double work! Ooouu....no)))

Add it to my 'Weekends todo list' 😉

w20k:
Couldn't find any problems with the new Isotope.js version. Works faster and smoother. Refreshing and resizing seems much better in the updated version.

_Status_: **Done**

Issue #3425, opened 2018-03-04
Adding a Reset backend Search button or (X)


the-specialist2005:
New Suggestion to add a button to clear search field in the back-end instead of having to delete the text letter by letter in order to clear search filtration.

LukeTowers:
@the-specialist2005 good idea, I'd welcome that; it would also serve as a solution for an issue that occurs when the search is saved in the user's session / cookies but doesn't get populated in the search box so the records are filtered by an invisible search. Make a PR and then ping me on it to review it

vosco88:
@LukeTowers I've started to work on this one. So far I've put the button to the left of the search like below as not to have the input field too crowded when closed.
![image](https://user-images.githubusercontent.com/14372933/49343138-78e16500-f664-11e8-9124-95453c1bd9f9.png)
Other placement I was thinking is to replace the search icon with the 'X' icon but I did not want to loose the search icon. So wanted to check with you what would actually be preferable.

LukeTowers:
@vosco88 I don't mind having the X replace the search icon, what are your concerns with that?

vosco88:
@LukeTowers probably just overthinking it :) ok will work on that, will ping you when done

PR #3975, opened 2018-12-09
Add reset search (X) to toolbar search input

vosco88:
as referenced in #3425 

This is how it would look
image

@LukeTowers feel free to check this (as this is my first bigger PR I do not know if I have taken correct approach by using just the october js framework)
Right now the X is always visible - I was thinking to have it appear only when there is input present but that would not solve the issue with the invisible search that you mentioned.

### DIFF
--- a/modules/backend/widgets/search/partials/_search.htm
+++ b/modules/backend/widgets/search/partials/_search.htm
@@ -3,7 +3,7 @@
         placeholder="<?= $placeholder ?>"
         type="text"
         name="<?= $this->getName() ?>"
-        value="<?= $value ?>"		
+        value="<?= $value ?>"
         data-request="<?= $this->getEventHandler('onSubmit') ?>"
         data-request-complete="if($(this).val().length) {$(this).next().show()} else {$(this).next().hide()};"
         <?= !$searchOnEnter ? 'data-track-input' : '' ?>
@@ -11,9 +11,12 @@
         data-load-indicator-opaque
         class="form-control <?= $cssClasses ?>"
         autocomplete="off" />
-    <button class="clear-input-text" 
-            value=""
-			style="<?= empty($value) ? 'display: none;' : ''; ?>"
-			onclick="$(this).prev().val('');$(this).prev().request();"
-		><i class="icon-times"></i></button> 
+    <button 
+        class="clear-input-text" 
+        value=""
+        style="<?= empty($value) ? 'display: none;' : ''; ?>"
+        onclick="$(this).prev().val('');$(this).prev().request();"
+    >
+        <i class="icon-times"></i>
+    </button> 
 </div>
--- a/modules/backend/widgets/search/partials/_search.htm
+++ b/modules/backend/widgets/search/partials/_search.htm
@@ -5,7 +5,7 @@
         name="<?= $this->getName() ?>"
         value="<?= $value ?>"		
         data-request="<?= $this->getEventHandler('onSubmit') ?>"
-		data-request-complete="if($(this).val().length) {$(this).next().show()} else {$(this).next().hide()};"
+        data-request-complete="if($(this).val().length) {$(this).next().show()} else {$(this).next().hide()};"
         <?= !$searchOnEnter ? 'data-track-input' : '' ?>
         data-load-indicator
         data-load-indicator-opaque
--- a/modules/backend/widgets/search/partials/_search.htm
+++ b/modules/backend/widgets/search/partials/_search.htm
@@ -13,7 +13,7 @@
         autocomplete="off" />
     <button class="clear-input-text" 
             value=""
-			style="<?php if($value == '') echo 'display:none' ?>"
+			style="<?= empty($value) ? 'display: none;' : ''; ?>"
 			onclick="$(this).prev().val('');$(this).prev().request();"
 		><i class="icon-times"></i></button> 
 </div>
--- a/modules/backend/widgets/search/partials/_search.htm
+++ b/modules/backend/widgets/search/partials/_search.htm
@@ -10,4 +10,10 @@
         data-load-indicator-opaque
         class="form-control <?= $cssClasses ?>"
         autocomplete="off" />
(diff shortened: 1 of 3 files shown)


vosco88:
@w20k I have not yet commited the recompiled assets as I have some other things there I am testing and would need to clear them first, that might cause it not working properly. I just wanted to share this if I am going the right direction.

w20k:
@vosco88, let me know when I could start testing your PR. Thanks!

vosco88:
@w20k I've cleared my additional data and recompiled storm.css. Hopefully you should be able to test now. Thanks

w20k:
@vosco88, could you show clear button only when search input is not empty? For me, it just stays there 😄

vosco88:
@w20k I was thinking about it, but in the original issue Luke mentioned that sometimes the input does not get filled with the session data and the list is then filtered by an invisible search. In that case the empty input would me no reset search button. So I wanted to wait on what Luke thinks. But for sure I can implement it. Where would be a good place to put the javascript code for this? In which file?

w20k:
@vosco88, If that's the case you'd have to add clear button only when you do a 'search' request.

Not 100% sure, might wait for @LukeTowers on this one, but could be inside `system/assets/ui/js/filters.js`.

LukeTowers:
@vosco88 that's not actually an issue, the issue is that it does get populated however it's not immediately apparent to the user that the search box actually has stuff in it from the session that is affecting the search.

vosco88:
Oh, I understand now. So I might do it differently then. @LukeTowers  any hint on where to put the javascript code for showing the button only when there is an input filled? would it be fine as per suggestion above into the filter.js?

LukeTowers:
@vosco88 you could use the `oninput` HTML attribute on the search box itself to check it's value and hide / show the clear button based on that.

vosco88:
@LukeTowers used that, works ok, only thing that remains to check on page load if there is any input and show the button if there is. `onload` on the input does not work. Where could I place the document onload check for this? Thanks for the help

vosco88:
@LukeTowers why haven't I thought of that? :) ok now it is done

w20k:
@vosco88, could you add make left padding 1px bigger, I can see the search icon: `padding-left: 6px;`. And also remove outline from the button.

```    padding-left: 6px;

**Comment by vosco88** [comment]
@w20k Just done it.

**Comment by w20k** [comment]
@vosco88, looks good!

dec-13-2018 15-49-39

File: modules/backend/widgets/search/partials/_search.htm [hunk 1]
```diff
@@ -3,18 +3,17 @@
...
         data-load-indicator-opaque
         class="form-control <?= $cssClasses ?>"
         autocomplete="off" />
     <button class="clear-input-text" 
             value=""
-			data-request="<?= $this->getEventHandler('onSubmit') ?>"
-			data-load-indicator
-			data-load-indicator-opaque
-			data-request-success="$(this).prev().val('');"
+			style="<?php if($value == '') echo 'display:none' ?>"
```
LukeTowers: change to `style="<?= empty($value) ? 'display: none;' : ''; ?>"`

modules/backend/widgets/search/partials/_search.htm
```diff
@@ -3,18 +3,17 @@
         placeholder="<?= $placeholder ?>"
         type="text"
         name="<?= $this->getName() ?>"
-        value="<?= $value ?>"
+        value="<?= $value ?>"		
         data-request="<?= $this->getEventHandler('onSubmit') ?>"
+		data-request-complete="if($(this).val().length) {$(this).next().show()} else {$(this).next().hide()};"
```
LukeTowers: Use spaces, not tabs - and please remove trailing whitespace

modules/backend/widgets/search/partials/_search.htm
vosco88: done

modules/backend/widgets/search/partials/_search.htm
vosco88: hopefully also done

modules/system/assets/ui/less/toolbar.less
```diff
@@ -27,6 +27,23 @@
...
+				top: 2px;
+				background-color: #fff;
+				padding-left:6px;
+				height:30px;
+				border: 0;
+			}
+			
+		.clear-input-text:focus {
+			outline:none;
+		}
```
LukeTowers: Please fix the indentation on these styles, should be four spaces and inline with the others in the same nested context

modules/backend/widgets/search/partials/_search.htm
```diff
@@ -3,11 +3,17 @@
         placeholder="<?= $placeholder ?>"
         type="text"
         name="<?= $this->getName() ?>"
-        value="<?= $value ?>"
+        value="<?= $value ?>"
```
LukeTowers: Still has trailing whitespace

modules/backend/widgets/search/partials/_search.htm
```diff
@@ -3,11 +3,17 @@
...
         <?= !$searchOnEnter ? 'data-track-input' : '' ?>
         data-load-indicator
         data-load-indicator-opaque
         class="form-control <?= $cssClasses ?>"
         autocomplete="off" />
+    <button class="clear-input-text" 
+            value=""
+			style="<?= empty($value) ? 'display: none;' : ''; ?>"
+			onclick="$(this).prev().val('');$(this).prev().request();"
+		><i class="icon-times"></i></button>
```
LukeTowers: Please change to 
```html
<button
    class="clear-input-text"
    value=""
    style="<?= empty($value) ? 'display: none;' : ''; ?>"
    onclick="$(this).prev().val('');$(this).prev().request();"
>
    <i class="icon-times"></i>
</button>
```

vosco88:
@LukeTowers I hope it should be fine now, Sorry for so many changes needed.

PR #3976, opened 2018-12-12
Don't error out if multiple rows found in a hasOne relation

Flynsarmy:
## The Issue

The Lists widget will error out if you define a hasOne relation column but multiple related records are found in the database.

## In Depth
Say I have the following two tables:
```
foo (
  id
  bar_id
  name
)

bar (
  id
)
```
My _Bar_ model defines a _hasOne_ relation to the _Foo_ model and my Bar List is displaying the relation like so:
```yaml
foo:
    label: Name
    type: relation
    relation: bar_id
    select: name
```
This will work perfectly fine unless the _foo_ table has multiple records with the same _bar_id_ in which case the Lists widget will generate the following SQL:

```sql
select 
	`bar`.*, 
	(select name from `foo` where `bar`.`id` = `foo`.`bar_id`) as `name` 
from `bar` 
order by `id` desc
```
resulting in the SQL error:
> Subquery returns more than 1 row

The list will fail to render with no visible error to the end user - it merely says no records found even if there are hundreds of records in the DB.

## Solution

This PR just checks for a hasOne relation type and limits the subquery to the first result, fixing the SQL error by returning the first matching row. This PR is backwards compatible and will not break existing installations - in fact it may slightly improve efficiency on existing installations as the DB engine doesn't need to keep looking after it's found its first result.

### DIFF
--- a/modules/backend/widgets/Lists.php
+++ b/modules/backend/widgets/Lists.php
@@ -498,6 +498,10 @@ class Lists extends WidgetBase
                     ? DbDongle::raw("group_concat(" . $sqlSelect . " separator ', ')")
                     : DbDongle::raw($sqlSelect);
 
+                if ($relationType == 'hasOne') {
+                    $countQuery->limit(1);
+                }
+
                 $joinSql = $countQuery->select($joinSql)->toSql();
 
                 $selects[] = Db::raw("(".$joinSql.") as ".$alias);


LukeTowers:
@Flynsarmy under what conditions would that occur? If you have defined a hasOne relationship that could theoretically return more than one result, then I would argue you have a problem with your data or relationship definition. Could you explain more about the real world situation encountered?

Flynsarmy:
Users are idiots. I'm writing a plugin that lets you set up Channels (essentially tables) then attach Fields to them (think the Builder plugin but entirely dynamic). If they set up a belongsTo relation on one side, and a hasOne on the other, it's very easy to relate a record twice by accident as there's no enforcement on the belongsTo side that they're only ever selecting something once. I'd rather the Lists widget work with what it has rather than just silently error out leaving the end user scratching their head wondering if they've suddenly lost all their data because the widget is now saying 'No records found'.

In any case, even if you thought my use case was too edge, the PR is a performance improvement in and of itself for existing, correct configurations as mentioned in my PR description.

LukeTowers:
@Flynsarmy hmm, I'm still not convinced this is the best way of doing it. It might be a performance improvement (I'd have to see numbers on that), but I'm mostly concerned about the incorrect behaviour silently working. Let's say that they do have the relationship misconfigured and they've now got incorrect data. I don't think just picking what is essentially a random related record to be the authoritative record for that relationship (especially when if they continue with their setup that record would change based on more data being entered). This problem is probably better solved with validation / throwing informative exceptions. Thoughts?

Flynsarmy:
That's definitely an option too. A bit beyond me though unfortunately so I won't send a PR for making it happen if you decided to go that route.

LukeTowers:
@Flynsarmy I meant that the validation would occur on your end within your plugin 😉

LukeTowers:

PR #3980, opened 2018-12-13
Datatable dropdown usability tweaks

bennothommo:
Adds a couple of usability tweaks to the dropdown cell type in the data table widget, to more closely mimic a native dropdown field.

1. Pressing the up or down arrow keys when the cell is focused but with the dropdown closed will select the previous or next item automatically and set it as the cell value. This does prevent the usual table function of going to the previous or next row when focused on a dropdown cell, but I think it's a worthwhile trade-off. When the dropdown is open, the up and down arrows work the same as previously implemented.

2. Typing out characters will initiate a search and select the first matching option automatically and set it as the cell value. For example, for the following options:
- Apples
- Oranges
- Bananas
Typing out `o` and `r` on the keyboard will automatically select the Oranges option.

### DIFF
--- a/modules/backend/widgets/table/assets/css/table.css
+++ b/modules/backend/widgets/table/assets/css/table.css
@@ -395,9 +395,7 @@ html.cssanimations .control-table td[data-column-type=dropdown] [data-view-conta
   cursor: pointer;
   outline: none;
 }
-.table-control-dropdown-list li:hover,
-.table-control-dropdown-list li:focus,
-.table-control-dropdown-list li.selected {
+.table-control-dropdown-list li:focus {
   background: #34495e;
   color: white;
 }
--- a/modules/backend/widgets/table/assets/js/build-min.js
+++ b/modules/backend/widgets/table/assets/js/build-min.js
@@ -279,7 +279,7 @@ if(ev.keyCode==68&&ev.altKey&&this.options.deleting){this.deleteRecord()
 this.stopEvent(ev)
 return}
 for(var i=0,len=this.options.columns.length;i<len;i++){var column=this.options.columns[i].key
-this.cellProcessors[column].onKeyDown(ev)}
+if(this.cellProcessors[column].onKeyDown(ev)===false){return}}
 if(this.navigation.onKeydown(ev)===false){return}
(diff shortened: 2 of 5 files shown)


LukeTowers:
Could you provide some gifs of this in action?

bennothommo:
Sorry for the delay - trying to work out video editing on Linux :P

Example of pressing the up and down arrow keys on the dropdown cells in a data table:
Up Down Arrow Example

Example of typing out the specific data type in the dropdown cells:
Typing Example

bennothommo:
@w20k Done :)

LukeTowers:
@w20k let me know when you're done testing this. @bennothommo looks great, this is something that's bugged me for a long time!

w20k:
@LukeTowers, will have to look through the commit and that should be it. Tomorrow, I guess 😉

w20k:
@bennothommo found few issues with the drop-down.

**Steps**:
- Open Dropdown, search for closest of the selected element and then move your mouse to select any
**Selectable Madness:**
_(Selected - Integer, searched for a Date. Pressed D-A, and mouseover to Timestamp)_
screenshot 2018-12-15 at 13 30 29

@bennothommo, one more thing, if you don't mind. Could you look at why you can not close drop-down if you select outside of the list? Thanks!

bennothommo:
@w20k Thanks heaps for the feedback. I've changed the behaviour of the dropdown somewhat - now only one item in a dropdown list will be focused at any one time, depending on the last input method used. For example, if you hover over the items, the hovered item will be "focused" and highlighted. If you then press the up or down arrow, or start typing in a search query, the focus will then shift away from the hovered item to the respective new item. This works similar to native dropdown menus.

Example:
Dropdown new behaviour

With regards to the dropdown not hiding when clicked outside the cell, it does hide when you click another cell, but doesn't hide if you click the toolbar or the empty space underneath (even though both are a part of the "table" element). I've added an "unfocusTable" to the click event of the table widget if you click outside its cells, which should hopefully resolve this issue.

w20k:
Awesome! Thanks, @bennothommo!

@LukeTowers, Testing Status - ✅

Issue #1021, opened 2015-03-24
Create a z-index manager


tresbach:
While editing something on richeditor inside a relation form popup, clicking on the "table button" renders the "insert table" popup behind the relation popup.

trying to insert a table:
![captura de tela de 2015-03-24 13 08 08](https://cloud.githubusercontent.com/assets/774244/6806434/6cda6ece-d227-11e4-8a47-79dab3a50549.png)

after closing the relation form:
![captura de tela de 2015-03-24 13 08 13](https://cloud.githubusercontent.com/assets/774244/6806435/6cddee1e-d227-11e4-9f4c-b48b3dce864e.png)

daftspunk:
There are conflicting z-index priorities in the backend. The above is a good example of one. We need to develop a z-index manager that can assign frequencies to layered elements.

daftspunk:
See #992 for an issue with the Color Picker

7talents:
For anybody who needs an immediate fix (hacky): you put write your own css file and load it in your plugins controller. The z-index of the relation modal ist 1050, so you will have to change the z-index of the <code>#redactor-modal-overlay</code>, <code>#redactor-modal-box</code> and <code>#redactor-modal</code> to something higher.

Since the redactors css gets loaded after your plugins css, you will have to tweak your own css a bit so that it overrides the original css. Since the original css uses <code>!important</code> already, you will have to use it as well, and prepend the id with body so that your css will take precedence even though it was loaded prior to the original. 

In short: this is the css that will work:
<code>body #redactor-modal-overlay, body #redactor-modal-box, body #redactor-modal{z-index:1060 !important}</code>

tplessis:
Thx @7talents for the rapid fix. Works well.

daftspunk:
Related issue: #1145 - Test for color picker neede

daftspunk:
The z-indexes have been reorganised via the ["Windex" system in Storm UI](https://github.com/octobercms/october/blob/develop/modules/system/assets/ui/less/windex.variables.less
). It appears to have resolved most, if not all issues. A z-index manager based on JS may still be required in future if requirements become more complex. Downgrading this issue until then.

PR #3996, opened 2018-12-20
Colorpicker: value input support in modals

toymaster:
**Issue**
When using colorpicker field in the relationship manager modal window it is impossible to set value manually in the colour input field because it is locked. This make the widget useless if you need to set a custom colour in the modal window.

**Background**
The modal container has the tabindex attribute set to "-1" to prevent access to items outside the model using tab. Colorpicker creates a parent container (.sp-container) to place custom colour selection code including colour value field. By default it is appended to body. _Since it is outside the modal window it cannot be edited because of tabindex value._
The issue is a part of Spectrum, which is sitting behind Colorpicker. [They recommend to use appendTo option when working with modal dialog.](https://bgrins.github.io/spectrum/#options-appendTo) However appenTo option is not configurable through October Colorpicker. 

Note: the issue has no relation to z-index definitions as in [992](https://github.com/octobercms/october/pull/992) or [1021](https://github.com/octobercms/october/issues/1021)

**Proposed Behavior**
By default Colorpicker container is appended to body. If relationship manager modal dialog is detected the container will be appended to the modal container. If more complicated modal structure involved developer may define appendTo option in the appropriate Colorpicker field of field.yaml.
```
fields:
	bg_colour:
		label: 'Select Colour'
		showAlpha: 1
		appendTo: '#mymodal'
		type: colorpicker
```
Setting appendTo is not required for standard October backend environment including relationship manager modal. It will be handled behind the scene.

**Solution**

1. Update colorpicker.js, ColorPicker.php and _colorpicker.htm to support appendTo option
2. Use a special function in colorpicker.js to set standard or custom parent container

**Testing**
It is pretty simple update. I have tested it on my stuff and it works reliably. The solution has no conflicts with my previous pull request [3940](https://github.com/octobercms/october/pull/3940). 
With both updates, the colorpicker is going to be finally usable in any backend context.

### DIFF
--- a/modules/backend/formwidgets/ColorPicker.php
+++ b/modules/backend/formwidgets/ColorPicker.php
@@ -43,6 +43,12 @@ class ColorPicker extends FormWidgetBase
      */
     public $showAlpha = false;
 
+    /**
+     * @var string  Append Colorpicker to a parent container
+     * Use css selector to define the container
+     */
+    public $appendTo = 'body';    
+    
     //
     // Object properties
     //
@@ -61,6 +67,7 @@ class ColorPicker extends FormWidgetBase
             'availableColors',
             'allowEmpty',
             'showAlpha',
+            'appendTo',
         ]);
     }
 
@@ -83,6 +90,7 @@ class ColorPicker extends FormWidgetBase
         $this->vars['availableColors'] = $availableColors = $this->getAvailableColors();
         $this->vars['allowEmpty'] = $this->allowEmpty;
         $this->vars['showAlpha'] = $this->showAlpha;
+        $this->vars['appendTo'] = $this->appendTo;
         $this->vars['isCustomColor'] = !in_array($value, $availableColors);
     }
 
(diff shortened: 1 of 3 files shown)


w20k:
@toymaster, could you also provide either steps to reproduce or screenshots(gifs): before/after.

toymaster:
If you want to reproduce the issue you need to take quite a few steps to build a working relation manager and colorpicker field in the form of the belonging model.

The nature of the issue between Spectrum and Bootsrap modals is discussed [here](https://github.com/bgrins/spectrum/issues/150) or [here](https://github.com/bgrins/spectrum/issues/161)
It was just overlooked when colorpicker was created.

I have attached a gif to show 2 colorpicker fields in the relation manager modal dialog. 

The left one is rendered inside the modal container after applying the fix and I can edit input value manually or with controls. 

The right one is rendered outside of the modal container (out of box behaviour) and I cannot manually edit input value. I still can edit it using controls but it is practically impossible to set a precise value in this way. It makes colorpicker hardly usable in modal windows.

colorpicker-resized-github

By the way, this gif shows as well that colorpicker saves rgba values as per [3940](https://github.com/octobercms/october/pull/3940). This does not happen out of box.

w20k:
@toymaster, could you also rebuild JS assets.

toymaster:
> @toymaster, could you also rebuild JS assets.

I made minimal necessary changes to colorpicker.js. Could you clarify what you mean by rebuilding JS assets?

LukeTowers:
Fixed by https://github.com/octobercms/october/pull/4199

PR #4010, opened 2018-12-25
In-browser filtering/searching of files in MediaManager

MilosStanic:
By applying this change, MediaManager would filter (search) among the already loaded files in the current directory that are listed in the DOM, instead of polling the server for each search term change. 
In my use case, when searching through a directory that contains more than a thousand files, the server breaks.

Edit:
Just to point out, my server is a dedicated Core i7 with 8 cores, 32GB RAM and SSD disk, and it breaks on filtering files. So this is much more efficient.

### DIFF
--- a/modules/backend/widgets/mediamanager/assets/js/mediamanager-browser-min.js
+++ b/modules/backend/widgets/mediamanager/assets/js/mediamanager-browser-min.js
@@ -335,7 +335,9 @@ if(this.lastSearchValue!==undefined&&this.lastSearchValue==value)
 return
 this.lastSearchValue=value
 this.clearSearchTrackInputTimer()
-this.searchTrackInputTimer=window.setTimeout(this.proxy(this.updateSearchResults),300)}
+var regex1=new RegExp("^((?!"+value+").)*$")
+this.$el.find('[data-item-type="file"]').show()
+this.$el.find('[data-item-type="file"]').filter(function(){return $(this).attr('data-public-url').match(regex1)}).hide()}
 MediaManager.prototype.deleteItems=function(){var items=this.$el.get(0).querySelectorAll('[data-type="media-item"].selected')
 if(!items.length){$.oc.alert(this.options.deleteEmpty)
 return}
--- a/modules/backend/widgets/mediamanager/assets/js/mediamanager.js
+++ b/modules/backend/widgets/mediamanager/assets/js/mediamanager.js
@@ -893,7 +893,11 @@
 
         this.clearSearchTrackInputTimer()
 
-        this.searchTrackInputTimer = window.setTimeout(this.proxy(this.updateSearchResults), 300)
+        var regex1 = new RegExp("^((?!" + value + ").)*$")
+        this.$el.find('[data-item-type="file"]').show()
+        this.$el.find('[data-item-type="file"]').filter(function(){
+            return $(this).attr('data-public-url').match(regex1)
+        }).hide()
     }
 
     //


w20k:
@MilosStanic, could you also provide compiled assets. (`php artisan october:util compile js`)

MilosStanic:
@w20k Yes, of course, I just don't know how to add them to this pull request. 
Can you help me? Not really masterful with git.

w20k:
@MilosStanic, just clone your branch to your local machine and run command. 
1. Clone your forked repo. 
2. Switch to your `patch-3` branch. 
3. Compile assets.
4. `git add YourCompilledFile`, if you don't have any changes, despite compiled - just `git add .`
5. `git commit -m "your message` & `git push`

w20k:
@MilosStanic, I'll revise your solution in Jan. 
I do think we should re-use or create a separate function to call, as it was with `updateSearchResults()`. But before that, I've to test it locally 😉

MilosStanic:
I apologize for posting to wrong branch. 
@w20k I really don't have an opinion on how you should implement this solution. It's a matter of application design, and I surely don't know anything about it. 
I just had a pressing issue, and I had to make a solution. So I thought it would be nice to contribute back to the project that brought me so much. 
Thank you all again for all your hard work.

w20k:
@MilosStanic, I was just thinking out loud 😄No worries)) 
Thanks a lot for your contribution!

w20k:
@bennothommo @LukeTowers what do you guys think if we could add a checkbox for MediaManager to switch from front-end search to back-end? Would be useful? 
For now, back-end could query through all your folders and subs, where Front-end searches only for files in the current directory.

bennothommo:
@w20k I'm not sure about a checkbox on the front-end, but certainly a configuration value might be a good idea - with the idea that the developer is given the option to use frontend or backend searching, knowing the pros and cons of both.

LukeTowers:
@w20k @bennothommo how about a client side way of triggering a full serverside search if their clientside search didn't return what they wanted? Something like the reload button or if there's something better suited UX wise.

Samuell1:
@LukeTowers Maybe better way will be to cache content of media folder to array and then search trough it not trough real files, folders? and refresh button to refresh cache?

LukeTowers:
@Samuell1 I like the idea of a cached paths array, that way we solve the original problem of the search being slow without adding another search layer on. Care to take a stab at implementing one?

MilosStanic:
Well, my pull request is actually "harvesting" the files listed in the controller via DOM, and then filtering through it.
So it is already implementing the search through filtering DOM elements (files), which are kind of cached in the DOM. 
Adding a refresh button that would redownload the files list and rerender it would solve the problem of keeping the list of files up to date if need be.

bennothommo:
You all might be on to something here...

I like @MilosStanic's way of checking the DOM first to search files that are in the current list. If the person cannot find what they are looking for there, perhaps above or underneath the results list, it could give the user a link to "Include sub-folders" in the search. This search would trigger the server-side search and look everywhere for the user's queried file. In this case, @Samuell1's suggestion of caching the file structure would be very useful.

w20k:
Will assign it to me, for now, want to play around with ideas and provide few options. From UX perspective re-running a query doesn't sound a proper approach to me and checkbox as well.

daftspunk:
Thanks all. This is a novel idea, however, it does seem to have too many trade-offs in the current format. I've made a note of it in our internal tracking system and this will be addressed in the next iteration of the media manager

PR #4020, opened 2018-12-30
Redirect user to last available page in list widget if current page is unavailable

### LINKED ISSUE #4008: List Widget: unable to achieve previous page when last record is deleted
##### Expected behavior
When I delete last record in current page (page > 1), i expect:
1. automatical redirect to previous page OR
2. any navigation button to achieve previous page.

##### Actual behavior
I see empty record list and I have no way to achieve previous page. Maybe delete cookie only.

##### Reproduce steps
lists-widget

##### October build
445

bennothommo:
Fixes #4008.

If the user's current page number (via session or via POST) is no longer available in a list widget, this fix will override the current page number to use the last available page number instead.

This should prevent people from being unable to navigate to a previous page of records if they delete the last record in a page or change the number of records per page.

### DIFF
--- a/modules/backend/behaviors/ImportExportController.php
+++ b/modules/backend/behaviors/ImportExportController.php
@@ -210,7 +210,7 @@ class ImportExportController extends ControllerBehavior
         catch (Exception $ex) {
             $this->controller->handleError($ex);
         }
-        
+
         $this->vars['sourceIndexOffset'] = $this->getImportSourceIndexOffset($importOptions['firstRowTitles']);
 
         return $this->importExportMakePartial('import_result_form');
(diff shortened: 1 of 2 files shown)


LukeTowers:
@bennothommo I've just noticed that the `prepareModel` method is horribly named. Could you rename it (and relevant uses of `$model` holding it's return value) to `prepareQuery()`? We'd also need the prepareModel method retained but with a deprecated warning like in https://github.com/octobercms/october/blob/master/modules/cms/classes/MediaLibrary.php#L20. If you think it would make more sense, make the naming changes in a separate PR and then rebase this PR on top of that work instead.

bennothommo:
@LukeTowers Yep, agreed. Will make the changes in a separate PR.

PR #4029, opened 2019-01-02
Finishes out the 'october:util purge uploads' placeholder function

austinderrick:
Finishes out the _october:util purge uploads_ placeholder function.


modules/system/console/OctoberUtil.php
```diff
@@ -294,8 +295,47 @@ protected function utilPurgeUploads()
...
+            if ($files = File::glob($targetDir.'/*')) {
+                foreach ($files as $file) {
+                    if (!is_file($file)) {
+                        continue;
+                    }
+                    
+                    $model = FileModel::where('disk_name', basename($file));
+                    if ($model->count()) {
+                        continue;
+                    }
```
LukeTowers: No need to have `$model` as a separate variable here, and besides, it's not a Model, it's a QueryBuilder. Please just call the `->count()` directly on the where and put it all in the if statement

austinderrick: Updated!

LukeTowers:
@bennothommo is this something you'd be interested in testing / reviewing?

modules/system/console/OctoberUtil.php
```diff
@@ -294,8 +295,46 @@ protected function utilPurgeUploads()
...
+                    
+                    if (FileModel::where('disk_name', basename($file))->count()) {
+                        continue;
+                    }
+                    
+                    $this->info('Purged: '. basename($file));
+                    $totalCount++;
+                    @unlink($file);
```
bennothommo: @austinderrick Could you check if the file is writable and can be deleted before saying it is purged and trying to unlink it, and indicate to the user if it could not be deleted? It is currently saying it has purged files that may not have actually been removed.

bennothommo: Also, if you don't mind, could you also make the same change to the `utilPurgeThumbs()` method? :)

bennothommo:
@austinderrick @LukeTowers Works well, although I did notice that it does not delete folders that have been emptied, so it can potentially leave a lot of orphaned directories.

LukeTowers:
@austinderrick @bennothommo could we use `deleteEmptyDirectory()` on the File model to clean up empty directories: https://github.com/octobercms/library/blob/master/src/Database/Attach/File.php#L779?

w20k:
Works great - confirmed ✅

LukeTowers:
@w20k We still need @bennothommo and @austinderrick to respond to my last question

w20k:
@LukeTowers just had few hours of spare time, run through all my todos, or half ;)

bennothommo:
@LukeTowers I've assigned it to myself to have a look at the method @LukeTowers suggested, as it appears @austinderrick has abandoned it. Hopefully will give it another look in the next couple of days.

bennothommo:
This PR is now being finalised in #4518.

PR #4034, opened 2019-01-04
[WIP] Inline form field validation

bennothommo:
This is a WIP / Proposal to implement inline form field validation for forms built with the Form widget.

One of the shortcomings of validation in October currently is that the form is validated on submission, and only one error message is shown via a flash message. While this is fine for smaller forms, when larger forms are produced, and the possibility of multiple fields having issues increases, it could cause frustration getting messages one at a time.

This pull request aims to provide immediate feedback on errors with form field contents by:

- Adding a validation call when a field has lost focus or has changed.
- Validating the linked model for the current form and retrieving any errors for the changed input in the context of the entire form contents. (ie. if a field is invalid due to the contents of another field, this too needs to be detected)
- Showing the validation error in a friendly way to the user (perhaps as a message underneath the field, or as an icon next to the label and using a tooltip message to display the error)

At the moment, the scope is to only handle simple form fields, such as text inputs and textareas, select dropdowns, radio options and checkboxes, but could potentially extend to more advanced widgets if they can be easily validated in the context of a model.

### Tasks:

- [x] Add onValidate event for fields in Widget
- [ ] Add JS bindings for all simple types of form fields
  - [x] Text-field based fields (text, textarea, number and password)
  - [x] Dropdowns and balloon selectors
  - [ ] Checkboxes, switches and checkbox lists
  - [ ] Radio lists
- [x] Add error messages to form
- [ ] Prevent form submission if any fields have errors
- [x] Add boolean to `form_config.yaml` to enable/disable validation
- [ ] Add hooks to allow plugins to extend field validation
- [ ] Add unit tests


modules/backend/widgets/form/assets/js/october.form.js
```diff
@@ -220,6 +221,54 @@
             .nextUntil('.section-field').hide()
     }
 
+    /*
+     * Bind validation checks on available fields, triggered when
+     * the fields is blurred or changed
+     */
+    FormWidget.prototype.bindValidation = function () {
+        let fields = this.getFieldElements(),
```
LukeTowers: Current code base in October uses `var` not `let`, do you have an argument to switch?

modules/backend/widgets/form/assets/js/october.form.js
```diff
@@ -220,6 +221,54 @@
...
+            form = this.$form;
```
LukeTowers: When a variable contains a jQuery reference it's common to prefix it with `$` to indicate it's a jQuery handled element instead of just a plain JS element.

modules/backend/widgets/form/assets/js/october.form.js
```diff
@@ -220,6 +221,54 @@
...
+
+        for (let field of fields) {
```
LukeTowers: I think October is still supposed to support at the very least IE 11?

modules/backend/widgets/form/assets/js/october.form.js
```diff
@@ -220,6 +221,54 @@
...
+            let fieldClasses = field.className.split(/\s+/).map(function (className) {
+                var fieldClass = className.match(/^([a-z\-]+)-field$/);
```
LukeTowers: I know it's annoying, but current convention in October JS files is no semicolons at the end of lines. Feel free to provide input on this though

modules/backend/widgets/form/assets/js/october.form.js
```diff
@@ -220,6 +221,54 @@
...
+            });
+
+            if (fieldClasses.length === 0) {
+                continue;
+            }
+
+            // Load validator depending on field type
+            switch (fieldClasses[0]) {
+                case 'text':
+                case 'password':
```
LukeTowers: `number` could probably be added here

modules/backend/widgets/form/assets/js/october.form.js
```diff
@@ -220,6 +221,54 @@
...
+                    this._validateText(form, field);
+                    break;
+            }
+        }
+    };
+
+    FormWidget.prototype._validateText = function (form, field) {
+        let ele = this.$el;
```
LukeTowers: Should be prefixed with `$` and would prefer `$elem` over `ele`

modules/backend/widgets/form/assets/js/october.form.js
```diff
@@ -220,6 +221,54 @@
...
+
+        field.querySelector('input').addEventListener('blur', function (ev) {
+            $(ele).request('onValidateField', {
```
LukeTowers: No need to pass `ele` to `$()`, it's already a jQuery object, could just do `$elem.request` directly

modules/backend/widgets/Form.php
```diff
@@ -452,6 +453,50 @@ public function onRefresh()
...
+    /**
+     * Validates a single field in the context of the entire form contents.
+     *
+     * @return array `valid` (bool) whether the field input is valid
+     *               `message` (string) the validation error message, if any
+     */
+    public function onValidateField()
+    {
+        $saveData = $this->getSaveData();
+        $field = Input::get('fieldName');
```
LukeTowers: Typically we just use the `input()` or `post()` helper functions directly instead of the facade. (see https://github.com/octobercms/october/blob/master/modules/backend/widgets/Form.php#L1130)

LukeTowers (changes requested):
Overall looks pretty nifty, just a few comments about JS style

LukeTowers:
Do you want to post some screenshots of the different options of displaying the error messages to the user? Keep in mind that not all fields have labels all of the time and we'd need it to work well on mobile too.

modules/backend/widgets/form/assets/js/october.form.js
bennothommo: I prefer `let` as it restricts the scope of the variable to the specific block, and I believe IE11 supports it as well, but I've switched to `var` to keep it consistent.

modules/backend/widgets/form/assets/js/october.form.js
bennothommo: I wasn't sure on which browsers October was still targeting, as I couldn't find anything in the docs. I'll keep in mind that IE11 is still needing to be supported.

modules/backend/widgets/form/assets/js/october.form.js
bennothommo: I had seen semicolons on some lines, ie: https://github.com/octobercms/october/blob/1e3e49ce08405b2b9cf1e60c9d399a1c04007ec1/modules/backend/widgets/form/assets/js/october.form.js#L324 so I figured I could get away with using them, but I will take them out to maintain consistency.

bennothommo:
@LukeTowers Not a problem. Will try and mock those up over the weekend.

modules/backend/widgets/form/assets/js/october.form.js
```diff
@@ -220,6 +221,102 @@
...
+     */
+    FormWidget.prototype.bindValidation = function () {
+        var fields = this.getFieldElements(),
+            $form = this.$form
+
+        for (var i in fields) {
+            var field = fields[i]
+
+            if (!field || !field.className) {
+                continue;
```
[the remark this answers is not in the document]
bennothommo: Oops :P

w20k:
@bennothommo, ping me when you need major testing to be done 😉

modules/backend/widgets/Form.php
```diff
@@ -1,6 +1,7 @@
 <?php namespace Backend\Widgets;
 
 use Lang;
+use Input;
```
LukeTowers: Unused reference

bennothommo: Removed :)

bennothommo:
@LukeTowers Do you have any thoughts on the styling of the error messages per https://github.com/octobercms/october/pull/4034#issuecomment-451716513? I might try and finalise that first before hooking it up to the other field types.

LukeTowers:
@bennothommo I'm leaning towards the inline messages right now, they're the simplest for the user to understand what's going on, perhaps with some iconography for colour accessibility? This [article](https://medium.com/@andrew.burton/form-validation-best-practices-8e3bec7d0549) makes a lot of excellent points.

bennothommo:
@LukeTowers Thanks for that. I think you're right, and those articles make compelling points. I'll proceed with the simple error message underneath the field.

w20k:
@bennothommo, would be awesome if you add JUnit tests 😄

w20k:
@bennothommo, I think unit tests would work, don't feel like functional testing is needed here.

modules/backend/classes/FormField.php
```diff
@@ -329,7 +334,9 @@ protected function evalConfig($config)
         if (isset($config['containerAttributes'])) {
             $this->attributes($config['containerAttributes'], 'container');
         }
-
+        if (isset($config['inlineValidation'])) {
+            $this->inlineValidation = boolval($config['inlineValidation']);
```
LukeTowers: why `boolval` instead of just typecasting to boolean with `(bool) $config['inlineValidation']`?

bennothommo: No particular reason - just the first function that came to mind.

LukeTowers (approved):

w20k:
@bennothommo do you think I could be any use? Would love to help you to finish this PR ;)

bennothommo:
@w20k I'd love any help you're willing to provide :). What would you be interested in having a look at? You could either set up some handlers for checkboxes and radio inputs, or you could maybe set up a PR for the test plugin to test this feature? Essentially, to test this feature, you just need to add `inlineValidation: true` to any form config or field config inside a form config to enable the check.

w20k:
@bennothommo I'll start from making a PR to the test-plugin. So, that I would be easier to test, already implemented features, and understand what's there missing :)

Would also ask you to push your branch to OctoberCore, if @LukeTowers don't mind. So, that I could provide PR's to help you with the coding part as well. Little by little getting back to my daily duties 😄

modules/backend/widgets/form/assets/js/october.form.js
```diff
@@ -220,6 +251,157 @@
...
+        })
+    }
+
+    FormWidget.prototype._validateDropdown = function ($form, field) {
+        var $elem = this.$el,
+            $field = $(field),
+            widget = this
+
+        var $innerField = $(field.querySelector('select'))
+        if (!$innerField) {
```
w20k: @bennothommo I think you have a bug here :) `$innerField` will be an empty object, but present.
You could either change it to `!!$innerField` or `$innerField.length === 0`.

bennothommo: Good catch :)

modules/backend/widgets/form/assets/js/october.form.js
```diff
@@ -220,6 +251,157 @@
...
+            $form = this.$form
+
+        for (var i in fields) {
+            var field = fields[i]
+
+            if (!field || !field.className) {
+                continue
+            }
+
+            var fieldClasses = field.className.split(/\s+/).map(function (className) {
```
w20k: @bennothommo I think this part is overly complicated :) 
If I've understood it correctly you need to find the field which has a class corresponded to the Regexp - `/^([a-z\-]+)-field$/` and that's it.

Two options: either use `indexOf()` to cut down all those not needed fields or change Regex to be more accurate (before&after spaces, dirty try: `(\s+)?([a-z\-]+)-field(\s+)?`).

bennothommo: @w20k Yeah I did initially have just a regex, but I thought this way was more readable, albeit a bit more complex. Your regex works just as well though. :)

bennothommo:
This PR is now being developed in #4322.

PR #4045, opened 2019-01-09
Added favicon management to back-end customisation

Farrow:
Added a field to the "customise back-end" area to allow for a custom favicon to be displayed in the backend instead of the default October logo.

If no file is specified, the existing October logo is used by default.

Note that this only customises the icon used in the back-end, not the front-end favicon.

### DIFF
--- a/modules/backend/lang/en/lang.php
+++ b/modules/backend/lang/en/lang.php
@@ -416,6 +416,8 @@ return [
         'brand' => 'Brand',
         'logo' => 'Logo',
         'logo_description' => 'Upload a custom logo to use in the back-end.',
+        'favicon' => 'Favicon',
+        'favicon_description' => 'Upload a custom favicon to use in the back-end',
         'app_name' => 'App Name',
         'app_name_description' => 'This name is shown in the title area of the back-end.',
         'app_tagline' => 'App Tagline',
--- a/modules/backend/layouts/_head.htm
+++ b/modules/backend/layouts/_head.htm
@@ -7,7 +7,7 @@
 <meta name="backend-timezone" content="<?= e(Backend\Models\Preference::get('timezone')) ?>">
 <meta name="backend-locale" content="<?= e(Backend\Models\Preference::get('locale')) ?>">
 <meta name="csrf-token" content="<?= csrf_token() ?>">
-<link rel="icon" type="image/png" href="<?= Backend::skinAsset('assets/images/favicon.png') ?>">
+<link rel="icon" type="image/png" href="<?= e(Backend\Models\BrandSetting::getFavicon()) ?>">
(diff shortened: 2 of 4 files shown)


Farrow:
I've added support for brand.faviconPath as requested.

LukeTowers:
Great, @w20k can you test and confirm it's working as expected?

w20k:
@LukeTowers, @Farrow  - will re-test it locally to be sure, but something is not working 😄

w20k:
@LukeTowers,
Status - ✅ 

Should it also work on the Auth page? If so, could you fix it @Farrow (`modules/backend/layouts/auth.htm`)?

Farrow:
@w20k 

I agree that the auth page's favicon ought to be the same as the front-end or back-end favicon instead of always being the October default, but I think it's debatable which is more appropriate.

In my own case, we want the front-end favicon to be different to the back-end favicon. It could be argued that since it's the auth page for the back-end it ought to be the back-end favicon, but you could also argue that since you're on the front-end (i.e. on a page accessible by anyone) maybe it ought to be the front-end favicon.

Maybe it should be a customisation option itself to specify which to use, but that may be overkill.

w20k:
@Farrow, ain't `/backend` is sort of a home page for backend? 😄
Lets here what @LukeTowers thinks.

LukeTowers:
@Farrow @w20k My vote is for the backend favicon. The auth page is a part of the backend, it's irrelevant that it's accessible without being logged in.

w20k:
@Farrow, could you fix the auth page?

Farrow:
I've updated the auth page as requested

w20k:
@Farrow - thanks!
@LukeTowers, tested ;)

SaifurRahmanMohsin:
Is there a way to access the same via the frontend with ease? I'm using this code as of now to access it:

```
function onStart() {
    $this['favicon'] = Backend\Models\BrandSetting::getFavicon();
}
```
And then on my page using it as:
`<link rel="shortcut icon" type="image/png" href="{{ favicon }}">`

But is there like a pre-defined variable like `{{ this.controller.some_brand_setting_variable.favicon }}` for easy access? I checked the documentation already by doing a search for favicon but couldn't find anything.

LukeTowers:
@SaifurRahmanMohsin just use my plugin: https://octobercms.com/plugin/luketowers-essentialvars

SaifurRahmanMohsin:
Oh ok. I was hoping to see a more straight-forward way than installing a plugin. I already have it solved using the method I shared above, just looking for a better alternative. Thanks anyway, I will check out your plugin as well.

LukeTowers:
@SaifurRahmanMohsin nope, no other way than adding it yourself. The plugin is just a really tiny helper to be able to add it in to themes easily. The argument against adding it by default is because that would involve an extra DB call on every request for something that not every theme author would use; hence the provision of that functionality through a plugin.

PR #4055, opened 2019-01-12
Fix incorrect MediaLibrary links when base path is not the domain root

### LINKED ISSUE #4039: Quick notice of a broken mediafinder version from 443 to 446 when upgrading
First of all, I would like to announce that my English is not good.
I don't know if upgrading the version caused a broken mediafinder. Today (06.01.2019), I received an automatic notification of a version update recommendation. Fortunately, my second project is in the development phase. Then other plugins that use the mediafinder will not display the image in the backend. Image link is broken. For example, the media url should be "http: //localhost/KAAR/storage/app/media/assets/more_image.jpg", but all links point to double root:
"http://localhost/KAAR/KAAR/.../.../etc /..."

I also looked at the mediamanager, then the download url is also broken. Screenshots speak better for themselves:
screenshot 2019-01-06 at 21 05 10

screenshot 2019-01-06 at 21 06 56

I do not know if the failure of the mediafinder is related to the upgrade? I want to listen to you!

adsa95:
This PR fixes #4039, a bug that was introduced in PR #3536.

Currently the app root/base path is added to the `storagePath` variable in `MediaLibrary`, which is then used to generate the links with the `getPathUrl()` function. Since PR #3536 this is then also passed through the `Url::to()`function to adhere to the link policy, however the base path is added again in this method and therefore we get duplicated base paths in the final URL (only noticeable if your base path is not the domain root). This PR solves the issue by not adding the base path to `storagePath` in the `MediaLibrary`.

`storagePath` is a protected variable and not used anywhere else within the `MediaLibrary` so this should not cause any other problems, it would however affect anyone extending the `MediaLibrary` class and using this variable.

Unfortunately I have no experience using tests, and no the needed time to learn this within a reasonable timeframe for the need of this PR. Ping @w20k

### DIFF
--- a/modules/system/classes/MediaLibrary.php
+++ b/modules/system/classes/MediaLibrary.php
@@ -73,10 +73,6 @@ class MediaLibrary
         $this->storageFolder = self::validatePath(Config::get('cms.storage.media.folder', 'media'), true);
         $this->storagePath = rtrim(Config::get('cms.storage.media.path', '/storage/app/media'), '/');
 
-        if (!starts_with($this->storagePath, ['//', 'http://', 'https://'])) {
-            $this->storagePath = Request::getBasePath() . $this->storagePath;
-        }
-
         $this->ignoreNames = Config::get('cms.storage.media.ignore', FileDefinitions::get('ignoreFiles'));
 
         $this->ignorePatterns = Config::get('cms.storage.media.ignorePatterns', ['^\..*']);


LukeTowers:
@w20k could you put together a test case for this?

w20k:
@adsa95, could you look at #4065, seems like there is one more issue with path duplication... Odd one! Thanks 😉

PR #4058, opened 2019-01-14
Changed z-index

VoroninWD:
Before:
image

After:
image

### DIFF
--- a/modules/system/assets/ui/less/popover.less
+++ b/modules/system/assets/ui/less/popover.less
@@ -21,7 +21,7 @@ div.control-popover {
     background-clip: content-box;
     left: 0;
     top: 0;
-    z-index: @zindex-popover - 1;
+    z-index: @zindex-popover;
     visibility: hidden;
 
     &.in, &.fade {


w20k:
Recompile pliz CSS assets `php artisan october:util compile less`!

w20k:
@LukeTowers - tested.

Status - ✅

Before (had to mess up with css height to make it visible):
screenshot 2019-01-26 at 23 01 43

After:
screenshot 2019-01-26 at 23 05 06

Found a minor UI bug: while in a popup if you scroll down, your options list will tag along 😄

LukeTowers:
@w20k how about in nested popups or outside of popups?

w20k:
@LukeTowers never ever used nested pop-ups)) Will look at it, think should be fine. Because it should be relative to pop-up window.
Outside - works as it should!

LukeTowers:
Alright. I hate mucking about with zindex changes, it always ends up breaking something somewhere in unexpected ways.

Issue #4053, opened 2019-01-12
Placeholder for select dropdown does not work without an empty option


GinoPane:
- OctoberCMS Build: 446

Placeholder for select dropdown defined using "form-control custom-select" does not work without an empty option specified in the select options. 

For example, we have such select:

```
<select class = "form-control custom-select" data-placeholder="Select something">
	<option value="1">1<option>
	<option value="2">2<option>
	<option value="3">3<option>
</select>
```

It will initially look like this even when we don't select anything:

![image](https://user-images.githubusercontent.com/3897579/51075291-9578b000-169a-11e9-85f2-6485ffdba0ed.png)

On the other hand if we define it like this:

```
<select class = "form-control custom-select" data-placeholder="Select something">
	<option value=""><option>
	<option value="1">1<option>
	<option value="2">2<option>
	<option value="3">3<option>
</select>
```

It will work as expected:

![image](https://user-images.githubusercontent.com/3897579/51075319-e5f00d80-169a-11e9-9b1a-79e3ca87711f.png)

The reason of the issue is that option `allowClear: true` is missing from select2 initialization.

I'm going to submit a PR for this.

GinoPane:
I've changed option name in description, because I used `allowEmpty` instead of `allowClear`.

w20k:
It miss lead me a bit :)

PR #4060, opened 2019-01-14
Fix for #4053

GinoPane:
For more info about select2 placeholders: https://select2.org/placeholders.

Though its documentation says something against empty option for selects with `multiple` attribute, `allowClear` seems to behave correctly for tag lists, for example:
image

It adds an option to remove all tags by one click, which is pretty convenient, though it is a side effect 😄


w20k:
@GinoPane could you recompile JS assets. `php artisan october:util compile js`

GinoPane:
@w20k done 👍

modules/system/assets/ui/js/select.js
```diff
@@ -113,6 +113,7 @@
             var placeholder = $element.data('placeholder')
             if (placeholder) {
                 extraOptions.placeholder = placeholder
+                extraOptions.allowClear = true;
```
w20k: Please remove semicolon ;)

GinoPane: Oops, done. Didn't pay attention that semicolons are not used there

LukeTowers:
@w20k ping me once you're done testing. Refs: #4053

w20k:
@LukeTowers I don't think this PR is related to #4053 .

PR is about clearButton. New option - allowClear
The issue is about enabling empty element in dropdown list aka placeholder. New option - allowEmpty 😉

w20k:
I guess, works as expected, @LukeTowers.
Status - ✅
Screenshots:
screenshot 2019-01-26 at 22 52 27
screenshot 2019-01-26 at 22 52 48

LukeTowers:
@GinoPane @w20k if that changes the existing behaviour by allowing dropdowns to be have a state of not having any option selected then that needs to be gated by a config property on the dropdown. Does it allow the selection to be cleared when the dropdown doesn't have a placeholder?

LukeTowers:
@w20k so when a placeholder is not set, does that mean that the clear button isn't present?

w20k:
@LukeTowers as for now, from js side - yes. 
Both options are set when placeholder is defined.

LukeTowers:
@w20k Sweet. So is this ready to be merged now?

w20k:
@LukeTowers Yep - ready!

PR #4061, opened 2019-01-14
Add the counter to the Main Menu

fansaien:
- Add the counter to the Main Menu, same as the Side Menu;
- Hide the counter when counter is 0 or empty string.


LukeTowers:
Implemented in https://github.com/octobercms/october/commit/cb878020806ff4b64d23ea4d1d138350b695df3b

PR #4062, opened 2019-01-15
Fix default value translation

acasar:
This is a fix for an issue reported here: https://github.com/octobercms/october/commit/a1fb23a98446579659ddb73836e5b12f9bcf76e8#diff-d7b489cccf21c7a4400772ca5d6968c6

Basically, we can have default values that are not strings (e.g. for repeater fields). And in these cases we now get an error. This PR ensures that we only try to translate default values when they are strings.

### DIFF
--- a/modules/backend/widgets/Form.php
+++ b/modules/backend/widgets/Form.php
@@ -1072,10 +1072,13 @@ class Form extends WidgetBase
         }
 
         $defaultValue = !$this->model->exists
-            ? trans($field->getDefaultFromData($this->data))
+            ? $field->getDefaultFromData($this->data)
             : null;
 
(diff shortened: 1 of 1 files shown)


LukeTowers:
@acasar could you post an example of what you're talking about? Field config and everything?

acasar:
@LukeTowers Here's an example:

```
characters:
    label: Characters
    type: repeater
    default:
        0:
            name: Jim Moriarty
            phone: 1234554321
        1:
            name: Irene Adler
            is_enabled: 1234554321
```

LukeTowers:
That's neat, didn't even realize you could do that. I would probably use placeholders for that specific example, but still cool.

Issue #4067, opened 2019-01-16
Reset search in toolbar not implemented correctly everywhere


Rike-cz:
- OctoberCMS Build: 446

### Description:
Reset search (X) in toolbar search input not implemented correctly everywhere (#3425). Implementation in Relation Manager toolbar missing.
![relation-manager](https://user-images.githubusercontent.com/4247582/51263631-ef3beb80-19b4-11e9-8fcf-226acf9da34e.png)

In Relation toolbar and List toolbar it's not perfect too (Mozilla/5.0 (Windows NT 6.1; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/70.0.3538.102 Safari/537.36 **OPR/57.0.3098.116**):
![relation](https://user-images.githubusercontent.com/4247582/51264205-43939b00-19b6-11e9-98ca-91761774e5d9.png)
![lists](https://user-images.githubusercontent.com/4247582/51264209-47272200-19b6-11e9-9457-0e251f8bc416.png)

Ping: @vosco88, @w20k

w20k:
@vosco88, Could you look at this issue and provide a PR to fix it? 
Thanks!

vosco88:
@w20k Sure, will look at it.
@Rike-cz The pictures of the list toolbar are from which browser? Because when I run it in Firefox it looks good, in Chrome a bit of padding needs to be added (that I think corresponds to the second picture) . But I am not sure about first picture and where to replicate it.

LukeTowers:
@vosco88 The first picture is a recordfinder widget with showSearch: true

Rike-cz:
@vosco88 My usecase is Relation - View (https://octobercms.com/docs/backend/relations#has-many).

vosco88:
@LukeTowers Sorry I meant the second picture :) I think I have already figured out the first picture and third. I don't know where to replicate the second.

vosco88:
I have created PR for this. 
@w20k maybe also @Rike-cz could you test it please? Seems to be working for me

PR #4068, opened 2019-01-16
Fix reset search button issues

vosco88:
Fix for #4067.

- pulled reset search button css out of the toolbar css
- height and padding adjusted

### DIFF
--- a/modules/system/assets/ui/less/toolbar.less
+++ b/modules/system/assets/ui/less/toolbar.less
@@ -113,7 +113,9 @@ html.mobile {
      display: inline-block;
      position: absolute;
      right: 1px;
-     top: 2px;
+     top: 0;
+     bottom: 0;
+     margin: auto;
      background-color: #fff;
      padding-left:7px;
      height:28px;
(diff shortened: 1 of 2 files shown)


w20k:
@Rike-cz, could you test this fix, and ping @LukeTowers if everything is fine.
Please, add screenshots. 
Thanks!

vosco88:
I can do some screenshots tommorow

Rike-cz:
**List widget** is good.
list

**Relation list** has padding only 5px, as you can see here: https://github.com/octobercms/october/blob/2dbc56aead2ef2355135349b83860457f48f6235/modules/backend/behaviors/relationcontroller/assets/less/relation.less#L20-L23
So result is little different.
relation

**Relation manager/Record finder** still missing.
record-finder

vosco88:
@Rike-cz strange, because also the relation manager was working for me. Will check it in the evening, but it was definitely working :)

Rike-cz:
Sorry, my bad! You removed also `.control-toolbar`. A will edit post above.

vosco88:
I've tinkered with it a bit more here are the results:

Regular list:
image

Relation list:
image

Record Finder:
image

Feel free to comment and test @Rike-cz :)

Rike-cz:
Now it's ok for me, thank you!

vosco88:
@w20k @LukeTowers seems to be working fine for me and also @Rike-cz

Issue #4037, opened 2019-01-05
Backend tabs and tables no longer horizontally scrollable


joseph-d:
- OctoberCMS Build: 444-446
- PHP Version: 7.2

### Description:

Since the latest 444-446 build update, it is no longer possible to horizontally scroll backend tabs and tables on mobile view.

### Steps To Reproduce:

Go to and backend view with tabs or tables that extend beyond the mobile viewport and try to swipe to scroll them horizontally as was previously possible prior to the Build 444 update. Swiping no longer results in the tabs or tables being scrolled.

LukeTowers:
@ayumihamsaki can you look into this? @joseph-d could you provide more information on the browser used?

joseph-d:
The issue is present on the following browsers that I use:

**Mobile:** 
Chrome 71.0.3578.99 on Android 8.0.0; SM-G955F Build/R16NW
Firefox 64.0.1 on Android 8.0.0; SM-G955F Build/R16NW

**Desktop in mobile emulation:** 
Chrome Version 71.0.3578.98 (Official Build) (64-bit) on Ubuntu Linux 18.10

I have also just tested it on Firefox Quantum 64.0 on Ubuntu 18.10 Desktop using a mobile responsive width and tabs and tables _do_ still appear to scroll vertically in Build 446 but I suspect this is just because Firefox on Desktop isn't doing mobile emulation properly.

If I change the storm files in /modules/system/assets/ui/ back to Build 443 versions, vertical scrolling of tabs and tables works properly on all browsers listed above.

w20k:
I also confirm bug, tested it on the:
- IPhone latest build 12.0.x
- Chrome 71.x

I think it could be reproduced with the chrome device mode.

LukeTowers:
@ayumihamsaki @ayumihamsaki2 @Teranode could you guys look into this and see if you can find the issue? @joseph-d are you able to try incrementally applying changes from the 446 version of storm.css to find the issue?

w20k:
@LukeTowers found the cause, will make PR tomorrow.

w20k:
@LukeTowers, will wait for CSS minifier revert, and then push.

joseph-d:
@w20k Sure, I'm ready to test as soon as you do the PR.

AugmentBLU:
[...]

PR #4069, opened 2019-01-16
Tabs and tabels scrolling fix

w20k:
Fix for: https://github.com/octobercms/october/issues/4037

Changes:
- Updated modernizr with touch events and force touch
- Minor CSS fixes
- Added `touch-action: none;`, chrome and other browsers need this rule to  disable touch scrolling

@ayumihamsaki, @joseph-d, could you re-test it on the real devices. Would appreciate it! 
Side note: Only tested tables with chrome emulator.


ayumihamsaki:
@w20k Seems ok have tested it on the following browsers:

Chrome V71 and V73
Firefox V66
Opera V57
Edge V18
IE 11
Safari TP 299

joseph-d:
@w20k I've tested this in the following mobile browers:

Chrome 71.0.3578.99 on Android 8.0.0; SM-G955F Build/R16NW
Firefox 64.0.1 on Android 8.0.0; SM-G955F Build/R16NW

The tables and tabs both scroll correctly, but there are some issues that are now present which were not present prior to build 444.

### Wordbreaking issue in tables:

Prior to Build 444, the data in the tables would _wordwrap_ but not _wordbreak_ -- it isn't necessary for data to _wordbreak_ if we have horizontal scrolling and _wordbreaking_ also makes it difficult to read data. It seems that _wordbreaking_ was added globally to tables in Build 444 in Line 73 of storm.css

Here are images to show the difference:

Before Build 444, we had _wordwrapping_ but not _wordbreaking_ : 

screenshot from 2019-01-17 10-15-00

From Build 444 onwards we also have _wordbreaking_ which makes data difficult to read: 

screenshot from 2019-01-17 10-13-53

### Scrolling tabs works but whole container moves

Prior to Build 444, scrolling the tabs would only move the tabs, but now it seems like the whole container moves. It is also possible to move the container by swiping in the empty black area at the top of the screen as in the following gif animation, it only happens on pages with overflowing/scrolling tabs so I think it's because the tabs are overflowing the container a little bit.

peek 2019-01-17 10-20

w20k:
@joseph-d thanks a lot for your detailed  output. 

Could you provide a fix with screenshots for the table 'word-break'? Your not the only one that found this bug. I've being pinged few times already about this change.

As for the scrolling, will look at it more closely after I'd get back.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

joseph-d:
@w20k It looks like the wordbreak issue was introduced recently by https://github.com/octobercms/october/commit/c5d0a467d87da588a49fd1a9d34d62de4e895b93#diff-81523c582c5bd5d138d1a736a6151340 which is obviously a good fix for that particular problem. Do you have any suggestion for how it could be fixed without recreating the issue in #3834 @matthew188 @LukeTowers as I'm not sure what to do.

Issue #2465, opened 2016-11-07
Why use awful <?php in template?


vincoweb:
Use <?php in template is for me step back. I have used pure laravel and this code is awful when I am trying October. This (<?php ) is not template engine:

<?php Block::put('breadcrumb') ?>
    <ul>
        <li><a href="<?= Backend::url('blabla/blog/posts') ?>">Posts</a></li>
        <li><?= e($this->pageTitle) ?></li>
    </ul>
<?php Block::endPut() ?>

Why do not use blade syntax like this (or similiar)?

{{ Block::put('breadcrumb') }}
    <ul>
        <li><a href="<?= Backend::url('blabla/blog/posts') ?>">Posts</a></li>
        <li><?= e($this->pageTitle) ?></li>
    </ul>
{{ Block::endPut()  }}

or

{{{ Block::put('breadcrumb') }}}
    <ul>
        <li><a href="<?= Backend::url('blabla/blog/posts') ?>">Posts</a></li>
        <li><?= e($this->pageTitle) ?></li>
    </ul>
{{{ Block::endPut()  }}}

or

{{% Block::put('breadcrumb') %}}
    <ul>
        <li><a href="<?= Backend::url('blabla/blog/posts') ?>">Posts</a></li>
        <li><?= e($this->pageTitle) ?></li>
    </ul>
{{% Block::endPut()  %}}

Thank you very much

LukeTowers:
October uses Twig for its frontend templating language, and PHP for its backend. Blade wouldn't add much to the backend templates. 

See https://octobercms.com/forum/post/why-twig and https://github.com/octobercms/october/issues/863 for more information.

LukeTowers:
@daftspunk this can probably be closed.

vincoweb:
Php tags are not twig (<?php ... ?>). I sayd octobercms could use template engine, not pure php tags. It is mess. I did not write use blade (twig is good), but for this purpose it is better to do it in similiar way (create special brackets for using PHP objects in twig). I do not understand why use pure php tags. For fast development, usibility and readability it could be in some template engine brackets. Also my IDE do not understand if this file is twig or php, or whatever, because of using both in same file. 

TY

LukeTowers:
If @daftspunk is game for it, you could probably look at adding support for using Blade in the backend templates yourself and then make a PR.

[...]

PR #4074, opened 2019-01-18
Escape output to prevent XSS injection

nathan-van-der-werf:
Another commit with some escaping to prevent XSS injections.

**Repeat from my last XSS pull request (#3924):**

I came across an (old) issue today (#2465). I totally agree that we should use a template engine like Twig in the backend since it prevents XSS injections in many cases (by default). (not saying it's the holy grail)

At this moment many input values / variables are not escaped in the backend, I am sure there are more XSS vulnerabilities in OctoberCMS after this & previous commit I made.


nathan-van-der-werf:
> @nathan-van-der-werf while I agree that a templating engine would be more secure by default, it's unlikely to happen given that would cause major backwards compatibility issues. In the meantime, I appreciate you keeping an eye open for these issues. If you could provide any recommendations for changes to the docs as well to help educate the developer base of the importance of escaping their output, that would be great too!

I'll check the docs when I've got some more time.

Issue #4078, opened 2019-01-21
Morph one and relation not working for backend list columns


SeriousKen:
I have a model with a morph one relation but the fields are empty when the list is displayed. The columns definition is:

``` yaml
columns:
    id:
        label: ID
    meta_url:
        label: URL
        relation: meta
        select: url
```

The Meta model holds common metadata for all page types and needs to be polymorphic.

``` PHP
class Meta extends Model 
{
    public $table = 'serious_content_meta';

    public $morphTo = [
        'content' = [],
    ];
}
```

Each page type has a morphOne back to the Meta model

``` PHP
class Page extends Model
{
    public $table = 'serious_content_pages';

    public $morphOne = [
        'meta' => [Meta::class, 'name' => 'content'],
    ];
}
```

The query that gets generated to populate the list is

``` SQL
SELECT "serious_content_pages".*, (SELECT url FROM "serious_content_meta" WHERE "serious_content_pages"."id" = "serious_content_meta"."content_id" and "serious_content_meta"."content_type" = ?) as meta_url FROM "serious_content_pages" ORDER BY "meta_url" DESC LIMIT 20 offset 0
```

The problem seems to stem from the ? not being filled in with the content type. I'm currently testing with an SQLite database which seems to run this query without error but the URL column is empty.

PR #4079, opened 2019-01-21
Support for morphOne relation in backend column list

SeriousKen:
Fix for issue https://github.com/octobercms/october/issues/4078

### DIFF
--- a/modules/backend/widgets/Lists.php
+++ b/modules/backend/widgets/Lists.php
@@ -350,6 +350,7 @@ class Lists extends WidgetBase
         $selects = [$primaryTable.'.*'];
         $joins = [];
         $withs = [];
+        $bindings = [];
 
         /**
          * @event backend.list.extendQueryBefore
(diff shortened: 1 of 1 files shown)


w20k:
@LukeTowers testing is done. Works as expected! 
Status - ✅ 

_Test plugin -> Post Model -> columns.yaml -> add_
```
review_content:
        label: Review Content
        relation: review
        select: content
```
Screenshot:
screenshot 2019-01-25 at 16 59 42

LukeTowers:
@w20k did you check to see if it broke any other types of relations or columns with no relations at all?

LukeTowers:
@SeriousKen @w20k does this add support for `morphTo` relations too? (the `@todo` comment in there)

SeriousKen:
@LukeTowers not sure if it does or not, but it might. I'll take a look.

SeriousKen:
@LukeTowers No, it doesn't work for morphTo. The model's `makeRelation` method specifically returns `null` when the relation is `morphTo` so the following code in the Lists widget fails as it calls `getTable` on `null`.

``` php
$table =  $this->model->makeRelation($column->relation)->getTable();
```

PR #4083, opened 2019-01-24
Allow multiline changelog messages on plugin page and preserve correct order of updates

GinoPane:
Some prerequisites for this.

Plugin development documentation shows versions.yaml file example:
image

And the system expects such order in `modules/system/controllers/Updates.php::getPluginVersionFile`:

```
$updates = is_array($updates) ? array_reverse($updates) : [];
```

So it reverses the order of updates to show with the latest update on the top.

On the other hand, the documentation does not tell that we **must** follow ascending order of updates, because both orders work fine and `VersionManager` actually sorts versions to get updates in proper order.

This leads to situation, when plugin updates on plugin changelog tab are listed starting from the oldest, like here:

before
 
This PR fixes such behavior and makes sure that the order of updates on that page is from the newest to the oldest no matter which order you use in `versions.yaml`. Also, I don't really like a limitation with single changelog message per update, therefore this PR also explodes update message by ";" and shows resulting messages on new lines which allows to simulate multiple messages. In my case the list looks like this after PR:

after

### DIFF
--- a/modules/system/controllers/Updates.php
+++ b/modules/system/controllers/Updates.php
@@ -181,7 +181,7 @@ class Updates extends Controller
         $contents = [];
 
         try {
-            $updates = (array)Yaml::parseFile($path.'/'.$filename);
+            $updates = (array) Yaml::parseFile($path.'/'.$filename);
 
             foreach ($updates as $version => $details) {
                 if (!is_array($details)) {
--- a/modules/system/controllers/Updates.php
+++ b/modules/system/controllers/Updates.php
@@ -176,21 +176,28 @@ class Updates extends Controller
         }
     }
 
-    protected function getPluginVersionFile($path, $filename)
+    protected function getPluginVersionFile($path, $filename): array
     {
         $contents = [];
 
         try {
-            $updates = Yaml::parseFile($path.'/'.$filename);
-            $updates = is_array($updates) ? array_reverse($updates) : [];
+            $updates = (array)Yaml::parseFile($path.'/'.$filename);
 
             foreach ($updates as $version => $details) {
-                $contents[$version] = is_array($details)
+                $changelogForVersion = is_array($details)
                     ? array_shift($details)
                     : $details;
+
+                $contents[$version] = array_filter(
+                    array_map('trim', explode(';', $changelogForVersion))
+                );
             }
         }
         catch (Exception $ex) {}
+
+        uksort($contents, function ($a, $b) {
+            return version_compare($b, $a);
+        });
 
         return $contents;
     }
--- a/modules/system/controllers/Updates.php
+++ b/modules/system/controllers/Updates.php
@@ -176,7 +176,7 @@ class Updates extends Controller
         }
     }
 
-    protected function getPluginVersionFile($path, $filename): array
+    protected function getPluginVersionFile($path, $filename)
     {
         $contents = [];
 
--- a/modules/system/controllers/Updates.php
+++ b/modules/system/controllers/Updates.php
@@ -184,13 +184,16 @@ class Updates extends Controller
             $updates = (array)Yaml::parseFile($path.'/'.$filename);
 
             foreach ($updates as $version => $details) {
-                $changelogForVersion = is_array($details)
-                    ? array_shift($details)
-                    : $details;
+                if (!is_array($details)) {
+                    $details = (array)$details;
+                }
 
-                $contents[$version] = array_filter(
-                    array_map('trim', explode(';', $changelogForVersion))
-                );
+                //Filter out update scripts
+                $details = array_filter($details, function($string) use ($path) {
+                    return !File::exists($path . '/updates/' . $string);
+                });
+
+                $contents[$version] = $details;
             }
         }
         catch (Exception $ex) {}
--- a/modules/system/controllers/Updates.php
+++ b/modules/system/controllers/Updates.php
@@ -190,7 +190,7 @@ class Updates extends Controller
 
                 //Filter out update scripts
                 $details = array_filter($details, function($string) use ($path) {
-                    return !File::exists($path . '/updates/' . $string);
+                    return !preg_match('/^[a-z_\-0-9]*\.php$/i', $string) || !File::exists($path . '/updates/' . $string);
                 });
 
                 $contents[$version] = $details;
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -18,6 +18,11 @@
             <directory>./vendor/october/rain/tests</directory>
         </testsuite>
     </testsuites>
+    <filter>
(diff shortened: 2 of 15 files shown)


w20k:
@GinoPane, could you also add a junit test for `getPluginVersionFile`method?

GinoPane:
@w20k do you mean PHP Unit test? I looked for existing tests for this method, found nothing and decided to leave it as it is.
I'll take a look, I've never touched October's tests before.

modules/system/controllers/Updates.php
```diff
@@ -176,22 +176,29 @@ public function details($urlCode = null, $tab = null)
         }
     }
 
-    protected function getPluginVersionFile($path, $filename)
+    protected function getPluginVersionFile($path, $filename): array
```
LukeTowers: Leave the return type hinting alone for now, we'll implement that across the entire code base all at once when we decide to do that.

LukeTowers:
@GinoPane I would prefer that the syntax for multiple change log entries per version is as follows:

```yaml
1.0.1:
- 'message'
- message 2
- message 3
- migration.php
```

Do the detection of the migration files with ends_with(.php) or a simple file exists check or something like that

GinoPane:
@LukeTowers yeah, I can look into this. Obviously, it's a preferred solution, but it's a potentially breaking change, therefore I preferred to play on a safe side 😄

modules/system/controllers/Updates.php
GinoPane: Ok

GinoPane:
@LukeTowers also, it would be nice to add something similar to marketplace, but I don't know if it's possible to update it

LukeTowers:
@GinoPane I think the marketplace just uses the same code anyways. It'll be a separate issue if not.

GinoPane:
My new comment includes a fix for `VersionManager`'s methods as well, because:
1) it treated all lines except the first one as possible scripts;
2) it added DB "script" entry even for non-existing files.

I used preliminary regex check with for possible files and passed only them into `applyDatabaseScript`, therefore it allowed to keep existing error reporting.

GinoPane:
@w20k could you please clarify what's going on with `composer install` for Octorber? When I try to install dependencies I have thousands of changed files (not in vendor directory, because it is ignored, but a lot of system files, assets, etc) and obviously I don't want to commit them:

image

LukeTowers:
@GinoPane October's core modules (backend, cms, system) are loaded via composer, so when you run composer install you're overwriting all of those files with whatever version is getting loaded by Composer. Pick the version you want to use (you should be using `dev-develop`, look at the docs for how to set that up, run `composer update` and then make your changes.

GinoPane:
@LukeTowers I ended up with commenting out those dependencies from composer at all 😄 Maybe it is a PHPStorm bug, but when those dependencies overwrite the code my repository gets screwed up.

@w20k The latest commit contains unit test changes.

GinoPane:
@w20k Is anything still required from me here?

w20k:
@GinoPane don't think so, need to review and run tests! Thanks ;)

w20k:
@GinoPane could you fix the test?

Either remove from `version.yaml` - `create_blog_settings_table.php` or add to the test ;)

GinoPane:
@w20k is anything failing? Travis does not report any failing tests as well as everything is running smoothly for me.

w20k:
Do you mean local Travis? :) It's not running pull tests, yet. If I'm not mistaken.

Your test:
```$expectedVersions = [
            '1.0.5' => [
                'Create blog settings table',
                'Another update message',
                'Yet one more update message'
            ],
            '1.0.4' => [
                'Another fix'
            ],
            '1.0.3' => [
                'Bug fix update that uses no scripts'
            ],
            '1.0.2' => [
                'Create blog post comments table',
                'Multiple update messages are allowed'
            ],
            '1.0.1' => [
                'Added some upgrade file and some seeding',
                'some_upgrade_file.php', //does not exist
                'some_seeding_file.php' //does not exist
            ]
        ];
```
Your version.yaml (bold lines fail):
```
1.0.5:
    - Create blog settings table
    - Another update message
    - Yet one more update message
    **- create_blog_settings_table.php**
1.0.4: Another fix
1.0.3: Bug fix update that uses no scripts
1.0.2:
    - Create blog post comments table
    - Multiple update messages are allowed
    **- create_comments_table.php**
```

GinoPane:
@w20k Why local? I meant Travis CI configured for the repo, it runs everything under `phpunit.xml`, including new tests. Console output is not very informative so here are the test results from local run:

image

Therefore I'm not pretty sure what's failing for you and where.

w20k:
@GinoPane sorry my bad, missed two files inside updates folder. Merging PRs by hand ... not a good option 😄

@LukeTowers tested - works fine.

modules/system/controllers/Updates.php
```diff
@@ -181,17 +181,27 @@ protected function getPluginVersionFile($path, $filename)
         $contents = [];
 
         try {
-            $updates = Yaml::parseFile($path.'/'.$filename);
-            $updates = is_array($updates) ? array_reverse($updates) : [];
+            $updates = (array)Yaml::parseFile($path.'/'.$filename);
```
LukeTowers: space between `(array)` and `Yaml` please.

phpunit.xml
```diff
@@ -18,6 +18,11 @@
             <directory>./vendor/october/rain/tests</directory>
         </testsuite>
     </testsuites>
+    <filter>
+        <whitelist processUncoveredFilesFromWhitelist="false">
+            <directory suffix=".php">./modules/</directory>
+        </whitelist>
+    </filter>
```
LukeTowers: Could you explain the purpose of this?

LukeTowers:
@GinoPane looking good so far. We'll also need an update to the documentation to add examples for the newly supported version formats.

phpunit.xml
GinoPane: I can remove this, because it's not required. This is a config change to enable code coverage report generation.

modules/system/controllers/Updates.php
GinoPane: Done

GinoPane:
I've added the docs update as well. Please check if it makes sense

w20k:
@LukeTowers - done with testing.

Status - all good! ✅ 

Screenshots:
```
1.0.1: First version of Test
1.0.2:
  - Create tables
  - Create tables
  - Create tables
  - Create tables
  - create_tables.php
1.0.3:
  - Seed tables
  - Seed tables
  - seed_tables.php
```
screenshot 2019-02-19 at 18 29 38

screenshot 2019-02-19 at 18 30 32
_One failed test is knowen issue_

PR #3879, opened 2018-10-18
Service Workers Invalid security token and Clear Site Data Security Headers

ayumihamsaki:
If a website has a Service Worker installed it would load and register before a User tries to login to the backend causing a "Invalid security token" message. This PR unregisters any installed Service Worker when a User opens the backend Signin webpage.

I have also added the NEW Security Headers to add Protection to October's Cache and Cookies. This includes two new Middleware that first clears any bad cached data before a User tries to login and the second Middleware will clear all the sensitive User Data when a User signs out of the Backend.

For more info on the new Security Header 'Clear Site Data' you can see the spec found here: https://www.w3.org/TR/clear-site-data/

This PR relates to the Github Issue found here: https://github.com/octobercms/october/issues/3707

Issue #3707, opened 2018-08-19
Invalid security token and Service Workers


ayumihamsaki:
I have read all the past issues with regards to the error message "Invalid security token" however none talk about Service Workers causing this issue.

When you add a Service Worker to October CMS you will get the following issues:

(**Please note**: In Firefox and Google Chrome you can turn off the Service Worker however you will still get issues)

1. Login to backend wont work first time and you have to login 2 or 3 times.
2. Saving files, you get constant problems trying to save files in the backend.
3. Having two or more tabs opened at same time while working on the same website will cause issues!

## Solution:

I'm using Google's Workbox - Service Worker Library and I have spoken to some Googler's about this issue. There is a solution to actually turn off the Service Worker's for the admin section. However, there is a security issue that needs to be taken into consideration! If you have setup the backend URL to a custom one to add an extra layer of security, you would not want to list that URL in Javascript and show people. So I would recommend using PHP to block Service Worker from caching the backend URL's.

So I open this issue for a feature request to be coded in php (hiding the backend custom URL) to make October CMS fully supportive of Service Workers and PWA websites by automatically turning off the Service Worker for admin webpages. (Hopefully I make sense?!)

So that's: 
Type: Bug
Type: Conceptual Enhancement

Further information can be found here:

https://github.com/GoogleChrome/workbox/issues/1164
https://github.com/w3c/ServiceWorker/issues/1257

w20k:
Hi, @ayumihamsaki , 

Had the same problem while testing jQuery 3.x, it was a minor bug in the octobercms.min.js file. 
Maybe you had an error in console? 

If not, might give it a try at the end of the week, and post my 'workflow'.

ayumihamsaki:
@w20k hi ya, I have my console dev tools screen open all the time. There is no error displayed. Only error is saying no manifest file can be loaded in the backend. Yah if you could look into it that would be great. It's a highly annoying bug.

w20k:
@ayumihamsaki , Hi, didn't have time to play with a fully PWA ready solution (or close to fully workable solution), but I did small testing, and thought about an idea. Haven't yet digged in the Workbox docs, so this might sounds silly, but Isn't that easier to just specify "white listed routes" ?

ayumihamsaki:
@w20k 

> Isn't that easier to just specify "white listed routes"

[...]

Issue #4076, opened 2019-01-19
Invalid security token after install and go to backend dashboard


patcaulfield:
- OctoberCMS Build: ### 
- PHP Version: PHP 7.0.3 / mariaDB 10.0
- Database Engine: 
- Plugins Installed: 

### Description:

I have a problem when I try install. 
After install, Click to the backend link url 
alert message 'Invalid security token' 
localhost install id OK but I'm using web hosting server. tow database name. Is it problem?

ayumihamsaki:
I get the same issue every time, just refresh the screen and will work second time. I did create a PR fix for it found here: https://github.com/octobercms/october/pull/3879 but mod's and admins never did anymore and merge it.

patcaulfield:
Oh, I see. I'm glad you have a same problem and you solved. Good. but me it's too difficult to understand your way to solved way because I don't know well about php. Thank you!

ayumihamsaki:
I've gone through all of the github repo and could only find closed issues relating to this issue, see here:

https://github.com/octobercms/october/issues/3640
https://github.com/octobercms/october/issues/3266

alekseyp:
I was logged in for couple weeks and started getting those errors out of no where.
Only happens on dashboard, if I navigate to any other admin page it loads, but won't let me save anything.

reed-josh:
I see a fix merged possibly there? I updated the octobercms install. Is there anything else we need to do to apply this fix?

ayumihamsaki:
@reed-josh There is two separate PR's fixing this, one to allow the middleware to work for the other PR. So I would say to add both merges and give it a try. p.s. not sure where the other merge is you will need to search for it.

PR #4088, opened 2019-01-25
Service Workers Invalid security token and Clear Site Data HTTP Header

ayumihamasaki2:
This is an updated PR to the old PR found here: https://github.com/octobercms/october/pull/3879

In this PR I have totally removed all the Middleware!

Instead we now have the following things happening:

1. Sign-in screen uninstalls any Service Worker and also completely clears any old cache which may contain old cookies!

* uninstall Service Worker is done by `modules/backend/assets/js/auth/unistall-sw.js` file

* completely clear any old cache and cookies done by this HTTP Header:
`Cache-Control: no-cache, no-store, must-revalidate`

For more info see here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control

2. Sign-out I have set the NEW Security HTTP Header 'Clear Site Data' to remove all Sensitive Data, to keep webmasters and admins more safe, when using "shared devices".

* add NEW Security HTTP Header 'Clear Site Data' by this HTTP Header:
`Clear-Site-Data: "cache", "cookies", "storage", "executionContexts"`

For more info on the new Security Header 'Clear Site Data' you can see the spec found here: https://www.w3.org/TR/clear-site-data/
and
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Clear-Site-Data#Sign-out

This PR relates to the Github Issue found here: https://github.com/octobercms/october/issues/3707

## Testing

I have tested this PR on a demo website and it seems to fix all the problems now. However there is one issue left. The Sign-in screen has the following HTTP Headers:

cache

It would be nice not to have the same Cache Control Header being loaded twice!

@w20k Let me know your thoughts.


ayumihamsaki:
PR linked to this issue as well: https://github.com/octobercms/october/issues/4076

ayumihamsaki:
p.s. I don't know where the HTTP header:

`Cache-Control: no-cache, private`

Is being installed from? If you can find the location maybe we could just update it from there? 

It should be:

`Cache-Control: no-cache, no-store, must-revalidate`

## no-cache:

Forces caches to submit the request to the origin server for validation before releasing a cached copy.

## no-store:

The cache should not store anything about the client request or server response.

## must-revalidate: (This is the important one to fix the Invalid security token Issue)
The cache must verify the status of the stale resources before using it and **_expired ones should not be used_**.

modules/backend/controllers/Auth.php
```diff
@@ -48,6 +50,9 @@ public function index()
      */
     public function signin()
     {
+        // Add HTTP Header to Clear Cache and Cookies before Signin, see github issue: #3707
+        header("Cache-Control: no-cache, no-store, must-revalidate");
```
LukeTowers: We can't have headers being sent manually from October, everything must tie into the global response object that Laravel deals with. A middleware would be a better option here, it would also resolve the double headers.

w20k:
As a first try I'm getting an error...

`Error We're sorry, but something went wrong and the page cannot be displayed.`
Will move changes to the local machine for testing. Will let you know @ayumihamasaki

LukeTowers:
@ayumihamasaki2 you might be better off using the closure middleware method here: https://laravel.com/docs/5.5/controllers#controller-middleware instead of trying to have entirely separate classes for them.

bennothommo:
@LukeTowers love a challenge! Will have look after I sort out the datatables issue.

mjauvin:
@ayumihamsaki in your middleware closure, I think you need to do this:

```
$response = $(next)($request);
$response->headers->set(...);
return $response;
```

ayumihamsaki:
Thanks for suggestion, have updated commit - same issue when testing:

```html
Call to undefined method Backend\Controllers\Auth::middleware()
```

mjauvin:
Yes, as @LukeTowers said, OctoberCMS Controller does not extend the base Laravel Routing Controller class. @bennothommo is looking into this, AFAICT

mjauvin:
I think you can write your signout() function as below:

```
    public function signout() 
    {   
        BackendAuth::logout();
        $response = Backend::redirect('backend');
        $response->headers->set('Clear-Site-Data', 'cache, cookies, storage, executionContexts');
        ￼
        return $response;
    }
```

mjauvin:
Did you try the following in the constructor:
```
        Route::any('backend/auth/signin')->middleware(*insert login closure code here*);
        Route::any('backend/auth/signout')->middleware(*insert logout closure code here*);
```

mjauvin:
for the signin method headers, try this:

```
diff --git a/modules/backend/classes/Controller.php b/modules/backend/classes/Controller.php
index fa7b1227..32cfbf2a 100644
--- a/modules/backend/classes/Controller.php
+++ b/modules/backend/classes/Controller.php
@@ -249,7 +249,11 @@ class Controller extends Extendable
             return $result;
         }
 
-        return Response::make($result, $this->statusCode);
+        $response = Response::make($result, $this->statusCode);
+        if ($action === 'signin') {
+            $response->header('Cache-Control', 'no-cache, no-store, must-revalidate', true);
+        }
+        return $response;
     }
 
     /**
```

mjauvin:
And for the signin headers, you need to modify the parent class as shown below:

```
diff --git a/modules/backend/classes/Controller.php b/modules/backend/classes/Controller.php
index fa7b1227..f8141e27 100644
--- a/modules/backend/classes/Controller.php
+++ b/modules/backend/classes/Controller.php
@@ -249,7 +249,11 @@ class Controller extends Extendable
             return $result;
         }
 
-        return Response::make($result, $this->statusCode);
+        $response = Response::make($result, $this->statusCode);
+        if ($action === 'signin') {
+            $response->header('Cache-Control', 'no-cache, no-store, must-revalidate');
+        }
+        return $response;
     }
 
     /**
```

ayumihamsaki:
@mjauvin thanks for your help, have added your suggestions.

mjauvin:
You'll need to restore the original signin() function, this is now done in the parent class...

ayumihamsaki:
@mjauvin Thanks, just tested it and working perfectly now on my test website!

@LukeTowers @w20k Working now!

w20k:
@ayumihamsaki did quick testing and got a bunch of errors.
Don't have time to dig in why. You could get those after logout in the console.

```Clear-Site-Data header on 'https://localhost.com:8443/backend/backend/auth/signout': Unrecognized type: cache.
17:02:08.118 backend:1 Clear-Site-Data header on 'https://localhost.com:8443/backend/backend/auth/signout': Unrecognized type: cookies.
17:02:08.118 backend:1 Clear-Site-Data header on 'https://localhost.com:8443/backend/backend/auth/signout': Unrecognized type: storage.
17:02:08.118 backend:1 Clear-Site-Data header on 'https://localhost.com:8443/backend/backend/auth/signout': Unrecognized type: executionContexts.
17:02:08.119 backend:1 Clear-Site-Data header on 'https://localhost.com:8443/backend/backend/auth/signout': No recognized types specified.```

**Comment by bennothommo** [comment]
@mjauvin One minor thing: The parent controller edit suggestion you provided would affect *any* controller in the backend that uses an action called `signin`. While I doubt it would be an action name that would be used often, it may still cause an unwanted side effect for a plugin.

**Comment by ayumihamsaki** [comment]
## Some test Results:

### Firefox (v67a):

```html
clear-site-data: "cache", "storage", "executionContexts"
```

Doesn't load the web page - forever trying to load.

```html
clear-site-data: cache, storage, executionContexts
```

Works fine!

### Chrome V71:

```html
clear-site-data: "cache", "storage", "executionContexts"
```

Loads the `cache` and `storage` but does not understand `executionContexts`

```html
clear-site-data: cache, storage, executionContexts
```

Gives these errors:

```html
Unrecognized type: cookies.
Unrecognized type: storage.
Unrecognized type: executionContexts.
No recognized types specified.
```

### Chrome V74:

```html
clear-site-data: "cache", "storage", "executionContexts"
```

Works fine!

```html
clear-site-data: cache, storage, executionContexts
```

Works fine!

### Conclusion

Seems to be a Browser issue and the perfect result with the latest Browsers is:

```html
clear-site-data: cache, storage, executionContexts
```

(No quotation marks)

mjauvin:
@bennothommo You're quite right! This should be much better:

```
diff --git a/modules/backend/controllers/Auth.php b/modules/backend/controllers/Auth.php
index 849a2675..ad5dd85e 100644
--- a/modules/backend/controllers/Auth.php
+++ b/modules/backend/controllers/Auth.php
@@ -35,6 +35,11 @@ class Auth extends Controller
         $this->layout = 'auth';
     }
 
+    public function run($action = null, $params = [])
+    {
+        return parent::__construct()->header('Cache-Control', 'no-cache, no-store, must-revalidate');
+    }
+
     /**
      * Default route, redirects to signin.
      */

```

And remove previous changes to the parent controller's run() method.

ayumihamsaki:
I think this is ok now, the only issue is with the browser and that seems to be fixed in Chrome V73 and V74. It just says an error message in the console log but the logout function still works and there is no infinite loop with Firefox when using the new Clear Site Data HTTP Header.

I request this to be reviewed now, for possible merge.

mjauvin:
@ayumihamsaki please revert change to parent class and apply commit from last comment...

ayumihamsaki:
@mjauvin Is it ok up to 5215e23 commit or need more changes? Your last comment confused me.

mjauvin:
hold on, something is not working with the local run() method...

mjauvin:
OK, was missing the method arguments...

```
diff --git a/modules/backend/controllers/Auth.php b/modules/backend/controllers/Auth.php
index 849a2675..2e2cbcac 100644
--- a/modules/backend/controllers/Auth.php
+++ b/modules/backend/controllers/Auth.php
@@ -35,6 +35,11 @@ class Auth extends Controller
         $this->layout = 'auth';
     }
 
+    public function run($action = null, $params = [])
+    {
+        return parent::run($action, $params)->header('Cache-Control', 'no-cache, no-store, must-revalidate');
+    }
+
     /**
      * Default route, redirects to signin.
      */
```

mjauvin:
Also, your JS asset filename has a typo "unIstall" instead of "unINstall" ...

ayumihamsaki:
Thanks, I just changed the keyboard here! I used the other one out.

mjauvin:
Looks good!

LukeTowers:
While the workaround provided is admirable, I'd prefer we waited until we fixed support for the `middleware()` method on October controllers so that we can implement a proper fix using that method.

ayumihamsaki:
Would that mean I need to revert it back to commit: 07dab21 and fix the js typo?

LukeTowers:
@ayumihamasaki basically

modules/backend/controllers/Auth.php
```diff
@@ -32,6 +32,8 @@ class Auth extends Controller
     public function __construct()
     {
         parent::__construct();
+        // Add JS File to un-install SW to avoid Cookie Cache Issues when Signin, see github issue: #3707
+        $this->addJs('../../../modules/backend/assets/js/auth/uninstall-sw.js');
```
LukeTowers: Can you use the `~` symbol to represent the application root instead?

modules/backend/controllers/Auth.php
```diff
@@ -32,6 +32,8 @@ class Auth extends Controller
     public function __construct()
     {
         parent::__construct();
+        // Add JS File to un-install SW to avoid Cookie Cache Issues when Signin, see github issue: #3707
+        $this->addJs('~/../../../../modules/backend/assets/js/auth/uninstall-sw.js');
```
LukeTowers: why is this so excessive with the `../` directory links?

ayumihamsaki: It doesn't work without them ?

ayumihamsaki: Screenshot on my test machine:

eg

ayumihamsaki: This:

```html
~/modules/backend/assets/js/auth/uninstall-sw.js
```

Gives me this:

```html
http://localhost/admincp/backend/auth/~/modules/backend/assets/js/auth/uninstall-sw.js
```

Does that mean something is wrong?

bennothommo:
@ayumihamsaki I've submitted PR #4106 which will provide middleware support in the controllers. If you want to take advantage of this, you will need to make the following adjustments to the `Auth` controller, as middleware must be defined in the constructor as per https://laravel.com/docs/5.5/controllers#controller-middleware.

```
diff --git a/modules/backend/controllers/Auth.php b/modules/backend/controllers/Auth.php
index 5e5b9187..eb9f00ac 100644
--- a/modules/backend/controllers/Auth.php
+++ b/modules/backend/controllers/Auth.php
@@ -32,6 +32,14 @@ class Auth extends Controller
     public function __construct()
     {
         parent::__construct();
+
+        $this->middleware(function ($request, $next) {
+            $response = $next($request);
+            // Clear Cache and any previous data to fix Invalid security token issue, see github: #3707
+            $response->headers->set('Cache-Control', 'no-cache, no-store, must-revalidate');
+            return $response;
+        })->only('signin');
+
         // Add JS File to un-install SW to avoid Cookie Cache Issues when Signin, see github issue: #3707
         $this->addJs('~/../../../../modules/backend/assets/js/auth/uninstall-sw.js');
         $this->layout = 'auth';
@@ -50,13 +58,6 @@ class Auth extends Controller
      */
     public function signin()
     {
-        $this->middleware(function ($request, $next) {
-            $response = $next($request);
-            // Clear Cache and any previous data to fix Invalid security token issue, see github: #3707                        
-            $response->headers->set('Cache-Control', 'no-cache, no-store, must-revalidate');
-            return $response;
-        });    
-       
         $this->bodyClass = 'signin';
 
         try {
```

bennothommo:
@ayumihamsaki And this for signout ;)

```
diff --git a/modules/backend/controllers/Auth.php b/modules/backend/controllers/Auth.php
index eb9f00ac..ec92e342 100644
--- a/modules/backend/controllers/Auth.php
+++ b/modules/backend/controllers/Auth.php
@@ -40,6 +40,13 @@ class Auth extends Controller
             return $response;
         })->only('signin');
 
+        $this->middleware(function ($request, $next) {
+            $response = $next($request);
+            // Add HTTP Header 'Clear Site Data' to remove all Sensitive Data when signout, see github issue: #3707
+            $response->headers->set('Clear-Site-Data', 'cache, cookies, storage, executionContexts');
+            return $response;
+        })->only('signout');
+
         // Add JS File to un-install SW to avoid Cookie Cache Issues when Signin, see github issue: #3707
         $this->addJs('~/../../../../modules/backend/assets/js/auth/uninstall-sw.js');
         $this->layout = 'auth';
@@ -114,13 +121,6 @@ class Auth extends Controller
      */
     public function signout()
     {
-        $this->middleware(function ($request, $next) {
-            $response = $next($request);
-            // Add HTTP Header 'Clear Site Data' to remove all Sensitive Data when signout, see github issue: #3707
-            $response->headers->set('Clear-Site-Data', 'cache, cookies, storage, executionContexts');
-            return $response;
-        });
-
         BackendAuth::logout();
         return Backend::redirect('backend');
     }
```

modules/backend/controllers/Auth.php
LukeTowers: hmm, no it's just not parsing it. Could you just use the `url()` function instead?

ayumihamsaki:
p.s. Thanks for everyone's help!

modules/backend/controllers/Auth.php
```diff
@@ -103,7 +120,7 @@ public function signin_onSubmit()
      * Logs out a backend user.
      */
     public function signout()
-    {
+    {
```
LukeTowers: Extra whitespace, please remove

LukeTowers:
Waiting on completion of #4106

w20k:
Hi @ayumihamsaki, found few more errors :) While testing @bennothommo solution.

Screenshot:
One is for uninstall-sw.js.
And second, something related to HTTP connection.
screenshot 2019-02-25 at 16 46 54

ayumihamsaki:
@w20k You can't use Service Workers under HTTP. Try again using HTTPs and it should not say that error. lol

w20k:
@ayumihamasaki I know that :) but your script would be used for everyone not only for those services where https is enabled by default. That's what I mean :)

ayumihamsaki:
@w20k According to Google's chrome security conference that happened this week, they said that over 90% of the internet now uses HTTPs, I think October CMS should now focus in that direction. The only time you should ever use HTTP with October CMS is on localhost (never in a production environment).

What I mean is that **both API's are for HTTPs only**, if a User is not using HTTPs then you get an error message in dev tools, but the code will still run and the User will still get logged out of October (I saying it won't lock up and stop the User from logging out of a HTTP environment).

In this situation I feel that those two error messages are more of a caution and not an error.

I would be more worried about the caution message saying that October CMS is not securing the password information properly:

> This page includes a password or credit card input in a non-secure context.

That should be fixed and is more important in my mind, as October CMS is leaking secure information out which a hacker could take advantage of.

ayumihamsaki:
@w20k According to Google's chrome security conference that happened this week, they said that over 90% of the internet now uses HTTPs, I think October CMS should now focus in that direction. The only time you should ever use HTTP with October CMS is on localhost (never in a production environment).

What I mean is that **both API's are for HTTPs only**, if a User is not using HTTPs then you get an error message in dev tools, but the code will still run and the User will still get logged out of October (I'm saying it won't get locked up and stop the User from logging out of the HTTP environment).

In this situation I feel that those two error messages are more of a caution than an error message in dev-tools.

I would be more worried about the caution message saying that October CMS is not securing the password information properly:

> This page includes a password or credit card input in a non-secure context.

That should be fixed and is more important in my mind, as October CMS is leaking secure information out which a hacker could take advantage of.

ayumihamsaki:
(If it's an issue I could create a wrapper to test if the website is using HTTPs or not and then execute the uninstall-sw.js code for HTTPs only).

w20k:
> This page includes a password or credit card input in a non-secure context.
@ayumihamasaki it's related to a password being processed within non-secured context (non-https).

My idea is to let users decide what to use. But I'd like to hear @LukeTowers opinion on this topic.

ayumihamsaki:
Cool, will watch this space and see what is decided and will adjust the code accordingly...

LukeTowers:
@ayumihamsaki just add a check to prevent the error message from being triggered unnecessarily.

w20k:
@ayumihamsaki ping me when you're done ;) Will re-test.

ayumihamsaki:
@w20k Updated the commit, let me know if that fixes the issue.

ayumihamsaki:
@w20k Ok ready for you to do some testing!

- Added HTTPS wrapper to both now.

w20k:
@ayumihamsaki I think you rushed a bit :)

`if (location.protocol === 'https:')` and `if(isset($_SERVER["HTTPS"]) && $_SERVER["HTTPS"] !== "on")`

ayumihamsaki:
@w20k 

Sorry I didn't test what I did, I will do the following:

1. Change the commit to add:

`if (location.protocol === 'https:')` and `if(isset($_SERVER["HTTPS"]) && $_SERVER["HTTPS"] === "on")`

2. I think you are talking about the issue with `"` and not using `'` I will take a wild guess and say that the current setup should work for Firefox (as per their issue on this). However, Google Chrome seems to be doing the opposite and I believe they have a bug issue open for this! I am hoping Google Chrome will change their ways for this issue. Currently in Chrome the current setup works - but gives an error message. If I switch it around you get an Infinite Loading Loop (which is far worse). 

So to conclude, I think point two is best to leave it as it is and wait for Chrome to update their browser.

**Lastly, small disclaimer**: [Clear Site Data](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Clear-Site-Data) and the [Turn off Service Worker](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerContainer/getRegistrations) are both "experimental" code according to MDN. So we should expect some changes to the code.

I would be happy to change the commit in point one and leave point two because Google Chrome seems to have fixed that issue in V73 and above.

Sorry for long comment - hopefully helps you to make a decision!

modules/backend/assets/js/auth/uninstall-sw.js
```diff
@@ -1,7 +1,11 @@
...
-navigator.serviceWorker.getRegistrations().then(
-    function(registrations) {
-        for (let registration of registrations) {  
-            registration.unregister();
-        }
-});
\ No newline at end of file
+// Wrapper to run the code only for HTTPs Connection
+if (location.protocol != 'https:')
+{
```
LukeTowers: curly brace on first line, also please capitalize the last `s` in `HTTPS` in the comment, lowercase `Connection` and pluralize it, comment should become `Only run on HTTPS connections`

ayumihamsaki: Done - Updated both as per your request.

LukeTowers:
@w20k will wait for testing result

modules/backend/controllers/Auth.php
```diff
@@ -40,9 +40,8 @@ public function __construct()
             return $response;
         })->only('signin');
 
-		// Wrapper to run the code only for HTTPs Connection	
-		if(isset($_SERVER["HTTPS"]) && $_SERVER["HTTPS"] === "on")
-        {
+		// Only run on HTTPS connections
+		if(isset($_SERVER["HTTPS"]) && $_SERVER["HTTPS"] === "on") {
```
w20k: @ayumihamsaki space after `if` :)

ayumihamsaki: Fixed.

(What tool are you using to check the typos?)

w20k: Eyes 😄 Hopefully will add travis-ci job, very soon, to check style.

ayumihamsaki: haha

w20k:
@LukeTowers testing is done.
Status - ✅ 

Chrome version: Version 72.0.3626.119 (64-bit)
_Still throwing erros, didn't test on Canary_
Firefox version: 66.0b11 (64-bit)

Screenshots:
screenshot 2019-03-01 at 16 56 15
screenshot 2019-03-01 at 16 56 20
screenshot 2019-03-01 at 16 57 10
screenshot 2019-03-01 at 16 57 21
screenshot 2019-03-01 at 16 57 28

PR #4090, opened 2019-01-27
RecordFinder nameFromRelation allows to query a related model

alpapan:
Hello

WRT the RecordFinder widget of the Backend Forms

Even though the current implementation of RecordFinder allows for columns of related models to be shown in the Find Record modal, there was no way to show this value in the Name field when selected.

This PR allows for just that by specifying the related model (as defined in the Model) in the fields.yaml

```
fields:
      top_level_version:
        tab: 'General info'
        label: Some label
        nameFromRelation: related_model
        nameFrom: name
        list: $/alpapan/biobench/models/mainmodel/columns.yaml
        type: recordfinder
        required: 1
        span: auto
```

where the model is:
```
public $belongsTo = [
      'related_model'=>[
        'alpapan\biobench\Models\RelatedModel',
        'table' => 'alpapan_biobench_related_model',
        'key' => 'seq_top_id'
      ],
```

This configuration will show the _name_ of the RelatedModel Model instead of the _name_ of the MainModel

NB: The existing code already allows for relationships in columns, just not the recordfinder.

```
columns:
    related_model_name:
      label: Some name
      type: text
      select: name
      relation: related_model
```

### DIFF
--- a/modules/backend/formwidgets/RecordFinder.php
+++ b/modules/backend/formwidgets/RecordFinder.php
@@ -277,7 +277,7 @@ class RecordFinder extends FormWidgetBase
         $nameFrom = explode('.', $this->nameFrom);
         foreach ($nameFrom as $attribute) {
             if (!empty($name->{$attribute})){
-              $name = $name->{$attribute};
+                $name = $name->{$attribute};
             }
         }
         return $name;
--- a/modules/backend/formwidgets/RecordFinder.php
+++ b/modules/backend/formwidgets/RecordFinder.php
@@ -17,7 +17,6 @@ use Backend\Classes\FormWidgetBase;
  *        prompt: Click the Find button to find a user
  *        keyFrom: id
  *        nameFrom: name
- *        nameFromRelation: null
  *        descriptionFrom: email
  *        conditions: email = "bob@example.com"
  *        scope: whereActive
@@ -43,14 +42,9 @@ class RecordFinder extends FormWidgetBase
     public $keyFrom = 'id';
 
     /**
-     * @var string Relation column to display for the name
+     * @var string Relation column to display for the name. Can use the dot notation to query a related model
      */
     public $nameFrom = 'name';
-
-    /**
-     * @var string Get nameFrom using a model relationship
-     */
-    public $nameFromRelation;
 
     /**
      * @var string Relation column to display for the description
@@ -139,7 +133,6 @@ class RecordFinder extends FormWidgetBase
             'prompt',
             'keyFrom',
             'nameFrom',
-            'nameFromRelation',
             'descriptionFrom',
             'scope',
             'conditions',
@@ -280,8 +273,9 @@ class RecordFinder extends FormWidgetBase
             return null;
         }
 
-        if ($this->nameFromRelation && $this->relationModel->{$this->nameFromRelation}){
-          return $this->relationModel->{$this->nameFromRelation}->{$this->nameFrom};
+        list($relation,$attribute) = explode('.', $this->nameFrom,2);
+        if (!empty($attribute) && !empty($relation)){
+          return $this->relationModel->$relation->$attribute;
         }
 
         return $this->relationModel->{$this->nameFrom};
--- a/modules/backend/formwidgets/RecordFinder.php
+++ b/modules/backend/formwidgets/RecordFinder.php
@@ -273,7 +273,7 @@ class RecordFinder extends FormWidgetBase
             return null;
         }
 
-        list($relation,$attribute) = explode('.', $this->nameFrom,2);
+        list($relation,$attribute) = array_pad(explode('.', $this->nameFrom,2), 2, null);
         if (!empty($attribute) && !empty($this->relationModel->$relation)){
           return $this->relationModel->$relation->$attribute;
         }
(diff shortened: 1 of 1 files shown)


LukeTowers:
@alpapan I'm not going to go with this approach, but I would accept having a dot syntax for nested values in the `nameFrom` property. (So to access student name from an enrollment record: `nameFrom: student.name`)
It would look something like this:
```php
$nameFrom = explode('.', $this->nameFrom);

$name = $this->relatedModel;
foreach ($nameFrom as $attribute) {
    $name = $name->{$attribute};
}
```

alpapan:
Thanks 
I will do it slightly differently as I'd rather not overwrite variables.

```
public function getNameValue()
[...]

 list($relation,$attribute) = explode('.', $this->nameFrom,2);
 if (!empty($attribute) && !empty($relation)){
     return $this->relationModel->{$relation}->{$attribute};
  }

[...]
```

LukeTowers:
>  I'd rather not overwrite variables.

Could you explain why? My example was provided in that way for a very specific purpose: allowing infinite nesting of the records: `relation.subrelation.subsubrelation.propertythatreturnsanobject.name`

alpapan:
Because nameFrom is a string in the rest of the class and becomes an array in the example you gave. 

Multiple dots:
In that case how about something like this:

```
        $attribute = {$this->nameFrom};
        $related = $this->relationModel;
        while (strpos($attribute,'.') !== FALSE){
          list($relation,$attribute) = array_pad(explode('.', $attribute,2), 2, null);
          if (!empty($attribute) && !empty($related->$relation)){
            $related = $related->$relation;
          }
          return $related->$attribute;
        }
```

I still have to think what to do if $related->$relation evaluates to empty (i.e. a missing relation). An exception?

alpapan:
Ok implemented that way!

LukeTowers:
@alpapan was just rereading the original PR. Could you instead implement the `select` option from the relation dropdown?

alpapan:
Not sure I understand. I need the column information as shown in the record finder, especially given 100s of records.

LukeTowers:
Hmm. Okay then. If you can submit a PR to the docs too documenting this new ability I can merge this.

LukeTowers:
@alpapan any updates on this?

alpapan:
Sorry I didn't have time to learn to edit the docs. Will look into it

alpapan:
Please see https://github.com/octobercms/docs/pull/372

mjauvin:
I think this and related PR to the docs (https://github.com/octobercms/docs/pull/372) fell through the cracks... ;)

msimkunas:
Any news regarding merging this PR? I was pretty surprised to find out that this functionality is not included in the default October installation.

LukeTowers:
@msimkunas check in again in a few weeks after we launch 466, we're focused on finalizing that right now.

modules/backend/formwidgets/RecordFinder.php
```diff
@@ -273,7 +273,14 @@ public function getNameValue()
             return null;
         }
 
-        return $this->relationModel->{$this->nameFrom};
+        $name = $this->relationModel;
+        $nameFrom = explode('.', $this->nameFrom);
+        foreach ($nameFrom as $attribute) {
+            if (!empty($name->{$attribute})){
+              $name = $name->{$attribute};
```
LukeTowers, line 280:
```suggestion
                $name = $name->{$attribute};
```

modules/backend/formwidgets/RecordFinder.php
```diff
@@ -273,7 +273,14 @@ public function getNameValue()
             return null;
         }
 
-        return $this->relationModel->{$this->nameFrom};
+        $name = $this->relationModel;
+        $nameFrom = explode('.', $this->nameFrom);
+        foreach ($nameFrom as $attribute) {
+            if (!empty($name->{$attribute})){
```
LukeTowers, line 279:
```suggestion
            if (!empty($name->{$attribute})) {
```

modules/backend/formwidgets/RecordFinder.php
```diff
@@ -273,7 +273,14 @@ public function getNameValue()
...
         }
 
-        return $this->relationModel->{$this->nameFrom};
+        $name = $this->relationModel;
+        $nameFrom = explode('.', $this->nameFrom);
+        foreach ($nameFrom as $attribute) {
+            if (!empty($name->{$attribute})) {
+                $name = $name->{$attribute};
+            }
+        }
```
LukeTowers, line 282: FYI there's definitely a bug here in that if I say `nameFrom: property.location.name`, and `property.location.name` is empty but `property.name` isn't then it will select the wrong value.

msimkunas: Wouldn't a simple `break` statement in case of an empty value be enough here?

Something along the lines of:

```
foreach ($nameFrom as $attribute) {
    if (!empty($name->{$attribute})) {
        $name = $name->{$attribute};
    } else {
        $name = null;
        break;
    }
}
```

LukeTowers: @msimkunas probably yes, if @alpapan doesn't want to continue this PR then perhaps you can resubmit it.

daftspunk:
This would have to use html[array][name] syntax. Dot notation created too many problems internally (it conflicted too much with Laravel's internals) which is why we opted for this route

Let me know if you still need this feature and we can take a look at it for you. Thanks

PR #4092, opened 2019-01-29
Fix CloudFlare Rocket Loader break the script loading order.

itpcc:
Due to CloudFlare Rockey Loader, CF will automatically "try" to optimize page loading speed by changing script type attribute. This breaks the script execution order and makes the user unable to upload the file(s) in the backend using "FileUpload" widget.

However, [CloudFlare allows adding "data-cfasync"](https://support.cloudflare.com/hc/en-us/articles/200169436-How-can-I-have-Rocket-Loader-ignore-specific-JavaScripts-) to prevent this. And it was used in the [commit #3841](https://github.com/octobercms/october/pull/3841/files). So, I think it should be the way to go and fix.


LukeTowers:
@itpcc this would probably be better implemented in the addJs method of the AssetMaker trait in the System module as a default attribute that can be overrided if necessary. That way it fixes every use case all at once in the same place.

itpcc:
> @itpcc this would probably be better implemented in the addJs method of the AssetMaker trait in the System module as a default attribute that can be overrided if necessary. That way it fixes every use case all at once in the same place.

I, in fact, agree with you. However, since it was used by many projects, applying attribute in every single tag like that, in my opinion, should be in the major release and by the main contributor. Because it needs to do a major test.

BTW, thank you for your comment about this PR. 😄

LukeTowers:
@itpcc I'm fine with making that change in October (and not as a major release, October doesn't use semver) given that the only thing that it could break is causing a slight degradation in backend script loading performance amongst those users that use CloudFlare's rocketloader. However, given that a lot of backend scripts break because of the RocketLoader I'd rather have stability by disabling the rocketloader by default (while still providing a way for developers to enable it on specific scripts if desired). I'll check with @daftspunk too though.

LukeTowers:
@itpcc the mighty @daftspunk has spoken:

> Change it to cache => false
> Find the code that outputs the <script> and <style> tags, have it rewrite 'cache' to whatever the data-flavour-of-the-day is
> The intention is to disable caching, and so that's how it should be described

itpcc:
> Change it to cache => false

Hold on, it this issue relate with the caching issue?

LukeTowers:
@itpcc no, he's just saying call the attribute 'cache' for flexibility and then have the output function convert that attribute to the necessary data- attributes. Does that make sense?

LukeTowers:
We would be modifying addJs to add a default attribute of cache with the value of false and then modify the actual output method for the results of addJs to convert that attribute of cache to render the data-cfasync attribute on the generated script tag.

foralien:
It's not clear if this fix is going to be applied, but it's needed badly: with Cloudflare Rocket some part of backend are completely broken due to incorrect script loading order and there is no clear and safe way to fix it without touching core files. Will be much appreciated to have it in future updates.

LukeTowers:
@foralien feel free to submit a PR of your own implementing https://github.com/octobercms/october/pull/4092#issuecomment-458423428

itpcc:
@LukeTowers I've move attribute(s) into `addJs` as @daftspunk was recommended earlier. Could you please check the code again?

Thanks in advance.

modules/system/traits/AssetMaker.php
```diff
@@ -127,6 +127,21 @@ public function addJs($name, $attributes = [])
...
+        // If needed no cache, prevent Cloudflare Rocker Loader to add additional attributes.
+        // @See https://github.com/octobercms/october/pull/4092
+        if(!$attributes['cache']){
+            $attributes['data-cfasync'] = 'false';
+        }
+
+        // Unset unnecessery attribute(s) to be printed.
+        $attributes = array_except($attributes, [
+            'cache'
+        ]);
```
LukeTowers: This can be simplified to the following:

```php
// Prevent CloudFlare's Rocket Loader from breaking stuff
// @see octobercms/october#4092, octobercms/october#3841, octobercms/october#3839
if (isset($attributes['cache']) && $attributes['cache'] == false) {
    $attributes['data-cfasync'] = 'false';
    unset($attributes['cache']);
}

Comment (itpcc) [reply modules/system/traits/AssetMaker.php]: Okay. Changed.
At first, I'd like to make it easy to implement for issues in the future.  But that's fine.

File: modules/backend/formwidgets/FileUpload.php [hunk 2]
```diff
@@ -345,7 +345,10 @@ public function onSaveAttachmentConfig()
     protected function loadAssets()
     {
         $this->addCss('css/fileupload.css', 'core');
-        $this->addJs('js/fileupload.js', 'core');
+        $this->addJs('js/fileupload.js', [
+            'build' => 'core',
+            'data-cfasync'  => 'false'
```
LukeTowers: This should be `'cache'`

modules/backend/widgets/Form.php
```diff
@@ -159,7 +159,10 @@ public function bindToController()
      */
     protected function loadAssets()
     {
-        $this->addJs('js/october.form.js', 'core');
+        $this->addJs('js/october.form.js', [
+            'build' => 'core',
+            'cache'  => 'false'
```
LukeTowers: Make this consistent with the other definition please `'false'` vs `false`

foralien:
Oh! It's merged - so wonderful! Thank you a lot )

LukeTowers:
@itpcc @foralien we will use https://github.com/heathdutton/cloudflare/pull/4 & https://github.com/octobercms/october/commit/e246427463ac5edcc6b7115b33838a4367a4f9a9 to fix this issue going forward.

PR #4097, opened 2019-01-30
Fix dropdown custom options method argument order

### LINKED ISSUE #4095: Dropdown custom options method arguments wrong order
According to docs https://octobercms.com/docs/backend/forms#field-dropdown

> The fourth method uses a specific method declared in the model's class. In the next example the listStatuses method should be defined in the model class. **This method receives all the same arguments as the getDropdownOptions method**, and should return an array of options in the format key => label.

The custom options getter method of the dropdown formwidget should receive the same arguments as `getDropdownOptions` method, however the order of arguments does not match. 

See https://github.com/octobercms/october/blob/master/modules/backend/widgets/Form.php#L1263
and https://github.com/octobercms/october/blob/master/modules/backend/widgets/Form.php#L1278

mjauvin:
Fix #4095

Credit @iotch

### DIFF
--- a/modules/backend/widgets/Form.php
+++ b/modules/backend/widgets/Form.php
@@ -1276,7 +1276,7 @@ class Form extends WidgetBase
                 ]));
             }
 
-            $fieldOptions = $this->model->$fieldOptions($field->value, $field->fieldName, $this->data);
+            $fieldOptions = $this->model->$fieldOptions($field->fieldName, $field->value, $this->data);
         }
 
         return $fieldOptions;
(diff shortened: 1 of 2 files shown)


mjauvin:
Can someone test this with the unit test? I still don't have a suitable test environment to do this.

w20k:
@mjauvin will do! No problemo)

mjauvin:
Not required anymore, this issue/PR will get dropped as only documentation
needs updating.

--
Marc

PR #4098, opened 2019-01-31
Fix setting default values for repeater in update context

CptMeatball:
This PR is a completion of #3766 which seems abandoned.

When setting a default value for a repeater group field, it does not get filled when the context is anything other than `create`. This PR copies the code from #3766 and adds the changes suggest by @LukeTowers in https://github.com/octobercms/october/pull/3766#issuecomment-426720874.

How to (from old PR):
_You can replicate this in the test plugin by adding a default value for some field in the config/repeater_fields.yaml_

_When you create a new item using this repeater field, the default value will be set.
But when you update an item, the default value won't be set when adding a new group._

### DIFF
--- a/modules/backend/formwidgets/Repeater.php
+++ b/modules/backend/formwidgets/Repeater.php
@@ -268,7 +268,7 @@ class Repeater extends FormWidgetBase
         $config->alias = $this->alias . 'Form'.$index;
         $config->arrayName = $this->getFieldName().'['.$index.']';
         $config->isNested = true;
-        $config->shouldFetchDefaultValues = self::$onAddItemCalled;
+        $config->enableDefaults = self::$onAddItemCalled;
 
         $widget = $this->makeWidget('Backend\Widgets\Form', $config);
         $widget->bindToController();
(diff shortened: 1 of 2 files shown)


LukeTowers:
Looks good @CptMeatball. Could you submit a PR to the docs?

CptMeatball:
Where should the `enableDefaults` config field be set? When setting it inside either the `repeater` field definition or the `group` definition to a falsey value, the value is still populated. 

Removing the config value from the Repeater formwidget class disables the functionality, so it definitely should be set there. I'm just having a hard time overwriting the value of `enableDefaults` from within the yaml field definitions.

Flothus:
I get the same issue for default values in the nestedform widget when used inside of a repeater. Default values are filled in `create` context but not in `update` context.  

My Build is 455. The issue is probably related to this, somebody want's to look into it, or should I fill a new bug report?

w20k:
@Flothus create a new issue, would be easier to track.

Flothus:
@w20k 
#4380 plugin to reproduce the issue is attached

Issue #4089, opened 2019-01-25
Local and Global events exhibit different behavior 


SeriousKen:
There's an issue here https://github.com/OFFLINE-GmbH/oc-clockwork-plugin/issues/1 that seems to be related to the way local and global events are handled in `\System\Traits\EventEmitter` and `\October\Rain\Events\Dispatcher`. With the following class I did some tests.

``` php
class EventTest
{
    use \System\Traits\EventEmitter;
}
```

If I run the following code

``` php
// @return void
Event::listen('eventtest.event', function () {
    
});

// @return string
Event::listen('eventtest.event', function () {
    return 'Global event';
});

$test = new \October\Demo\Classes\EventTest();

// @return void
$test->bindEvent('event', function () {
        
});

// @return string
$test->bindEvent('event', function () {
    return 'Local event';
});

$response = $test->fireSystemEvent('eventtest.event', [], false);
dump($response);
```

I get

``` php
array:3 [
  0 => "Local event"
  1 => null
  2 => "Global event"
]
```

It seems that the null comes from the global event, but the processing of local events ignores any null responses and they don't end up in the result.

There also seems to be a discrepency between `fireSystemEvent` and `fireEvent` when halt is `true` and all of the event handlers return `null`.

For `fireEvent` you get `null` but for `fireSystemEvent` you get and empty array. I would expect `fireSystemEvent` when halting to return a single value ie `null` the same as `fireEvent`.

tobias-kuendig:
See https://github.com/octobercms/october/pull/4105

PR #4105, opened 2019-02-01
Discard empty event results

tobias-kuendig:
There is a bug when having a wildcard event listener the `onRefresh` method of the form widget stops working:

> Symfony\Component\Debug\Exception\FatalThrowableError: Unsupported operand types in /project/modules/backend/widgets/Form.php:450

The problem here is that a wildcard event will return an array with a single `null` value in it. 

This will trigger an exception when trying to merge the results with `$eventResult + $result` (null + [])

You can reproduce the bug by adding a simple wildcard event listener in any plugin and then use a `dependsOn` option on any field.

```php
\Event::listen('*', function() {});
```

When changing the field's value the `onRefresh` method is triggered and the exception is thrown.

My proposed solution is to filter out empty array values before merging.

### DIFF
--- a/modules/backend/widgets/Form.php
+++ b/modules/backend/widgets/Form.php
@@ -445,7 +445,7 @@ class Form extends WidgetBase
          */
         $eventResults = $this->fireSystemEvent('backend.form.refresh', [$result], false);
 
-        foreach ($eventResults as $eventResult) {
+        foreach (array_filter($eventResults) as $eventResult) {
             $result = $eventResult + $result;
         }
 


mjauvin:
It would be interesting to know why it's actually return null though. Did you look into that? Is it only returning null when the closure in your event listener is empty as in your example?

tobias-kuendig:
I've been tracking down the error for over an hour today with no real source of the problem. I currently do not have time to look into it any further.

@SeriousKen 	in issue #4089 	might be on to something though.

mjauvin:
I guess any listener could return null anyway, so your solution probably makes sense.

mjauvin:
And you're saying this only happens if your event listener uses a wildcard?

tobias-kuendig:
I do believe it's only wildcards, yes. On the other hand it could happen with other listeners too but the `*` events are just triggered in enough cases where this problem could occur. :thinking:

w20k:
@tobias-kuendig could you test solution that @SeriousKen provided, most probably that should fix everything event related :)

Library PR: https://github.com/octobercms/library/pull/377

tobias-kuendig:
Thanks @w20k, I'll take a look at this next week.

tobias-kuendig:
The fix from octobercms/library#377 is working in my case. The error no longer appears. :+1:

w20k:
@tobias-kuendig should this PR be closed then? Or it's a separate issue? :)

tobias-kuendig:
I think this PR can be closed now since the dispatcher checks for empty event responses. Tell me if you thing otherwise. Thanks @SeriousKen 	for fixing the issue!

PR #4106, opened 2019-02-01
Allow controller middleware in backend controllers

bennothommo:
Related PR: https://github.com/octobercms/october/pull/4088

Adds support for controller middleware in constructor method of backend controllers.

The PR rebases the `Controller` class in the Backend to use Laravel's Controller class, which allows for middleware to be inserted. A custom loader is added to the main `BackendController` constructor which retrieves applicable middleware and copies it into that class, as it is this class that is fed through Laravel's request pipeline.


modules/backend/classes/BackendController.php
```diff
@@ -49,6 +49,26 @@ class BackendController extends ControllerBase
      */
     public function __construct()
     {
+        // Find requested controller to determine if any middleware has been attached
+        $pathParts = explode('/', Request::path());
```
LukeTowers: Does this support installations that are in subfolders? (i.e. not attached to the root domain)

bennothommo: It should now. :)

modules/backend/classes/BackendController.php
```diff
@@ -49,6 +49,26 @@ class BackendController extends ControllerBase
      */
     public function __construct()
     {
+        // Find requested controller to determine if any middleware has been attached
+        $pathParts = explode('/', str_replace(Request::root() . '/', '', Request::url()));
+        if ($pathParts >= 2) {
+            // Drop off preceding "backend" URL part
+            array_shift($pathParts);
+            $path = implode('/', $pathParts);
```
LukeTowers: Does the above work correctly when the `backendUrl` is set to an empty string? (Usually for sites that don't use the CMS module, purely backend based)

bennothommo: I didn't realise that was possible. I doubt it will at this stage but will make allowances for that.

LukeTowers: Sure is possible, required a little bit of tweaking here and there to make it seamless but it works great for building dedicated web applications powered by OctoberCMS

bennothommo: All good, managed to work it out - needed to remove the CMS module from Composer and `config/cms.php`. :)

bennothommo:
@LukeTowers This should be ready for review now, to unblock #4088.

LukeTowers:
@w20k can you review?

w20k:
@LukeTowers will try to deal with it tomorrow morning. 
Chaotic month :) If you have priory tickets, just ping me on those! Thank! ;)

w20k:
@bennothommo shouldn't this import `use October\Rain\Extension\Extendable;` be removed from Controller.php?

w20k:
@LukeTowers @bennothommo I've tested PR in par with #4088 and it works 👍 
But if you both don't mind, will do a few more. Hopefully tomorrow, also in the morning :) 
And will inform watchers!

bennothommo:
@w20k Removed that line. Thanks heaps for testing :)

w20k:
@LukeTowers @bennothommo - testing is finished. Works great, couldn't find any bugs 😄 Getting old))
Status - ✅

modules/backend/classes/BackendController.php
```diff
@@ -203,4 +258,17 @@ protected function parseAction($actionName)
...
+    /**
+     * Determine if the given options exclude a particular method.
+     *
+     * @param  string  $method
+     * @param  array  $options
+     * @return bool
+     */
+    protected static function methodExcludedByOptions($method, array $options)
+    {
+        return (isset($options['only']) && ! in_array($method, (array) $options['only'])) ||
```
w20k: @bennothommo could you use only one-way: space after `!` or none at all 😄  (68,136: `!is_null`, 62: `!empty`, 270: `! in_array`)

bennothommo: @w20k In my defence, I copied Laravel's homework :P
https://github.com/laravel/framework/blob/aa36e209480de3bb36036a959fc3b114a21e7ebf/src/Illuminate/Routing/ControllerDispatcher.php#L76

w20k: Thanks, @bennothommo!

filipac:
@w20k @LukeTowers I think this introduced a new bug.

I am having a repeater that includes some codeeditor fields. When I try to save, I get this error after I apply this commit:
```[2019-03-12 11:24:22] local.ERROR: October\Rain\Exception\SystemException: There is no an authenticated user to load or save preferences for. in /Users/filipac/Code-Meevo/pillbox/modules/backend/models/UserPreference.php:33```

**Comment by filipac** [comment]
If I go to a commit before the one this was applied, the code works, after it breaks.

**Comment by filipac** [comment]
Tracked it down to this line:
```php
$controllerClass = '\\'.$author.'\\'.$plugin.'\Controllers\\'.$controller;
```
$author and $plugin is lowercase. So it results in a not found controller.

w20k:
@filipac if you fix, like this, will this monkey patch work (I'm PRO in monkey patching 😄 )?

`$controllerClass = '\\'. ucfirst($author).'\\'. ucfirst($plugin).'\Controllers\\'.$controller;`

filipac:
@w20k ucfirst for controller too.. But it won't work if controller has another uppercase letter...

bennothommo:
@filipac I believe PHP classes are case-insensitive, so the casing of letter should have no bearing on the line. Would you be able to try the changes in PR #4190 and see if that makes a difference?

w20k:
@bennothommo I think so also, but just trying to cover some unimaginable cases :D

Issue #3439, opened 2018-03-10
DateTime picker in mode:time stores DATE and TIME I need just TIME to be stored


roulendz:
##### Expected behavior
Using repeater for storing in JSON day of the week and working hours.

Somehow saves not just hours/time but also date.
```
businessHours:
            label: 'Classroom working days and hours'
            prompt: 'Add new item'
            span: left
            type: repeater
            tab: 'Classroom working hours'
            form:
                fields:
                    dow:
                        label: Dropdown
                        options:
                            - Sunday
                            - Monday
                            - Thursday
                            - Wednesday
                            - Thursday
                            - Friday
                            - Saturday
                        span: left
                        default: '1'
                        emptyOption: '--'
                        type: dropdown
                    start:
                        label: 'Start time'
                        mode: time
                        span: left
                        default: '10:00'
                        type: datepicker
                    end:
                        label: 'End time'
                        type: datepicker
                        mode: time
                        span: auto
                        default: '18:00'
```
#### I can not figure out why date is stored along ;(
```
[
   {
      "dow":"0",
      "end":"2018-03-10 18:00:00",
      "start":"2018-03-10 10:00:00"
   },
   {
      "dow":"1",
      "end":"2018-03-10 18:00:00",
      "start":"2018-03-10 10:00:00"
   },
   {
      "dow":"2",
      "end":"2018-03-10 18:00:00",
      "start":"2018-03-10 10:00:00"
   },
   {
      "dow":"3",
      "end":"2018-03-10 18:00:00",
      "start":"2018-03-10 10:00:00"
   },
   {
      "dow":"4",
      "end":"2018-03-10 18:00:00",
      "start":"2018-03-10 10:00:00"
   },
   {
      "dow":"5",
      "end":"2018-03-10 18:00:00",
      "start":"2018-03-10 10:00:00"
   },
   {
      "dow":"6",
      "end":"2018-03-10 18:00:00",
      "start":"2018-03-10 10:00:00"
   }
]
```
##### Reproduce steps
Maybe the issue is connected that I am storing all that repeater data in one JSON field - bussinesHours

##### October build
434

roulendz:
![HTML markup](https://fv16.failiem.lv/thumb_show.php?i=r72kwgu3&view)


#### I think the problem is that date and time are stored in one field, just rendered to separate fields. 

### So when data is stored, in that field also today date is coming along.

Maybe to now overwork DateTime fields, there is way on saving/crating to extract that time and remove today date.

[...]

PR #4111, opened 2019-02-04
Set the datetime value with time 00:00:00 when using the date mode

fansaien:
Set the datetime value with time 00:00:00 when using the date mode in Date picker.

### DIFF
--- a/modules/system/assets/ui/js/datepicker.js
+++ b/modules/system/assets/ui/js/datepicker.js
@@ -141,7 +141,7 @@
     DatePicker.prototype.onSelectDatePicker = function(pickerMoment) {
         var pickerValue = pickerMoment.format(this.dbDateFormat)
 
-        var timeValue = this.getTimePickerValue()
+        var timeValue = this.options.mode === 'date' ? '00:00:00' : this.getTimePickerValue()
 
         var momentObj = moment
             .tz(pickerValue + ' ' + timeValue, this.dbDateTimeFormat, this.timezone)
@@ -316,7 +316,8 @@
         format: null,
         yearRange: 10,
         firstDay: 0,
-        showWeekNumber: false
+        showWeekNumber: false,
+        mode: 'datetime',
     }
 
     // PLUGIN DEFINITION
--- a/modules/system/assets/ui/js/datepicker.js
+++ b/modules/system/assets/ui/js/datepicker.js
@@ -317,7 +317,7 @@
         yearRange: 10,
         firstDay: 0,
         showWeekNumber: false,
-        mode: 'datetime',
+        mode: 'datetime'
     }
 
     // PLUGIN DEFINITION
--- a/modules/backend/formwidgets/DatePicker.php
+++ b/modules/backend/formwidgets/DatePicker.php
@@ -1,5 +1,6 @@
 <?php namespace Backend\FormWidgets;
 
+use Config;
 use Carbon\Carbon;
 use Backend\Classes\FormField;
 use Backend\Classes\FormWidgetBase;
@@ -120,8 +121,14 @@ class DatePicker extends FormWidgetBase
         if ($value = $this->getLoadValue()) {
 
             $value = DateTimeHelper::makeCarbon($value, false);
+            if ($this->mode === 'date' ) {
+                $backendTimeZone = \Backend\Models\Preference::get('timezone');
+                $value->setTimezone($backendTimeZone);
+                $value->setTime(0,0,0);
+                $value->setTimezone(Config::get('app.timezone'));
+            }
+            $value = $value->toDateTimeString();
 
-            $value = $value instanceof Carbon ? $value->toDateTimeString() : $value;
         }
 
         $this->vars['name'] = $this->getFieldName();
--- a/modules/backend/formwidgets/datepicker/partials/_datepicker.htm
+++ b/modules/backend/formwidgets/datepicker/partials/_datepicker.htm
@@ -13,6 +13,7 @@
         <?php if ($formatMoment): ?>data-format="<?= $formatMoment ?>"<?php endif ?>
         <?php if ($minDate): ?>data-min-date="<?= $minDate ?>"<?php endif ?>
         <?php if ($maxDate): ?>data-max-date="<?= $maxDate ?>"<?php endif ?>
+        data-mode="<?= $mode ?>"
         <?php if ($yearRange): ?>data-year-range="<?= $yearRange ?>"<?php endif ?>
         <?php if ($firstDay): ?>data-first-day="<?= $firstDay ?>"<?php endif ?>
         data-show-week-number="<?= $showWeekNumber ?>"


w20k:
@fansaien could you describe in detail why do you need it? With an example?

fansaien:
Hello @w20k ,
I need to use the datepicker with the date mode, like this config:
```YAML
fields:
    start_date:
        label: Start Date
        type: datepicker
        mode: date
        span: left
        default: now
```

Right now, the "hidden" input value's time is when the user open this control's time, and can not be changed, like this:
```HTML
<!-- Data locker -->
<input type="hidden" name="start_date" id="Form-field-start_date" value="2019-02-04 22:30:50" data-datetime-value="">
```



I think  if the input value's time always can be set 00:00:00 that will be much better , and also it depends on the user's time zone.

For example, my current timezone is UTC - 6, the HTML is this: 
```HTML
<input type="hidden" name="start_date" id="Form-field-start_date" value="2019-02-04 06:00:00" data-datetime-value="">
```
With this value, I can easy save the date in my controller like this:
```PHP
$model->start_at = DateTime::makeCarbon($data['start_date']);
$model->end_at = DateTime::makeCarbon($data['end_date']);
$model->end_at = $model->end_at->addDay(); 
```

LukeTowers:
@w20k note that he means time set to 0 in accordance with the application's timezone (which should always be UTC, but I digress.

w20k:
@fansaien I'm also against changes inside the core. I'd suggest looking at the JS side, where you could define your '0:00:00' time with ease.

Like so, an example:`$("#datepicker_id_or_class").data('datetimepicker').setLocalDate(new Date(year, month, day, 00, 00));`

LukeTowers:
@w20k note that I'm biased in this issue, @fansaien is my coworker and we're working on a project that would utilize this. I would say that the current behavior of sending a time value isn't ideal when the mode is `date` (just like there's an issue with the date being sent when the mode is `time`). My preferred solution to this would probably be to send only the date itself, but that can get tricky to handle in a robust manner considering timezones and dates potentially being stored in a timestamp or datetime schema column.

fansaien:
Hello @w20k  Thank you for your response and suggestion. I really understand it is not easy to modify the core. 

I think that may be a bug, because it stored a wrong time.

For example:
If a user open the date mode datepicker at 2019-02-05 14:00:00 (UTC), and the begin_date will be stored with that value, the end_date will be 2019-02-06 14:00:00. Then others try to find some data in 2019-2-5, the condition will be 2019-02-05 14:00:00 <= x < 2019-02-06 14:00:00. That can not get the right results.

I suggest the mode option looks like this:

- datetime:  Stored the date and time
- date:         Only stored the date, and time is 0
- time:         Only stored the time 

I also found a related issue, that he only needs the time value with datepicker when the mode is time, https://github.com/octobercms/october/issues/3439

w20k:
@LukeTowers, @fansaien:
Here is a small example, why it could cause a lot of issues.

Example:
Timezone is GMT-6 -> 23.01.2019 with time set to 0:00:00
app.timezone is UTC 0 -> 23.01.2019 with time set to 6:00:00

But if your zone is `+`, like mine:
Timezone is GMT+2 -> 23.01.2019 with time set to 0:00:00
app.timezone is UTC 0 -> 22.01.2019 with time set to 22:00:00 (**You will get a different day.**)

I might be wrong If Carbon doesn't convert date when timezone is set.
```
  $backendTimeZone = \Backend\Models\Preference::get('timezone');
  $value->setTimezone($backendTimeZone);
  $value->setTime(0,0,0);
  $value->setTimezone(Config::get('app.timezone'));
```

fansaien:
Hello @w20k ,

UTC 22.01.2019 with time set to 22:00:00 is not the different day. Because we usually compare with the date time not just date.

If I need to get the results from +2 timezone on 2019-01-23 day,
It will search like this: 2019-01-22 22:00:00(UTC) <= x < 2019-01-23 22:00:00(UTC)

2019-01-22 22:00:00(UTC)  = 2019-01-23 00:00:00 (UTC +2)

w20k:
Hello, @fansaien fair enough.

@LukeTowers will re-run tests once more, and PR is good to go. Though I really do sway toward sending only the date. 

Side note: There is only one corner case, can't imagine who would use it, but when your Backen and DB have different time zones.

modules/system/assets/ui/js/datepicker.js
```diff
@@ -316,7 +316,8 @@
         format: null,
         yearRange: 10,
         firstDay: 0,
-        showWeekNumber: false
+        showWeekNumber: false,
+        mode: 'datetime',
```
w20k: @fansaien could you remove trailing comma.

fansaien: Sure, I submit that.

LukeTowers:
@w20k we explicitly tell people to not have a different timezone between your DB and app.timezone config, so that shouldn't be a problem.

w20k:
@fansaien Thank a lot for helping ;)
@LukeTowers testing is done. Could be merged.

nizzac:
This pull request is causing an issue for me in the UI where the Backend Preferences timezone is set to (UTC -00:00) Europe/London.

After a discussion with @w20k it seems the issue is related to timezones that have DST. (Daylight Savings Time)

So when I pick a date with the datepicker that is after the date the clocks go forward... the value set on the hidden input is being set as an hour earlier, thus changing the day...

For example, if I choose 1st April 2019.... rather than the value being set to `2019-04-01 00:00:00` it's being set to `2019-03-31 23:00:00`

LukeTowers:
@fansaien can you comment on the above?

fansaien:
Yes, I will do more test and reply to him.

fansaien:
@nizzac  I tested on both my local timezone -6 and your timezone +1 (After March 31)

For my timezone UTC-6, if I choose 2019-04-02, the value is 2019-04-02 06:00:00(UTC). 

```HTML
<input type="hidden" name="start_date" id="Form-field-start_date" value="2019-04-02 06:00:00" data-datetime-value="">
```
For your timezone London, it is UTC + 1.
The HTML is:
```HTML
<input type="hidden" name="start_date" id="Form-field-start_date" value="2019-04-01 23:00:00" data-datetime-value="">
```
I think that's correct, because the value of input element is not your local timezone, it is the UTC timezone value.

You can also try this website : https://www.timeanddate.com/worldclock/converted.html?iso=20190401T000000&p1=136

2019-04-01 00:00:00(London local Timezone)  = 2019-03-31 23:00:00(UTC)

nizzac:
If you save that data to the database and then try reloading the page. The date set on the date picker is now wrong. With Europe/London timezone...

This is for fields that are set to mode: date

Isnt the issue here that you wantes to set times to 00:00:00 for date pickers? There shouldnt be a need to manipulate the time at all? Especially when doing so changes the date picked.

fansaien:
@nizzac 

I tested with Europe/London timezone on mysql date column and timestamp column.

date column: The date picker is wrong, because the time has been lost.
timestamp column: The date picker is still right.

I suggest using timestamp to save the date value. The datepicker is designed to support datetime, date and time. It will post UTC datetime to the server. So, if the database save timestamp if will be fine.

nizzac:
> date column: The date picker is wrong, because the time has been lost.

This is a problem though right? People will use date columns in the database when they want to store only a date.

Also, if you're using a datepicker with `mode: date` you'd expect the date to be set in the database as you've selected it.

I might be misunderstanding how date and time data is handled here but to me the 4th of April 2019 is  4th of April 2019, regardless of what timezone you're in.

fansaien:
@nizzac 

Hello, please think about the start date and end date or events in the calendar, that should be different.

With the date mode, the previous version will save the current time, and this time can not be changed. In the way, the value should be "2019-04-02 08:54:01", "2019-04-02 15:03:23" or whatever else.

I can not just remove the time of the date time value because I do not know others how to use the time value. My solution is just keep the time value correct.

My suggestions are:
1. If you would like the date column, you can deal with the UTC datetime before you save it.
2. Please create a PR with a new mode name, only keep the date in the value.

w20k:
@fansaien might be related to this PR, could you spend few minutes of your time, if it's not - just ignore it. Thanks a lot!
Issue: https://github.com/octobercms/october/issues/4222

LukeTowers:
@fansaien it's not, just a user error, you can ignore that issue.

w20k:
@fansaien, sorry my bad, rushed with assumption ;) not an issue!

fansaien:
@w20k No worries 👍

nizzac:
> Hello, please think about the start date and end date or events in the calendar, that should be different.

If I was saving an event in a calendar that needed start and end times times then I wouldn't use a DATE column type in the database, I would use a data type that handled dates & times such as a timestamp.

> With the date mode, the previous version will save the current time, and this time can not be changed. In the way, the value should be "2019-04-02 08:54:01", "2019-04-02 15:03:23" or whatever else.

If you need to alter the time of an attribute, then you should use mode: datetime picker.

> I can not just remove the time of the date time value because I do not know others how to use the time value. My solution is just keep the time value correct.

Again, you're talking about time, but this issue is relating to the `mode: date` option for the picker. If you wanted to manipulate time, you should use datetime right?

> @fansaien it's not, just a user error, you can ignore that issue.

I don't think this is a user error.
A backend users timezone setting should make a difference to a value if the datepicker mode is set to date only.

LukeTowers:
@fansaien I think we need to find an alternative solution to the original problem. This is causing issues even on the project that it was originally intended for, see the shipments start date field, it changes after saving

w20k:
@nizzac might cause issues, and should probably work 😄
```
// Set both timezones to UTC to disable converting between them
        if (this.ignoreTimezone) {
            this.appTimezone = 'UTC'
            this.timezone = 'UTC'
        }
```

nizzac:
I don't know what you mean

fansaien:
@LukeTowers , the datepicker will post a UTC datetime string to the server, no matter the mode is datetime, date or time. This datetime string will be converted to PHP Carbon/Datetime type.

If just want to post date or time string to the backend, I think should try to add new mode for that.

nizzac:
That doesn't make any sense. Why create a new mode for something that there is already a mode for?
If the mode is just date, or just time - then the timezone shouldn't come into the equation. 

It only needs to be taken into account when the datepicker is in mode "datetime"

I can't think of a situation where you would want to choose just a date with no ability to change the time of the value and want it to be affected by timezones.

LukeTowers:
@fansaien can we go back to the original problem? Why is it necessary to set the time part of the date when in mode `date` to 00:00:00?

fansaien:
@LukeTowers I do not think we can go back. 

The original version will save the time as the user open this datepicker's value. So the time can be any one: 8:34, 13:21...... To set the 0:0:0, we can know the exact date of users's timezone.

nizzac:
Just out of curiosity, is there a reason you cant add some logic in a beforeSave method that sets the time to 00:00:00 for your particular situation?

LukeTowers:
@fansaien that's not what I meant. Why does the time have to be 0:0:0? I'm starting to think that a beforeSave implementation is a better solution for your original issue.

fansaien:
@LukeTowers 0:0:0 is user's timezone value. When user select one date, the js will convert 0:0:0 to UTC and store into the input value, just like  

<input type="hidden" name="start_date" id="Form-field-start_date" value="2019-04-02 06:00">

I think it is the right value. We can handle this UTC value in the server side.

Samuell1:
I think this PR should be revisited i think it feel little off we setting 00:00:00 even for user timezone it should add hours to it then conver to UTC and save 00:00:00 not in reverse order. @LukeTowers

LukeTowers:
@Samuell1 if you could submit a PR with your proposed fix and a real world example of the problem this caused, then that would help out a lot. I hate dealing with timezone issues, it sucks so much.

bennothommo:
@LukeTowers I've just had a look through this thread. I'm with you - I don't believe the time should play any part whatsoever in a `date` mode datepicker widget, and feel it would be better to strip this out completely and assume that the user is doing the sensible thing and using a `DATE` column to store it in the DB.

Samuell1:
@bennothommo Looks like this line is doing funky stuff because it shouldnt be for user 00:00:00 but converted to timezone https://github.com/octobercms/october/pull/4111/files#diff-c204464daa968cdaa706a81562b46c69R144

bennothommo:
Let's continue the conversation in https://github.com/octobercms/october/issues/5072 - @multiwebinc has submitted an issue that is pretty much covering what's being discussed here.

multiwebinc:
@bennothommo I should have done a search first.

bennothommo:
@multiwebinc That's fine, this was buried deep so it might've been difficult to find. Your issue summarises the problem perfectly so we'll use that to track the bug.

PR #4113, opened 2019-02-05
Add configuration for loading related model in FormField

SeriousKen:
By default the FormField only gets the keys for a relation, but this makes it impossible to use NestedForm in conjuction with a relation. This enhancement adds a flag `useRelationData` which defaults to `false` for backwards compatibility and when `true` fetches the relation's data into the field value.

### DIFF
--- a/modules/backend/classes/FormField.php
+++ b/modules/backend/classes/FormField.php
@@ -690,7 +690,7 @@ class FormField
         foreach ($keyParts as $key) {
 
             if ($result instanceof Model && $result->hasRelation($key)) {
-                if ($key == $lastField) {
+                if ($key == $lastField && ! $this->getConfig('useRelationData', false)) {
                     $result = $result->getRelationValue($key) ?: $default;
                 }
                 else {


LukeTowers:
@SeriousKen examples?

SeriousKen:
@LukeTowers for example, you could have the following configuration for a user model with a one-to-one related profile model.

``` yaml
fields:
  username:
    label: Username
    type: text
  email:
    label: Email
    type: text
  profile:
    label: Profile
    type: nestedform
    form: $/october/demo/models/profile/fields.yaml
```

This currently works when saving but doesn't work when updating as the nested form is not filled from the profile model as `FormField::getFieldNameFromData` only returns the primary key of the relation. By adding a config option to `FormField` it makes it able to retrieve the actual related model if required, with the default being `false`. The above form config would then become:

``` yaml
fields:
  username:
    label: Username
    type: text
  email:
    label: Email
    type: text
  profile:
    label: Profile
    type: nestedform
    useRelationData: true
    form: $/october/demo/models/profile/fields.yaml
```

By having this in the `FormField` class makes it possible for this feature to be used for other types of form widgets.

LukeTowers:
@SeriousKen does that actually work? AFAIK updating related models with the NestedForm FormWidget is not officially supported; although it is a very interesting idea / use case. @munxar @bennothommo any input?

bennothommo:
@LukeTowers unfortunately I haven't had the chance to use the nested form widget yet, so I couldn't give an educated opinion on it.

At a glance, it looks like a neat idea, but there would be a lot of things to consider, such as what would happen for one-to-many relationships (I'd probably limit this to just one-to-one relationships), or what would happen if there is a nested form inside a nested form that both reference related models.

LukeTowers:
@seanthepottingshed any input on this?

seanthepottingshed:
@LukeTowers 

I'm currently in the UK, will have a proper look when I get back to work on Monday.

seanthepottingshed:
@LukeTowers 

Yep, I think this would be really useful. My only thoughts ( without testing ) when the `nestedForm` inherit the target model's validation rules / custom messages and `filterFields` method?

LukeTowers:
I have no idea. Haven't tested this at all. @SeriousKen any input?

SeriousKen:
@seanthepottingshed @LukeTowers The model's validation should work as normal when saving the parent model. I can test a few scenarios to see what happens and see if anything doesn't work as expected.

LukeTowers:
@SeriousKen any updates on this?

SeriousKen:
Sorry for not getting back sooner, had a quick look today and can confirm that the `filterFields` method is not called on the relation model by a nested form. Also, dropdowns are looking for the get___Options method on the parent model, not the relation. There is obviously some extra work involved in getting this to work in a way that would be acceptable.

paishin:
I believe working with nested relationships is my biggest painpoint when working with October's backend, it would be great if we could have something like this working; it opens up so many possibilities.

Tested this last night this is what I fount:
- Works fine for one-to-one relationships
- Fails to load relation widget in nested form
- custom widget's getSaveValue is not called upon saving

paishin:
I have created a new widget called RelatedForm that solves the issue of not being able to load relation widgets within that NestedForm does and also takes care of the relation data issue that was mentioned above without having to specifically specify the 'useRelationData' attribute.

I am still facing the initial issue of getSaveValue of child widgets not being triggered upon saving (strangely enough, getSaveValue does get called when I set dependsOn  and then change the value of depended formfield but not on save).

Anyone has an idea why this event may not be triggering?

paishin:
Nevermind the above, I found the solution. I just had to trigger getSaveValue for each property manually for everywidget, now everything seems to be working fine.

I will create a pull request for you to test

LukeTowers:
@SeriousKen what's the status on this?

daftspunk:
Good idea but the implementation is not quite right. @SeriousKen get in touch if you still need help with this

PR #4116, opened 2019-02-05
Fix form recordpicker disabled behaviour. Fix #3983

### LINKED ISSUE #3983: disabled:true on form recordpicker results in integrity constraint violation
##### Expected behavior
Adding `disabled: true` to a recordpicker in fields.yaml should not send the (unchanged) value to the FormController.

##### Actual behavior
The unchanged value is sent to the FormController, where it is set to null (because it is disabled), resulting in an attempt to override the record id with null.

##### Reproduce steps
Make a fields.yaml file with a `recordpicker` element and set `disabled: true`.

##### October build
434

It seems that this issue lies in the `getSaveData()` method in the `Backend\Widgets\Form` class. Adding the following code into the `foreach` on line 1166 appears to fix the issue but I haven't been able to run tests to confirm.

```
if ((isset($widget->config->disabled) && $widget->config->disabled) ||
    (isset($widget->config->hidden) && $widget->config->hidden)) {
        continue;
}
```

tmus:
Fixes #3983.

If a formwidget is disabled or hidden, the value is no longer added added to the form data to persist. This fixes an issue where a null value was being saved.

### DIFF
--- a/modules/backend/widgets/Form.php
+++ b/modules/backend/widgets/Form.php
@@ -240,7 +240,7 @@ class Form extends WidgetBase
     public function renderField($field, $options = [])
     {
         $this->prepareVars();
-        
+
         if (is_string($field)) {
             if (!isset($this->allFields[$field])) {
                 throw new ApplicationException(Lang::get(
@@ -1165,6 +1165,11 @@ class Form extends WidgetBase
          */
         foreach ($this->formWidgets as $field => $widget) {
             $parts = HtmlHelper::nameToArray($field);
+
+            if ((isset($widget->config->disabled) && $widget->config->disabled)
+                || (isset($widget->config->hidden) && $widget->config->hidden)) {
+                continue;
+            }
 
             $widgetValue = $widget->getSaveValue($this->dataArrayGet($result, $parts));
             $this->dataArraySet($result, $parts, $widgetValue);


bennothommo:
@LukeTowers definitely wouldn't hurt, although this fix is currently going to the `master` branch so my Repeater changes wouldn't be in that branch yet.

Should I make a separate PR or is this change supposed to go to the `develop` branch?

w20k:
This pull request has been made to the wrong branch. Please review the contributing guidelines as all PRs need to be made to the develop branch.

w20k:
@bennothommo fixed that ;)

tmus:
My bad, thanks @w20k ;)

bennothommo:
@w20k Awesome! @tmus, if you don't mind, would you be able to add the same fix to the `Repeater` form widget here :) https://github.com/octobercms/october/blob/6dbe9be6154cc328ce21e4bdeba015327a410ac7/modules/backend/formwidgets/Repeater.php#L206

LukeTowers:
@bennothommo is there any way that the Repeater could utilize its nested Form widgets to get the proper save processing done so that it doesn't have to replicate the functionality?

bennothommo:
@LukeTowers I do seem to recall trying the `getSaveData` method for each form widget in the repeater, and there being something that didn't work quite right. I'll give it another try in the next day or so - I agree that the code shouldn't be repeated unless absolutely necessary.

bennothommo:
@LukeTowers finally got around to following this up.

The reason I opted against `getSaveData` from the form widgets was to have the opportunity to clear out empty values from the widget data. The specific test scenario I was working with had datatables inside a repeater, that was itself inside another repeater. If one iteration of the inner repeater had 3 items, `getSaveData` was returned 3 items for all of them, even if a separate inner repeater had only 1 item. I believe it has something to do with the hidden `index` value that is used in repeaters.

LukeTowers:
@bennothommo that sounds like it's a separate bug, are you able to look into a fix for that?

bennothommo:
@LukeTowers I'll give it a go :)

LukeTowers:
@tmus now that @bennothommo's changes are merged in, what's left on this?

bennothommo:
@LukeTowers I think @tmus's changes just need to be tested and can be merged once confirmed working. My PR should now be using their form widget changes already.

tmus:
@LukeTowers @bennothommo yeah that's right. I did run it against the existing test suite without issue, and tested it manually, but I'm sure you guys will want to cast an eye over it. I've sort of stepped away from October for the immediate future, but let me know if there's anything else you need.

iboved:
We need the same check for **readOnly** option.
For example, checkbox doesn't support HTML `readonly` attribute. To make it readonly or disabled, we have to use `disabled` HTML attribute in the HTML markup, no matter what we use - `disabled: true` or `readOnly: true` configuration.

PR #4124, opened 2019-02-07
Reimplemented infinite loading bar with transforms instead of widths

Josniii:
This is a first attempt at reimplementing the base infinite loading bar included with OctoberCMS, using transforms instead of animating widths. This is better performance wise as mentioned in octobercms#4118.

Note that this is not a one-to-one reimplementation, the keyframes differ in percentages slightly. I thought the current implementation was needlessly hard to read, so I implemented a slightly simpler animation generator instead.


w20k:
@Josniii Thanks for making a PR. Will test it tomorrow ;)

w20k:
@LukeTowers - tested, works well. Though didn't find any GPU spikes, but performance vice this solution is much better.

PR #4128, opened 2019-02-10
Enable to change theme partial override on Component

Rike-cz:
### Situation
I have article in WYSIWYG editor (e.g. Static Page) and I want to place more than one photo album carousels as Snippets. First Snippet gets alias "_myCarousel_", second "_myCarousel1_" etc.
I have also component template override in theme partials folder called "_myCarousel_". First Snippet on page process this override, but second one is looking for "_myCarousel1_" folder in theme partials. As developer I do not know, how many carousels user will use... five? Ten?

### Solution
Component renderer looks for partial override in folder based on his alias. In this case I need something, what force component to find template partial in only one concrete folder. I added to Component the method `themePartialOverride`, that can be override by developer to force only one theme partial folder to use.

This pull request is reply on https://github.com/rainlab/pages-plugin/issues/295


LukeTowers:
@Rike-cz I'd prefer that the snippet looks for its alias override first, and then failing that it looks for an override for its default alias, and then falls back to the plugin provided partial.

Rike-cz:
@LukeTowers Great idea! I tried to implement it.

LukeTowers:
@Rike-cz what's the status of this?

Rike-cz:
There is problem with backward compactibility. I solved it with adding compactibility switch to config file, but I am not sure if it's good solution. But I am using it in my projects and it works.

Rike-cz:
What with this PR? I am still using it in production and is good feature.

LukeTowers:
@Rike-cz you'll have to resubmit it, the original repo it was submitted from was deleted.

PR #4132, opened 2019-02-12
Set the report widget alias from container and base

fansaien:
ReportContainer.php already get the unique widget alias, but forgot to set to the widget. 

In this way, we can get different widget alias with the same reportwidget, just like this:

report_container_dashboard_4
report_container_dashboard_5


LukeTowers:
Implemented in https://github.com/octobercms/october/commit/bbf398133ed30a3e616e874c59359c878875ea61

PR #4140, opened 2019-02-15
Add  popup-allow-focus for the second range input

filipac:
When list filters are in a modal, you cannot select the second field because it does not have  popup-allow-focus class.


w20k:
@filipac could you post before/after screenshots|gifs?

filipac:
Sure..
Before:
before

After:
after

w20k:
@LukeTowers this also - done ;) 
Status - ✅

PR #4142, opened 2019-02-15
Return 500 when asset exception caught

alekseyp:
Right now if you try to load assets that don't exist `200` is returned.
This PR will return `500` for any caught exception.

Before:
```
HTTP/1.1 200 OK
Server: nginx/1.15.0
Content-Type: text/html; charset=UTF-8
Transfer-Encoding: chunked
Connection: keep-alive
Cache-Control: no-cache, private
Date: Fri, 15 Feb 2019 20:57:38 GMT

/* The combiner file &#039;541208ee51c521bbfa180fe046910ad0&#039; is not found. */⏎          
```

After:
```
HTTP/1.1 500
Server: nginx/1.15.0
Content-Type: text/html; charset=UTF-8
Transfer-Encoding: chunked
Connection: keep-alive
Cache-Control: no-cache, private
Date: Fri, 15 Feb 2019 21:18:23 GMT

/* The combiner file &#039;725cc5a8ab3f7bb5d2f5104e047079bfdfda&#039; is not found. */⏎      
```

### DIFF
--- a/modules/system/classes/SystemController.php
+++ b/modules/system/classes/SystemController.php
@@ -4,6 +4,7 @@ use Lang;
 use ApplicationException;
 use Illuminate\Routing\Controller as ControllerBase;
 use Exception;
+use Response;
 
 /**
  * The is the master controller for system related routing.
@@ -36,7 +37,7 @@ class SystemController extends ControllerBase
 
         }
         catch (Exception $ex) {
-            return '/* '.e($ex->getMessage()).' */';
+            return Response::make('/* '.e($ex->getMessage()).' */', 404);
         }
     }
 }
--- a/modules/system/classes/SystemController.php
+++ b/modules/system/classes/SystemController.php
@@ -37,7 +37,7 @@ class SystemController extends ControllerBase
 
         }
         catch (Exception $ex) {
-            return Response::make('/* '.e($ex->getMessage()).' */', 404);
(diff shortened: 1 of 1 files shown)


LukeTowers:

LukeTowers:
@alekseyp shouldn't it return a 500 error, not a 404?

alekseyp:
@LukeTowers thanks for the branch fix.

IMO it should return 404, because 500 is expected to be a (temporary) error. Since it's not an error, but an old asset, I think 404 is more appropriate.

It will also help CDNs no to cache all those files that return 200.

I'm open to suggestions tho.

LukeTowers:
@alekseyp Exceptions can be thrown by a lot of other things than just a missing file, so I think this should be a 500 error.

w20k:
@alekseyp I agree with @LukeTowers there might be more than one type of exception.
But, I mean, you could catch `ApplicationException` when cache not found :) Might be something that you're looking for.

LukeTowers:
I still don't think that's specific enough. If the file doesn't exist, the developer has an error in how they've setup their code, so I would still say that it's a 500 error.

alekseyp:
Sure, feel free to update it the way you feel it suites octobercms the best.
In any case it shouldn't be 200.

LukeTowers:
@alekseyp submit the change to 500 and I'll merge it

alekseyp:
@LukeTowers done

PR #4145, opened 2019-02-16
Apply Model Filters Before Data Saved

### LINKED ISSUE #4144: Hidden Fields Not Saved
- OctoberCMS Build: 446
- PHP Version: 7.2
- Database Engine: mysql
- Plugins Installed: Builder

### Description:
Hidden fields are not saved, https://octobercms.com/forum/post/usage-of-hidden-form-field

### Steps To Reproduce:
- Add `hidden: true` to any form field in yaml configuration file.
- Then add:
```
my_field:
  hidden: true
  dependsOn:
    - some_other_field
```
- Add `filterFields` method to model class:
```
public function filterFields($fields) {
  $fields->my_field->hidden = false;
}
```
- `my_field` value will never **saved**!

MWalid:
Fixes #4144 
- and https://octobercms.com/forum/post/usage-of-hidden-form-field?page=1#post-26291

### TL;DR
Cannot save value of hidden field even after `$field->hidden = false` is called in `Model::filterFields`

### Long Story
- Add `hidden: true` to any form field in yaml configuration file.
- Then add:
```
my_field:
  hidden: true
  dependsOn:
    - some_other_field
```
- Add `filterFields` method to model class:
```
public function filterFields($fields) {
  $fields->my_field->hidden = false;
}
```
- `my_field` value will never be **saved**!

### DIFF
--- a/modules/backend/behaviors/FormController.php
+++ b/modules/backend/behaviors/FormController.php
@@ -232,6 +232,8 @@ class FormController extends ControllerBehavior
 
         $this->initForm($model);
 
+        $model->forceFill($this->formWidget->getSaveData());
+
         $this->controller->formBeforeSave($model);
         $this->controller->formBeforeCreate($model);
 
--- a/modules/backend/widgets/Form.php
+++ b/modules/backend/widgets/Form.php
@@ -240,7 +240,7 @@ class Form extends WidgetBase
     public function renderField($field, $options = [])
     {
         $this->prepareVars();
-        
+
         if (is_string($field)) {
             if (!isset($this->allFields[$field])) {
                 throw new ApplicationException(Lang::get(
@@ -1121,6 +1121,7 @@ class Form extends WidgetBase
     public function getSaveData()
     {
         $this->defineFormFields();
+        $this->applyFiltersFromModel();
 
         $result = [];
 


MWalid:
How long usually it takes to review and possibly merge the PR?

w20k:
@MWalid depends on PR and users activity. Also the availability of members.

LukeTowers:
@MWalid please add some detailed usage examples of why this is needed

MWalid:
@LukeTowers actually this is helpful in case you need to show/hide some fields based on other field's value, in my case it was the following situation:
if `car_count` is greater than `1` the user must enter `sub_packages` numbers, the `sub_packages` field is hidden by default, however in `filterFields` methods I'm showing it again - it shows on the user interface and the user enters `sub_packages` value, then they click on save,  `sub_packages`  value is not saved.

w20k:
@MWalid why not use a trigger? (https://octobercms.com/docs/backend/forms#field-trigger-events)

MWalid:
@w20k Triggers work only with checkboxes, at least this is what the documentation says.

w20k:
@MWalid why do think it only work with checkboxes? Value should be "true", but it supports three diferent types: **checked, unchecked, value[somevalue]**.

>`condition | determines the condition the specified field should satisfy for the condition to be considered "true". Supported values: checked, unchecked, value[somevalue].`

Here is a small example from the test-plugin:
```
tags:
    label: Tags by Relation
    type: taglist
    mode: relation
    nameFrom: name
    tab: Tags
    placeholder: Test

_tags_trigger:
    commentAbove: Hide if "a" or "b" are used as tags above
    tab: Tags
    trigger:
        action: hide
        field: tags[]
        condition: value[a][b]
```

MWalid:
Thanks for pointing that out,

I have seen this before, however in my case the condition was `value > 1` is it possible to achieve that with triggers?

w20k:
@MWalid sorry mate don't remember, but most probably - not ;)
Though I do think to add this feature to the triggers might be really useful!

https://github.com/octobercms/october/blob/master/modules/system/assets/ui/js/input.trigger.js#L70

MWalid:
@w20k true that would be useful, also showing hidden fields as stated here would be very useful, many people reported the issue and done hours of search like me:)

w20k:
@MWalid lets here what @LukeTowers thinks about that! And if he is good to go with this fix, will test it.

MWalid:
Looks good thanks a lot,

Regards,

Walid Ammar,

modules/backend/behaviors/FormController.php
```diff
@@ -232,6 +232,8 @@ public function create_onSave($context = null)
 
         $this->initForm($model);
 
+        $model->forceFill($this->formWidget->getSaveData());
```
LukeTowers: Why is this required?

MWalid: @LukeTowers  In case of creating new record, the model is initiated by [`$model = $this->controller->formCreateModelObject()`](https://github.com/octobercms/october/blob/master/modules/backend/behaviors/FormController.php#L230), at this point the model attributes are empty and when calling `Model::filterFields($attributes)` all attribute values are null. 

Calling `$model->forceFill($this->formWidget->getSaveData())` would fill the model with current form data and `$attributes` would have the actual values to be used inside `Model::filterFields`.

MWalid: @LukeTowers any chance we can get this reviewed soon?

LukeTowers: @MWalid I'm pretty busy right now, perhaps @w20k could take a look at it or you could offer a bounty on it to bump it up the priority list. Is forceFill how the model is populated in the update method?

daftspunk:

MWalid:
Since I did not get a reply on the issue thread #4144 I'm reposting the test plugin link here.

Test plugin is here: https://github.com/MWalid/test-plugin

1- Go to playground > People.

2- Open any record.

3- Change the name to walid

4- Blur name field, preferred name field should appear.

5- Enter any value.

6- Save & refresh the page.

7- Preferred name value is not saved.

Screen record: http://g.recordit.co/yjnxesDujX.gif

modules/backend/behaviors/FormController.php
bennothommo: @MWalid This is a potential security issue allowing `forceFill` here, as that bypasses mass-assignment rules that may be implemented in the model. I don't think we can accept this PR with this in place.

bennothommo:

iboved:
I have the same problem.  I have a hidden field:
```
title:
    label: Title
    type: text
    hidden: true
```
Data from hidden field is excluded from the saved data. It is clear. Depending on some condition, I show this field in **filterFields** method:
```
public function filterFields($fields, $context = null)
{
    $fields->title->hidden = false;
}
```
**BUT** data is still excluded and I can not save the new value.

mjauvin:
Guys, what is the real purpose of having a form field with hidden attribute?

Would it make sense to add a type "hidden" which would map to <input type="hidden"> and allow saving its value?

bennothommo:
@mjauvin I think there's enough events to allow someone to apply data to a form or a model programatically. I don't see any real need for a hidden type - TBH, I don't really see a need for a hidden attribute for fields either.

PR #4148, opened 2019-02-19
Add backend.list.extendRecords event for export

fansaien:
Provides an opportunity to modify and / or return the $results collection object before the controller export it.

The same as Lists widget getRecords function:
https://github.com/octobercms/october/blob/9581b23d1e85ccfdc283946894afa2df177b304f/modules/backend/widgets/Lists.php#L614


w20k:
@fansaien could you add a documentation for this event?

LukeTowers:
@w20k event is already documented in the Lists widget

Issue #1840, opened 2016-03-11
How to disable assets combining on development in OctoberCMS


rsobon:
Is there possibility to disable asset combining in development environment and enable in production without modifying the twig layout file?

Note: This has nothing to do with asset minification and `enableAssetMinify` flag in the `app/config/cms.php`

For example if there is debug set to true in config/app.php: `'debug' => true`, the twig array in the `<script>` tag should output scripts separately.

Twig array:

```
<script src="{{ [
    'assets/js/one.js',
    'assets/js/two.js',
    '@framework',
    '@framework.extras'
]|theme }}"></script>
{% scripts %}
```

Rendered result should be:
```
<script src="http://localhost/themes/demo/assets/js/one.js"></script>
<script src="http://localhost/themes/demo/assets/js/two.js"></script>
<script src="/modules/system/assets/js/framework.js"></script>
<script src="/modules/system/assets/js/framework.extras.js"></script>
<script src="/modules/system/assets/js/framework.extras.css"></script>
```

And if there is debug set to false in config/app.php: `'debug' => false`

Rendered result should be something like:
```
<script src="http://localhost/combine/2cdc17b704821cd2ffbd9be8e4d340f9-1457016128"></script>
```

Detailed issue on stackoverfow:
https://stackoverflow.com/questions/35892560/how-to-disable-assets-combining-on-development-in-octobercms

rsobon:
OK I managed to work this out but not with the `config/app.php` mentioned in the question.

Other solution requires to create `.env` file in the root directory of OctoberCMS. This file is by default in the `.gitignore` so you can have different .env files on production and development ([Piece of documentation][1])

Contents of this file should be:

    APP_ENV=dev

Then you can access variable in the twig:

    {% if this.environment == 'dev' %}
        <script src="{{ 'assets/vendor/jquery.min.js'|theme }}"></script>
        <script src="{{ 'assets/semantic/dist/semantic.min.js'|theme }}"></script>
        <script src="{{ 'assets/javascript/app.js' | theme }}"></script>
        {% framework extras %}
    {% else %}
        <script src="{{ [
         'assets/vendor/jquery.js',
         'assets/semantic/dist/semantic.js',
         'assets/javascript/app.js',
         '@framework',
         '@framework.extras'
         ]|theme }}"></script>
    {% endif %}


  [1]: http://octobercms.com/docs/setup/configuration#environment-config

daftspunk:
[...]

PR #4150, opened 2019-02-20
New Twig filters for assets

### LINKED ISSUE #289: Turn off asset combing on development level
Is it possible turn off asset combing while I'm developing

GinoPane:
Hi guys!
I want to suggest a couple of new Twig filters to resolve octobercms/october#289, octobercms/october#1840 (and maybe some other previously opened issues with the same reason).

The usage example:

```
{{ [
	'assets/js/jquery.js',
	'assets/js/popper.min.js',
	'assets/js/bootstrap.js'
] | themeJs }}
```

With attributes:

```
{{ [
	'assets/js/jquery.js',
	'assets/js/popper.min.js',
	'assets/js/bootstrap.js'
] | themeJs('text/javascript', ["foo='bar'"]) }}
```

The same is for CSS with `themeCss` filter.

Depending on a new config setting `disableAssetCombine` (`null` by default) the filters will either return a single html node with a combined asset or multiple script or link nodes with single asset per node.

Please let me know what you think.

### DIFF
--- a/config/cms.php
+++ b/config/cms.php
@@ -199,6 +199,20 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Determines if the asset combiner is disabled.
+    |--------------------------------------------------------------------------
+    |
+    | If the asset combining is disabled, themeJs, themeCss Twig filters won't
+    | combine arrays of assets into one single file and will leave them as
+    | they are and include assets per file individually. This setting
+    | won't make any effect to the basic Twig theme filter
+    |
+    */
+
+    'disableAssetCombine' => true,
+
+    /*
+    |--------------------------------------------------------------------------
     | Check import timestamps when combining assets
     |--------------------------------------------------------------------------
     |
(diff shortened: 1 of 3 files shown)


GinoPane:
I hope this won't get lost 😄

w20k:
Hopefully, not :D @GinoPane will this also work with minification?

GinoPane:
It uses system `combine` call under the hood, so the behaviour is the same as for simple `| theme` filter except the difference that was noted in the linked issues.

So I believe, it should.

w20k:
@GinoPane docs ;) Would be nice to have a test or few for your new functionality! And it's good to be tested 😄

LukeTowers:
@GinoPane is this still something that you're interested in?

GinoPane:
Hmm, as I see, the issues referenced here did not get any better solution, so yes.

LukeTowers:
@GinoPane have you seen the `decompileBackendAssets` feature added by @bennothommo? I feel like that's similar to what you're trying to achieve here. Perhaps you could take another stab at this using that as a reference?

GinoPane:
No, I haven't. Thanks for reference! Will check.

daftspunk:
Hi @GinoPane! This is a nice idea, however, we need to limit the amount of code that we add to the core. This issue could be solved with a partial

```
{% set assets = [
	'assets/js/jquery.js',
	'assets/js/popper.min.js',
	'assets/js/bootstrap.js'
] %}

{% partial 'render-assets' decompile=this.environment == 'local' %}
```

---

render-assets.htm

```
{% if decompile %}
    {% for asset in assets %}
        <script src="{{ asset }}"></script>
    {% endfor %}
{% else %}
    <script src="{{ assets|theme }}"></script>
{% endif %}
```

Hope this helps

PR #4152, opened 2019-02-21
fix: File name contains non-latin characters

gitlog:
Check for unsafe file extensions and name  throw exception https://github.com/octobercms/october/blob/develop/modules/system/classes/MediaLibrary.php#L492


w20k:
@gitlog could you example your `fix` a bit? I don't feel like it's a good idea, to invalidate every file name with non ascii characters in it.

LukeTowers:

PR #4154, opened 2019-02-21
Fix issue with list filters when more filters appear on page.

### LINKED ISSUE #4153: When having two lists with filters, the second one does not work
- OctoberCMS Build: b39eb03336faf4c30ddb059e1973eb313dbd68f6 (dev-develop)
- PHP Version: 7.2
- Database Engine: mysql
- Plugins Installed: None

### Description:
If you have two lists on the same page (say in two different tabs or one in modal and one in page) and both have filters, the second one has errors when applying the filters:
```
storm-min.js?v446:3205 Uncaught TypeError: Cannot read property 'data' of null
    at FilterWidget.filterScope (storm-min.js?v446:3205)
    at HTMLButtonElement.<anonymous> (storm-min.js?v446:3113)
    at HTMLDocument.dispatch (jquery.min.js?v446:2)
    at HTMLDocument.y.handle (jquery.min.js?v446:2)
```

### Steps To Reproduce:
1. Install a fresh october installation
2. Download october test plugin ( https://github.com/octoberrain/test-plugin )
3. To the post and comments tables add one filter for each. (https://github.com/filipac/oc-filters-reproduce-bug-plugin/commit/db81810f60883356f2a55a9d6eb578f6f214791c)
4. Try to apply filter for first table (posts tab) works ✅
5. Try to apply or reset fitlers for second (comments) ‼️ fail with the above error

The error occurs only when having more lists on page and is triggered because ```self.$activeScope``` and ```self.activeScopeName``` seems to be null.

I will make a new PR to check the solution I used on my project to make this work, and you can review it.

filipac:
Fixes #4153

I tried my best to describe the issue in #4153 but basically when having two lists per page, or even another one in a modal, filters of all but first one do not work. This is because the scope of the popover is somehow lost. It is better to listen for events only after popover is show and only on that target (event.relatedTarget). 

What I did to ensure scope is not lose was to move event listeners to `show.oc.popover` callback and bind the listeners to event.relatedTarget returned from the event.

What is the procedure of minifying the Storm library? Do I have to submit storm.min.js too in this PR (so I need to run october:util compile js?)? Or you will do this.

### DIFF
--- a/modules/system/assets/ui/js/filter.dates.js
+++ b/modules/system/assets/ui/js/filter.dates.js
@@ -40,8 +40,22 @@
     FilterWidget.prototype.initFilterDate = function () {
         var self = this
 
-        this.$el.on('show.oc.popover', 'a.filter-scope-date', function () {
+        this.$el.on('show.oc.popover', 'a.filter-scope-date', function (event) {
             self.initDatePickers($(this).hasClass('range'))
+
+            $(event.relatedTarget).on('click', '#controlFilterPopoverDate [data-trigger="filter"]', function (e) {
+                e.preventDefault()
+                e.stopPropagation()
+
+                self.filterByDate()
+            })
+
+            $(event.relatedTarget).on('click', '#controlFilterPopoverDate [data-trigger="clear"]', function (e) {
+                e.preventDefault()
+                e.stopPropagation()
+
+                self.filterByDate(true)
+            })
         })
 
         this.$el.on('hiding.oc.popover', 'a.filter-scope-date', function () {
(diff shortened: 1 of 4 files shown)


w20k:
@filipac could you re-compile assets: `php artisan october:util compile js`

filipac:
@w20k done

LukeTowers:
@linhtc does this PR solve your problem?

alxy:
I had the same issue with opening a modal twice with filters applied to it - can confirm that this solves the issue.

filipac:
@alxy of course it does, haha, I already use the fix locally until merged.

alxy:
I also tested the setup with multiple lists on one page - this should be mergable :)

PR #4159, opened 2019-02-22
Overriding env DB parameters on plugin tests

czerwonyd:
Overriding env DB parameters in `phpunit.xml` when we do not want to use SQLite stored in memory.

For example developers can use this feature when they use raw SQL statements that works for their engine and SQLite does not support them.


LukeTowers:
@czerwonyd won't this cause the values to be pulled from `.env` if the PHPUnit environment variables aren't setup?

czerwonyd:
@LukeTowers yes it will, you are right. That's probably why it's hardcoded to use SQLite stored in memory. This is the default behaviour of Laravel (described [here](https://laravel.com/docs/5.5/testing)) as well.

On the other hand in OctoberCMS the default behavior should be using SQLite in memory. Right now there are two ways of overriding the default testing DB params. I moved it to config files, but without `config/testing/database.php` params again were taken from `config/database.php`. As a consequence I decided to add another environmental variable `DB_OVERRIDE_DEFAULT`. This has to be set to true in order to override the default behavior of using SQLine in memory. It's a prevention from using `dev` DB in tests by default.

LukeTowers:
@czerwonyd instead of making it an environment variable by default, could you make it a configuration item in the database.php, something like `useConfigForTesting` with a thorough comment explaining it's purpose (defaulting to false of course, and also adding that item to the `october:env` command)

czerwonyd:
@LukeTowers, good idea. Done, check it now.

PR #4160, opened 2019-02-24
Use Form getSaveData method when saving form widgets within repeaters

bennothommo:
Related comment: https://github.com/octobercms/october/pull/4116#issuecomment-464469659

Uses the `Backend/Widgets/Form::getSaveData()` method to retrieve the value of an internal form widget within an item instance of a repeater. This improve the saving of widgets within repeaters.

~~Also includes a fix to return `null` if datatable widgets have no rows, to make it absent from the form data.~~ - No longer needed


LukeTowers:
@bennothommo Would that `null` behaviour prevent a datatable from being able to be emptied by the user and having that empty state persisted through the save data?

modules/backend/formwidgets/Repeater.php
```diff
@@ -202,16 +202,16 @@ protected function processSaveValue($value)
...
          */
-        foreach ($this->formWidgets as $field => $form) {
-            foreach ($form->getFormWidgets() as $formField => $widget) {
-                $parts = HtmlHelper::nameToArray($field . '[' . $formField . ']');
-
-                $widgetValue = $widget->getSaveValue($this->dataArrayGet($value, $parts));
-                if (empty($widgetValue) || !count($widgetValue)) {
-                    continue;
-                }
-                $this->dataArraySet($value, $parts, $widgetValue);
```
LukeTowers: Should these helper functions be removed now too?

bennothommo:
@LukeTowers I'm going to re-do this fix. Looking back at it, it is a bit hacky.

modules/backend/formwidgets/Repeater.php
bennothommo: Yep, removed :)

bennothommo:
@LukeTowers This is ready for review. I've simplified the fix significantly and it seems to have stopped the phantom repeater items.

w20k:
@bennothommo I like your changes. Removed half of the code and added one additional parameter 😄 That's what you call a clean code)

bennothommo:
@w20k I surprised even myself with how simple the fix *should* have been.

LukeTowers:
Waiting on testing, looks good @bennothommo!

Rike-cz:
I have problem with this line in version 448: https://github.com/octobercms/october/blob/dd53206a82ed59cd483aca5b5968b3993413b7a5/modules/backend/formwidgets/Repeater.php#L210
WidgetValue may not be countable, right? It fails with simple string from mediafinder.

bennothommo:
@Rike-cz This pull request should resolve that.
See: https://github.com/octobercms/october/pull/4160/files#diff-a8e657fa9654328055121a4e2cc4311bL210

Rike-cz:
@bennothommo Thanks.

yweyte:
@bennothommo Thank you very much! This fix the bug.

LukeTowers:
@bennothommo @w20k what's the status of this?

w20k:
@LukeTowers it's better to ask @bennothommo. He asked to wait for a little bit: https://github.com/octobercms/october/issues/4178#issuecomment-470979298.

bennothommo:
@LukeTowers a fair few people have reported that this PR fixed their errors. I haven't heard back from the person who reported data loss, but I believe their issue was fixed by the last commit. @seanthepottingshed's issue is not related to this as far as I can tell, so I believe this is ready to merge.

Balex70:
@LukeTowers we have faced same issue on few projects, and this PR fixed it!

AugmentBLU:
@bennothommo Sorry for the delay - the PR is working however I am having to manually change the file as composer is not downloading the right branch. I have my composer set to `dev-develop` but still downloads 448. 

@daftspunk is there a reason why I cannot download the latest dev branch both locally and on a staging server with DO? I always use the following in the main composer file and it has worked for a while but it's not pulling the latest dev branch so Repeaters are still broken.

```
        "october/rain": "dev-develop as 1.0",
        "october/system": "dev-develop",
        "october/backend": "dev-develop",
        "october/cms": "dev-develop",
```

LukeTowers:
@AugmentBLU it's because the marketplace server was recently moved to AWS and it was responsible for automatically running the script that splits to the composer packages: https://github.com/octobercms/library/blob/master/build/rain-split-full.sh. I've just manually run it so the packages should be up to date now. @daftspunk is looking into a new setup for it to be run automatically.

bennothommo:
@LukeTowers @daftspunk Might be worth looking into [GitHub Actions](https://developer.github.com/actions/) when it comes out of beta. This allows you to run commands inside a Docker instance on certain Git actions, ie. merges, tags, etc. We use a similar system in Gitlab internally and it saves a metric tonne of time.

LukeTowers:
Thanks for the pointer @bennothommo! Could you ping me again when it's out?

bennothommo:
Will do.

Issue #4168, opened 2019-02-27
Bug: Popup loading indicator not hidden after ajax done


drmzio:
![](https://cl.ly/ed926cc58cf8/Image%2525202019-02-27%252520at%2525204.36.18%252520PM.png)

- OctoberCMS Build: 447

### Description:
The loading indicator still shows in the background after completed ajax requests. Found out about this bug while creating my own popups. Haven't been able to find a fix to submit PR, maybe someone can help.

Edit: Found a fix for it #4169 

### Steps To Reproduce:
Just click __Check for updates__ within the backend updates page.

w20k:
@drmzio never seen this bug before :)

w20k:
@drmzio all completed issues will be automatically closed merge to master ;)

drmzio:
> @drmzio all completed issues will be automatically closed merge to master ;)

@drmzio Oh sorry! My mistake 😁

PR #4169, opened 2019-02-28
Fix popup loading indicator

drmzio:
Fixes issue #4168

### DIFF
--- a/modules/system/assets/ui/js/popup.js
+++ b/modules/system/assets/ui/js/popup.js
@@ -254,7 +254,7 @@
             setTimeout(function(){ self.$backdrop.addClass('loading'); }, 100)
         }
         else {
-            this.$backdrop.removeClass('loading');
+            setTimeout(function(){ self.$backdrop.removeClass('loading'); }, 100)
         }
     }
 
--- a/modules/system/assets/ui/storm-min.js
+++ b/modules/system/assets/ui/storm-min.js
@@ -3807,7 +3807,7 @@ this.$backdrop=null}}
 Popup.prototype.setLoading=function(val){if(!this.$backdrop)
 return;var self=this
 if(val){setTimeout(function(){self.$backdrop.addClass('loading');},100)}
-else{this.$backdrop.removeClass('loading');}}
+else{setTimeout(function(){self.$backdrop.removeClass('loading');},100)}}
 Popup.prototype.setShake=function(){var self=this
 this.$content.addClass('popup-shaking')
 setTimeout(function(){self.$content.removeClass('popup-shaking')},1000)}


w20k:
@drmzio could you provide a simple example to reproduce your issue?

LukeTowers:
JS will need to be compiled too I believe. Also @drmzio could you make the zindex of the loader indicator higher than the popup modal itself? It would make the issue more apparent if it happens again, as well as solve the issue of opening nested popups not showing the loading indicator beyond the first popup.

drmzio:
@LukeTowers Wouldn't compiling JS have marge conflicts with other PRs?

@w20k Here's the steps to reproduce:

1. Install OctoberCMS
2. Add this code to the home page __markup__ section:
```twig
{% put styles %}
    <link rel="stylesheet" href="{{ 'modules/system/assets/ui/storm.css'|app }}">
{% endput %}
{% put scripts %}
    <script src="{{ 'modules/system/assets/ui/storm-min.js'|app }}"></script>
    <script>
        var newPopup = $('<a />');
        newPopup.popup({ handler: 'onLoadContent' });
    </script>
{% endput %}
```
3. Add AJAX handler function to home page __code__ section:
```php
function onLoadContent()
{
    return 'Loaded';
}
```

Video demo: https://cl.ly/5b456d5319b4

LukeTowers:
@drmzio nope, whoever gets in first makes the rest of the PRs that would affect the same file have to rebase. Rarely happens though that we've got multiple active PRs touching the same file.

drmzio:
@LukeTowers Ok I added the compiled JS

w20k:
> **@LukeTowers commet** 
>@drmzio could you make the zindex of the loader indicator higher than the popup modal itself

@drmzio could you push changes for z-index and should be read!
Status - ✅

drmzio:
@w20k Hi, I don't think that is possible because the loading indicator is inside the modal backdrop. (See screenshot)

Test demo: https://octobercms.site000.com/

LukeTowers:
@drmzio if the loading indicator switched to a different position stack (i.e. fixed) than the modal backdrop wouldn't that work? Or would it cause other issues?

LukeTowers:
Actually, maybe it's the modal backdrop that needs to be bumped up when a nested popup gets opened?

drmzio:
Hi @LukeTowers this would require changing the default behavior of the popup library and dynamically updating the backdrop z-index on nested modal windows. For now, I think this should be fine. The only issue I was having was with the loading indicator still showing after AJAX completes.

LukeTowers:
Fair enough. @w20k can you test that this resolves the original issue?

techstationpro:
Bro please give index.html....

LukeTowers:
@techstationpro sounds like a spammer to me

w20k:
@LukeTowers couldn't reproduce the described issue, even with the provided steps, but I guess, fix should be an issue for other users or use cases.

PR #4171, opened 2019-02-28
Use 1px for 'min-size' layout rows to fix Firefox spacing issue

### LINKED ISSUE #4114: Minor display issue on the deleted user's page
- OctoberCMS Build: 447

### Description:
The bottom margin of the red alert box is not constant.

### Steps To Reproduce:
- Delete an administrator
- Go to administrator's page
- Navigate between tabs

bennothommo:
Fixes #4114.

With an element that is set to `display: table-row` and a height of `0` or `auto`, Firefox seems to allow it and other elements in the "table" to resize based on space used. This may cause some elements to take up more space than what they require. Setting the height of these elements to `1px` forces the element to only take up the space it needs.


w20k:
@bennothommo could you also provide before/after screenshots.
It's only Firefox related bug?

bennothommo:
@w20k As far as I can tell, it's only Firefox. Chrome works fine. I can't test IE/Edge as I use Linux.

Before fix:
screenshot from 2019-02-28 21-22-22
After fix:
screenshot from 2019-02-28 21-23-06

bennothommo:
@w20k Turns out I still have access to BrowserStack. Edge doesn't exhibit the issue either, so it's only Firefox. All 3 browsers work fine after the fix.

w20k:
Thanks for digging @bennothommo. Could you re-compile assets, and it's ready to be merged.

bennothommo:
@w20k Done. :)

w20k:
@bennothommo I think you use the non-dev version of the library. Compiled assets look odd.

bennothommo:
@w20k I compiled it using `october:util compile less`. Is that the correct command?

w20k:
@bennothommo command is correct, but my fix is still in the dev branch (https://github.com/octobercms/library/commit/1ff7f177922e112b08f6cc9e4e6a8f7a0d75b584). Could you use `edgeUpdates` or `composer-dev` and re-compile assets once more? 
Thanks heaps! 😄 I love this phrase now))

bennothommo:
@w20k Sorry, my `october/rain` library was out-of-date. Recompiled the CSS :)

w20k:
@bennothommo appreciated :) 

@LukeTowers  - done!

PR #4172, opened 2019-02-28
New fill trigger action

JLyons009:
Added a 'fill' trigger action that functionally is the opposite of 'empty.'  Primarily for a 'Select All' use case in a backend form.


LukeTowers:
Don't commit all of the language files. Only use `october:util compile js`. PR is unusable because of the other changes.

PR #4193, opened 2019-03-13
Reuse the model instance from Form widget

alxy:
This fixes https://github.com/rainlab/translate-plugin/issues/209

This is a potentially breaking change. In my tests, everything went smooth, however I am almost certain that it affects a lot of cases that I have not thought of.

### DIFF
--- a/modules/backend/behaviors/RelationController.php
+++ b/modules/backend/behaviors/RelationController.php
@@ -1062,7 +1062,7 @@ class RelationController extends ControllerBehavior
         $saveData = $this->manageWidget->getSaveData();
 
         if ($this->viewMode == 'multi') {
-            $model = $this->relationModel->find($this->manageId);
+            $model = $this->manageWidget->model;
             $modelsToSave = $this->prepareModelsToSave($model, $saveData);
             foreach ($modelsToSave as $modelToSave) {
                 $modelToSave->save(null, $this->manageWidget->getSessionKey());


LukeTowers:
I don't see how this would cause issues, but here's the history of that line:
https://github.com/octobercms/october/commit/a20712b73cd14a562fcc6ad949cabdff02c3a05e
https://github.com/octobercms/october/commit/04462f5b4c4258ba8afdf13da4d50f0736158b03#diff-63fc4fc41aa256f8d5bcfe7a37245ecfR470
https://github.com/octobercms/october/commit/001a9f92ffe1c34161863478160ffb499ccefc10#diff-63fc4fc41aa256f8d5bcfe7a37245ecfR465
https://github.com/octobercms/october/commit/896190645982789f3db68c416eb0507bd329d2f7#diff-63fc4fc41aa256f8d5bcfe7a37245ecfR426
https://github.com/octobercms/october/commit/dbc4043a90238db18d65aca1c7fd17dd79c314af#diff-63fc4fc41aa256f8d5bcfe7a37245ecfR443

Didn't notice anything that explained the rationale for the original approach.

mjauvin:
So far so good, have been playing with plugins using many relations and didn't experience a glitch yet.
Nice work!

mjauvin:
@daftspunk care to give your opinion on this change?

daftspunk:
TL;DR This should be merged as is

I've audited this change to the best of my abilities and can confirm it does not appear to break anything. The logic path shows manage widget is always a Form widget with the available model

If for whatever reason it becomes a problem in future, consider replacing the expression with something like:

```
$model = $this->manageWidget->model->exists
    ? $this->manageWidget->model
    : $this->relationModel->find($this->manageId);
```

I doubt that we'll ever need to...

Thanks!

PR #4200, opened 2019-03-18
Make default format options configurable

alxy:
If the default format mode is selected, there is no chance to configure the format options for delimiter, enclosure, escape and encoding. This considers a new config file item `formatOptions` and falls back to `null` if it is not specified.

I will also provide a PR to the docs.

### DIFF
--- a/modules/backend/behaviors/ImportExportController.php
+++ b/modules/backend/behaviors/ImportExportController.php
@@ -597,9 +597,9 @@ class ImportExportController extends ControllerBehavior
          */
         $defaultOptions = [
             'fileName' => $this->exportFileName,
-            'delimiter' => $this->getConfig('formatOptions[delimiter]', ','),
-            'enclosure' => $this->getConfig('formatOptions[enclosure]', '"'),
-            'escape' => $this->getConfig('formatOptions[escape]', '\\'),
+            'delimiter' => $this->getConfig('defaultFormatOptions[delimiter]', ','),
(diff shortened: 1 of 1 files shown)


alxy:
Documentation PR: https://github.com/octobercms/docs/pull/363

LukeTowers:
Can you rename this to `defaultFormatOptions` @alxy?

alxy:
Sure, however I prefer shorter configuration key names. It should be clear that in many cases this configuration is not needed (where the standard values just work) and can thus be completely omitted from the config file.
Can we hear a second opinion on the naming?

LukeTowers:
Shorter is usually better yes, but in my opinion the `default` part of this is necessary. `formatOptions` implies that you are providing the available format options, and it's not like the extra 7 characters turn it into a novel 😉

LukeTowers:
Feel free to ping @daftspunk on this if you really want a second opinion, although the likelihood of him responding is low.

alxy:
Okay, anyhow, will add the default then...

alxy:
@LukeTowers Done.

LukeTowers:
Great! @w20k could you test this please?

w20k:
@LukeTowers will try to deal with testing on this week. I'm a bit out of sync with testing PRs right now, trying to move on to my own company... takes time and effort :)

w20k:
@LukeTowers tested - done! Works great!

Issue #4201, opened 2019-03-18
Exception when trying to access non-existent backend URL when CMS module is disabled


LarBearrr:
- OctoberCMS Build: 447 
- PHP Version: 7.3
- Database Engine: MySQL
- Plugins Installed: RainLab.SiteMap-dev-master
RainLab.Builder-dev-master
Bedard.Debugbar-v2.0.0

### Description:
When trying to access a non-existent backend controller, an exception is thrown if the CMS module is disabled and there is no active theme set or installed:

![image](https://user-images.githubusercontent.com/28734844/54563850-5582c180-4998-11e9-80b2-40d651db7694.png)

### Steps To Reproduce:
1. Remove the active theme (demo theme by default)
2. Update `config/cms.php` to set `backendUri` to null (L39) and remove `'CMS'` from the `loadModules` property (L105).
3. Try to access a non-existent backend URL `/backend/author/plugin/foobar`

LarBearrr:
Per @alxy suggestion, it seems that "this might be a better check": (https://github.com/octobercms/october/blob/26173486d39cdf4312535ecf247e3d7a161713b4/modules/system/reportwidgets/Status.php#L98)[https://github.com/octobercms/october/blob/26173486d39cdf4312535ecf247e3d7a161713b4/modules/system/reportwidgets/Status.php#L98]

Indeed, changing (https://github.com/octobercms/october/blob/3118660d834f161d513da08477be92281a2eb96a/modules/backend/classes/BackendController.php#L71)[https://github.com/octobercms/october/blob/3118660d834f161d513da08477be92281a2eb96a/modules/backend/classes/BackendController.php#L71] to `if (in_array('Cms', Config::get('cms.loadModules', []))) {` will produce the desired result:

![image](https://user-images.githubusercontent.com/28734844/54564264-5405c900-4999-11e9-8e23-d72472a3cf4f.png)

**NOTE** I also removed the `activeTheme` from `config/cms.php`

LarBearrr:
See #4202

PR #4202, opened 2019-03-18
Display backend 404 for non-existent backend urls if CMS module is not loaded

LarBearrr:
There are projects that sometimes only require the backend feature's to be used. This can be achieved by [disabling the active theme](https://github.com/octobercms/october/blob/3118660d834f161d513da08477be92281a2eb96a/config/cms.php#L14), [setting the backend URI to null](https://github.com/octobercms/october/blob/3118660d834f161d513da08477be92281a2eb96a/config/cms.php#L39), and [removing the CMS module from the active modules](https://github.com/octobercms/october/blob/3118660d834f161d513da08477be92281a2eb96a/config/cms.php#L105) in [config/cms.php](https://github.com/octobercms/october/blob/master/config/cms.php).

However, the current check for routing backend requests [only checks if the CMS Controller exists](https://github.com/octobercms/october/blob/3118660d834f161d513da08477be92281a2eb96a/modules/backend/classes/BackendController.php#L71). This will result in an exception being thrown if the above steps have been implemented because the CMS Controller technically exists, but there is no active theme to make the view from.

@alxy Found [the solution to check the `loadModules` array for `Cms`](https://github.com/octobercms/october/blob/26173486d39cdf4312535ecf247e3d7a161713b4/modules/system/reportwidgets/Status.php#L98)%5Bhttps://github.com/octobercms/october/blob/26173486d39cdf4312535ecf247e3d7a161713b4/modules/system/reportwidgets/Status.php#L98%5D) which will produce the backend 404 page as desired:

This PR addresses #4201.

### DIFF
--- a/modules/backend/classes/BackendController.php
+++ b/modules/backend/classes/BackendController.php
@@ -96,7 +96,7 @@ class BackendController extends ControllerBase
      */
     protected function passToCmsController($url)
     {
-        if (class_exists('\Cms\Classes\Controller')) {
+        if (in_array('Cms', Config::get('cms.loadModules', [])))  {
             $this->cmsHandling = true;
             return App::make('Cms\Classes\Controller')->run($url);
         } else {
--- a/modules/backend/classes/BackendController.php
+++ b/modules/backend/classes/BackendController.php
@@ -96,7 +96,10 @@ class BackendController extends ControllerBase
      */
     protected function passToCmsController($url)
     {
-        if (in_array('Cms', Config::get('cms.loadModules', [])))  {
+        if (
+            in_array('Cms', Config::get('cms.loadModules', [])) &&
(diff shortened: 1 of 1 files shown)


LukeTowers:
Do both please with an `||`

GinoPane:
> Do both please with an `||`

Sounds like it should be "&&" if you need both. The original condition was `true` when it ended up with an exception, so you'd get the same exception with "or".

LarBearrr:
> Do both please with an `||`

@GinoPane is correct, using `||` throws the same exception as before. `&&` renders the backend 404.

LukeTowers:
Whoops, late night sleep deprived brain talking, glad you two are on the ball :)

LarBearrr:
Closes #4202

PR #4207, opened 2019-03-25
Allow List Widget configs to add CSS classes to head column cell

bennothommo:
Allows for the definition of custom CSS classes for the header cell of a column in a list widget. There is currently a way to specify custom CSS classes for the body cell of a column through the `cssClass` property, but this allows developers to also specify a CSS class for the header cell via the `headCssClass` property.

As an example, with this in place, we can do responsive list tables by hiding a column on smaller devices:

```
    version:
        label: system::lang.updates.plugin_version
        sortable: false
        cssClass: visible-lg
        headCssClass: visible-lg
```

Happy to update the docs if this change is approved.

### DIFF
--- a/modules/backend/classes/ListColumn.php
+++ b/modules/backend/classes/ListColumn.php
@@ -82,6 +82,11 @@ class ListColumn
     public $cssClass;
 
     /**
+     * @var string Specify a CSS class to attach to the list header cell element.
+     */
+    public $headCssClass;
+
+    /**
      * @var string Specify a format or style for the column value, such as a Date.
      */
     public $format;
@@ -137,6 +142,9 @@ class ListColumn
         }
         if (isset($config['cssClass'])) {
             $this->cssClass = $config['cssClass'];
+        }
+        if (isset($config['headCssClass'])) {
+            $this->headCssClass = $config['headCssClass'];
         }
         if (isset($config['searchable'])) {
             $this->searchable = $config['searchable'];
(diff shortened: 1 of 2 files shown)


w20k:
@bennothommo could you also add docs for that? :) 
Reviewed - ✅

bennothommo:
@w20k Docs PR here :) https://github.com/octobercms/docs/pull/367

LukeTowers:
@bennothommo does hiding a head cell actually hide the entire column or does it just move the rest of the head cells over?

w20k:
@LukeTowers I believe if you use both cssClass/headCssClass - should hide the column.
If not in the pair that would be a mess...

LukeTowers:
Are there any concerns with how this interacts with the invisible column property or how the user has defined their list setup?

w20k:
As I've commented already, the only huge issue might be with non-paired column/header...

Sorting - no prob, can't see can't interact.
Searchable: true - also an issue, but devs should understand that.
Invisible - might be a problem, too. It all depends on how you use it. Currently, it uses `visibility: hidden` and in pair with, something like, `display: none` - also could lead to `trouble` 😄

bennothommo:
@LukeTowers I just used the example above as a demonstration of my particular use case for this feature - it's not intended to be a replacement for the `invisible` property. @w20k is right, you'd need to use both `cssClass` and `headCssClass` in my example to get the proper effect.

Ultimately though, it's giving power to the developers to add a class to the list table head cells, as they can already do so for the content cells. It shouldn't affect anyone's list setups currently, as it's a completely new property.

PR #4217, opened 2019-03-27
add oc.inputPreset.beforeUpdate plugin hook

mjauvin:
If locale data attributes are found, avoid applying source input changes to all target translations

fix for https://github.com/rainlab/translate-plugin/issues/429

Note: don't forget to recompile JS assets when applying this change

Credits to @munxar and @w20k for this fix!

### DIFF
--- a/modules/system/assets/ui/js/input.preset.js
+++ b/modules/system/assets/ui/js/input.preset.js
@@ -229,7 +229,6 @@
             var timeout = event.type === 'paste' ? 100 : 0 
             var updateValue = function(self, el, prefix) { 
                 if (el.data('update') === false) {
-                    el.data('update', true)
                     return
                 }
                 el   
--- a/modules/system/assets/ui/js/input.preset.js
+++ b/modules/system/assets/ui/js/input.preset.js
@@ -228,8 +228,13 @@
 
             var sourceLocale = $(this).siblings('.ml-btn[data-active-locale]').text()
             var targetLocale = $el.data('locale-value')
+            var targetActiveLocale = $el.siblings('.ml-btn[data-active-locale]').text()
+            
             if (sourceLocale && targetLocale && sourceLocale !== targetLocale)
                 return
+
+            if (targetLocale && targetActiveLocale && targetActiveLocale !== targetLocale)
+                $el.siblings('.ml-btn[data-active-locale]').text(targetLocale)
 
             $el
                 .val(prefix + self.formatValue())
@@ -242,8 +247,13 @@
 
             var sourceLocale = $(this).siblings('.ml-btn[data-active-locale]').text()
             var targetLocale = $el.data('locale-value')
+            var targetActiveLocale = $el.siblings('.ml-btn[data-active-locale]').text()
+
             if (sourceLocale && targetLocale && sourceLocale !== targetLocale)
                 return
+
+            if (targetLocale && targetActiveLocale && targetActiveLocale !== targetLocale)
+                $el.siblings('.ml-btn[data-active-locale]').text(targetLocale)
 
             setTimeout(function() {
                 $el
--- a/modules/system/assets/ui/js/input.preset.js
+++ b/modules/system/assets/ui/js/input.preset.js
@@ -226,6 +226,11 @@
             if (self.cancelled)
                 return
 
+            var sourceLocale = $(this).siblings('.ml-btn[data-active-locale]').text()
+            var targetLocale = $el.attr('data-locale-value')
+            if (sourceLocale && targetLocale && sourceLocale !== targetLocale)
+                return
+
             $el
                 .val(prefix + self.formatValue())
                 .trigger('oc.inputPreset.afterUpdate')
@@ -233,6 +238,11 @@
 
         this.$src.on('paste', function() {
             if (self.cancelled)
+                return
+
+            var sourceLocale = $(this).siblings('.ml-btn[data-active-locale]').text()
+            var targetLocale = $el.attr('data-locale-value')
+            if (sourceLocale && targetLocale && sourceLocale !== targetLocale)
                 return
 
             setTimeout(function() {
--- a/modules/system/assets/ui/js/input.preset.js
+++ b/modules/system/assets/ui/js/input.preset.js
@@ -243,6 +243,10 @@
                     .trigger('oc.inputPreset.afterUpdate')
             }, timeout)
         })
+        
+        this.$el.on('change', function() {
+            self.cancelled = true
+        })
     }
 
     InputPreset.prototype.formatNamespace = function() {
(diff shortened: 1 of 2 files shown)


w20k:
@Rike-cz - mind if I ask you to test this PR. Should probably fix your issue with translate plugin :)

mjauvin:
> @Rike-cz - mind if I ask you to test this PR. Should probably fix your issue with translate plugin :)

You will of course need to recompile your js assets with artisan...

Rike-cz:
@w20k Works for me, thanks!

w20k:
Cool, thanks @Rike-cz!

mjauvin:
The only thing I noticed is that if the target element is currently selected in a different language than the source element language, its value will show the copied input from the source even though the language selector button still shows the other language. 

Still, the behavior underneath does what it's supposed to do.
The only way to not see this would be to trigger a language change in the target to match the one on the source, but I don't know how to do this or if this is really needed.

Opinions?

alxy:
Do not forget to test this if the translate plugin is not installed.

w20k:
@alxy It shouldn't be a problem here, you won't have more than two input fields, pretty much every time $src and $el will be present.
`sourceLocale && targetLocale` -> will be undefined 😄

w20k:
@mjauvin this should do the trick - ✅. 
Currently you chagne value of a single field that corresponds to source locale. So, changing `text`(lang code) in button is the best solution here ;)

Example:
```javascript
var sourceLocale = $(this).siblings('.ml-btn[data-active-locale]').text()
var targetActiveLocale = $el.siblings('.ml-btn[data-active-locale]').text()
var targetLocale = $el.data('locale-value')
if (sourceLocale && targetLocale && sourceLocale !== targetLocale)
    return

if (targetLocale && targetActiveLocale && targetActiveLocale !== targetLocale)
    $el.siblings('.ml-btn[data-active-locale]').text(targetLocale)
```

munxar:
@mjauvin thank a lot for the implementation. the only thing that bothers me is the fakt that we make the system dependant to the translate plugin. there are other places where this happens and I personally don't like it. This stuff should be in the translate plugin and the system must be extendable that the translate plugin can modify in a way that it works.
If we continue this route it will be unmaintainable in the future and against the mdularity of october cms.

On the other hand I see the need for this fix, because right now it breaks quite a lot of sites (especially the paes plugin site title to slug logic). Could slack together after this is mered to find a way in the future to move this kind of code into the translate plugin?
Again tnx for your work!

w20k:
@munxar we could fix it in the Translate Plugin as well, but core code should be changed before that. 
This part, for sure:  `this.$src.on('input paste'....`. Should be a class, id or data specififc.

mjauvin:
@munxar I wouldn't say "dependant" but it's definitely an exception FOR the translate plugin. Whether the translate plugin is installed or not does not break the code.

But, yes, making a change in core to allow plugins to change the behavior would be preferable.

w20k:
People are ignoring me 😄 Will state it once again, the core was written so it shouldn't support multi-lang behavior by default or better to say multiple inputs.

LukeTowers:
I would prefer if this could be implemented as the input.preset.js firing an event that could be hooked into by the translate plugin. Adding these specific exceptions directly to core negatively impact code readability and unnecessarily muddy the waters.

mjauvin:
Ok, here's a new proposal which should please everyone, albeit potential implementation bugs:

```
        this.$src.on('input paste', function(event) {
            if (self.cancelled)
                return

            var timeout = event.type === 'paste' ? 100 : 0
            var updateValue = function(self, el, prefix) {
                if (self.cancelled)
                    return
                el  
                    .val(prefix + self.formatValue())
                    .trigger('oc.inputPreset.afterUpdate')
            }

            var src = $(this)
            setTimeout(function() {
                $el.trigger('oc.inputPreset.beforeUpdate', [self, src])
                setTimeout(updateValue, 100, self, $el, prefix)
            }, timeout)
        })
```

I am calling updateValue() with a 100ms delay to avoid racing problems with the 'oc.inputPreset.beforeUpdate' event handler

mjauvin:
And we need to add this to Rainlab.Translate plugin multilingual.js:

```
    $('[data-input-preset]').on('oc.inputPreset.beforeUpdate', function(event, inputPreset, src) {
        var sourceLocale = src.siblings('.ml-btn[data-active-locale]').text()
        var targetLocale = $(this).data('locale-value')
        var targetActiveLocale = $(this).siblings('.ml-btn[data-active-locale]').text()
            
        if (sourceLocale && targetLocale && sourceLocale !== targetLocale)
            inputPreset.cancelled = true

        if (inputPreset.cancelled === false && targetLocale && targetActiveLocale && targetActiveLocale !== targetLocale)
            $(this).siblings('.ml-btn[data-active-locale]').text(targetLocale)
    })  
```

mjauvin:
I tested this and it works very well, with or without the translate plugin active

LukeTowers:
That looks good to me @mjauvin, submit the code for testing please

mjauvin:
Accompanying Rainlab.Translate PR:

https://github.com/rainlab/translate-plugin/pull/458

w20k:
@mjauvin Not it looks better! Thanks ;) Will re-test, hopefully, today! Thanks!

Rike-cz:
I found a problem. When I write to field which is listened slowly, it works. When I write quickly, it falls to previous bad behavior.
ml-slug-bug

w20k:
Hm... will test it today, might not be a prob at all 😄

mjauvin:
confirmed @Rike-cz discovery, applied a fix to both related PRs to address this.

it's now working as expected for fast typers... ;)

Rike-cz:
@mjauvin Woderful! For me it works with and without Rainlab.Translate plugin great!

w20k:
@mjauvin doesn't work for me with the latest build... not sure, yet, why.

mjauvin:
did you forget to rebuild your js assets? did you make sure your cleared
your cache to get the new multilingual.js file?

--
Marc

w20k:
Nope, I didn't...Url field language switch doesn't change... will debug it a little bit.

mjauvin:
I'm pretty sure it's a caching issue, that works really well now...

--
Marc

w20k:
Do you use chrome devtools with `cache: false`?

mjauvin:
I don't know, but I force a cache clear after each changes with
ctrl-shift-del

--
Marc

w20k:
@mjauvin latest build, fresh install with everything - doesn't work, that's confirmed. Will look deeper, why, and let you know.

w20k:
@LukeTowers this one should be good for merge.

w20k:
@mjauvin could you re-build assets, or it's not needed for `input.preset.js`?

mjauvin:
@w20k We need the built assets to be checked-in!? I thought the install process did this, no? If not, wouldn't it make sense it did?

LukeTowers:
@mjauvin no, built assets need to be checked in. Please ping me once that's complete.

mjauvin:
added compiled assets @LukeTowers

PR #4218, opened 2019-03-27
Fixes #4209.

### LINKED ISSUE #4209: Twig 2.7.4 breaks OctoberCMS w/ RainLab.Pages

LukeTowers:
Fixes #4209, replaces #4215

### DIFF
--- a/modules/cms/classes/CmsCompoundObject.php
+++ b/modules/cms/classes/CmsCompoundObject.php
@@ -9,8 +9,8 @@ use Cms\Twig\Extension as CmsTwigExtension;
 use Cms\Components\ViewBag;
 use System\Twig\Extension as SystemTwigExtension;
 use October\Rain\Halcyon\Processors\SectionParser;
-use Twig_Source;
-use Twig_Environment;
+use Twig\Source as TwigSource;
+use Twig\Environment as TwigEnvironment;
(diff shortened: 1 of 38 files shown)


tobias-kuendig:
This is working for me :+1: Tested on two installations with RainLab.Pages.

PR #4227, opened 2019-04-01
Optionally copy files instead of symlinking them while mirroring the public directory

msimkunas:
As discussed with @LukeTowers 

This pull request adds a `--copy` switch to the `october:mirror` command and copies the files and directories instead of symlinking them into the new public directory.

The functionality is completely optional. It works by creating a copy of all specified asset files and directories and then updates the mirrored `index.php` file with the updated paths to the `app.php` and `autoload.php` files.

### Why?

The `--copy` switch makes it possible to use the public folder deployment configuration on web hosts that do not support symbolic links.

Effectively, this change also allows shared hosting deployments to utilize the public folder configuration (although such use cases may be impractical because the `october:mirror` command would have to be run on the developer's machine).

The downsides of this approach is that it creates a copy of all published assets in the new public directory. However, I see this as a non-issue because the additional storage space required is an extremely minor inconvenience when compared with the advantage of broader web host support.

Moreover, in terms of file management copying is as good as symlinking in this case because the `october:mirror` command still needs to be run [after each system update](https://github.com/octobercms/docs/blob/f22b49c86bde81457a74c630a9636a966793189b/setup-configuration.md#using-a-public-folder).

I've tested these changes in my local development environment but they still need more testing to make sure everything works as expected. Any ideas on possible test cases?

### DIFF
--- a/modules/system/console/OctoberMirror.php
+++ b/modules/system/console/OctoberMirror.php
@@ -112,6 +112,10 @@ class OctoberMirror extends Command
             $this->mirrorWildcard($wildcard);
         }
 
+        if ($this->option('copy')) {
+            $this->updateIndexPaths();
+        }
+
         $this->output->writeln('<info>Mirror complete!</info>');
     }
 
@@ -178,7 +182,33 @@ class OctoberMirror extends Command
             }
         }
 
-        symlink($src, $dest);
+        if ($this->option('copy')) {
+            if(File::isFile($src)) {
+                File::copy($src, $dest);
+            } else if(File::isDirectory($src)) {
+                File::copyDirectory($src, $dest);
+            }
+        } else {
+            symlink($src, $dest);
+        }
+    }
+
+    protected function updateIndexPaths()
+    {
+        $this->output->writeln('<info> - Updating index.php paths</info>');
+
+        $dest = $this->getDestinationPath().'/index.php';
+        $contents = File::get($dest);
+
+        $src = base_path().'/bootstrap/app.php';
+        $relativePath = $this->getRelativePath($dest, $src);
+        $contents = str_replace('/bootstrap/app.php', "/$relativePath", $contents);
+
+        $src = base_path().'/bootstrap/autoload.php';
+        $relativePath = $this->getRelativePath($dest, $src);
+        $contents = str_replace('/bootstrap/autoload.php', "/$relativePath", $contents);
+
+        File::put($dest, $contents);
     }
 
     protected function getDestinationPath()
@@ -236,6 +266,7 @@ class OctoberMirror extends Command
     {
         return [
             ['relative', null, InputOption::VALUE_NONE, 'Create symlinks relative to the public directory.'],
+            ['copy', null, InputOption::VALUE_NONE, 'Copy files and directories instead of making symlinks.'],
         ];
     }
 }
--- a/modules/system/console/OctoberMirror.php
+++ b/modules/system/console/OctoberMirror.php
@@ -183,9 +183,9 @@ class OctoberMirror extends Command
         }
 
         if ($this->option('copy')) {
-            if(File::isFile($src)) {
+            if (File::isFile($src)) {
                 File::copy($src, $dest);
-            } else if(File::isDirectory($src)) {
+            } else if (File::isDirectory($src)) {
                 File::copyDirectory($src, $dest);
             }
         } else {


LukeTowers:
@bennothommo @daftspunk what's your thoughts on this? I'm not sure I'm a huge fan of this since while you should re-run the mirroring command on every update to pickup any new directories at least updates to existing files will be passed through; however this method will not do that which could lead to issues. I'm not sure we want to support this on hosting providers that are already so bad they don't support symlinks.

bennothommo:
@LukeTowers I definitely agree its scope of use is quite low, but I can think of a couple of examples:

- Docker instances - traditionally they've had flaky symlink support, although I believe it's better now.
- Windows hosts - not sure if it's changed now, but Windows historically had no concept of symlinks.

I suppose if a developer has full control of their upgrade process, and perhaps have their storage folder kept on an external file system, then this would be perfectly safe to use.

LukeTowers:
@bennothommo my main argument against it is the use case provided in the original post, specifically shared hosts that don't support symlinks / terminal access. Thus I'm against it since that would just lead to further issues on those sorts of hosts (which shouldn't really be used in the first place). Thoughts?

bennothommo:
@LukeTowers I agree with you on not *recommending* those sorts of hosts, but we do support Windows hosts implicitly, as we are supporting SQL Server. This solution would definitely help with Windows hosts.

bennothommo:
However, maintenance burden wise, it's probably going to be painful to maintain when people start talking about permissions and missed changes to theme or plugin assets. So yeah, I'm probably leaning towards not including it.

LukeTowers:
I could be fine including it if we were very explicit in the documentation about the downsides of using this flag and all of the ways things could potentially break when using this flag that we don't want to see issues about.

daftspunk:
I'm 50/50 on this, everyone makes good points. Will revisit after some pondering

modules/system/console/OctoberMirror.php
```diff
@@ -178,7 +182,33 @@ protected function mirror($src, $dest)
             }
         }
 
-        symlink($src, $dest);
+        if ($this->option('copy')) {
+            if(File::isFile($src)) {
+                File::copy($src, $dest);
+            } else if(File::isDirectory($src)) {
```
LukeTowers:
```suggestion
            if (File::isFile($src)) {
                File::copy($src, $dest);
            } else if (File::isDirectory($src)) {
```

bennothommo:
@github-actions run

<details>

```js
(async () => {
  // Get issue/PR number

  // Add Testing Needed label
    owner: context.repo.owner,
    repo: context.repo.repo,
    issue_number: number,
    labels: ['Testing Needed'],
  });

  // Comment body
    + '\n\n'
    + '\n\n'
    + '```\n'
    + '```';

  // Add comment
})();
```

LukeTowers:
With the recent addition of image resizing in the core, this made me rethink this a bit. Copying files would also include the media, public storage, and resized images folder, which means if your site uses images at all, you could very easily end up in the situation of doubling your file storage requirements every time you run the command. If we do go ahead with including it I would want a confirmation step that explains all of the downsides of using the flag and strongly recommends switching to a host that does provide support for symlinks.

bennothommo:
I'm actually leaning more towards not needing this anymore. I've tested symlinks on Docker and they work fine if you are using a base image that supports symlinks (which pretty much every PHP image does). I've also, through the Windows testing implemented recently, worked out that Windows servers support symlinks just fine since Windows Server 2008.

Only Windows Server 2013 and above are supported by Microsoft now, which means that Windows is fine for symlinks.

Windows *not* supporting symlinks was the only reason I felt this sort of change might be necessary - with the latest knowledge, I don't believe this is required.

msimkunas:
Perhaps the only remaining use case is running October on shared hosting which may or may not be something you'd like to support. 

Personally, I'd rather not use shared hosting for my October projects but I imagine that for someone running a simple low maintenance site it could be a reasonable choice. Your call.

LukeTowers:
@msimkunas I think at this point we'll close it for now, until someone actually needs it for their use case, rather than integrate it for the theoretical usecase of someone that's not technical enough to use better hosting but is technical enough to want to use a public folder and can use the terminal. At that point we can look at adding it again, but with the confirmation step. Thanks for your work on it up to this point!

msimkunas:
Well, in some cases shared hosting may be chosen for reasons other than the developer not knowing any better.

For simple sites, VPS maintenance is overkill and paying for managed hosting is just too expensive if the site barely has any visitors at all and is intended to be used solely for its backend features.

But I'm not going to debate this decision any further, I do see where you're coming from and understand that this is very low priority. Thanks for taking the time to wrap this up.

Samuell1:
@msimkunas VPS is overkill for websites that have so low visitors if i can recommend if you building just static site you can deploy html, css, js to Netlify (Its has free tier) or just run cheapest instance on AWS if you need php on serverside.

msimkunas:
@Samuell1 A low visitor count does not imply uselessness – you could be building an exclusively backend-oriented app or a private community forum website, for example. In all of these cases October could be a valid CMS choice.

bennothommo:
@msimkunas For what it's worth, the most common shared hosting environment uses cPanel/WHM, and I am certain they support symlinks, so I feel it is a non-issue. As @LukeTowers said, if someone indicates a use case where it's definitely required, we'll happily revisit it.

PR #4102, opened 2019-02-01
Allow datatables to work within a repeater

bennothommo:
Fixes #4063.

Allows datatable widgets (and other widgets) to enact their `getSaveValue` method inside a repeater, and fixes the POST data structure from a datatable to allow it to be repeated in a repeater.

PR #4231, opened 2019-04-03
Fix JS error when calling getDropdownOptions handler in DataTable widget

bennothommo:
Hat tip @Samuell1.

The change to the DataTable widget alias to make it work in repeaters (https://github.com/octobercms/october/pull/4102) caused the onGetDropdownOptions handler to fail. This handler appears to be used if a dropdown column is specified without options.

The issue results from the table alias containing square brackets, which will not work with AJAX framework handler as it throws the following error:
```
Error: Invalid handler name. The correct handler name format is: "onEvent".
```

This fix should prevent that error. I have tested this fix with the repeating datatables and they still appear to work with this change in place.

### DIFF
--- a/modules/backend/formwidgets/DataTable.php
+++ b/modules/backend/formwidgets/DataTable.php
@@ -150,7 +150,7 @@ class DataTable extends FormWidgetBase
 
         $config->dataSource = 'client';
         if (isset($this->getParentForm()->arrayName)) {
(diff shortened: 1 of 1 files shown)


LukeTowers:
@Samuell1 can you test and confirm the fix?

w20k:
@LukeTowers I've already asked him, probably he will manage to test in the evening ;)

Samuell1:
Tested it and it works! @LukeTowers

Issue #4237, opened 2019-04-05
Search widget sometimes shows results for parts of the words from previous ajax request


CSNWEB:
- OctoberCMS Build: v1.0.442 
- PHP Version: 7.1
- Database Engine: Postgres
- Plugins Installed: Martin.SSOLogin, various unpublished private ones.

### Description:
While typing in the search widget of a backend controller list it sometimes happens that a previous search request takes longer to finish then the last one. This leads to an earlier one beeing used, showing you results from the first part of the word you typed after already seeing the one for the full word shortly. I think when a new search request is sent, all older ones should be cancelled by the client. Happy to help implementing a fix if someone gives me a hint where that could be possible. 

### Steps To Reproduce:
I think one needs a bigger list in our case ~20 searchable columns and ~9000 entries in order to see this bug. And one must type faster than the requests take clearly. If that conditions are met as said above go to that list in the backend and type something in the searchbar.

PR #4238, opened 2019-04-05
Abort old track input requests

CSNWEB:
This aborts old request triggered by the data-track-input feature. Without this older requests might overwrite the results of newer ones. See #4237 for more details.

### DIFF
--- a/modules/system/assets/js/framework-min.js
+++ b/modules/system/assets/js/framework-min.js
@@ -134,8 +134,8 @@ var interval=$(this).data('track-input')
 if(!interval)
 interval=300
 var self=this
-this.dataTrackInputTimer=window.setTimeout(function(){if(this.lastDataTrackInputRequest){this.lastDataTrackInputRequest.abort();}
-this.lastDataTrackInputRequest=$(self).request();},interval)})
+this.dataTrackInputTimer=window.setTimeout(function(){if(self.lastDataTrackInputRequest){self.lastDataTrackInputRequest.abort();}
+self.lastDataTrackInputRequest=$(self).request();},interval)})
 $(document).on('submit','[data-request]',function documentOnSubmit(){$(this).request()
 return false})
 $(window).on('beforeunload',function documentOnBeforeUnload(){window.ocUnloading=true})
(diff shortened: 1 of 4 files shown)


w20k:
@CSNWEB - nice one! Will try to test this out somehow)

CSNWEB:
@w20k thanks! You can see that the abort is working fine without having the issues I described in #4237, as long as your requests take longer than 300ms, typing a little bit slower will let you see the aborts in the network tab of of the dev tools (I used chrome, but I think every browser has this)

w20k:
@CSNWEB don't have any issues with your code, should work - nothing fancy 😄 Just need to make sure it works as expected.
While you're still here could you re-compile the assets and commit those as well: `php artisan october:util compile js`

CSNWEB:
If you want something to complain 😄 : The this bounds to the global window object, I think, thats not super nice, but I did not find a better solution. 
Okay will compile the asses that didn't knew this was needed. Thanks for pointing out!

w20k:
>Do not edit compiled asset files such as october.css, framework.css, etc. directly. Instead, edit the LESS files inside the less/ directory and then run php artisan october:util compile assets from the project root and commit the changed LESS files as well as the recompiled asset files.

This part in contributing guidelines: https://github.com/octobercms/october/blob/master/.github/CONTRIBUTING.md
@CSNWEB first will test, then look if it could be `optimized` ;)

CSNWEB:
Oh yes sorry didn't read that again and forgot about it. However added them now.

LukeTowers:
@CSNWEB 
> The this bounds to the global window object, I think, thats not super nice, but I did not find a better solution.

Are you sure about that? Shouldn't `this` be referencing the element that triggered the request?

LukeTowers:
Ah, wait I see. Could you use `self` instead?

CSNWEB:
@LukeTowers yes that worked, what an obvious choice, thanks 😀! @w20k For testing it helped me to set the connection speed to slow 3g in chrome.

LukeTowers:
Just need someone to test now

bennothommo:
@LukeTowers @CSNWEB @w20k I've tested it and confirmed it working.

Screenshot from 2019-04-19 22-06-21

w20k:
@bennothommo thanks for helping out! 😉
@LukeTowers could be merged, I guess.

PR #4204, opened 2019-03-19
[WIP] Relation manage filters

alxy:
**This PR is a WIP**

It aims at making [list filters](https://octobercms.com/docs/backend/lists#list-filters) available for the RelationController, specifically the manage widget.

Example usage in `config_relation.yaml`:
```
posts:
    label: Posts
    list: ~/plugins/october/test/models/post/columns.yaml
    form: ~/plugins/october/test/models/post/fields_simple.yaml
    manage:
        showSearch: true
        filter:
            scopes:
                published_at:
                    label: Date
                    type: daterange
                    conditions: created_at >= ':after' AND created_at <= ':before'
                username:
                    label: Name
                    type: text
                    conditions: name = :value
                    size: 2
                status:
                    label: Status
                    modelClass: October\Test\Models\Attribute
                    conditions: status_id in (:filtered)
                    nameFrom: name
```

Peek 2019-03-20 11-29

It does work, however there are some issues that need to be fixed. These are mostly interface/javascript based:
- [x]  The `date` and `daterange` pickers appear to open behind the modal
- [x] Clicking apply/clear on a filter popup triggers a JS error
- [x] Figure out why the positioning of the popups is a bit off when using a container element

I debugged the JS errors a bit and they do happen if the modal is opened a second time without a har refresh of the page. It appears, in this case the filter is not correctly re-initialized. Here are the errors I am getting in the console once I clear the Apply/Clear buttons on a filter widget popover:

```
storm-min.js?v447:31699 Uncaught TypeError: Cannot read property 'find' of null
    at FilterWidget.updateScopeDateSetting (storm-min.js?v447:31699)
    at FilterWidget.filterByDate (storm-min.js?v447:31765)
    at HTMLButtonElement.<anonymous> (storm-min.js?v447:31498)
    at HTMLDocument.dispatch (jquery.min.js?v447:5201)
    at HTMLDocument.elemData.handle (jquery.min.js?v447:5009)
FilterWidget.updateScopeDateSetting @ storm-min.js?v447:31699
FilterWidget.filterByDate @ storm-min.js?v447:31765
(anonymous) @ storm-min.js?v447:31498
dispatch @ jquery.min.js?v447:5201
elemData.handle @ jquery.min.js?v447:5009
3storm-min.js?v447:31699 Uncaught TypeError: Cannot read property 'find' of null
    at FilterWidget.updateScopeDateSetting (storm-min.js?v447:31699)
    at FilterWidget.filterByDate (storm-min.js?v447:31765)
[...]

PR #4241, opened 2019-04-06
Add support for filters on the RelationController

alxy:
This is a clean-up of https://github.com/octobercms/october/pull/4204

Demo: Peek 2019-03-20 11-29

Example usage in `config_relation.yaml`:
```
posts:
    label: Posts
    list: ~/plugins/october/test/models/post/columns.yaml
    form: ~/plugins/october/test/models/post/fields_simple.yaml
    manage:
        showSearch: true
        filter:
            scopes:
                published_at:
                    label: Date
                    type: daterange
                username:
                    label: Name
                    type: text
                    conditions: name = :value
                    size: 2
                status:
                    label: Status
                    modelClass: October\Test\Models\Attribute
                    conditions: status_id in (:filtered)
                    nameFrom: name
```

Documentation PR: https://github.com/octobercms/docs/pull/368


Rike-cz:
Tested for my use cases and it works great!

w20k:
@alxy - works great!
@LukeTowers testing is finished! If you have any comments to add, feel free ;)

LukeTowers:
@alxy can you add support to the view config too? I don't imagine it will be too much more work to do that and I want both modes to support it if we're going to have support for it present.

LukeTowers:
@alxy if you can merge https://github.com/octobercms/october/pull/3558 with this then I'd be happy to merge this PR in

alxy:
@LukeTowers @w20k I have incorporated the new changes, however, the filters now need new testing. It should work now for view and manage modes.

Rike-cz:
I have problem with checkboxes in filters. When I click on checkbox in manage mode, view mode checkbox is triggered.
relationmanager

alxy:
@Rike-cz Can you try to debug this? I'd think its entirely JS related...

Rike-cz:
@alxy As you can see, I have the same filter in view and in manage mode. And because ids of html input elements are not prefixed, they have the same id. Checkbox in view mode is `scope-available` just as in manage mode. And Bootstrap (?) JS for custom checkboxes finds view mode before manage mode, so view mode checkbox is triggered.

LukeTowers:
@alxy sounds like the filter widget needs to include it's alias in the generated IDs for those form elements

alxy:
@LukeTowers yeah, this might be true, but is actually an issue of the FilterWidget itself then...

LukeTowers:
One that will need to be fixed before this can be merged

alxy:
@LukeTowers @Rike-cz I added a commit to fix this issue. It adds the widget alias to each filter scope - I wonder if it has any other consequences...

Rike-cz:
I think it's ok. Works for me and my tested usecases.

LukeTowers:
@bennothommo could you take a look at this?

bennothommo:
@LukeTowers Will do.

modules/backend/behaviors/RelationController.php
```diff
@@ -1329,12 +1398,32 @@ public function relationExtendManageWidget($widget, $field, $model)
...
     {
     }
 
+    /**
+     * Provides an opportunity to manipulate the manage filter widget.
+     * @param \Backend\Widgets\Filter $widget
+     * @param string $field
+     * @param \October\Rain\Database\Model $model
+     */
+    public function relationExtendManageFilterWidget($widget, $field, $model)
```
bennothommo: @alxy I might suggest renaming this to `relationExtendManageFilterScopes` and changing the first attribute to `$filter` so it is consistent with the list widget filter extensions (https://octobercms.com/docs/backend/lists#extend-filter-scopes)
```suggestion
    public function relationExtendManageFilterScopes($filter, $field, $model)
```

modules/backend/behaviors/RelationController.php
```diff
@@ -1329,12 +1398,32 @@ public function relationExtendManageWidget($widget, $field, $model)
...
+    {
+    }
+
+    /**
+     * Provides an opportunity to manipulate the view filter widget.
+     * @param \Backend\Widgets\Filter $widget
+     * @param string $field
+     * @param \October\Rain\Database\Model $model
+     */
+    public function relationExtendViewFilterWidget($widget, $field, $model)
```
bennothommo: @alxy As above :)
```suggestion
    public function relationExtendViewFilterScopes($filter, $field, $model)
```

bennothommo:
@alxy @Rike-cz @LukeTowers Tested it quite thoroughly, all seems to be working well. I've left a couple of suggestions on the callback methods available for extending the filter widget. You may need to include some details about these callback methods in the docs PR as well.

modules/backend/behaviors/RelationController.php
alxy: I made these callbacks consistent with the other hooks, e.g. `relationExtendManageWidget`.
I also have to say that the callback is indeed more generic and does not only allow for extending the filter scopes (although this might be the most common use case), but for modifying the entire Filter widget. This is why I think the name is ok.

bennothommo: Technically, a dev also has the same scope of access with the `extendListFilterScopes` method as well, being that the first parameter there is also an instance of `\Backend\Widgets\Filter`. I figured that renaming this method might also add some separation from the similarly named `relationExtendManageWidget` method.

@w20k @LukeTowers do you guys have any thoughts on this?

w20k: We could change only an attribute name and call it, something like, - `$filterWidget`. And that should be it 😄

`public function relationExtendManageFilterWidget($filterWidget, $field, $model)`

bennothommo: Thinking about it, I think it's fine to keep the name as is. It's probably `extendListFilterScopes` that should be renamed :)

bennothommo:
@alxy Once you've added the docs for the filter widget callback methods to your docs PR, I think this will be ready to merge.

host138:
In which version will it be implemented?

bennothommo:
@host138 It should already be available. It was released with build 455.

host138:
I have this version, but it doesn't work stable. 
For example:
- i cannot set focus in the first number range
- in the text type filter doesn't work 
- label is strange

in the config_relation.yaml i added: 
```
manage:
        list: ~/plugins/itlab/conference/models/user/columns.yaml
        form: ~/plugins/itlab/conference/models/user/fields.yaml
        recordsPerPage: 20
        filter:
            scopes:
                id:
                    label: ID
                    type: numberrange
                    conditions: id >= ':min' and id <= ':max'
                phone:
                    label: Phone
                    type: text
                    conditions: phone LIKE concat("%", :value, "%")
                    size: 5
                name:
                    label: Name
                    type: text
                    conditions: name LIKE concat("%", :value, "%")
                    size: 5
                email:
                    label: Email
                    type: text
                    conditions: email LIKE concat("%", :value, "%")
                    size: 5
                created_at:
                    label: Created
                    type: daterange
```
Result:
relationmanager_filter

Strange label:


Finally, will be nice to use the format for the daterange type: 
```
created_at:
    label: Created
    type: daterange
    format: d.m.Y h:i:s
```

host138:
I have this version, but it doesn't work stable. 
For example:
- i cannot set focus at the first number range
- in the "text" type filter doesn't work 
- label is strange

in the config_relation.yaml i added: 
```
manage:
        list: ~/plugins/author/plugin/models/user/columns.yaml
        form: ~/plugins/author/plugin/models/user/fields.yaml
        recordsPerPage: 20
        filter:
            scopes:
                id:
                    label: ID
                    type: numberrange
                    conditions: id >= ':min' and id <= ':max'
                phone:
                    label: Phone
                    type: text
                    conditions: phone LIKE concat("%", :value, "%")
                    size: 5
                name:
                    label: Name
                    type: text
                    conditions: name LIKE concat("%", :value, "%")
                    size: 5
                email:
                    label: Email
                    type: text
                    conditions: email LIKE concat("%", :value, "%")
                    size: 5
                created_at:
                    label: Created
                    type: daterange
```
Result:
relationmanager_filter

Strange label:


Finally, will be nice to use the format for the daterange type: 
```
created_at:
    label: Created
    type: daterange
    format: d.m.Y h:i:s
```

alxy:
@host138 It's weird, I have had the same issues for the date/daterange filters as well (Wasnt able to focus). I think we have fixed this [here](https://github.com/octobercms/october/commit/a2315624df0406a78d12ff262faac4d8443c05d6#diff-d5530d68aa66cdd0d3bca6bd092da290).
The text filter worked for me, try with a minimal example please (i.e. remove all other filters, try a different filter name, etc)
Labels are probably odd because there is too few space for all these filters. Feel free to add some CSS to fix this.

PR #4234, opened 2019-04-04
Simplify Repeater Logic

LukeTowers:
Refs: #4230

PR #4242, opened 2019-04-08
Additional fixes to #4234

bennothommo:
Additional fixes to #4234 (Simplify Repeater Logic) to resolve issues brought up here: https://github.com/rainlab/translate-plugin/pull/466#issuecomment-480583540

Related: https://github.com/rainlab/translate-plugin/pull/467

### DIFF
--- a/modules/backend/formwidgets/Repeater.php
+++ b/modules/backend/formwidgets/Repeater.php
@@ -78,6 +78,13 @@ class Repeater extends FormWidgetBase
     protected $groupDefinitions = [];
 
     /**
+     * Determines if repeater has been initialised previously
+     *
+     * @var boolean
+     */
+    protected $loaded = false;
+
+    /**
      * @inheritDoc
      */
     public function init()
@@ -93,6 +100,11 @@ class Repeater extends FormWidgetBase
 
         if ($this->formField->disabled) {
             $this->previewMode = true;
+        }
+
+        // Check for loaded flag in POST
+        if ((bool) post($this->alias . '_loaded') === true) {
+            $this->loaded = true;
         }
 
         $fieldName = $this->formField->getName(false);
@@ -197,8 +209,17 @@ class Repeater extends FormWidgetBase
      */
     protected function processItems()
     {
+        $currentValue = ($this->loaded === true)
+            ? post($this->formField->getName())
+            : $this->getLoadValue();
+
+        if ($currentValue === null) {
+            $this->indexCount = 0;
+            $this->formWidgets = [];
+            return;
+        }
+
         $groupMap = [];
-        $currentValue = post($this->formField->getName(), $this->getLoadValue());
 
         // Ensure that the minimum number of items are preinitialized
         // ONLY DONE WHEN NOT IN GROUP MODE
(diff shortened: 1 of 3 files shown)


bennothommo:
@LukeTowers @w20k This branch is also ready for review.

I had added a reorder event handler to the main Repeater widget for the reordering fix in the MLRepeater widget, but that fix was no longer needed. I would be happy to keep it in there as @CptMeatball was interested in seeing a `sync` option added for MLRepeater down the line and this handler would be of use for this, but please let me know if you want me to remove it in the mean time.

This branch also contains some maintenance on index handling when processing items, to augment the work @LukeTowers has already done.

LukeTowers:
@bennothommo let's remove the reorder stuff in the meantime, it'll be easier to add it in when we full understand that feature.

bennothommo:
@LukeTowers No probs, reverted.

The only thing left that I've noticed during testing is that if I have previously saved a repeater with an item and then reloaded the page, deleted that item and created a new item, the index starts at 1. This is due to [this line](https://github.com/octobercms/october/pull/4242/files#diff-a8e657fa9654328055121a4e2cc4311bR201). If there is no POST data for the field, which is the case when the repeater has no items, it instead uses the `loadData` which contains the original repeater item.

I'm thinking that once the form is loaded initially, there is no need to rely on the `loadData` any further, so it might be better to just use the POST data exclusively from that point on, but I thought I'd check with you first to see if there's anything that I might have missed in this approach before I make that change.

LukeTowers:
Is there anyway that we could infer from the request an explicit empty state so we know how to tell the difference between a request that has all items removed and an initial page load? What would your proposed change look like?

bennothommo:
@LukeTowers A repeater widget returns no value if it is empty, effectively disappearing from the POST data. I'm not sure that's reliable enough to presume an empty state.

I can think of 2 potential fixes:

1. Have a hidden field added to the form when a repeater is empty, with an empty value, making the field appear in the POST data. When an item is added, remove this hidden field.
- Pro: Quick fix
- Con: May break any systems that are relying on the repeater being hidden in POST if empty.

2. Add an explicit value to the AJAX calls when a repeater item is added or deleted, or the repeater is refreshed, and only use POST data when this value is present.
- Pro: Shouldn't break any systems like option 1
- Con: Lot more work, would need to make sure that the AJAX request is actually affecting that repeater.

LukeTowers:
@bennothommo and what would the pros cons look like for just ignore loadValue after initial load?

bennothommo:
@LukeTowers It might be difficult to determine that it is in fact the initial load, especially for forms that open up in an Ajax request.

LukeTowers:
@bennothommo wouldn't `!Request::ajax()` suffice?

LukeTowers:
Ah, never mind, I see what you're saying with the second part of your comment.

LukeTowers:
Option 1 with a custom form name (i.e. name = `$this->alias . '_is_empty'`) so it doesn't get lumped into the regular POST array for the widget and is instead accessed directly would probably work the best

bennothommo:
Sounds good!

bennothommo:
@LukeTowers Done the fix. Your comment gave me an idea about how to detect the initial load - simply have a hidden field that marks when a repeater is loaded. If that's in the POST, then it's been initialized previously and POST data should be used exclusively from then on.

Issue #4243, opened 2019-04-08
Add method/event after sorting


jexme:
It may be worth adding an event, or a method in the model, after the records are sorted (ReorderController, onReoder method).

As an example, this would be very useful if the model data is cached and you need to clear the cache after a change in sorting.

LukeTowers:
Closing as it has been over a month since any activity on this occurred and we are trying to figure out what issues are still relevant. If this is still something that you would like to see through to fruition please respond and we can get the ball rolling.

Issue #4025, opened 2018-12-31
afterBoot event not firing


SeriousKen:
##### Expected behavior
An event handler bound to model.afterBoot should get triggered when the model is booted

##### Actual behavior
The event is never triggered

##### Reproduce steps
Bind an event handler to any model eg:

```php
Example\Plugins\Model\MyModel::extend(function ($model) {
    $model->bindEvent('model.afterBoot', function () {
        \Log::info('Model booted');
    });
});
```

##### October build
443

The cause seems to be in the constructor of `October\Rain\Database\Model`

```php
/**
 * Constructor
 */
public function __construct(array $attributes = [])
{
    parent::__construct();

    $this->bootNicerEvents();

    $this->extendableConstruct();

    $this->fill($attributes);
}
```

The parent constructor has alreadt fired the `booted` Eloquent event so when `bootNicerEvents` registers the handler for `booted` it is too late. Calling `bootNicerEvents` before calling the parent constructor seems to fix this but I don't know if this has other implications or not.

jexme:
@LukeTowers @SeriousKen I confirm the problem, at the moment this functionality does not work, but you cannot call `bootNicerEvents` before the constructor.

Yes, in this case, the `$model-> afterBoot()` method will work, but the event itself will still not be listened to, because trait Emitter, in which the creation of new events takes place (`bindEvent` method), will not be called yet

PR #4245, opened 2019-04-08
Added method for model before/after sorting records #4243

jexme:
Added such an option to call the method before / after sorting records. Adding an event for a model, I think it would be too costly to solve this problem, because when sorting, more than one record will change.

Perhaps you can offer a better option?


LukeTowers:
Local events would be the same as a model method. I would prefer that `model.beforeReorder` and `model.afterReorder` local events were added instead, although you will need to provide inline docs for them (see https://github.com/octobercms/library/blob/master/src/Database/Model.php#L178-L188 for an example) and you will need to provide an example case for both events to vouch for the need for their existence.

LukeTowers:
Refs: #4243

jexme:
> Local events would be the same as a model method. I would prefer that `model.beforeReorder` and `model.afterReorder` local events were added instead, although you will need to provide inline docs for them (see https://github.com/octobercms/library/blob/master/src/Database/Model.php#L178-L188 for an example) and you will need to provide an example case for both events to vouch for the need for their existence.

accepted, will do

but at the moment I noticed that the method you cited in the example (afterBoot) does not work, such a problem has already been created #4025

PR #4248, opened 2019-04-09
Added event and method in model after reordered records

jexme:
Added `model.afterReorder` event and `afterReorder` method in model after reordered records.

PS: due to lack of experience with PR, I apologize for duplication. #4245

### DIFF
--- a/modules/backend/behaviors/ReorderController.php
+++ b/modules/backend/behaviors/ReorderController.php
@@ -153,6 +153,23 @@ class ReorderController extends ControllerBehavior
                     break;
             }
         }
+
+        /**
+         * @event model.afterReorder
+         * Called after the model is reordered
+         *
+         * Example usage:
+         *
+         *     $model->bindEvent('model.afterReorder', function () use (\October\Rain\Database\Model $model) {
+         *         \Log::info("{$model->name} has been reordered!");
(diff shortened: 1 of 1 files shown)


bennothommo:
@jexme Apologies for the extreme delay in getting this PR looked at, however, the event as it is defined now is a bit unclear - it appears to be a model event, but the `$model` that the event is being fired on is not necessarily the model that's actually been re-ordered.

From the `ReorderController` class, you can see that when the model is a SimpleTree model being re-ordered, the method called just accepts an [array of IDs](https://github.com/octobercms/october/blob/9961c7363d048d20a9104f68f46a4b7fafef883a/modules/backend/behaviors/ReorderController.php#L125) to modify, whereas, when the model is a NestedTree model being re-ordered, it's not the `$model` instance that is re-ordered, but a [`$sourceNode` instance](https://github.com/octobercms/october/blob/9961c7363d048d20a9104f68f46a4b7fafef883a/modules/backend/behaviors/ReorderController.php#L138-L153) that is re-ordered instead. Your event appears to be more synonymous with a controller event, which means you can't use the example that you have included in the docblock for the event.

Could you either change this to a controller event, or instead, change the event to be fired with the correct model instances?

PR #4262, opened 2019-04-16
Added strip_tags to exportFromList

ChrisBijsters:
When using date/time column types October creates date/time html tags around the column values, these are not wanted during export as it will generate invalid csv.


LukeTowers:
See the `raw` config option under `useList:`

Issue #1084, opened 2015-04-25
Media Manager doesn't work on Safari


SaifurRahmanMohsin:
The media manager doesn't seem to work on Safari. It encounters and error so the rest of the JS doesn't load, so I cannot perform any actions. I tested on Chrome and Firefox and it works well. Here's the error I get

![Safari Media Manager Issue](http://i.imgur.com/RkCQlEl.png)

hybridvision:
Which version of Safari? It works fine for me and I have v7.0.6.

I had some javascript errors initially in Chrome but it was a cache problem so make sure you've cleared your cache.

SaifurRahmanMohsin:
I'm running Safari Version 8.0.5 (10600.5.17)

hybridvision:
Glad it worked :)

I should probably update my Safari at some point!

Issue #1077, opened 2015-04-21
Report widget layout


scottbedard:
I've got a fairly complicated dashboard set up, and i'm encountering a bug with how the widgets are positioned. The first screenshot shows how I've set up the dashboard, the second shows what is displayed when I refresh the backend. As you can see, the widgets are not maintaining their position. I'll try and poke around a bit tonight and see if I can find the cause of the issue.

![screenshot 2015-04-21 01 41 13](https://cloud.githubusercontent.com/assets/7980426/7246222/beed08ba-e7c7-11e4-9662-80c82357a2ac.png)

![screenshot 2015-04-21 01 41 29](https://cloud.githubusercontent.com/assets/7980426/7246240/e5a6ff06-e7c7-11e4-91f5-ba7c14bc7386.png)

daftspunk:
Fixed by #1100

scottbedard:
Uhg, this nuisance appears to be back...
https://www.dropbox.com/s/udu9ax0qvu1rq09/Screenshot%202015-05-16%2009.15.00.png?dl=0

Please re-open this, I will look into this again later today.

PR #4263, opened 2019-04-16
Update Raphael V2.2.7 to V2.2.8

ayumihamsaki2:
Update Raphael V2.2.7 to V2.2.8 shouldn't be a problem to update to October as as minor semantic version update.

Side note: The V2.2.7 was dated two years ago and V2.2.8 this year, as many features have been updated to the internet this new version brings some useful features and upgrades to October CMS.

ref: https://github.com/octobercms/october/issues/3867


ayumihamsaki2:
Patch File: https://github.com/ayumihamsaki/patches/blob/master/raphael/0001-Update-Raphael-V2.2.7-to-V2.2.8.patch

ayumihamsaki2:
@w20k If you got any free time, would you like to test this ? Thanks in advance!

LukeTowers:
@ayumihamsaki the difference between the two versions in terms of the actual code is quite dramatic, it looks like it's now some sort of compiled webpack crap? Is there an option without that mess?

ayumihamsaki:
@LukeTowers Yah I hate Webpack as well makes javascript files way too big in size! I had a look for a none-webpack version and they are all webpack versions now.

The only choices on V2.2.8 are:

- `raphael.min.js` (includes `eve` and it's minified)
- `raphael.js` (includes `eve` and it's not minified)
- `raphael.no-deps.js` (doesn't include `eve` it's not minified)
- `raphael.no-deps.min.js` (doesn't include `eve` it's minified)

*All four versions are bundled webpack.

I took the following version (because we use `Eve`):

- `raphael.js` (includes `eve` and it's not minified)

---

Not sure what to do - to get you a non-webpack version though.

ayumihamsaki:
Side note: they were using webpack V4 to compile this version.

LukeTowers:
What exactly do we gain by upgrading?

ayumihamsaki:
V2.2.8

- Downgraded qunit to v1 to make tests work again
- Updated webpack to build dist
- Updated supportsTouch query from modernizr [#1084](https://github.com/DmitryBaranovskiy/raphael/pull/1084) (**possible improvement to october with touchscreens**)
- Fix drag offsets passed to "drag.start" from touch drags in elproto.drag [#1101](https://github.com/DmitryBaranovskiy/raphael/pull/1101)  (**possible improvement to october with touchscreens**)
- Fix memory leak in Element.removeData() [#1077](https://github.com/DmitryBaranovskiy/raphael/pull/1077)  (**possible bug fix for october cms**)
- Adding react-raphael to related projects [#1072](https://github.com/DmitryBaranovskiy/raphael/issues/1072)
- Added icons back to raphael's website
- Updated version in banner and raphael's website
- Updated site refs to https

This is my guess.

LukeTowers:
Hmm. What do you think about manually porting only the relevant fixes over?

ayumihamsaki:
I'm thinking this version and all the future versions they plan to use webpack now. Looking at the code file it is pure garage to bug fix for october. The question comes to mind do we need to bug fix this file or will the errors be in the october files instead?

I don't mind leaving this PR open and not adding it to October, this package related to things like the scoreboard see here: https://octobercms.com/docs/ui/scoreboard

Right now I don't see any major errors or issues with it and I have lots of custom plugins using these things everyday! If there was a big issue I would have filed it.

So personally I think just leave this pr open and decide at a later date. The jQuery [V3.4.0](https://github.com/octobercms/october/pull/4259) that is quite important as it has a major security fix and quite a few updates to some things needed for october. I will add that list to that pr now for ref.

LukeTowers:
I'll close this PR for now, we can look at making any fixes if they ever come up as issues here.

PR #4266, opened 2019-04-17
 Add option to disable UpdateManager after backend login 

### LINKED ISSUE #3471: All caches are flushed on backend login

Samuell1:
Fixes #3471

### DIFF
--- a/config/cms.php
+++ b/config/cms.php
@@ -104,7 +104,7 @@ return [
     |
     */
 
-    'runMigrationsOnLogin' => true,
+    'runMigrationsOnLogin' => null,
 
     /*
     |--------------------------------------------------------------------------
--- a/config/cms.php
+++ b/config/cms.php
@@ -98,7 +98,7 @@ return [
     | Determines if logging in backend should run UpdateManager
     |--------------------------------------------------------------------------
     |
-    | If value is true it runs UpdateMananger after login in backend.
+    | If value is true, UpdateMananger will be run on logging in to the backend.
     | It's recommended to set this value to 'false' in production enviroments
     | because it cleares cache on every login in backend.
     |
--- a/config/cms.php
+++ b/config/cms.php
@@ -104,7 +104,7 @@ return [
     |
     */
 
-    'runUpdateManagerAfterBackendLogin' => true,
+    'runMigrationsOnLogin' => true,
 
     /*
     |--------------------------------------------------------------------------
--- a/config/cms.php
+++ b/config/cms.php
@@ -365,7 +365,7 @@ return [
     | Twig Strict Variables
     |--------------------------------------------------------------------------
     |
-    | If strict_variables is disabled, Twig will silently ignore invalid 
+    | If strict_variables is disabled, Twig will silently ignore invalid
     | variables (variables and or attributes/methods that do not exist) and
     | replace them with a null value. When enabled, Twig throws an exception
     | instead. If set to null, it is enabled when debug mode (app.debug) is
--- a/config/cms.php
+++ b/config/cms.php
@@ -100,7 +100,7 @@ return [
     |
     | If value is true, UpdateMananger will be run on logging in to the backend.
     | It's recommended to set this value to 'false' in production enviroments
-    | because it cleares cache on every login in backend.
+    | because it clears the cache every time a user logs in to the backend.
     |
     */
 
--- a/config/cms.php
+++ b/config/cms.php
@@ -95,6 +95,19 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Determines if logging in backend should run UpdateManager
+    |--------------------------------------------------------------------------
+    |
+    | If value is true it runs UpdateMananger after login in backend.
+    | It's recommended to set this value to 'false' in production enviroments
+    | because it cleares cache on every login in backend.
+    |
+    */
+
+    'runUpdateManagerAfterBackendLogin' => true,
+
+    /*
+    |--------------------------------------------------------------------------
     | Determines which modules to load
     |--------------------------------------------------------------------------
     |
--- a/config/cms.php
+++ b/config/cms.php
@@ -99,8 +99,9 @@ return [
     |--------------------------------------------------------------------------
     |
     | If value is true, UpdateMananger will be run on logging in to the backend.
-    | It's recommended to set this value to 'false' in production enviroments
+    | It's recommended to set this value to 'null' in production enviroments
     | because it clears the cache every time a user logs in to the backend.
+    | If set to null, this setting is enabled when debug mode (app.debug) is enabled.
     |
     */
 
--- a/config/cms.php
+++ b/config/cms.php
@@ -359,7 +359,7 @@ return [
     */
 
     'forceBytecodeInvalidation' => true,
-    
+
     /*
     |--------------------------------------------------------------------------
     | Twig Strict Variables
@@ -372,7 +372,7 @@ return [
     | enabled.
     |
     */
-    
+
     'enableTwigStrictVariables' => false,
 
     /*
--- a/config/cms.php
+++ b/config/cms.php
@@ -101,7 +101,8 @@ return [
     | If value is true, UpdateMananger will be run on logging in to the backend.
     | It's recommended to set this value to 'null' in production enviroments
     | because it clears the cache every time a user logs in to the backend.
-    | If set to null, this setting is enabled when debug mode (app.debug) is enabled.
+    | If set to null, this setting is enabled when debug mode (app.debug) is enabled
+    | and disabled when debug mode is disabled.
     |
     */
 
--- a/config/cms.php
+++ b/config/cms.php
@@ -95,7 +95,7 @@ return [
 
     /*
     |--------------------------------------------------------------------------
-    | Determines if logging in backend should run UpdateManager
+    | Automatically run migrations on login
     |--------------------------------------------------------------------------
     |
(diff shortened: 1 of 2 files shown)


Samuell1:
Another issue with flushing cache on login is when cache is flushed but OPcache isnt its starts doing weird errors of missing files in cache system.

`ErrorException: file_get_contents(/var/www/html/storage/cms/cache/da/34/default.htm.php): failed to open stream: No such file or directory in /var/www/html/modules/cms/classes/CodeParser.php:298`
and
`Symfony\Component\Debug\Exception\FatalThrowableError: Class 'Cms5cc4c983a111d838254349_89df23eac16847138d745dd0ae254294Class' not found in /var/www/html/modules/cms/classes/CodeParser.php:181`

Samuell1:
Any news on this? @daftspunk @LukeTowers

config/cms.php
```diff
@@ -93,6 +93,19 @@
 
     'backendSkin' => 'Backend\Skins\Standard',
 
+    /*
+    |--------------------------------------------------------------------------
+    | Determines if logging in backend should run UpdateManager
```
bennothommo: @Samuell1 Might be better to say "Automatically check for plugin updates on login".

config/cms.php
```diff
@@ -93,6 +93,19 @@
...
+    |--------------------------------------------------------------------------
+    |
+    | If value is true it runs UpdateMananger after login in backend.
+    | It's recommended to set this value to 'false' in production enviroments
+    | because it cleares cache on every login in backend.
+    |
+    */
+
+    'runUpdateManagerAfterBackendLogin' => true,
```
bennothommo: @Samuell1 That might be an unnecessarily large setting name. Could it be shrunk to something like `updatePluginsOnLogin`?
```suggestion
    'updatePluginsOnLogin' => true,
```

config/cms.php
```diff
@@ -93,6 +93,19 @@
 
     'backendSkin' => 'Backend\Skins\Standard',
 
+    /*
+    |--------------------------------------------------------------------------
+    | Determines if logging in backend should run UpdateManager
+    |--------------------------------------------------------------------------
+    |
+    | If value is true it runs UpdateMananger after login in backend.
```
bennothommo: @Samuell1 Grammar fix.
```suggestion
    | If value is true, UpdateMananger will be run on logging in to the backend.
```

config/cms.php
```diff
@@ -93,6 +93,19 @@
...
+    | It's recommended to set this value to 'false' in production enviroments
+    | because it cleares cache on every login in backend.
```
bennothommo: @Samuell1 Grammar fix.
```suggestion
    | because it clears the cache every time a user logs in to the backend.
```

modules/backend/controllers/Auth.php
```diff
@@ -100,12 +100,14 @@ public function signin_onSubmit()
             'password' => post('password')
         ], $remember);
 
-        try {
-            // Load version updates
-            UpdateManager::instance()->update();
-        }
-        catch (Exception $ex) {
-            Flash::error($ex->getMessage());
+        if (config('cms.runUpdateManagerAfterBackendLogin', true)) {
```
bennothommo: @Samuell1 If you agree with the config name change above, you'll need to change it here too.
```suggestion
        if (config('cms.updatePluginsOnLogin', true)) {
```

config/cms.php
Samuell1: @bennothommo I think it will be good to name it like that because updatemanager doesnt only updates plugins but everything.

config/cms.php
Samuell1: @bennothommo It doesnt automatically check for plugin updates, its just installs latest migrations that were not executed before.

bennothommo: @Samuell1 Fair enough. What about "Automatically run migrations on login"?

I would prefer if it said what UpdateManager does as opposed to just saying "run UpdateManager".

config/cms.php
bennothommo: @Samuell1 What about `runMigrationsOnLogin`?

modules/backend/controllers/Auth.php
```diff
@@ -100,12 +100,14 @@ public function signin_onSubmit()
             'password' => post('password')
         ], $remember);
 
-        try {
-            // Load version updates
-            UpdateManager::instance()->update();
-        }
-        catch (Exception $ex) {
-            Flash::error($ex->getMessage());
+        if (config('cms.runMigrationsOnLogin', true)) {
```
LukeTowers: I would prefer that you used the Config facade here.

modules/backend/controllers/Auth.php
```diff
@@ -100,12 +100,14 @@ public function signin_onSubmit()
...
+            try {
+                // Load version updates
+                UpdateManager::instance()->update();
+            }
+            catch (Exception $ex) {
```
LukeTowers: this should be on the same line as the closing brace for the try

config/cms.php
```diff
@@ -93,6 +93,19 @@
...
+    | Automatically run migrations on login
+    |--------------------------------------------------------------------------
+    |
+    | If value is true, UpdateMananger will be run on logging in to the backend.
+    | It's recommended to set this value to 'false' in production enviroments
+    | because it clears the cache every time a user logs in to the backend.
+    |
+    */
+
+    'runMigrationsOnLogin' => true,
```
LukeTowers: @bennothommo I wonder if we should have this set to null by default to represent pulling its value from the value of app.debug (debug enabled this enabled, debug disabled this disabled). Thoughts? Are users still able to have migrations run when installing new plugins or plugin updates through the backend when debug (i.e. this) is disabled in that case?

modules/backend/controllers/Auth.php
Samuell1: @LukeTowers changed, and changed for another too :)

modules/backend/controllers/Auth.php
```diff
@@ -90,8 +90,8 @@ public function signin_onSubmit()
             throw new ValidationException($validation);
         }
 
-        if (($remember = config('cms.backendForceRemember', true)) === null) {
-            $remember = (bool) post('remember');
+        if (($remember = Config::get('cms.backendForceRemember', true)) === null) {
+            $remember = (bool)post('remember');
```
LukeTowers: revert the space after type cast change please

config/cms.php
bennothommo: @LukeTowers Users can still use the *Check for updates* button in the *Plugins & Updates* area of the Settings to run migrations, so I agree this would be a good change to have. Migrations shouldn't automatically run on production.

LukeTowers: @bennothommo @Samuell1 could you make the change?

modules/backend/controllers/Auth.php
LukeTowers: And @Samuell1 can you make the tests pass by fixing the other coding style issues in this file? Right now it's just the one liner conditional in filterList() that needs to be addressed AFAIK

modules/backend/controllers/Auth.php
```diff
@@ -100,12 +100,19 @@ public function signin_onSubmit()
             'password' => post('password')
         ], $remember);
 
-        try {
-            // Load version updates
-            UpdateManager::instance()->update();
+
+        $runMigrationsOnLogin = Config::get('cms.runMigrationsOnLogin', null);
+        if ($runMigrationsOnLogin === null) {
```
LukeTowers: use is_null here instead please

config/cms.php
```diff
@@ -93,6 +93,19 @@
...
+    | Automatically run migrations on login
+    |--------------------------------------------------------------------------
+    |
+    | If value is true, UpdateMananger will be run on logging in to the backend.
+    | It's recommended to set this value to 'false' in production enviroments
+    | because it clears the cache every time a user logs in to the backend.
+    |
+    */
+
+    'runMigrationsOnLogin' => null,
```
LukeTowers: Please add a line explaining what the null value means

LukeTowers:
@Samuell1 please also fix the listed style problems: https://travis-ci.org/octobercms/october/jobs/552112097#L793-L796

Samuell1:
@LukeTowers they doesnt exists @bennothommo said i should ignore them

modules/backend/controllers/Auth.php
Samuell1: I used same structure like is used on other "null" config values.

Samuell1:
Any info on this? @LukeTowers

config/cms.php
```diff
@@ -93,6 +93,20 @@
...
 
+    /*
+    |--------------------------------------------------------------------------
+    | Automatically run migrations on login
+    |--------------------------------------------------------------------------
+    |
+    | If value is true, UpdateMananger will be run on logging in to the backend.
+    | It's recommended to set this value to 'null' in production enviroments
+    | because it clears the cache every time a user logs in to the backend.
+    | If set to null, this setting is enabled when debug mode (app.debug) is enabled.
```
LukeTowers: Add another line here "and disabled when debug mode is disabled"

PR #4267, opened 2019-04-17
Updated documentation for AJAX search (Select2)

jexme:
Current documentation is incorrect, because Most likely the type of answer for Select2 has been changed, you can see the documentation here: https://select2.org/data-sources/ajax


LukeTowers:
@jexme could you provide a PR to the [test plugin](https://github.com/octoberrain/oc-test-plugin) to demo how this should work?

bennothommo:
This will not be needed as https://github.com/octobercms/october/pull/4414 shall provide a JS-based workaround for the Select2 data format.

PR #4270, opened 2019-04-18
Add theme config cache

### LINKED ISSUE #4265: Active theme config is not cached
- OctoberCMS Build: 447
- PHP Version: 7.3.3

### Description:
Theme config is not cached and that means its calling Yaml parse on every request, more testing shows that it takes around **20ms** to parse _theme.yaml_.

obrázok

### Steps To Reproduce:
This can be reproduced by using {{ this.theme }} variable in templates.

Samuell1:
Fixes #4265

### DIFF
--- a/modules/cms/classes/Theme.php
+++ b/modules/cms/classes/Theme.php
@@ -465,6 +465,8 @@ class Theme
         $contents = Yaml::render($values);
         File::put($path, $contents);
         $this->configCache = $values;
+
+        self::resetCache();
     }
 
     /**
--- a/modules/cms/classes/Theme.php
+++ b/modules/cms/classes/Theme.php
@@ -338,7 +338,7 @@ class Theme
         }
 
         try {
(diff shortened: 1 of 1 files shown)


modules/cms/classes/Theme.php
```diff
@@ -484,6 +494,7 @@ public static function resetCache()
 
         Cache::forget(self::ACTIVE_KEY);
         Cache::forget(self::EDIT_KEY);
+        Cache::forget(self::CONFIG_KEY);
```
LukeTowers: You'd need to add a call to resetCache in writeConfig() to match the other logic

modules/cms/classes/Theme.php
```diff
@@ -336,7 +337,16 @@ public function getConfig()
             return $this->configCache = [];
         }
 
-        $config = Yaml::parseFile($path);
+        try {
+            $config = Cache::remember(self::CONFIG_KEY, 1440, function() use($path) {
```
LukeTowers: space between `use` and `($path)` please

Issue #4271, opened 2019-04-18
Performance Issue with storm-min.js when loading web pages


ayumihamsaki:
I have noticed this several times where you try to do a "**hard reload**" of a web page in the backend while I am working on building some new plugins, see the waterfall screenshot:

![bug](https://user-images.githubusercontent.com/17784082/56372547-21006080-61f7-11e9-9619-3ba95742fd4b.png)

Above: Here we can see it has taken Chrome 11 seconds to open `storm-min.js` the whole web page has a blank screen waiting for this asset to load before the render can display. This is no good!

It jumps into my mind straight away that a simple fix for preloading the `storm-min.js` would fix this issue:

From the top of my head there are two backend files:

- [auth.htm](https://github.com/octobercms/october/blob/master/modules/backend/layouts/auth.htm)
- [head.htm](https://github.com/octobercms/october/blob/master/modules/backend/layouts/_head.htm)

(I think - need to double check)

Anyway I think to fix the solution at in the <head> section of the backend some preloading of the important asset files. This will speed up October and prevent issues this this happening, e.g.

```html
<link rel="preload" href="/modules/system/assets/ui/storm-min.js" as="script">
```

I think you could preload the other important files:

- storm.min.js
- october-min.js
- framework.js
- storm.css
- october.css
- jquery.min.js
- jquery-migrate.min.js

I believe these are present on every backend web page so will be 100% loaded and 100% preloaded.

PR #4272, opened 2019-04-18
Added Preload and Version Numbers

ayumihamsaki:
Not sure why the `Auth.htm` was missing the version numbers on the loaded assets (I added them into the code). I also `preloaded` all main assets.

This relates to the github issue: https://github.com/octobercms/october/issues/4271

### DIFF
--- a/modules/backend/layouts/_head.htm
+++ b/modules/backend/layouts/_head.htm
@@ -35,11 +35,11 @@
 <link rel="preload" href="<?= Backend::skinAsset('assets/js/october-min.js') ?>?v<?= $coreBuild ?>" as="script">
 <link rel="preload" href="<?= Url::asset('modules/system/assets/js/lang/lang.'.App::getLocale().'.js') ?>?v<?= $coreBuild ?>" as="script">
 
-<script data-cfasync="false" src="<?= Backend::skinAsset('assets/js/october.flyout.js') ?>"></script>
-<script data-cfasync="false" src="<?= Backend::skinAsset('assets/js/october.tabformexpandcontrols.js') ?>"></script>
+<script data-cfasync="false" src="<?= Backend::skinAsset('assets/js/october.flyout.js') ?>?v<?= $coreBuild ?>"></script>
+<script data-cfasync="false" src="<?= Backend::skinAsset('assets/js/october.tabformexpandcontrols.js') ?>?v<?= $coreBuild ?>"></script>
 
-<link rel="preload" href="<?= Backend::skinAsset('assets/js/october.flyout.js') ?>" as="script">
-<link rel="preload" href="<?= Backend::skinAsset('assets/js/october.tabformexpandcontrols.js') ?>" as="script">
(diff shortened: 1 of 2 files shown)


LukeTowers:
Could you add some performance numbers in here and explain how preload works / browser support for it?

ayumihamsaki:
## Browser Support

- IE11 No
- Edge 17+
- Firefox 56+
- Chrome 50+
- Safari 11+
- Opera 37+
- Samsung Internet 5+

## What is Preload

Preload (spec) is a new web standard aimed at improving performance and providing more granular loading control to web developers. It gives developers the ability to define custom loading logic without suffering the performance penalty that script-based resource loaders incur.

## Performance (with and without Preload with a Waterfall Chart)

1_PSMeFcC3AXDUmdNf5l19Ug

### Without (Preload)

In the above image the assets are loaded when the Browser finds them.

### With (Preload)

In the above image the assets are loaded straight away by the Browser (reduced wait time)

## October CMS

With regards to October CMS the issue is that the `storm.min.js` file needs to be loaded before the rest of the assets and if it gets delayed then the browser can not load the rest of the file assets. When you add preload the assets are loaded before the rendering of the browser begins - hence the name **preload** (the assets are preloaded before the rendering begins).

---

Hope that explains and answers your question.

LukeTowers:
I don't recognize `as="style"` as valid, what's the browser support for that? Also, does every single script need to be loaded in that way?

ayumihamsaki:
> I don't recognize `as="style"` as valid, 

That is part of the Preload W3c spec see here for the spec details: https://www.w3.org/TR/preload/#as-attribute

It's a valid HTML attribute.

> what's the browser support for that?

That attribute will work fine in all browsers that support and don't support the `as` attribute. I know because I asked W3c that exact question in the past. So for example in IE11 that doesn't support that attribute, there won't be any errors and the browser just doesn't process the `link-rel preload` code.

> Also, does every single script need to be loaded in that way?

No, you should only preload the file assets that you are 100% sure they are going to be loaded when the user opens the web page.

For example, let's say I have a javascript that will run when the user scrolls down the web page. That shouldn't be preloaded and will show up in the console.log as a warning message saying:

`the following script was preloaded but not used within 1 second.`

Basically **I have preloaded only the assets that I know 100% will be loaded at the start of the browser render**.

There is a balance between preloading things and I have been using the preload spec for over 3 years and know which things need to be preloaded and which things will give a negative performance.

---

Happy to answer anymore questions on it. I'm almost an expert on that api, so can answer most questions on it.

LukeTowers:
Thanks @ayumihamsaki.

I think I know why:
> Not sure why the Auth.htm was missing the version numbers on the loaded assets (I added them into the code)

If you have the version number in that then anyone would be able to visit the login page for an OctoberCMS website and determine the exact build number that was used. It used to be considered best practice to obscure the version / software you were using for security however that's increasingly pointless with automated scanning.

PR #4276, opened 2019-04-19
Finish theme:sync command

bennothommo:
Finishes the `theme:sync` command as per @LukeTowers changes in #3908.


modules/cms/ServiceProvider.php
```diff
@@ -29,6 +29,7 @@ public function register()
         $this->registerComponents();
         $this->registerThemeLogging();
         $this->registerCombinerEvents();
+        $this->registerDatabaseLayerModels();
```
LukeTowers: Can you rename this to `registerHalcyonModels` @bennothommo ?

PR #4279, opened 2019-04-20
Update Dropzone V4.0.1 to V5.5.1

ayumihamsaki:
This is a major update from V4 to V5 and will need some testing to make sure October CMS works ok with this update.

ref: https://github.com/octobercms/october/issues/3867

Request some testers on this PR please.

### DIFF
--- a/modules/backend/assets/vendor/dropzone/dropzone.js
+++ b/modules/backend/assets/vendor/dropzone/dropzone.js
@@ -1,3 +1,16 @@
+/**
+ * DropZone V5.5.1 (non-minified) for testing on October CMS
+ */
+
+"use strict";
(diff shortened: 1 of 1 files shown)


ayumihamsaki:
Patch for testing can be found here: https://github.com/ayumihamsaki/patches/blob/master/dropzone/0001-Update-Dropzone-V4.0.1-to-V5.5.1.patch

w20k:
@ayumihamsaki found an easy way to create a patch from PR 😄Just add .patch to any PR, and here you go Git patch file.
Like this: `https://patch-diff.githubusercontent.com/raw/octobercms/october/pull/4279.patch`

PR #4280, opened 2019-04-20
Move autoloading to index to make it same with laravel 5.5

Samuell1:
Moving autoloader to index to make it equal with laravel: https://github.com/laravel/laravel/blob/5.5/public/index.php 

And it removes useless check for *compiledPath* file because it was removed in laravel 5.5.


Samuell1:
Closing needs to be updated with Laravel 6.0

bennothommo:
This has been partially implemented in https://github.com/octobercms/october/pull/4893. We've retained the `bootstrap/autoload.php` file for backwards compatibility.

PR #4038, opened 2019-01-05
Fix Sidebar and File Column bar in the code editor section

ayumihamsaki2:
See github Issue: #4005

Issue #4005, opened 2018-12-22
Fix Sidebar and File Column bar in the code editor section


ayumihamsaki:
I was testing this on a pr I am working on and notice that this is also broken on a normal cms version, see screenshot:

![untitled](https://user-images.githubusercontent.com/17784082/50375453-661fd780-05f5-11e9-8e06-14d3806ae401.png)

Therefore  I will open a new issue and this needs to get fixed separately.

This needs to be fixed before I can continue with the pr here: https://github.com/octobercms/october/pull/3974

It's being injected with this code:

`<div class="layout-cell w-350 hide-on-small" id="layout-side-panel" data-control="layout-sidepanel" style="left: 120px;top: 62px;height: 880px;">`

The `left:120px` needs to have a different value with the css media query under `991px`.

```css
@media (max-width: 991px) and (min-width: 768px)
.layout-sidenav-container {
    width: 100px;
}

.layout-sidenav-container {
    width: 120px;
}

@media (max-width: 767px)
.layout-sidenav-container {
    width: 80px;
}
```

Also I think under 768px the design of October in the sidebar is very poor! I can't even read the small text. Please note Google Developers quote that you should use 16px font in mobile devices! Please see the sidebar screenshot under 768px and try to read the font size:

![untitled](https://user-images.githubusercontent.com/17784082/50375568-2d80fd80-05f7-11e9-991d-c169b3a59544.png)

As you can see the sidebar font size is blurry to read because it is too small. That screenshot is on a computer, using a mobile device the dpi will make it even worse.

I would also suggest October using a default 16px (1rem) and also updating it's system to start using responsive fonts.

## Easy Solution

Having a quick think on this why not just remove this code:

```css
// Delete all this
@media (max-width: 991px) and (min-width: 768px)
.layout-sidenav-container {
    width: 100px;
}

@media (max-width: 767px)
.layout-sidenav-container {
    width: 80px;
}
```

There is no good reason why you need to reduce the sidebar and having a sidebar of 120px in width will not be problematic on a mobile device!

Then remove the font-size css which is getting smaller on smaller screen resolutions.

```css
// Delete all this
@media (max-width: 767px) {
  #layout-sidenav {
    font-size: 10px;
  }
}
@media (max-width: 991px) and (min-width: 768px) {
  #layout-sidenav {
    font-size: 12px;
  }
}
```
and

```css
// Remove this line
#layout-sidenav ul li a {
    padding: 1.429em .714em;
    display: block;
[...]

PR #4283, opened 2019-04-20
Fix Sidebar and File Column bar in the code editor section

ayumihamasaki2:
re-complied from pr https://github.com/octobercms/october/pull/4038

### DIFF
--- a/modules/backend/assets/css/october.css
+++ b/modules/backend/assets/css/october.css
@@ -667,7 +667,7 @@ nav#layout-mainmenu ul li a:focus {text-decoration:none;color:rgba(255,255,255,0
 nav#layout-mainmenu ul li a i {line-height:1;font-size:30px;vertical-align:middle}
 nav#layout-mainmenu ul li a img.svg-icon {height:30px;width:30px;margin-right:10px;position:relative;top:0}
 nav#layout-mainmenu ul.nav {display:inline-block}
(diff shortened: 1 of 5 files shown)


ayumihamsaki:
See github Issue: #4005

LukeTowers:
@ayumihamsaki you need to pull in the latest dev branch and compile the CSS again. Your current version is missing the no_icons navbar mode.

ayumihamsaki:
Quick question, what compiler do you use ? Are there any Gulp files I can use ?

w20k:
Don't think there is something like that right now @ayumihamsaki .

ayumihamsaki:
Would it be a good idea to add one ? Something for the `compiler` and `testing`. Just create an extra folder in this repo and store them in and then it might make things easier with regards to this subject and also standardize things.

w20k:
I don't think that util.gulp for this single use case (though you've got a point) is a good idea.
It would be better to create a proper builder process for JS files, and I'm not a big fan of gulp, If you ask me 😄

LukeTowers:
@ayumihamsaki the built in OctoberCMS asset compiler: `php artisan october:util compile less`

ayumihamasaki2:
Need to close this Github is refusing for me to push anything? Will open a new pr.

ayumihamsaki:
moved to here: https://github.com/octobercms/october/pull/4293

Issue #4277, opened 2019-04-19
Enhancement Idea: Adding Priority Hints API to October CMS


ghost:
## Introduction

I am about to start work on adding a new API to my October CMS websites. The API has been in the making for around 2 years now by Google and now can be used in a "production" website environment.

## About the API

When the browser loads assets they are given a Priority for the loading, see below an example:

![1](https://user-images.githubusercontent.com/17784082/56428699-2d98bd80-62b8-11e9-83f6-d9626e438930.png)

A developer can add an attribute to fine tune the loading of these assets to enhance the performance of the loading of the web page.

The API is made up of the following attributes:

The importance attribute accepts one of three values:

- **high**: The resource may be prioritized, if the browser's own heuristics don't prevent that from happening.
- **low**: The resource may be _de_prioritized, if the browser's heuristics permit.
- **auto**: Let the browser decide what priority is appropriate for a resource. This is the default value.

## Why I am filing an Issue here

The following places I have assets in October CMS are in three main locations:

1. Inside the `<head>` section.
2. Near the bottom of the web page code after the `<footer>` section.
3. Adding the October CMS AJAX code snippet (again at the bottom).

This gives us the following code example:

```html
<!doctype html>
<html>
<head>
<meta charset="utf-8">
<title>Example</title>

<!-- Position 1 -->
<script src="example-1.js"></script>
</head>
<body>

<!-- Position 2 -->
<script async src="example-2.js"></script>

<!-- Position 3 -->
{% framework extras %}
{% scripts %}  
</body>
</html>
```

For Positions 1 and 2 this is down to the developer and will be different on every website and that side of things is not part of this issue.

For Position 3 (the October CMS AJAX), this is the thing I want to discuss in this issue.

## Adding the new attributes to the October CMS AJAX by default

I want to add the new attributes to the October CMS AJAX by default.

An example of the new API would look like this:

```html
<script src="/js/app.js" defer importance="high"></script>
```

Please see the new attribute called `importance` which is set to `high`.

## Discussion

I would like to know first, what would be the best setting to give the October CMS AJAX, `low`, `high` or `auto` ?

## Creating a PR

When the discussion is over, I would like to inject this new attribute into the code snippet:

```html
{% framework extras %}
{% scripts %}  
```

Then October CMS will automatically add the new attributes into the code and the develop[er won't need to do anything!

## Another Github Issue (Ongoing)

[...]

Issue #4030, opened 2019-01-03
Speeding up the loading of web pages using October's AJAX by using rel-preload


ghost:
To speed up October CMS web pages that use October's AJAX code:

```html
{% framework extras %}
{% scripts %}  
```

A developer could add link `rel-preload` which **must be loaded inside the `<head>` tag** to speed up the loading of the web page(s) that use October AJAX by reducing the wait times for the injected files. Below shows the following method:

## Method to Speed up October's AJAX web pages

```html
<!doctype html>
<html>
<head>
<meta charset="utf-8">
<title>Correct Method 1</title>

<link rel="preload" href="https://example.com/modules/system/assets/css/framework.extras-min.css" as="style">

<link rel="preload" href="https://example.com/modules/system/assets/js/framework.combined-min.js" as="script">

</head>

<body>

..

{% framework extras %}
{% scripts %} 

</body>
</html>
```

which automatically turns into into:

```html
<!doctype html>
<html>
<head>
<meta charset="utf-8">
<title>Correct Method 2</title>

<link rel="preload" href="https://example.com/modules/system/assets/css/framework.extras-min.css" as="style">

<link rel="preload" href="https://example.com/modules/system/assets/js/framework.combined-min.js" as="script">

</head>

<body>

..

<script src="/modules/system/assets/js/framework.combined-min.js"></script>
<link rel="stylesheet" property="stylesheet" href="/modules/system/assets/css/framework.extras-min.css">

</body>
</html>
```

After talking to Luke, he suggests creating a PR and injecting the extra `rel-preload` code lines into the `<head>` section automatically.

Please note Luke said:

> `{% framework %}` is typically supposed to be the head anyways

However I know lots of developers actually adding `{% framework extras %}` and `{% scripts %}` to the bottom of the web page. So needs some extra discussion about that.

The `rel-preload` link tags must be added into the `<head>` section only.

Not sure how to code the pr for this right now, but for now have opened up the github issue for discussions.

ayumihamsaki:
Creating:

```html
{% framework extras preload %}
{% scripts preload %}
```  

Is a good idea, agree does place the burden on the developer.

Another idea is directly injecting the dom:

```html
<!doctype html>
<html>
<head>
<meta charset="utf-8">
<title>Test</title>
</head>

<body>

<script>
document.getElementsByTagName("head")[0].innerHTML += "<link rel=\"preload\" href=\"/modules/system/assets/css/framework.extras-min.css\" as=\"style\">";
[...]

PR #4285, opened 2019-04-21
Add new Performance API's to October CMS

ayumihamsaki:
This PR adds the new Priority Hints API, Preload API and async to the October CMS AJAX injected code and adds the new Priority Hints API to the loaded assets in the head section.

This is related to the github issues: https://github.com/octobercms/october/issues/4277 and https://github.com/octobercms/october/issues/4030

## Explanation

As performance becomes increasingly important, it's exciting to see browsers implement new features which give developers more control over resource loading. Resource Hints such as `rel=preload` and `rel=preconnect` give developers more control over resource loading and connections to cross-origin servers, respectively. `Client Hints` expose details of a user's device and preferences that developers can use to improve performance in nuanced ways. Continuing in this vein, a new api feature known as `Priority Hints`.

When a browser downloads a resource, the resource is assigned a priority. By default, priorities depend on the type of resource (e.g., script, image, etc.), and the location of the resource reference in the document. For example in Chrome, CSS loaded in typical fashion via the `<link>` element in the `<head>` will be assigned a priority of **highest**, as it blocks rendering. Images in the viewport may be assigned a priority of **high**, whereas images outside the viewport may be assigned a priority of **low**. A `<script>` loaded at the end of the document may receive a priority assignment of **medium** or **low**, but this can be influenced by defer and async.

## Priority Hints API

The API is rather simple and was created by Google. It features an attribute called: `importance` which you can assign either a `high`, `low` or `auto` value to it.

- **high** - The developer considers the resource as being high priority.
- **low** - The developer considers the resource as being low priority.
- **auto** - The developer does not indicate a preference. This also serves as the attribute's invalid value default and missing value default.

In this PR I have given the important assets in the `<head>` section a `importance="high"` and the October AJAX code a `importance="low"`. This is because the important assets are needed to render the web page and the assets for October CMS AJAX code can be loaded after the render of the web page. This increases the load performance of the web page in October CMS.

## Useful Links

Priority Hints API Spec: https://wicg.github.io/priority-hints/


ayumihamsaki:
Note: There are conflicts here because I have taken the code from this PR https://github.com/octobercms/october/pull/4285 (which hasn't fully been updated yet).

ayumihamsaki:
**Note**: There are conflicts here because I have taken the code from this PR #4285 (which hasn't fully been updated yet).

ayumihamsaki:
**Note**: There are conflicts here because I have taken the code from this PR https://github.com/octobercms/october/pull/4272 (which hasn't fully been updated yet).

LukeTowers:
@ayumihamsaki please resolve conflicts to reflect my latest changes to those files.

ayumihamsaki:
@LukeTowers Thanks, I didn't release you update the code, I will fix this pr accordingly.

I got another quick question, see the following screenshot:

qs

Notice the following filenames:

```
auth.htm
default.htm
form-with-sidebar.htm
```

How come they don't have an underscore in front of them? To be like this:

```
_auth.htm
_default.htm
_form-with-sidebar.htm
```

Then all the filenames will match the same style.

I know this is a minor thing, I'm just use to working with ISO level standard things where everything matches and we have to add `continually` to everything on purpose.

ayumihamsaki:
@LukeTowers Please review my updated commit, I think this resolves the conflicts now ?

ayumihamsaki:
I've had a think on this and wondered if it would also be a good idea to add the `client hints` API as well at the same time.

My personal setup I use the following:

```
accept-ch: Device-Memory, DPR, Width, Viewport-Width
accept-ch-lifetime: 3600
```

These are HTTP Headers by the way, for more details see:

https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/client-hints/
https://developers.google.com/web/updates/2015/09/automating-resource-selection-with-client-hints

Spec: https://httpwg.org/http-extensions/client-hints.html

If it's ok I will add these headers into the two files:

```
modules/backend/layouts/_head.htm
modules/backend/layouts/auth.htm
```

w20k:
@ayumihamsaki I do think `hints` are a good idea to add, but not in this particular use case. 
It's mainly a website specific headers, which should be added manually and not automated.

ayumihamsaki:
@w20k When you say `hints` which one do you mean? There are two different API's here: `Client Hints` and `Priority Hints`. Or maybe you are referring to both. Not sure what you mean also by:

> It's mainly a website specific headers, which should be added manually and not automated.

Are you talking about the `Client Hints` HTTP Header in this case?

w20k:
I was talking about client headers hints, where priority hints is a really  good addition ;)

ayumihamsaki:
Yah that's what I thought. So you want me to remove the `Client Hints` HTTP Header? The reason for adding it is because I have seen some people's plugins that would benefit from the header. Such as a plugin that creates a gallery list of images. Let me know if you want me to delete it out of the code or not? I added it for people's plugins basically.

ayumihamsaki:
@w20k Done if you want to re-view this now, think ready for merge/review process now. Thanks.

w20k:
Done with testing! Might not able to merge it today, but will do it on weekends! Thanks, @ayumihamsaki!

Screenshots:
Screenshot 2019-06-14 at 16 20 51
Screenshot 2019-06-14 at 16 24 54

ayumihamsaki:
@w20k Thanks 🥇

modules/cms/twig/FrameworkNode.php
```diff
@@ -35,25 +35,35 @@ public function compile(TwigCompiler $compiler)
             $compiler
                 ->write("if (\$_minify) {" . PHP_EOL)
                 ->indent()
-                    ->write("echo '<script src=\"'. Request::getBasePath()
-                    .'/modules/system/assets/js/framework.combined-min.js\"></script>'.PHP_EOL;" . PHP_EOL)
+                    ->write("echo '<script async src=\"'. Request::getBasePath()
+                    .'/modules/system/assets/js/framework.combined-min.js\" importance=\"low\"></script>'.PHP_EOL;" . PHP_EOL)
+                    ->write("echo '<link rel=\"preload\" href=\"'. Request::getBasePath()
```
LukeTowers: Why were these added? Many instances of October load the framework assets at the end of the document where <link> would be invalid. Also, how does it have any effect at all appearing immediately after the asset is requested in the first place?

ayumihamsaki: I've tested this and the browser does preload them, have you tested it?

ayumihamsaki:
See Proof:

Untitled

LukeTowers:
Weird, I always thought that <link> had to be in the head. Rereading https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link it appears that's changed.

LukeTowers:
Thanks for the clarification @ayumihamsaki, looks good to me.

ayumihamsaki:
No problem.

I'm going to add some extra info for future reference just in case someone asks this question as well.

With regards to the new performance API the following should be applied (as I done in this PR).

The new performance API now gives us a two level preload rendering, so we get the following:

## Sync Javascript Example

I put the following code in the `<head>` 

```html
<link rel="preload" href="main.js" as="script" importance="high">
<script src="main.js" importance="high"></script>
```

## Async Javascript Example

I put the following code at the bottom of the web page just above the closing body tag

```html
<link rel="preload" href="main.js" as="script" importance="low">
<script async src="main.js" importance="low"></script>
```

Note: This may confuse people when looking at a timing diagram, but the browser will put both the two examples at the top of the waterfall. The Browser will then render preload with the `importance="high"` first and then the browser will render the preload with importance="low"` later.

Hope that helps people to understand.

w20k:
@ayumihamsaki thanks for a `backup` ;) Had a short weekend/vacation)!

ayumihamsaki:
@w20k Nice, hope you went somewhere good!

acasar:
This PR seems to break backwards compatibility.

I am including some custom scripts after the ``{% framework extras %}`` call. It used to work fine (framework scripts were loaded before my custom scripts) but after this PR i'm getting an error:

image

To provide more details - this was the situation before this PR:

image

And this is the situation after the PR, which causes the error above:

image

The fix is of course easy enough, but shouldn't we make sure that existing applications aren't broken? Maybe we should provide a different twig call to include Performance API parameters?

acasar:
Also, due to a mix up in tagging, this PR has already made it to stable release. See more details here: https://github.com/octobercms/october/issues/4465

ayumihamsaki:
Hi, I can't reproduce your error. I did the following:

Test 1:

```
{% framework extras %}
{% scripts %}
        
{# Scripts #}
<script async src="https://www.google-analytics.com/analytics.js" importance="low"></script>
<script async src="{{ 'assets/js/bottom.js'|theme }}" importance="low"></script>
```

Test 2:

```
{# Scripts #}
<script async src="https://www.google-analytics.com/analytics.js" importance="low"></script>
<script async src="{{ 'assets/js/bottom.js'|theme }}" importance="low"></script>

{% framework extras %}
{% scripts %}
```
        
Maybe someone selse can reproduce it? I wait for my more feedback.

ayumihamsaki:
Hi, I can't reproduce your error. I did the following:

Test 1:

```
{% framework extras %}
{% scripts %}
        
{# Scripts #}
<script async src="https://www.google-analytics.com/analytics.js" importance="low"></script>
<script async src="{{ 'assets/js/bottom.js'|theme }}" importance="low"></script>
```

Test 2:

```
{# Scripts #}
<script async src="https://www.google-analytics.com/analytics.js" importance="low"></script>
<script async src="{{ 'assets/js/bottom.js'|theme }}" importance="low"></script>

{% framework extras %}
{% scripts %}
```
        
Maybe someone else can reproduce it? I wait for my more feedback.

This is probably a stupid question, but do you have a conflict in your javascript instead. Because the preload function is loading that code in a different order now.

acasar:
I think the key is to try without ``async`` and ``importance="low"`` attributes on your custom scripts.

When using the attributes it of course works, but most existing October applications don't use them, that's when the issues happen.

ayumihamsaki:
Your code order from the screenshots you have given is like this:

Before your code is loading like this:

```
combine.js
framework.combined-min.js
framework.extras.min.css
..
```

Now after preloading the order is like this:

```
framework.combined-min.js
framework.extras-min.css
combine.js
..
```

ayumihamsaki:
Your code order from the screenshots you have given is like this:

Before your code is loading like this:

```
combine.js
framework.combined-min.js
framework.extras.min.css
..
```

Now after preloading the order is like this:

```
framework.combined-min.js
framework.extras-min.css
combine.js
..
```

Try preloading your `combine.js` file.

stretch

Let me know if this works or not?

acasar:
@ayumihamsaki the issue is in frontend.js, not combine.js , but I already said that it's an easy thing to fix and it's not the reason I posted my comment.

The main issue is that it breaks backwards compatibility. Usually just updating October to the latest release shouldn't require any code changes, but after this goes live, most applications that include custom javascript without any preloading attributes will need to update their code.

So to sum up -  I understand this change and I'm not against it, but we should be able to make this opt-in and not applied by default since it clearly breaks existing applications.

ayumihamsaki:
@acasar understand, sorry we should of added a comment in the update release about this breaking change.

acasar:
In my opinion this is to big of a change to simply solve it with a release note. At my company alone we'd need to update about 90 websites, but I believe thousands of websites will be affected overall as well as most of the themes available in the October marketplace.

My suggestion is that we leave ``{% framework extras %}`` intact so that it works as it always did and introduce a new tag (e.g. ``{% framework optimized-extras %}``) to include performance related attributes.

ayumihamsaki:
I know Luke won't go for a new tag called: `{% framework optimized-extras %}` we discussed this in the past.

But don't get me wrong, every pr I submit I always want to cater for backwards compatibility, will wait for feedback from one of the admins.

Samuell1:
Maybe better will be parameter, something like: `{% framework extras preload %}`?

@ayumihamsaki Btw why scripts in backend are loaded in head not at end of body to prevent it to block render?

LukeTowers:
@ayumihamsaki If you could make a PR to revert the change to `{% framework extras %}` to require an extra `async` (`{% framework extras async %}`) also do it for `{% framework %}` and `{% jquery %}` so we can roll it into 458 to make the builds stable that would be appreciated. @acasar is correct, this shouldn't have made it in by default, I didn't realize it would break sites in that way.

LukeTowers:
@bennothommo @daftspunk This is going to require fixing before we can make 457 stable.

ayumihamsaki:
@LukeTowers 

I think,

{% framework extras preload %}

is better than

{% framework extras async %}

The async issue is because of the browser version, using a combination of `async` and `importance="low"` and so it's behind a flag at the moment for older browsers - sorry I forgot to mention that part.

For newer browsers which don't need to turn on the flag the issue would be a `preload` order. So I agree with @Samuell1 with the naming.

LukeTowers:
@ayumihamsaki I don't, preload implies that you're going to make it load before the rest of the scripts which is exactly the opposite of the problem here. The problem is that the scripts no longer load in the same order that was previously expected because they are being loaded asynchronously. I get that the `preload` attribute may be the more important for performance but the `aysnc` attribute is what causes the issue so I think that's what we should emphasize.

ayumihamsaki:
@LukeTowers you're not correct, you are looking at the code from a normal async view and not using the importance API. It will be like this:

```html
<script async src=/modules/system/assets/js/framework.combined-min.js importance="low"></script>
<link preload href=/modules/system/assets/js/framework.combined-min.js as="script" importance="low">
```

### Importance API turned on (flag on) - lastest Google Canary:

The files are loaded `sync` not `async` due to combination of the `importance` API and `preload`.

### Importance API turned off (flag off) - stable Google Chrome:

The files are loaded `async` due to the importance API not turned on.

LukeTowers:
@ayumihamsaki fair enough. We can use `preload` then but in the docs we should make note of the fact that `preload` will load assets `async` in older browsers so that people are aware of the impacts of using it.

Samuell1:
@ayumihamsaki Btw why this change was made to "front end" twig parser (modules/cms/twig/FrameworkNode.php) ? Isnt for now best to remove it and developer should choose how he wants to import framework?

ayumihamsaki:
Yah good idea. Sorry this is totally my fault for using Google Canary and not thinking about the experiments flag turned on.

ayumihamsaki:
@LukeTowers sorry to keep pinning you, I just had a chat with Addy Osmani one of the importance API authors, he was saying that right now different browsers are doing different things with regards to `async` and `prioritized`.

He suggested the best idea woul;d be like this:

```
{% framework extras optimized %}
```

Then explain in the doc's about the preloading and async/sync actions by the browsers - modern and backwards compatibility.

ayumihamsaki:
I have created a PR fix here:

https://github.com/octobercms/october/pull/4463/commits/00f688e6b7ce95eea26c9875a888a57721761a2d

We now have the following three options:

```
{% framework %}
{% framework extras %}
{% framework extras_optimized %}
```

Very sorry to @acasar for the issue!

ayumihamsaki:
I have created a PR fix here: https://github.com/octobercms/october/pull/4463/commits/00f688e6b7ce95eea26c9875a888a57721761a2d

We now have the following three options:

```
{% framework %}
{% framework extras %}
{% framework extras_optimized %}
```

Very sorry to @acasar for the issue!

acasar:
@ayumihamsaki no problem at all :)

Thanks to everyone for looking into this so quickly 🥇

LukeTowers:
@ayumihamsaki why is it `_optimized` instead of a separate option entirely? I was thinking a separate option so that `{% framework optimized %}` and `{% jquery optimized %}` and `{% framework extras optimized %}` would all be options.

ayumihamsaki:
@LukeTowers it gave me an error when testing.

```
{% framework extras optimized %} = Error
{% framework extras-optimized %} = Error
{% framework extras_optimized %} = Passed
```

I can create the following tomorrow:

```
{% framework %}
{% framework_optimized %}
{% framework extras %}
{% framework extras_optimized %}
```

p.s. What do you mean by `{% jquery optimized %}`, are you thinking so just creating a minified version ? If so why not just minify all the files by default.

Let me know I will update the pr later.

bennothommo:
@ayumihamsaki I personally am not a fan of the underscores - what sort of error are you getting with `extras optimized`?

w20k:
Thanks for feedback @acasar! 
My bad, need to include more front-end tests for my test routine!

Samuell1:
@bennothommo Its returning some weird twig error `Unexpected token "name" of value "optimized" ("end of statement block" expected).`

ayumihamsaki:
Sorry late reply, had to get some sleep. yah exactly what Samuell1 said is what I was seeing. That's why I went for the underscore as computers read `extras_optimized` and `extrasoptimized` the same.

ayumihamsaki:
Adding extra security, I plan to add Subresource Integrity to the loaded jQuery files via the tags.

ayumihamsaki:
Adding extra security, I plan to add Subresource Integrity to the loaded jQuery files via the tags.

There is a small catch and I will do the following to not mess things up!

I will add the current version of `jQuery v3.3.1` and `jQuery Migrate v3.0.1`, to this pr.

As Luke has put the following pr on hold: https://github.com/octobercms/october/pull/4461

I will update the `FrameworkNode.php` file with a newer sri to match those new versions of `jquery` and `jquery migrate`, then everything will go smoothly and people won't even notice.

ayumihamsaki:
@bennothommo 

Sorry late reply, had to get some sleep, yah exactly what Samuell1 said is what I was seeing. That's why I went for the underscore as computers read `extras_optimized` and `extrasoptimized` the same.

If you want to fix the parser side to allow mutli-parms. I will update the pr to this:

```
{% framework %} 
{% framework optimized %}
{% framework extras %}
{% framework extras optimized %}
{% framework jquery %} = loads jquery sync only (not framework)
{% framework jquery optimized %} = loads jquery sync, high priorty and preload (not framework)
```

Updated commit found here: https://github.com/octobercms/october/pull/4463/commits/7aefdcf5b741c772f591cfa9842ec71a59b76ab8

ayumihamsaki:
Notes for people reading this in the future.

The twig tags load jQuery `sync` not `(async)`. If you want to load jQuery using `async` you can do so by manually adding the jquery into your web page without using the tags.

For further info how to setup jQuery `async` you can read this article: https://idiallo.com/javascript/async-jquery

bennothommo:
@ayumihamsaki No problem :). If a fix to the argument parser can be done purely within October code, then go for it. If it will require an update to the Twig framework directly, then don't worry about it - any changes to the Twig code may be lost with an update, and I would imagine they have their reasons for not allowing it.

Samuell1:
This twig parser file specific for framework tag needs to be changed somehow to allow passing $name and attributes too: https://github.com/octobercms/october/blob/develop/modules/cms/twig/FrameworkTokenParser.php

ayumihamsaki:
> If it will require an update to the Twig framework directly, then don't worry about it.

Going to set them to this:

```
{% framework %} 
{% framework optimized %}
{% framework extras %}
{% framework extras_optimized %}
{% framework jquery %} = loads jquery sync only (not framework)
{% framework jquery_optimized %} = loads jquery sync, high priorty and preload (not framework)
```

As per commit: https://github.com/octobercms/october/blob/7fd15432cf9af5e756de213cc8e453ef7053975c/modules/cms/twig/FrameworkNode.php

LukeTowers:
This has become too complicated and I'm not satisfied with the experience of using these "optimized" tags. @ayumihamsaki please create a PR to completely revert your changes to the framework, framework extras, and jquery tags. If we still want to create a way for developers to easily opt in to the optimized versions then that can be a separate issue / PR, but for now I just want the backwards incompatible issue solved so we can release 458. Any developers that absolutely need the optimized version can always include those files however they want to, these tags are just shortcuts.

ayumihamsaki:
@LukeTowers no problem, I will revert the changes now. If I hard code the files into my websites will there be any danger of filename changes or anything in the future?

LukeTowers:
@ayumihamsaki I doubt it. If there are they'd be noted in the release notes.

ayumihamsaki:
@LukeTowers 

Done:

Revert file is here: https://github.com/octobercms/october/blob/4a03be9f90456189ce619e762c3583e32f131f8b/modules/cms/twig/FrameworkNode.php

Gives the following:

```
{% framework %} 
{% framework extras %}
{% framework jquery %}
```

Updated doc's here: https://github.com/octobercms/docs/blob/ad0a47a9cf160c38c0d2ef331564bf1443dc07f4/ajax-introduction.md#framework-options

p.s. Going to try my best to finish off the large PR tomorrow, as I know you want to push this commit quickly.

LukeTowers:
@ayumihamsaki can you split those two changes into their own PRs? I only want to merge those changes right now, the big changes will take a lot more reviewing before they're ready to merge.

Samuell1:
Btw if framework tag changes will be made later it needs to reflect on combiner aliases too. https://octobercms.com/docs/markup/filter-theme#combiner-aliases 

And from my experience i mostly recommend importing scripts like that because then you can combine many requests to few files.

ayumihamsaki:
Nice tip!

Want to change the following doc's line:

```
@jquery | Reference to the jQuery library (v3.4.0) used in the back-end. (JavaScript)
```

into

```
@jquery | Reference to the jQuery library (highest available version) used in the back-end. (JavaScript)
```

That way I don't need update this doc's page every time a jquery update takes place.

LukeTowers:
@Samuell1 actually the aliases won't need to be touched, the "optimized" tag changes how the script / style / link elements are generated, not the actual content itself. If someone's using the combiner aliases then they should be doing those optimizations themselves. @ayumihamsaki I'd prefer we kept the version information there, but you could maybe add a note to the files in the october repo to make sure that we always update that information when updating the jQuery version.

ayumihamsaki:
> but you could maybe add a note to the files in the october repo to make sure that we always update that information when updating the jQuery version.

Yah going to make a note of that for future ref.

LukeTowers:
@ayumihamsaki @Samuell1 @acasar The fix has been implemented in https://github.com/octobercms/october/commit/99777f5977ef50d1b8757bf0fc1e539e5e8a5afb. We can discuss adding the optimized versions in a separate issue dedicated issue just for that, but seeing as it's just a shortcut and devs can always write that version manually into their themes themselves I don't really see a huge need for it right now.

ayumihamsaki:
I'm NOT up for adding the optimized versions! I just created a partial file with the optimized versions and  added that twig into my layout files instead.

LukeTowers:
@ayumihamsaki fair enough, that seems like the easiest solution for people that want to do it that way 😂 Thanks for your work on this so far though!

PR #4301, opened 2019-04-29
Prevent subject and password from being HTML encoded in the invite email

Pindagus:
This prevents that the subject and/or password are being HTML encoded in the invite email.

https://github.com/octobercms/october/commit/ac5bc866ff8f0e7dea3ed7f667b99848723eb268 should have fixed this for the password, but didn't. A fix for this is included in this request.

### DIFF
--- a/modules/backend/views/mail/invite.htm
+++ b/modules/backend/views/mail/invite.htm
@@ -1,4 +1,4 @@
-subject = "Welcome to {{ appName|raw }}"
+subject = "Welcome to {{ appName | raw }}"
 layout = "system"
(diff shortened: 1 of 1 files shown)


modules/backend/views/mail/invite.htm
```diff
@@ -1,4 +1,4 @@
-subject = "Welcome to {{ appName }}"
+subject = "Welcome to {{ appName|raw }}"
```
LukeTowers: space before and after `|` please

Pindagus: Added spaces for format

PR #4311, opened 2019-05-02
Add "onUpload" AJAX handler to FileUpload widget

alxy:
Potential fix for #4300

### DIFF
--- a/modules/backend/behaviors/RelationController.php
+++ b/modules/backend/behaviors/RelationController.php
@@ -299,6 +299,10 @@ class RelationController extends ControllerBehavior
             $field = post(self::PARAM_FIELD);
         }
 
+        if ($this->initialized) {
+            return;
+        }
+
         $this->config = $this->originalConfig;
         $this->model = $model;
         $this->field = $field;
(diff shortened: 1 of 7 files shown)


mjauvin:
Tested in thr playground and it definitely fix all the issues I was having playing with this yesterday.

MUCH cleaner approach than what we started experimenting with.

bennothommo:
@alxy Tested it and it works great. I've tried it in a few different configurations and didn't find any issues. Well done! Would you be up for applying the same fix to the MediaManager widget, as per @LukeTowers comment here: https://github.com/octobercms/october/issues/4300#issuecomment-487680319

alxy:
> @alxy Tested it and it works great. I've tried it in a few different configurations and didn't find any issues. Well done! Would you be up for applying the same fix to the MediaManager widget, as per @LukeTowers comment here: [#4300 (comment)](https://github.com/octobercms/october/issues/4300#issuecomment-487680319)

I am not sure if I'd like to do this. The mediamanager is complex (just looked at it) and there are multiple ways to upload files (Via the widget itself, via `mediafinder`, via `richedtitor`) and currently there are no issues I am aware of. So why should we change it?

w20k:
Later on today, will retest. Seems like it's ready.

w20k:
@bennothommo looks good to me as well! Could be merged, I guess :)

PR #4314, opened 2019-05-03
Rework the media manager to use "onUpload" handler

alxy:
This is in accordance with my other PR to the fileupload: https://github.com/octobercms/october/pull/4311

I found that the media manager is initialized and bound to the controller on every backend page (`ocmediamanager`) - I am not sure why this is necessary. But in this case it enables us to use the MediaManager right from the richeditor without any more setup work.


modules/backend/widgets/MediaManager.php
```diff
@@ -1472,24 +1470,11 @@ protected function getTargetDimensions($width, $height, $originalImagePath)
...
-     * Detect the upload post flag
-     * @return void
+     * Upload handler for the server-side processing of uploaded files
      */
-    protected function checkUploadPostback()
+    public function onUpload()
     {
-        if ($this->readOnly) {
-            return;
-        }
```
LukeTowers: This check needs to stay here @alxy

alxy: I would say we don't need this any longer, as the handler is no longer ran on any initialization (constructor) of the widget. But I will investigate this next weekend.

alxy:
@LukeTowers I wonder how a `mediafinder` widget is supposed to work in `readOnly` mode? In fact, I am a bit confused about the difference between `readOnly` (which in my tests didn't have much effect on the behavior of the fields, e.g. text fields can be modified as usual) and `disabled` (which clearly disables the field value from beeing changed). Can you shed some light into this?

LukeTowers:
@alxy I did up a table comparing the differences between the modes in a PR / issue comment somewhere, probably related to adding readOnly / disabled support to various formwidgets

alxy:
I would say they should either behave the same for this the mediafinder at least or `readOnly` may not have any effect t all (current behavior). I am okay with both options. Can you link the mentioned issue comment or tell me what I should do?

bennothommo:
@alxy From what I've been able to glean from the media manager widget whilst working on the multiple item selection, readOnly appears to be used to allow the media manager to select files in the media folder, but otherwise not be able to upload, rename, crop or delete files within the media manager widget. I can't find any specific instances where this is actually triggered or used, but I can think of use cases (for example, working on a remote filesystem with read access only, or perhaps a configuration or plugin that allows clients to select from set files without giving them access to modify them)

alxy:
@bennothommo In this case, we should just remove the buttons to upload or move files I'd say? Interestingly, the [MediaManager](https://github.com/octobercms/october/blob/master/modules/backend/widgets/MediaManager.php#L49) itself supports a `readOnly` mode - which does remove the upload buttons. However, the `mediafinder` form widget really only interfaces the globally available `ocmediamanager` widget via JS. As said earlier, I am not sure why it is globally available to every backend controller: If we would just initialize the widget if it is used, we could also set the `readOnly` property accordingly. Not sure if this can easily be accomplished....

bennothommo:
@alxy Yeah I'm not entirely sure why it needs to be initialised on every page too - your proposal makes sense in that it should be initialised only when needed, and allow any properties to be set through the `mediafinder` widget too to be passed through to the `mediamanager` widget on initialisation.

Feel free to make the change and we'll give it a good look through before merging. Let me know if you need any help with this.

alxy:
I have added the required changes, however I just realized that there are some merge conflicts due to some more recent changes. Will resolve these if this is accepted.

As stated earlier, I just initialized the media manager widget in the corresponding form widgets if needed.

bennothommo:
@alxy We'll need you to resolve the conflicts first before we can accept it, so we can be sure that it does not cause issues with any of the recent changes.

alxy:
@bennothommo Ok, I used the github interface to resolve the conflicts. Not tested though.

bennothommo:
@alxy Thanks - great work on this, really impressed.

Only thing I noticed was that the `readOnly` code that @LukeTowers flagged earlier is still missing from your latest changes. Whilst it probably is handled elsewhere on the frontend, I think the check needs to stay to prevent people from using crafted AJAX queries in the backend to circumvent read-only restrictions. Would you be able to restore that, perhaps using the `abortIfReadOnly` method?

alxy:
Done. However, I realized when testing the readOnly mode for form widgets (e.g. the mediafinder), while the buttons in the top toolbar of the popup are hidden in this mode (which is correct behavior), an insert button on the bottom right does still appear. Probably it should be hidden in readOnly mode as well?

bennothommo:
@alxy That button can stay. That's the button to insert the selected items into whichever field the mediafinder is connected to.

alxy:
Yeah, but the readOnly=true mode would suggest to me that we cannot change the selected item ;)

bennothommo:
@alxy I think that is what "disabled" is for ;)

LukeTowers:
Thanks for the reminder @bennothommo. @alxy readOnly in the MediaFinder means that you can view the media library and select images. Disabled means that you can't do either.

bennothommo:
@alxy Just checking in to see if you have finalised the `disabled` and `readOnly` states yet?

alxy:
@bennothommo yes, this is resolved. See the abortIfReadonly call in the handler.

bennothommo:
@alxy Cool, will test it all shortly.

modules/backend/classes/Controller.php
```diff
@@ -158,14 +158,6 @@ public function __construct()
          */
         $this->user = BackendAuth::getUser();
 
-        /*
-         * Media Manager widget is available on all back-end pages
-         */
-        if ($this->user && $this->user->hasAccess('media.*')) {
-            $manager = new MediaManager($this, 'ocmediamanager');
-            $manager->bindToController();
-        }
```
LukeTowers: @alxy @bennothommo doesn't this mean that the MediaManager widget is now bound to the controller once for every single instance of a field that uses it? That seems a bit excessive.

LukeTowers: Is there any way we could make a FormWidget trait for including the media manager in formwidgets so that it detects when the media manager widget has been already added to the controller and only adds it once if it hasn't already been added? That way we could also get rid of all the changes for dynamically changing the alias referenced.

bennothommo: @LukeTowers correct, but I figure it's no different than the Repeater widget doing the same thing.

bennothommo:
@alxy I have tested this on a clean install. It works for mediafinder fields in the Test plugin (including `readOnly` and `disabled` states), but seems to break for the rich editor, especially in the Static Pages plugin.

image

@LukeTowers has [made a suggestion](https://github.com/octobercms/october/pull/4314#discussion_r302125601) towards the widget binding to the controller which may be worth consideration.

modules/backend/classes/Controller.php
LukeTowers: It is a bit different actually. For FormWidgets, there should be a per-field instance of the FormWidget as each field could be different in its configuration. However, this is slightly different in that the Media Manager Widget (not formwidget) is a common tool that is used to access / manage the media library in October, and it just so happens that lots of different FormWidgets use this tool. Basically, there should be one "true" media manager, and however many instances of the FormWidgets that use it as necessary. Thoughts? @bennothommo

LukeTowers:
We'll also need to make sure the Media page itself still works, it might have been relying on the generic controller initialization.

bennothommo:
@LukeTowers I've confirmed that working all correctly in my tests. :)

modules/backend/classes/Controller.php
bennothommo: @LukeTowers My thinking was that given each form widget that uses the Media Manager having the potential to affect operation of the Media Manager in some way (mainly from the `readOnly` and `disabled` properties), it might be safer for them to operate with different instances. All that being said, I haven't really studied the effects of the differences between how @alxy has set up and the normal operation beyond just testing that it works on a user level.

alxy: But that would mean we need to initialize the "true" media manager "somewhere". Right now it is done in a central place for each and every backend request. I think its far more sensible to just initialize it when needed, even though that might mean we need to create multiple instances of the widget. (I don't see another nice solution to this problem.)

alxy:
@bennothommo thanks, I will look into this. Might take some time however.

bennothommo:
@alxy Not a problem. @LukeTowers, we might need to hold this off until the next version, I think.

modules/backend/classes/Controller.php
LukeTowers: @bennothommo I believe the readOnly property is passed to the media manager via JS so that shouldn't be a problem.

@alxy the idea is that there would be a common trait that every widget would call `$this->bindMediaManager()` and it would check the controller instance to make sure the media manager is only ever bound once. To be clear, my main concerns with the approach taken in this PR is the passing around of `mediaManagerAlias` everywhere and the repeating of that code; then as a secondary concern the added performance overhead of intializing it multiple times (although I haven't actually run any performance testing so I'm not going to be too pushy on that as changes for the sake of performance without actual numbers to go by are more harmful than helpful most of the time).

LukeTowers:
Replaced by https://github.com/octobercms/october/pull/5026, thanks @alxy for all your hard work on this!

PR #4317, opened 2019-05-03
Update jQuery from V3.4.0 to V3.4.1

ayumihamsaki2:
Update jQuery from V3.4.0 to V3.4.1

See github: ref: #3867 

Note: This is a quick update which probably means important bug fixes and this release was pushed out quickly instead of waiting to add it to their V4.0.0 release.

Below is the following commits from their release:

- Build: Update Sinon from 2.3.7 to 7.3.1, other updates
- Tests: Fix the core-js polyfill inclusion method
- Build: Update the master version to 4.0.0-pre
- CSS: Don't automatically add "px" to properties with a few exceptions
- Build: Update Sizzle from 2.3.3 to 2.3.4
- Release: update AUTHORS.txt
- Build: Fix unresolved jQuery reference in finalPropName
- Build: Test on Node.js 12, stop testing on Node.js 6 & 11
- Event: Fix handling of multiple async focus events
- Event: Prevent leverageNative from registering duplicate dummy handlers
- Data: Separate data & css/effects camelCase implementations
- Event: Stop shimming focusin & focusout events
- Tests: Fix the new focusin/focusout test in IE
- Core: Remove deprecated jQuery APIs
- Tests: Restrict an event test fallback to TestSwarm
- Core: Drop support for IE <11, iOS <11, Firefox <65, Android Browser and PhantomJS
- Build: Fix AMD dependencies in curCSS
- Effect: Fix a unnecessary conditional statement in .stop()

This update should not create any issues to the current version of October CMS as it is aminor semantic update version number.


ayumihamsaki:
Side Note: jQuery is really bad at DOM XSS Protection! Today I have ran October CMS through the new Trusted Types API with this version of jQuery and found 5 potential DOM XSS attacks routes. Though the current older versions October is running has just as many attack routes.

It maybe a good idea to add a custom jQuery version to October to patch these possible DOM XSS issues. This would also require adding them to a CSP Policy, which maybe added to October CMS.

Though this new security API is still in a trial, it would be a good idea to consider fixing this security hole, due to DOM XSS being such a common security issue.

The results from my test is that jQuery has 5 `innerHTML()` which are not sanitized - the new api would sanitize these.

ayumihamsaki:
Side Note: jQuery is really bad at DOM XSS Protection! Today I have ran October CMS through the new Trusted Types API with this version of jQuery and found 5 potential DOM XSS attacks routes. Though the current older versions October is running has just as many attack routes.

It maybe a good idea to add a custom jQuery version to October to patch these possible DOM XSS issues. This would also require adding them to a CSP Policy, which maybe added to October CMS.

Though this new security API is still in a trial, it would be a good idea to consider fixing this security hole, due to DOM XSS being such a common security issue.

## Test Results

The results from my test is that jQuery has 5 `innerHTML()` which are not sanitized - the new api would sanitize these.

csp

w20k:
@ayumihamsaki I'm not sure if DOMPurify is quite smart against other libraries. In the case of jQuery, just did a quick run, they use `innerHtml` for initialization of the component. I don't see any XSS issues with that 😉

ayumihamsaki:
@w20k I've got an open issue with the jQuery team right now and they have openly confirmed the problem. One of the lead coders over there gave me this reply:

> Yeah, this is a tough one. The team has discussed it for years and had issues determining how to make things safer without seriously redesigning the API. Most of our HTML goes through` jQuery.parseHTML()` and `jQuery.htmlPrefilter()` so it is currently possible to try and sanitize through that. There's the one fast path for .html(string) that assigns directly to `.innerHTML` but that already goes through the `prefilter` and could just be taken out if needed. The others are Sizzle feature checks for browsers we won't support in 4.0 anyway.
> 
> Perhaps we could recognize a Trusted Type and pass it through as if were HTML, but not actually do any manipulation with it, e.g. `$("div").append(someTrustedType`). (As an aside, I would hope that a Trusted Type doesn't return an HTML representation when you call `.toString()` on it.)
> 
> Seems like they've taken old browsers into account with a simple polyfill that just returns strings if the browser doesn't support Trusted Types.
> 
> But given where the implementations are at, it is probably too early to add any code.

As you can see they are seriously considering my idea of using the new Trusted Types API to fix the DOM XSS attacks. I'm tying to push them to sort out the DOM XSS problem in the V4 updates, just waiting for the green light from the rest of the jquery core team.

As for October CMS, maybe we could add this issue to one of the lists under the **_Projects_** tab and address it with a future update to jQuery. I think this issue would actually carry a `Priority: High` label in this repo, due to the security threat level.

LukeTowers:
@ayumihamsaki do you have a proof of concept attack utilizing this vector that you could send to me? Otherwise I'm not super concerned about it because it appears to require chaining vulnerabilities to be useful.

ayumihamsaki:
@LukeTowers 

I've not done a proof of concept attack yet, I only started playing around with this yesterday. If I was going to test it out I would try bypassing this:

> There's the one fast path for .html(string) that assigns directly to .innerHTML but that already goes through the prefilter and could just be taken out if needed.

I also did a search on the web for any related attacks like this to jQuery and nothing has been documented yet.

However, there is one attack that is currently possible with the current version of October CMS right now and that is a **"Prototype Pollution"**. jQuery have patched that attack with this current PR version. So this PR would fix that attack. For ref, their github PR that fixed it can be found here: https://github.com/jquery/jquery/pull/4333

---

**Short answer**: I think we can't do anything right now. But I will monitor what the jQuery team do and hopefully get a solution in the V4.0.0 or greater releases and will then re-talk about this when we update again at a later date. For now I leave all these comments as future notes.

ayumihamsaki:
Update: jQuery have accepted my git issue idea and now added it to their V4.0.0 milestone.

ayumihamsaki:
Testing Notes: I plan to create another PR to also update jQuery Migrate from V3.0.1 to V3.1.0 I will link it to this one. Please test both the updated jQuery and jQuery Migrate at the same time! (Just wanting to finish off my tab pr first before creating my next pr).

w20k:
FYI: @ayumihamsaki, @LukeTowers 
Will start testing PRs related to JS upgrades. 
Testing flow:
- [ ] Run through all the OCMS plugins, especially Test plugin
- [ ] Check, if it's working with corresponded widgets/modules
- [ ] Test for corner cases (like browser cache)

ayumihamsaki:
@w20k Don't start testing this pr until I done the jQuery Migrate PR please. I might just close this pr and create a joint PR with both the updated jQuery and jQuery Migrate together (need to test both at the same time).

LukeTowers:
@ayumihamsaki or just add both to this PR instead of creating a new one

ayumihamsaki:
ok will do that

w20k:
@ayumihamsaki ping me when it's ready ;)

ayumihamsaki2:
Closing this as no branch and other issues, please go to the updated PR: https://github.com/octobercms/october/pull/4461

PR #4320, opened 2019-05-06
Optimize the Install products page for large screens

gergo85:
Current layout:
layout-now

New Layout:
layout-new

Tablet layout:
layout-tablet

### DIFF
--- a/modules/system/controllers/updates/_install_plugins.htm
+++ b/modules/system/controllers/updates/_install_plugins.htm
@@ -20,7 +20,7 @@
 
     <div class="row">
 
-        <div class="col-md-7">
+        <div class="col-md-5">
 
             <!-- Installed plugins -->
             <div id="pluginList"
@@ -66,7 +66,7 @@
             </div>
 
         </div>
-        <div class="col-md-5">
+        <div class="col-md-7">
 
             <!-- Recommended extras -->
             <div class="suggested-products-container">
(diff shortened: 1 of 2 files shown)


bennothommo:
@gergo85 Nice work. This is a bit subjective, but I think the installed plugins list might look a tad cramped at the smallest `md` resolution (992px):

Before your change:
Screenshot from 2019-05-06 17-12-09

After your change:
Screenshot from 2019-05-06 17-09-43

Perhaps the two columns should be split 50/50 in `md` resolutions.

gergo85:
You're right. I changed the width of columns.

gergo85:
Wouldn't it be better to increase the number of recommended plugins and themes to 8?

bennothommo:
@gergo85 I'd be keen to see it increased to even 10 if it's in two columns, but dropped down to 7 for the lower resolutions. Maybe the last 3 can be visible only on `lg` devices.

@w20k @LukeTowers Any thoughts on this?

w20k:
If you ask me, `Install page` from a UI&UX perspective should be reworked a little bit.

1. The number of rows and font should be the same.
2. There should be the title, author name (italic and small underneath) and description
3. You should have an option to read `description` of a plugin you're going to install/uninstall
4. Same as for adding, you should have an easy way to remove, not search for a cross for uninstalling
5. Both rows (Rec&Inst) should work and be shown in the same manner for Desktop/Tablet/Mobile
6. Maybe also remove `hr`, right now it looks odd....

LukeTowers:
I don't really use this page too much (mostly just use composer) so not really any input from me. I don't mind @w20k's suggestions as long as it can still remain clean and simple (both visually and from a code perspective)

Issue #2130, opened 2016-06-17
Media library files selection: Shift and Ctrl


side-by-side:
Hello.
I have searched for the same issues or pull request, as it quite obvious, but found nothing. If i missed some, sorry.

In the Media Library we can't select several files using Ctrl or Shift keys, that is the basic functionality of any file manager. Am i wrong?

gabsource:
I can use drag to select multiple files.
Shift allows multiple selection too : 
- Shift+Click add a single file to the selection list
- Shift+Arrow Down/Up add the next file to the selection list

side-by-side:
Yes, thank you for the answer.
But it is still not the expected behaviour. Each of us know how to select range of files using Shift and how to select different files using Ctrl in Windows, in any system, in most of good web-based filemanagers too.

File manager in the October CMS realy good, but this is definetly an issue, I think.

LukeTowers:
Duplicate of https://github.com/octobercms/october/issues/1658

LukeTowers:
Support added in https://github.com/octobercms/october/commit/0c6735f624c82870b918da02c9a103eadf94d9a5

PR #4324, opened 2019-05-08
[WIP] Add support for maxItems in the mediafinder FormWidget

### LINKED ISSUE #1658: Add support for maxItems option to MediaFinder
Hello,
Media finder could not pick multiple images. 

Am i wrong?

LukeTowers:
> **NOTE:** This is still a WIP. Work is not completed on this yet.

This fixes #1658. #2130 should be implemented as a part of this too.

## Problems that will be solved
- Currently in order to select multiple images (or files in general) for purposes such as galleries developers have two options: Use an AttachMany relationship with a FileUpload widget (which prevents the MediaLibrary from being used) or use a Repeater with individual MediaFinder widgets (which is a clunky user experience). This enhancement would simplify that UX
- We will eventually be adding support for DB-based metadata on MediaLibrary items. The data structure used by this enhancement will pave the way for adding that support

1. **Progressive:** This enhancement will be fully progressive, i.e. existing installations and new installations will continue to have the default behavior of the MediaFinder take effect (i.e. one image selected stored as just the path to that image); only instances of the MediaFinder that set a `maxItems: null || integer` will see this feature set. 
2. **Extensible:** The data structure will be structured to ensure that this will enable future extensions to the MediaFinder / MediaLibrary, specifically as it relates to metadata

## Task list:
- [ ] Add support for Shift & Ctrl selection of items (#2130)
- [ ] Add `maxItems` property for the `mediafinder` FormWidget. Options: `null`: Infinite amount of items - `integer`: Limit selection to specified number of items - `false`: (Default value when option not provided) Behavior disabled completely, only store and load strings for single items
- [ ] Add unit tests for maintaining backwards compatibility
- [ ] Add examples to test plugin
- [ ] Add support to the media filter (or decide how to handle that)

## Comments on implementation:

The frontend compatibility is not difficult to solve as mentioned with a change to the media filter that detects an array and gets the correct value from it. One thing I'd like to point out is that my ideal data storage structure for this change is
```php
$items = [
    [
        'path' => '/path/to/image.jpg',
        ...
    ],
    [
        'path' => '/path/to/image2.jpg',
        ...
    ],
];
```
The above structure is to allow for the theoretical overriding of media item metadata (not implemented yet) on a per field basis.

The backend compatibility could be solved. VARCHAR columns are 255 characters. JSON encoding a mediafinder value would add 12 characters (`{['path:'']}`) to the value stored in the DB. That is relatively We could provide something like the following in the widget:
```
public function getLoadValue($value)
{
    if (is_string($value) && is_valid_json($value)) {
        $value = json_decode($value);
    } elseif (is_string($value)) {
        $value = [['path' => $value]];
    }
    return $value;
}
public function getSaveValue($value)
{
     // this would be slightly more complicated as we would need to determine if the property
     // it's being saved to is jsonable or not to decide whether or not to provide the value as a
     // JSON-encoded string or as the associative array that it will be.
}
```
This would maintain backwards compatibility by supporting loading the following data types:
1. Existing 1-item paths as strings stored in DB VARCHAR columns
2. Existing 1-item paths as JSON encoded data objects stored in DB VARCHAR columns
3. 1-or-many-item paths as JSON encoded data objects stored correctly in DB TEXT or JSON columns
and saving the following data types:
1. 1-item paths stored in DB VARCHAR columns as JSON encoded data
2. 1-or-many-item paths stored in DB TEXT/JSON columns as PHP arrays to be serialized by the model

`maxItems` can have the following values:
```
null: Infinite amount of items
integer: Limit selection to specified number of items
false: (Default value when option not provided) Behavior disabled completely, only store and load strings for single items
```
 
This would retain backwards compatibility without requiring anything to be changed in the frontend. If a developer doesn't change anything then their mediafinder data behaves exactly as expected. If a developer wants to take advantage of this functionality then they either set `maxItems` to `null` or an `integer` and that stores the data in the associative array format and requires them to loop over the values on the frontend.


bennothommo:
Development of this has moved to #4325 to take advantage of the updated branch.

Issue #3625, opened 2018-07-03
Adding multiple images inside rich text editor with a single drag


kojaktaha:
##### Expected behavior
So the richtext editor froala comes with a nice feature, adding images from the gallery, when I multiselect images and click insert ... all the images I selected should be added to my text. 

##### Actual behavior
Only one image is added

##### Reproduce steps
add a rich text editor to your model, when you are inserting the data try to add multiple images as mentioned in the (Expected behavior).

##### October build
431

![github](https://user-images.githubusercontent.com/19997163/42204595-d5181756-7eaa-11e8-8f53-03c292fb1b1b.PNG)

LukeTowers:
Related to #1658

kojaktaha:
Any possible work around until the enhancement is made?

LukeTowers:
@kojaktaha not really, you'll just have to insert them one by one

bennothommo:
Will be included in the work done in https://github.com/octobercms/october/pull/4325.

daftspunk:
Tested with 1.0 and 1.1, was unable to replicate this. Perhaps it has been fixed elsewhere already

1. Nav to Test Plugin > Posts > Rich Editor
2. Open MM
3. Select multiple images and clicking insert
4. Multiple images are inserted

Just ping my username if this is still an issue using the latest version

PR #4325, opened 2019-05-08
[WIP] Add support for maxItems in the mediafinder FormWidget

### LINKED ISSUE #1658: Add support for maxItems option to MediaFinder

bennothommo:
> **NOTE:** This is still a WIP. Work is not completed on this yet.

This PR is a recreation of #4324 with a new branch. This fixes #1658. #2130 should be implemented as a part of this too. Credit of original PR and work to @LukeTowers.

## Problems that will be solved
- Currently in order to select multiple images (or files in general) for purposes such as galleries developers have two options: Use an AttachMany relationship with a FileUpload widget (which prevents the MediaLibrary from being used) or use a Repeater with individual MediaFinder widgets (which is a clunky user experience). This enhancement would simplify that UX
- We will eventually be adding support for DB-based metadata on MediaLibrary items. The data structure used by this enhancement will pave the way for adding that support

1. **Progressive:** This enhancement will be fully progressive, i.e. existing installations and new installations will continue to have the default behavior of the MediaFinder take effect (i.e. one image selected stored as just the path to that image); only instances of the MediaFinder that set a `maxItems: -1 || integer` will see this feature set. 
2. **Extensible:** The data structure will be structured to ensure that this will enable future extensions to the MediaFinder / MediaLibrary, specifically as it relates to metadata

## Task list:
- [x] Add support for Shift & Ctrl selection of items (#2130)
- [ ] Add `maxItems` property for the `mediafinder` FormWidget. Options: `-1`: Infinite amount of items - `integer`: Limit selection to specified number of items - `false`: (Default value when option not provided) Behavior disabled completely, only store and load strings for single items
- [ ] Add unit tests for maintaining backwards compatibility
- [x] Add examples to test plugin (https://github.com/octoberrain/test-plugin/pull/67)
- [ ] Add support to the media filter (or decide how to handle that)
- [ ] Allow multiple images to be selected for Froala editor. (#3625)

## Comments on implementation:

The frontend compatibility is not difficult to solve as mentioned with a change to the media filter that detects an array and gets the correct value from it. One thing I'd like to point out is that my ideal data storage structure for this change is
```php
$items = [
    [
        'path' => '/path/to/image.jpg',
        ...
    ],
    [
        'path' => '/path/to/image2.jpg',
        ...
    ],
];
```
The above structure is to allow for the theoretical overriding of media item metadata (not implemented yet) on a per field basis.

The backend compatibility could be solved. VARCHAR columns are 255 characters (191 if using UTF8MB4 on MySQL). JSON encoding a mediafinder value would add 12 characters (`{['path:'']}`) to the value stored in the DB. We could provide something like the following in the widget:
```
public function getLoadValue($value)
{
    if (is_string($value) && is_valid_json($value)) {
        $value = json_decode($value);
    } elseif (is_string($value)) {
        $value = [['path' => $value]];
    }
    return $value;
}
public function getSaveValue($value)
{
     // this would be slightly more complicated as we would need to determine if the property
     // it's being saved to is jsonable or not to decide whether or not to provide the value as a
     // JSON-encoded string or as the associative array that it will be.
}
```
This would maintain backwards compatibility by supporting loading the following data types:
1. Existing 1-item paths as strings stored in DB VARCHAR columns
2. Existing 1-item paths as JSON encoded data objects stored in DB VARCHAR columns
3. 1-or-many-item paths as JSON encoded data objects stored correctly in DB TEXT or JSON columns
and saving the following data types:
1. 1-item paths stored in DB VARCHAR columns as JSON encoded data
2. 1-or-many-item paths stored in DB TEXT/JSON columns as PHP arrays to be serialized by the model

`maxItems` can have the following values:
```
-1: Infinite amount of items
integer: Limit selection to specified number of items
false: (Default value when option not provided) Behavior disabled completely, only store and load strings for single items
```
 
This would retain backwards compatibility without requiring anything to be changed in the frontend. If a developer doesn't change anything then their mediafinder data behaves exactly as expected. If a developer wants to take advantage of this functionality then they either set `maxItems` to `-1` or an `integer` and that stores the data in the associative array format and requires them to loop over the values on the frontend.


bennothommo:
@LukeTowers I've been experimenting with a different way of selecting multiple files/images that I thought I'd run by you. Instead of using the highlighted files to indicate multiple selections, we could perhaps provide a checkbox alongside each file when in multiple selection mode, with a "Selections" area on the sidebar. Ticking a file would add it to the "Selections" area, which would then be used to populate the files in the field.

The main benefit of this would mean that users could select multiple files across different directories, instead of being limited to the one directory. The main disadvantage is that it's an extra action required to select items and it's a different selection mechanism to single file mode where the user just has to highlight the file they wish to add.

LukeTowers:
@bennothommo could we have a "selections" section with the highlighted rows instead of having a checkbox? A checkbox is an alright solution, but it's not a typical tool for the user to perform multiple file selection

bennothommo:
@LukeTowers Sounds good to me.

multiwebinc:
There's a sentence in the original post that is incomplete:

> That is relatively 

I'm wondering what you were trying to say there.

bennothommo:
@multiwebinc Unfortunately I'm not sure what the sentence was, it was copied from @LukeTowers original PR implementing this feature.

bennothommo:
I have changed the behaviour of the SHIFT and CTRL keys as part of this PR, to more closely match file browsers on Windows, OS X and Linux:

- Holding down CTRL and then:
  - *Clicking an item:* Will add or remove that item from a multi-selection of items.
  - *Pressing Up/Left Arrow or Down/Right Array:* Will select the previous or next items respectively to the last selected item. If no item was previously selected, this will instead select the first item in the list and work from there.

- Holding down SHIFT and then:
  - *Clicking an item:* Will select all items between the clicked item and the last selected item, inclusive. If no item was previously selected, this will simply select the clicked item.
  - *Pressing Up/Left Arrow or Down/Right Array:* Will operate the same as holding down the CTRL key.

LukeTowers:
@multiwebinc not sure what that was either, probably an initial attempt at describing the data handling logic.

@bennothommo SHIFT & CTRL behaviour as describes sounds good to me, that's what I was going for with my initial attempts.

LukeTowers:
@bennothommo could we also address https://github.com/octobercms/october/issues/3625 as a part of this?

bennothommo:
@LukeTowers no worries - should be able to include it with the other changes.

PR #4328, opened 2019-05-10
Fix theme cache key null value

Samuell1:
Fix issue when config cache key is null.

One thing iam not sure how i should change is in tests: https://github.com/octobercms/october/blob/master/tests/unit/cms/classes/ThemeTest.php#L13

Second issue is that there are functions that are static and other are not static than i cant use `$this->resetCache()`

### DIFF
--- a/modules/cms/classes/Theme.php
+++ b/modules/cms/classes/Theme.php
@@ -222,7 +222,7 @@ class Theme
      */
     public static function setActiveTheme($code)
     {
-        self::resetCache();
+        self::resetCache($code);
 
         Parameter::set(self::ACTIVE_KEY, $code);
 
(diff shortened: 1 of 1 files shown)


LukeTowers:
Why is this needed?

w20k:
@LukeTowers, @Samuell1 added cache for the themes, but it's buggy right now. Cache for the theme won't be cleared on reactivation only with hard cache reset. I've asked him to look at it, but current solution doesn't sounds great to me. Need to investigate, of we won't break anything with non static call.

Samuell1:
@LukeTowers Mostly issue is with flushing cache on changing theme doesnt work because `(new self)->getDirName()` is null

LukeTowers:
@Samuell1 is this still needed?

Samuell1:
Yes but i need to rework this to other PR. Because i need to change static methods somehow.

PR #4329, opened 2019-05-10
Dashboard Updates (Improving the Down-Scaling Aspect and Styling)

ayumihamsaki:
The CSS and LESS is finished and I have done the easy Javascript.

But the `reportcontainer.js` needs a lot of work done to it. **_I request help with regards to updating that file!_**

The `reportcontainer.js` has been temporary turned off in the `ReportContainer.php` to show that the present code does work! See code line: https://github.com/ayumihamasaki2/october/blob/b259ebdcb0b93473478406e0d010555f3c34305d/modules/backend/widgets/ReportContainer.php#L128

However, the javascript in `reportcontainer.js` when turned on messes everything up and gives poor results.

For now you can test this first commit and see it working and report any small issues! However, I request help to finish this PR off as I feel updating the `reportcontainer.js` is going to require a lot of work and not so easy to do.

This PR relates to the issue found here: https://github.com/octobercms/october/issues/3958

Below, demo's this PR and shows the down-scaling, updated styling to improve WCAG 2.0 AA Web Standards making the styling, contrasts and font sizing easier to read and much less blurry. Also added CSS code to cut the words and force the writing into the boxes. Made the images responsive and lots of other general tidy up's:

aaaaaaa

* The only thing not working is the drag and drop due to the `reportcontainer.js` being turned off.

### DIFF
--- a/modules/backend/widgets/ReportContainer.php
+++ b/modules/backend/widgets/ReportContainer.php
@@ -224,7 +224,7 @@ class ReportContainer extends WidgetBase
 
         $widgets = $this->getWidgetsFromUserPreferences();
 
-        $num =  count($widgets);
+        $num = count($widgets);
         do {
             $num++;
             $alias = 'report_container_'.$this->context.'_'.$num;
@@ -399,8 +399,8 @@ class ReportContainer extends WidgetBase
                 8  => '8 ' . Lang::choice('backend::lang.dashboard.columns', 8),
                 9  => '9 ' . Lang::choice('backend::lang.dashboard.columns', 9),
                 10 => '10 ' . Lang::choice('backend::lang.dashboard.columns', 10),
-				11 => '11 ' . Lang::choice('backend::lang.dashboard.columns', 11),
-				12 => '12 ' . Lang::choice('backend::lang.dashboard.columns', 12)
+                11 => '11 ' . Lang::choice('backend::lang.dashboard.columns', 11),
+                12 => '12 ' . Lang::choice('backend::lang.dashboard.columns', 12)
             ]
         ];
         $result[] = $property;
--- a/modules/backend/widgets/ReportContainer.php
+++ b/modules/backend/widgets/ReportContainer.php
@@ -4,7 +4,6 @@ use File;
 use Lang;
 use Flash;
 use Request;
-use BackendAuth;
 use Backend\Classes\WidgetBase;
 use Backend\Classes\WidgetManager;
 use Backend\Models\UserPreference;
@@ -126,7 +125,7 @@ class ReportContainer extends WidgetBase
     {
         $this->addCss('css/reportcontainer.css', 'core');
         $this->addJs('vendor/isotope/jquery.isotope.min.js', 'core');
-        $this->addJs('js/reportcontainer.js', 'core');
+        //$this->addJs('js/reportcontainer.js', 'core');
     }
 
     //
@@ -146,10 +145,6 @@ class ReportContainer extends WidgetBase
 
     public function onMakeLayoutDefault()
     {
-        if (!BackendAuth::getUser()->hasAccess('backend.manage_default_dashboard')) {
-            throw new ApplicationException("You do not have permission to do that.");
-        }
-
         $widgets = $this->getWidgetsFromUserPreferences();
 
         SystemParameters::set($this->getSystemParametersKey(), $widgets);
@@ -229,15 +224,12 @@ class ReportContainer extends WidgetBase
 
         $widgets = $this->getWidgetsFromUserPreferences();
 
-        $num = count($widgets);
+        $num =  count($widgets);
         do {
             $num++;
             $alias = 'report_container_'.$this->context.'_'.$num;
         }
         while (array_key_exists($alias, $widgets));
-
-        // Ensure that the widget's alias is correctly set for this request
-        $widget->alias = $alias;
 
         $sortOrder = 0;
         foreach ($widgets as $widgetInfo) {
@@ -339,8 +331,7 @@ class ReportContainer extends WidgetBase
         $configuration['alias'] = $alias;
 
         $className = $widgetInfo['class'];
-        $availableReportWidgets = array_keys(WidgetManager::instance()->listReportWidgets());
-        if (!class_exists($className) || !in_array($className, $availableReportWidgets)) {
+        if (!class_exists($className)) {
             return;
         }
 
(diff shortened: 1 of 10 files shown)


ayumihamsaki:
@w20k done, here you go.

w20k:
@ayumihamsaki thank you!)

modules/backend/widgets/reportcontainer/assets/js/reportcontainer.js
```diff
@@ -41,7 +41,7 @@
         $(window).resize($.proxy(this.updateWidth, this))
         this.updateWidth()
 
-        if (!Modernizr.touchevents) {
+        if (!Modernizr.touch) {
```
w20k: @ayumihamsaki touch event was removed from the `Modernizr`, could you review your PR? Maybe there are some old leftovers :)

w20k:
One more thing, or a wise question to ask, @ayumihamasaki do you really think we need to use a pure CSS approach instead of an isotope? I feel like we should tweak Isotop (reportcontainer.js) somehow to do the same as you've suggested in this PR.

modules/backend/widgets/ReportContainer.php
```diff
@@ -146,10 +145,6 @@ public function onResetWidgets()
 
     public function onMakeLayoutDefault()
     {
-        if (!BackendAuth::getUser()->hasAccess('backend.manage_default_dashboard')) {
```
w20k: Why do you need to remove `BackendAuth  access`?

modules/backend/widgets/reportcontainer/assets/js/reportcontainer.js
ayumihamsaki: Have updated a new commit and also spotted a couple of typo's in the code

ayumihamsaki:
@w20k to answer your question:

> do you really think we need to use a pure CSS approach instead of an isotope?

I only did the pure CSS way, because I thought it will be 100% Browser compatible even with IE11. Yes, agree it would be easier to update the isotope instead (more official way).

w20k:
@ayumihamsaki finally got a minute to look through my task list. I think your CSS approach is not correct in this particular case.

Here is a small demo, how it should behave with isotop.js, my guess it should be mainly fixed in reportcontainer.js.
https://codepen.io/desandro/pen/mEinp

ayumihamsaki:
@w20k just looked at your demo and adjust the screen, I seem to get a big gap on the right hand side, see screenshot:

isotope

Personally I almost question why we use this terrible library at all. I'd much prefer `flexbox` with `wrap`. I know October was created before the age of flexbox and css grids.

I just wonder if there is a better library that can do the grad and drop feature.

ayumihamsaki:
p.s. If it's ok with you, my plan is to finish off the tab pr and then create and finish off an updated version of jQuery and jQuery Migrate and then do this.

ayumihamsaki:
@w20k just looked at your demo and adjust the screen, I seem to get a big gap on the right hand side, see screenshot:

isotope

Personally I almost question why we use this terrible library at all. I'd much prefer `flexbox` with `wrap`. I know October was created before the age of `flexbox` and `css grids`.

I just wonder if there is a better library that can do the drag and drop feature.

---

[edit] Also if you change your CSS and set this:

```css
.grid-item--width2 { width: 600px; }
.grid-item--height2 { height: 600px; }
```

The boxes hang out of the screen.

---

I rather code it properly with CSS flexbox and use a library like: https://shopify.github.io/draggable/

See here an example: https://shopify.github.io/draggable/examples/flexbox.html

Resize the screen and see how smooth it is, compared to isotope.

p.s. If it's ok with you, my plan is to finish off the tab pr and then create and finish off an updated version of jQuery and jQuery Migrate and then do this.

w20k:
@ayumihamsaki mainly what's my concern is that we're changing the core behavior, if it's okay to move on toward Responsive/FlexBoxes. I'm good to go, but we'd likey remove JS part `whatsoever`, it's not needed 😉

ayumihamsaki:
I was looking at the networks files:

```
{"X_OCTOBER_ASSETS":{
"js":[
  "modules/backend/widgets/mediamanager/assets/js/mediamanager-browser-min.js?v456",
  "modules/backend/widgets/reportcontainer/assets/vendor/isotope/jquery.isotope.min.js?v456",
  "modules/backend/widgets/reportcontainer/assets/js/reportcontainer.js?v456"
],

"css":[
  "modules/backend/widgets/mediamanager/assets/css/mediamanager.css?v456",
  "modules/backend/assets/css/dashboard/dashboard.css?v456",
  "modules/backend/widgets/reportcontainer/assets/css/reportcontainer.css?v456"
],
"rss":[]
}}
```

I'm thinking of doing the following:

1. Load them properly via HTML and not via `json` and also preload the assets.

2. Delete `jquery.isotope.min.js` and use a more modern light-weight drag n drop library.

3. Add the new flexbox code to the `dashboard.css`

---

For now don't worry about this PR. I will create a fresh PR with all this updated code. Maybe if a few years time we could go even further and use the Browser native drag n drop but right now, see here: https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API

But support is not great, right now, see here: https://developer.mozilla.org/en-US/docs/Web/API/DataTransferItem#Browser_compatibility

ayumihamsaki:
I was looking at the networks files:

```
{"X_OCTOBER_ASSETS":{
"js":[
  "modules/backend/widgets/mediamanager/assets/js/mediamanager-browser-min.js?v456",
  "modules/backend/widgets/reportcontainer/assets/vendor/isotope/jquery.isotope.min.js?v456",
  "modules/backend/widgets/reportcontainer/assets/js/reportcontainer.js?v456"
],

"css":[
  "modules/backend/widgets/mediamanager/assets/css/mediamanager.css?v456",
  "modules/backend/assets/css/dashboard/dashboard.css?v456",
  "modules/backend/widgets/reportcontainer/assets/css/reportcontainer.css?v456"
],
"rss":[]
}}
```

I'm thinking of doing the following:

1. Load them properly via HTML and not via `json` and also preload the assets.

2. Delete `jquery.isotope.min.js` and use a more modern light-weight drag n drop library.

3. Add the new flexbox code to the `dashboard.css`

---

For now don't worry about this PR. I will create a fresh PR with all this updated code. Maybe in a few years time we could go even further and use the Browser native drag n drop, see here: https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API

But support is not great, right now, see here: https://developer.mozilla.org/en-US/docs/Web/API/DataTransferItem#Browser_compatibility

w20k:
@ayumihamsaki better to ask @LukeTowers and @bennothommo what they think about moving from `isotope.js` to a totally different approach.

LukeTowers:
@ayumihamsaki @w20k as long as it remains functionally the same (and nothing else uses isotope, check the reorder controller to be sure) I don't mind switching to something that works better as long as its thoroughly tested and actually demonstrably works better.

ayumihamsaki:
Will do, my goal will be to make sure users can tell no difference between the old and new code.

@w20k please can you remove this from your To Do list and the Milestone label.

I will work on this issue in my spare time and try and test as much as possible to confirm it's a good idea.

For now leave this pr open, I will close it later on. When hopefully I create a better PR to address this issue.

w20k:
@ayumihamsaki if you need a hand with JS - ping me ;)

PR #4331, opened 2019-05-11
Add check to make sure event result is an array in Form widget

### LINKED ISSUE #1: Enabling this plugin make backend dependsOn to crash
Hello,

I firstly posted the issue on october cms issue queue, but it seems that it's caused by your plugin. Disabling it resolved the issue. [See my original post here](https://github.com/octobercms/october/issues/4028).

SeriousKen:
Wildcard listeners will add null responses to the event result array and causes an error when they are merged into the result. This PR adds a check to see if each event result is an array before adding it to the result array. This should fix https://github.com/OFFLINE-GmbH/oc-clockwork-plugin/issues/1 and not introduce any backward compatability issues.

### DIFF
--- a/modules/backend/widgets/Form.php
+++ b/modules/backend/widgets/Form.php
@@ -449,6 +449,10 @@ class Form extends WidgetBase
         $eventResults = $this->fireSystemEvent('backend.form.refresh', [$result], false);
 
         foreach ($eventResults as $eventResult) {
+            if (!is_array($eventResult)) {
+                continue;
+            }
+            
             $result = $eventResult + $result;
         }
 


bennothommo:
I have tested this and null values are no longer being sent to the `backend.form.refresh` event. @SeriousKen can you confirm this fixes the issue with the Clockwork plugin as well, before I merge?

tobias-kuendig:
I can confirm this PR fixes [the issue that came up](https://github.com/OFFLINE-GmbH/oc-clockwork-plugin/issues/1) when using the clockwork integration plugin.

bennothommo:
Awesome, thanks @tobias-kuendig!

PR #4332, opened 2019-05-11
add methods to allow extending/viewing layoutPath

mjauvin:
With this, it is now possible to do the following in a plugin definition file:

```
\Backend\Classes\Controller::extend(function($controller) {
     list($author, $plugin) = explode('\\', strtolower(get_class()));
     $controller->addLayoutPath(sprintf('$/%s/%s/layouts', $author, $plugin));
});
```

And creating a layout view/partial file in the plugin's layouts folder, thus, overriding the backend layout.

### DIFF
--- a/modules/system/traits/ViewMaker.php
+++ b/modules/system/traits/ViewMaker.php
@@ -44,6 +44,32 @@ trait ViewMaker
     public $suppressLayout = false;
 
     /**
+     * Prepends a path on the available layout path locations.
+     * @param string|array $path
+     * @return void
+     */
+    public function addLayoutPath($path)
+    {
+        $this->layoutPath = (array) $this->layoutPath;
+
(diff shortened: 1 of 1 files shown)


mjauvin:
Ok, so I know you can define custom Backend Skins to do the same, but I find this makes it simpler and more flexible. 

When defining new skins, one has to create a new skin class and change the cms.backendSkin setting to point to this new class, etc...

What do you guys think?

bennothommo:
@mjauvin What would happen in the case of several plugins adding layout paths? I think the benefit of the `cms.backendSkin` setting means that only one theme is going to affect the backend at any given time, so there's no confusion about where changes are being introduced from.

mjauvin:
@bennothommo that's a valid point. would you say the same reasoning might be valid regarding the "addViewPath()" method?

PR #4334, opened 2019-05-12
Do not reset groupMap indexes

datune:
The change introduced in https://github.com/octobercms/october/commit/9b5bd83f10a73e92e5b88a9b9f23230d8bf374eb?diff=split causes grouped repeater widgets to lose the correct "index to widget" correlation once they have been sorted.

Can easily be reproduced:
- Create a grouped repeater config, add two widgets of your choosing
- Create a page and add two widgets, save
- Change the order of the widgets and save
- Reload the backend page and both widgets will now appear to have lost it's content

Before sorting:
blocks[0][textarea_content] = "Textarea 1 Content"
blocks[0][_group] = "textarea"
blocks[1][textarea_content] = "Textarea 2 Content"
blocks[1][_group] = "textarea"

After sorting:
blocks[1][textarea_content] = "Textarea 2 Content"
blocks[1][_group] = "textarea"
blocks[0][textarea_content] = "Textarea 1 Content"
blocks[0][_group] = "textarea"

@bennothommo Is there a particular reason why you opted to reset the indexes when creating the groupMap? I certainly don't mean to introduce a breaking change, but my proposed change seems to fix the problem and seems to not break anything. Perhaps you could have a look too?

### DIFF
--- a/modules/backend/formwidgets/Repeater.php
+++ b/modules/backend/formwidgets/Repeater.php
@@ -237,8 +237,8 @@ class Repeater extends FormWidgetBase
         }
 
         if (is_array($currentValue) && count($currentValue)) {
-            foreach ($currentValue as $value) {
-                $groupMap[] = array_get($value, '_group');
+            foreach ($currentValue as $index => $value) {
+                $groupMap[$index] = array_get($value, '_group');
             }
         }
 
--- a/modules/backend/formwidgets/Repeater.php
+++ b/modules/backend/formwidgets/Repeater.php
@@ -233,9 +233,12 @@ class Repeater extends FormWidgetBase
                 }
             }
         }
-        
+       
         collect($currentValue)->each(function ($value, $index) {
-            $this->makeItemFormWidget($index, array_get($value, '_group'));
+            $groupCode = array_get($value, '_group', false);                    
+            if (!$groupCode) return;
+
+            $this->makeItemFormWidget($index, $groupCode);
         });
         
         $this->indexCount = max(count($currentValue), $this->indexCount);
--- a/modules/backend/formwidgets/Repeater.php
+++ b/modules/backend/formwidgets/Repeater.php
@@ -234,7 +234,9 @@ class Repeater extends FormWidgetBase
             }
         }
         
-        if (!is_array($currentValue)) return;
+        if (!is_array($currentValue)) {
+            return;
+        }
        
         collect($currentValue)->each(function ($value, $index) {
             $this->makeItemFormWidget($index, array_get($value, '_group', false));
--- a/modules/backend/formwidgets/Repeater.php
+++ b/modules/backend/formwidgets/Repeater.php
@@ -233,6 +233,8 @@ class Repeater extends FormWidgetBase
                 }
             }
         }
+        
+        if (!is_array($currentValue)) return;
        
         collect($currentValue)->each(function ($value, $index) {
             $groupCode = array_get($value, '_group', false);                    
--- a/modules/backend/formwidgets/Repeater.php
+++ b/modules/backend/formwidgets/Repeater.php
@@ -239,7 +239,7 @@ class Repeater extends FormWidgetBase
         }
        
         collect($currentValue)->each(function ($value, $index) {
-            $this->makeItemFormWidget($index, array_get($value, '_group', false));
+            $this->makeItemFormWidget($index, array_get($value, '_group', null));
         });
         
         $this->indexCount = max(count($currentValue), $this->indexCount);
--- a/modules/backend/formwidgets/Repeater.php
+++ b/modules/backend/formwidgets/Repeater.php
@@ -234,7 +234,7 @@ class Repeater extends FormWidgetBase
             }
         }
         
-        collect($currentValue)->each(function($value, $index) {
+        collect($currentValue)->each(function ($value, $index) {
             $this->makeItemFormWidget($index, array_get($value, '_group'));
         });
         
--- a/modules/backend/formwidgets/Repeater.php
+++ b/modules/backend/formwidgets/Repeater.php
@@ -217,9 +217,7 @@ class Repeater extends FormWidgetBase
             $this->indexCount = 0;
             $this->formWidgets = [];
             return;
-        }
-
-        $groupMap = [];
+        }       
 
         // Ensure that the minimum number of items are preinitialized
         // ONLY DONE WHEN NOT IN GROUP MODE
@@ -235,20 +233,11 @@ class Repeater extends FormWidgetBase
                 }
             }
         }
-
-        if (is_array($currentValue) && count($currentValue)) {
-            foreach ($currentValue as $index => $value) {
-                $groupMap[$index] = array_get($value, '_group');
-            }
-        }
-
-        if (!count($groupMap)) {
-            return;
-        }
-
-        foreach ($groupMap as $index => $groupCode) {
-            $this->makeItemFormWidget($index, $groupCode);
-        }
+        
+        collect($currentValue)->each(function($value, $index) {
+            $this->makeItemFormWidget($index, array_get($value, '_group'));
+        });
+        
         $this->indexCount = max(count($currentValue), $this->indexCount);
     }
 
--- a/modules/backend/formwidgets/Repeater.php
+++ b/modules/backend/formwidgets/Repeater.php
@@ -237,10 +237,7 @@ class Repeater extends FormWidgetBase
         if (!is_array($currentValue)) return;
        
         collect($currentValue)->each(function ($value, $index) {
-            $groupCode = array_get($value, '_group', false);                    
-            if (!$groupCode) return;
-
-            $this->makeItemFormWidget($index, $groupCode);
+            $this->makeItemFormWidget($index, array_get($value, '_group', false));
         });
         
         $this->indexCount = max(count($currentValue), $this->indexCount);


bennothommo:
@datune Thanks for letting us know. The commit you mentioned is part of a larger group of changes @LukeTowers and I did to the repeater widget to effectively remove the need for indexes - the ordering of repeater items is determined by the order the data comes in POST. Note that [this line](https://github.com/octobercms/october/blob/4e92686c1a7de662840fa160b7348003f5ba817d/modules/backend/formwidgets/Repeater.php#L203) effectively removes the index keys of the repeater items on save. I'll double check the functionality now but as far as I was aware, it was all working, even in group mode.

datune:
@bennothommo The reordering does not work though! If you were to save, what happens on Post is ALL content is lost, because the content of the widgets is not loaded anymore. You can very easily reproduce this. I‘m back in my office later, I‘ll try to debug it further if you want...

bennothommo:
@datune You're right, and your fix seems to work. Thanks heaps for that. I might need another set of eyes on this too. @w20k, would you mind giving this a test and seeing if this works for you?

datune:
I have investigated this a bit further. I feel like the whole method processItems needs a code review. 

For example, the following snippet:

```php
        if (is_array($currentValue) && count($currentValue)) {
            foreach ($currentValue as $index => $value) {
                $groupMap[$index] = array_get($value, '_group');
            }
        }
```

could be rewritten to
```php
        collect($currentValue)->each(function($value, $index) use(&$groupMap) {
            $groupMap[$index] = array_get($value, '_group');
        });
```

However, I would take it one step further even. I do not understand the need for $groupMap or any of the following code:
```php
        if (is_array($currentValue) && count($currentValue)) {
            foreach ($currentValue as $key => $value) {
                $groupMap[$key] = array_get($value, '_group');
            }
        }

        if (!count($groupMap)) {
            return;
        }

        foreach ($groupMap as $index => $groupCode) {
            $this->makeItemFormWidget($index, $groupCode);
        }
```

I think all of the above code could be rewritten with 3 lines: 
```php
        collect($currentValue)->each(function($value, $index) use(&$groupMap) {
            $this->makeItemFormWidget($index, array_get($value, '_group'));           
        });
```
This completely gets around having to recreate any indexes, and should not interfere with the correct indexes being taken from a POST request.
@LukeTowers @bennothommo can you please review and let me know if you agree? If so, I can update this PR.

bennothommo:
@datune I like those changes, however we'd have to ensure that `$currentValue` is an array and has items before converting it to a collection, and that initialisation of the widget stops if `$currentValue` is not an array, or has no items.

The `$groupMap` array was set up to basically link an index to a group code, which is then passed through to a form widget with the specified group form fields. The error you encountered was because when the repeater was re-ordered, the index of the value no longer matched the index in the `$arrayKey` property in the form widgets for each repeater item.

datune:
@bennothommo All your points are already adressed with those 3 lines of code. There will always be an array, it will never enter the iteration when there are no items. I understand why you created $groupMap, but it is not necessary ;-)

modules/backend/formwidgets/Repeater.php
```diff
@@ -235,20 +233,11 @@ protected function processItems()
...
-
-        if (!count($groupMap)) {
-            return;
-        }
-
-        foreach ($groupMap as $index => $groupCode) {
-            $this->makeItemFormWidget($index, $groupCode);
-        }
+        
+        collect($currentValue)->each(function($value, $index) {
```
bennothommo: @datune Please add a space after `function`

datune: @bennothommo Done.

bennothommo:
@datune I would still prefer a check to make sure that `$currentValue` is an array (or at least is iterable) and has items before being converted to a collection. I've noticed that Laravel's `Collection` initializer will convert anything to an array, even a string or an integer, which may result in phantom items being added.

datune:
@bennothommo There is literally no need to check if $currentValue is an array, because there will always be an array. And iterating will only commence if there is something to be iterated over. We could argue that we may want to check if WHAT we are iterating needs to be validated, but this has not been done before either...

At the beginning of the method the value for $currentValue is retreived, if you want to add validation to ensure that $currentValue contains data that we expect (other than of course not null and it being an iterable array, which is already taken care of by the collect() and each() methods), you would have to add it here. But, I feel that is a topic on it's own.

seanthepottingshed:
@bennothommo @datune 

I've tested this and the indices are now updated as expected. Basically I'm just trying to get to a point where I can extend either a **Static** or **CMS** page with a repeater and validate the repeater groups individually based upon their type.

This following may well be out of the scope of this discussion and I'm more than happy to open a new issue / corresponding PR - now that this works as expected with a **Static** page I'm trying to get it to also work with a **CMS** page, as such locally I've amended my local PR as below:

```php
<?php

namespace October\Test;

use Backend;
use System\Classes\PluginBase;

/**
 * Test Plugin Information File
 */
{
    /**
     * Returns information about this plugin.
     *
     * @return array
     */
    public function pluginDetails()
    {
        return [
            'name'        => 'October Tester',
            'description' => 'Used for testing the Relation Controller behavior and others.',
            'author'      => 'Alexey Bobkov, Samuel Georges',
            'icon'        => 'icon-child',
            'homepage'    => 'https://github.com/daftspunk/oc-test-plugin',
        ];
    }

    public function registerNavigation()
    {
        return [
            'test' => [
                'label'    => 'Playground',
                'url'      => Backend::url('october/test/people'),
                'icon'     => 'icon-child',
                'order'    => 200,
                'sideMenu' => [
                    'people'    => [
                        'label' => 'People',
                        'icon'  => 'icon-database',
                        'url'   => Backend::url('october/test/people'),
                    ],
                    'posts'     => [
                        'label' => 'Posts',
                        'icon'  => 'icon-database',
                        'url'   => Backend::url('october/test/posts'),
                    ],
                    'users'     => [
                        'label' => 'Users',
                        'icon'  => 'icon-database',
                        'url'   => Backend::url('october/test/users'),
                    ],
                    'countries' => [
                        'label' => 'Countries',
                        'icon'  => 'icon-database',
                        'url'   => Backend::url('october/test/countries'),
                    ],
                    'reviews'   => [
                        'label' => 'Reviews',
                        'icon'  => 'icon-database',
                        'url'   => Backend::url('october/test/reviews'),
                    ],
                    'galleries' => [
                        'label' => 'Galleries',
                        'icon'  => 'icon-database',
                        'url'   => Backend::url('october/test/galleries'),
                    ],
                    'trees'     => [
                        'label' => 'Trees',
                        'icon'  => 'icon-database',
                        'url'   => Backend::url('october/test/trees'),
                    ],
                    'pages'     => [
                        'label' => 'Pages',
                        'icon'  => 'icon-database',
                        'url'   => Backend::url('october/test/pages'),
                    ],
                ],
            ],
        ];
    }

    public function registerFormWidgets()
    {
        return [
            'October\Test\FormWidgets\TimeChecker' => [
                'code' => 'timecheckertest',
            ],
        ];
    }

    public function boot()
    {
        \Event::listen('backend.form.extendFields', function ($widget) {
            if (!$widget->isNested) {
                $pluginManager        = \System\Classes\PluginManager::instance();
                $isStaticIndexPage    = $pluginManager->hasPlugin('RainLab.Pages') && $widget->model instanceof \RainLab\Pages\Classes\Page && $widget->model->url === '/';
                $isCmsDemoPluginsPage = $widget->model instanceof \Cms\Classes\Page && $widget->model->settings['url'] === '/demo/plugins';
                if ($isStaticIndexPage || $isCmsDemoPluginsPage) {
                    $this->extendViewBagFields($widget);
                    $this->bindRepeaterValidationRules($widget->model);
                }
            }
        });
    }

    protected function extendViewBagFields($widget)
    {
        $widget->addFields([
            'viewBag[test_repeater]' => [
                'prompt' => 'Add Data',
                'type'   => 'repeater',
                'groups' => [
                    'textarea' => [
                        'name'        => 'Textarea',
                        'description' => 'Basic text field',
                        'icon'        => 'icon-file-text-o',
                        'fields'      => [
                            'text_area' => [
                                'label' => 'Text Content',
                                'type'  => 'textarea',
                                'size'  => 'large',
                            ],
                        ],
                    ],
                    'quote'    => [
                        'name'        => 'Quote',
                        'description' => 'Quote item',
                        'icon'        => 'icon-quote-right',
                        'fields'      => [
                            'quote_position' => [
                                'span'    => 'auto',
                                'label'   => 'Quote Position',
                                'type'    => 'radio',
                                'options' => [
                                    'left'   => 'Left',
                                    'center' => 'Center',
                                    'right'  => 'Right',
                                ],
                            ],
                            'quote_content'  => [
                                'span'  => 'auto',
                                'label' => 'Details',
                                'type'  => 'textarea',
                            ],
                        ],
                    ],
                    'image'    => [
                        'name'        => 'Image',
                        'description' => 'Pick something from the media library',
                        'icon'        => 'icon-photo',
                        'fields'      => [
                            'img_upload'   => [
                                'span'        => 'auto',
                                'label'       => 'Image',
                                'type'        => 'mediafinder',
                                'mode'        => 'image',
                                'imageHeight' => 260,
                                'imageWidth'  => 260,
                            ],
                            'img_position' => [
                                'span'    => 'auto',
                                'label'   => 'Image Position',
                                'type'    => 'radio',
                                'options' => [
                                    'left'   => 'Left',
                                    'center' => 'Center',
                                    'right'  => 'Right',
                                ],
                            ],
                        ],
                    ],
                ],
                'span'   => 'full',
                'tab'    => 'Test',
            ],
        ], 'primary');
    }

    protected function bindRepeaterValidationRules($model)
    {
        $model->rules += [
            'viewBag.test_repeater' => 'required',
        ];
        $model->bindEvent('model.beforeValidate', function () use ($model) {
            if (!empty($model->test_repeater)) {
                foreach ($model->test_repeater as $index => $repeater) {
                    $repeaterGroup       = $repeater['_group'];
                    $repeaterGroupPrefix = 'viewBag.test_repeater.' . $index . '.';
                    switch ($repeaterGroup) {
                        case 'textarea':
                            $model->rules[$repeaterGroupPrefix . 'text_area']                   = 'required';
                            $model->customMessages[$repeaterGroupPrefix . 'text_area.required'] = 'Please enter text area';
                            break;
                        case 'quote':
                            $model->rules[$repeaterGroupPrefix . 'quote_position']                   = 'required';
                            $model->rules[$repeaterGroupPrefix . 'quote_content']                    = 'required';
                            $model->customMessages[$repeaterGroupPrefix . 'quote_position.required'] = 'Please select quote position';
                            $model->customMessages[$repeaterGroupPrefix . 'quote_content.required']  = 'Please enter quote content';
                            break;
                        case 'image':
                            $model->rules[$repeaterGroupPrefix . 'img_upload']                     = 'required';
                            $model->rules[$repeaterGroupPrefix . 'img_position']                   = 'required';
                            $model->customMessages[$repeaterGroupPrefix . 'img_upload.required']   = 'Please upload image';
                            $model->customMessages[$repeaterGroupPrefix . 'img_position.required'] = 'Please select image position';
                            break;
                    }
                }
            }
        });
    }
}
```

If you visit `/demo/plugins` the validation seems to be a step behind what is going off in the DOM.

datune:
I'm still pondering this. Really, what you are worried about, is that the CONTENTS of the array are not what we expect it to be, hence possibly creating phantom items. One way to get around that is to validate every iteration against something that we require for it to be valid. 

For example, we could implement as follows:
```php
        collect($currentValue)->each(function($value, $index) {
            $groupCode = array_get($value, '_group', false);
            if (!$groupCode) return;
            
            $this->makeItemFormWidget($index, $groupCode);
        });
```
This ensures that it no phantom items are added. Would this work?

bennothommo:
@datune Not quite. See, given that by default the Repeater is using `getLoadValue` from `FormWidgetBase` to retrieve the value, there is no garuntee that the value will be an array as that method is used for all form widgets. Nor is there any garuntee that there will be an array from POST if someone were to mess around with the POST data before sending.

I would rather remove all doubt that we are dealing with an array before trying to create form widgets and initialising items in the Repeater.

datune:
@bennothommo I understand that the value of $currentValue MAY not be an array, but, using collect($currentValue) will always ensure that an iterable collection is created, even if $currentValue contains only a string! So, there is no need to check if we have an array. Now, in order for phantom items to be prevented (which was never implemented before either), I have posted a solution in my previous post.

modules/backend/formwidgets/Repeater.php
```diff
@@ -235,20 +233,13 @@ protected function processItems()
...
-
-        if (!count($groupMap)) {
-            return;
-        }
-
-        foreach ($groupMap as $index => $groupCode) {
-            $this->makeItemFormWidget($index, $groupCode);
-        }
+        
+        if (!is_array($currentValue)) return;
```
bennothommo: @datune Would you mind adding braces around the conditional block, as per [PSR-2](https://www.php-fig.org/psr/psr-2/) guidelines?
```suggestion
        if (!is_array($currentValue)) {
            return;
        }
```

modules/backend/formwidgets/Repeater.php
```diff
@@ -235,20 +233,15 @@ protected function processItems()
...
+        if (!is_array($currentValue)) {
             return;
         }
-
-        foreach ($groupMap as $index => $groupCode) {
-            $this->makeItemFormWidget($index, $groupCode);
-        }
+       
+        collect($currentValue)->each(function ($value, $index) {
+            $this->makeItemFormWidget($index, array_get($value, '_group', false));
```
bennothommo: @datune Sorry, one last thing. The default value if `_group` isn't available should be `null`, not `false`.
```suggestion
            $this->makeItemFormWidget($index, array_get($value, '_group', null));
```

w20k:
@bennothommo tested it as much as I could, seems like it fixes the original issue!

bennothommo:
@w20k @datune I've tested it again with a clean copy and it seems to be all working.

PR #4337, opened 2019-05-16
Make plugin dependency checks case-insensitive

tobias-kuendig:
Currently, if I add a `$require = ['Rainlab.Blog']` to my plugin class, the backend will always report that the blog plugin is missing. This is because the correct plugin code is `RainLab.Blog`, note the upper-case `L`.

Other methods of checking plugins in the PluginManager (`exists`, `findByIdentifier` etc) are case-insensitive. 

This PR makes the dependency check case-insensitive as well.

One thing I noticed: The `normalizeIdentifier` method currently iterates over all plugins to compare a lower-case representation of the plugin code. This could be optimized by keeping a map of lower-case to "normalized" plugin codes and simply use `isset()`.

If you guys feel this would be a good optimization I am happy to submit a PR for it. Since the PluginManager code is run on every request I feel like it should be as performant as possible. On the other hand, most installations won't have that many plugins and together with opcache this might be regarded as "micro-optimization".


bennothommo:
@tobias-kuendig I have added a test case for the particular issue you are fixing here. Please see https://github.com/octobercms/october/pull/4337

Would you mind merging `develop` into your branch, testing your fix against this case, and adjusting the test case as necessary?

tobias-kuendig:
I had to change the return values of the `hasPlugin` method with wrong case since those are now detected as available:

https://github.com/octobercms/october/pull/4337/commits/cc5ac727ea999edde17d8a5ee5dad3c1ff71a051

bennothommo:
@tobias-kuendig Cool - I was expecting that to occur :)

Would you mind also fixing these code quality errors as well, and then I'll be happy to merge this?
https://travis-ci.org/octobercms/october/jobs/553168742#L793-L795

tobias-kuendig:
Done!

PR #4340, opened 2019-05-20
Make contributions easier by having an online IDE (Gitpod) for October

meysholdt:
hi there 👋!

I got inspired by your statement 
> October is [...] whose sole purpose is to make your development workflow simple again.

and I thought I can contribute to that mission: Eliminate the setup effort for first-time visitors, contributors and developers. I believe this can make it much easier for contributors to get the source up-and-running and make their first contribution. 

The core idea here is to specify the dev-environment as code and store it in git, so that it's available to everybody who wants to work with the code. This PR contributes a dev-environment-as-code, namely the files `.gitpod.yml` and `.gitpod.dockerfile`. The IDE of [Gitpod](https://www.gitpod.io) has full language support for PHP.

To try, click on 
https://gitpod.io/#https://github.com/meysholdt/october

and see
1. a new dev environment (a docker container) launches on gitpod.io
2. git-clone checks out this repo. (GitHub OAuth required)
3. `composer install` runs
4. `apachectl start` runs
5. a preview opens to show the October Demo.

this is how it looks like:

image

Disclosure: I work on Gitpod.

I noticed that `composer install` takes several minutes to run... but we can avoid that people have to wait for it. Gitpod can run it on git-push, just like Travis/Jenkins would do it, when you configure the [Gitpod App](https://github.com/apps/gitpod-io/) on this repo. Then, when you open a workspace, `composer install` will already have completed and all dependencies are available so that the demo can launch instantaneously. 

Let me know what you think.

PR #4341, opened 2019-05-21
Make contributions easier by having an online IDE (Gitpod) for October (reuse Dockerfile)

meysholdt:
In https://github.com/octobercms/october/pull/4340 I proposed an online IDE ([gitpod.io](http://www.gitpod.io)) for October and @LukeTowers asked:

>  Does the docker file have to be the one you specified or could you use [aspendigital/octobercms](https://hub.docker.com/r/aspendigital/octobercms/)?

I would like to answer with this PR: 
yes, we can absolutely use your Dockerfile instead of using [Gitpod's workspace-full image](https://github.com/gitpod-io/workspace-images/blob/master/full/Dockerfile).

Some small modifications make it usable in Gitpod:
* don't do the `git clone` in the Dockerfile, because we should not need a new docker-build for every commit/branch/tag. Gitpod runs `git clone` on a volume that mounted into the workspace-container that that's backed by an SSD drive.
* move the command that deal with the source code into `.gitpod.yaml`, so they are executed after `git clone`
* create a Linux user named `gitpod`
* configure Apache. 

See this PR's commits for details.

Gitpod builds the workspace-container on-demand, but only when the Dockerfile has changed. 

To try, click on this link. GitHub OAuth is needed for git-clone.
https://gitpod.io/#https://github.com/meysholdt/october/tree/docker

This is how it looks like:
image

Let me know what you think :)

### DIFF
--- a/.gitpod/apache/ports.conf
+++ b/.gitpod/apache/ports.conf
@@ -1,0 +1,2 @@
+# In Gitpod, we want tu run Apache on port 8080
+Listen 8080
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,0 +1,21 @@
+image:
+  file: .gitpod/Dockerfile
+  context: .gitpod
+ports:
+  - port: 8080
+    onOpen: open-preview
+tasks:
+- init: |
+    composer install --no-interaction --prefer-dist --no-scripts
+    composer clearcache
+    cp -R /home/gitpod/octobercms-config-docker config/docker
+    echo 'config/docker' >> .git/info/exclude
+    echo 'plugins/october/drivers' >> .git/info/exclude
+    echo "APP_ENV=docker" > .env
+    echo "OCTOBERCMS_DEVELOP_COMMIT=$(git rev-parse HEAD)" >> .env
+    touch storage/database.sqlite
+    chmod 666 storage/database.sqlite
+    php artisan october:up
+    php artisan plugin:install october.drivers
+  command: |
+    apachectl start
--- a/.gitpod/Dockerfile
+++ b/.gitpod/Dockerfile
@@ -1,0 +1,75 @@
+FROM php:7.3-apache
+
+RUN apt-get update && apt-get install -y cron git-core jq unzip vim zip \
+  libjpeg-dev libpng-dev libpq-dev libsqlite3-dev libwebp-dev libzip-dev && \
+  rm -rf /var/lib/apt/lists/* && \
+  docker-php-ext-configure zip --with-libzip && \
+  docker-php-ext-configure gd --with-png-dir --with-jpeg-dir --with-webp-dir && \
+  docker-php-ext-install exif gd mysqli opcache pdo_pgsql pdo_mysql zip
+
+RUN { \
+    echo 'opcache.memory_consumption=128'; \
+    echo 'opcache.interned_strings_buffer=8'; \
+    echo 'opcache.max_accelerated_files=4000'; \
+    echo 'opcache.revalidate_freq=2'; \
+    echo 'opcache.fast_shutdown=1'; \
+    echo 'opcache.enable_cli=1'; \
+  } > /usr/local/etc/php/conf.d/docker-oc-opcache.ini
+
+RUN { \
+    echo 'log_errors=on'; \
+    echo 'display_errors=off'; \
+    echo 'upload_max_filesize=32M'; \
+    echo 'post_max_size=32M'; \
+    echo 'memory_limit=128M'; \
+  } > /usr/local/etc/php/conf.d/docker-oc-php.ini
+
+RUN pecl install xdebug
+
+RUN { \
+    echo "#zend_extension=$(find /usr/local/lib/php/extensions/ -name xdebug.so)" ; \
+    echo 'xdebug.remote_enable=on'; \
+    echo 'xdebug.remote_autostart=off'; \
+    echo 'xdebug.remote_host=host.docker.internal'; \
+  } > /usr/local/etc/php/conf.d/docker-xdebug-php.ini
+
+ENV COMPOSER_ALLOW_SUPERUSER=1
+
+RUN curl -sS https://getcomposer.org/installer | php -- --install-dir=/usr/local/bin --filename=composer && \
+  /usr/local/bin/composer global require hirak/prestissimo
+
+RUN a2enmod rewrite
+
+COPY config/docker /usr/src/octobercms-config-docker
+
+ENV OCTOBERCMS_TAG develop
+ENV OCTOBERCMS_DEVELOP_CHECKSUM e7ca6f6c6651cfb040f73c5a6dc3ea0633345e17
+ENV OCTOBERCMS_DEVELOP_COMMIT a9e198c8b4e4eaee2438cd89d2e3d6e3e6e93bd5
+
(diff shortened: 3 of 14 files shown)


LukeTowers:
Why do you have to copy the docker file? Can't you just base a new file on the other one? Also, why do you have to replicate the config files?

meysholdt:
> Why do you have to copy the docker file? Can't you just base a new file on the other one?

we could also extend [aspendigital/octobercms](https://hub.docker.com/r/aspendigital/octobercms/) and ignore or delete the `git clone` that's already in there. It's probably not harmful, so let me know if that's a better approach. Gitpod will create a fresh, clean `git clone` when you open a workspace. That's quick because it doesn't rebuild any containers and ensures you're on the right branch and have all the latest commits.

I kept the commit history of this PR clean, so commit https://github.com/octobercms/october/pull/4341/commits/99e60b59e9f60ed966f720cb2509b06b8af89de1 unveils the exact changes I made. See some commands moving from `Dockerfile` to `.gitpod.yml`.


> Also, why do you have to replicate the config files?

I did it because I found [october/config/dev](https://github.com/octobercms/october/tree/master/config/dev) to be empty. 

Alternatively, we could use 
a) [october/config](https://github.com/octobercms/october/tree/master/config) with custom `.env`.
b) if we extend [aspendigital/octobercms](https://hub.docker.com/r/aspendigital/octobercms/), use the ones from there. 

Do you have a recommendation?

petehalverson:
Gitpod looks very cool. Nice work! Thanks for taking the time to put together a PoC on how to leverage it for October CMS!

I created [labs.octodock.com](https://labs.octodock.com) over a year ago and have made it available to maintainers for testing Issues, PRs and plugins. It leverages our Docker image and uses some basic orchestration tools to make instances available for testing and sharing. 

While our solution is tailored to integrate October's plugin repos, it lacks an IDE and GitHub integration which Gitpod is doing well! `theia-ide` looks extremely well done. I'd like to try to integrate it into our solution.

Our goal has been to leverage what we've built to provide some service offerings (plugin + theme preview sites, short lived 'try now' sites and eventually web hosting) to help fund further development and maintenance of October CMS, which is why (I believe) @LukeTowers suggested we base it off the same Docker image.

petehalverson:
@meysholdt the aspendigital/octobercms image has the [`APP_ENV` set to `docker` by default](https://github.com/aspendigital/docker-octobercms/blob/master/Dockerfile.template#L52). The `docker/config` files are setup to leverage environment vars passed to the container at runtime, see [App Environment](https://github.com/aspendigital/docker-octobercms#app-environment)

petehalverson:
Also, there's some nuance to the [install process with `composer install` and `git reset`](https://github.com/aspendigital/docker-octobercms/blob/master/Dockerfile.develop.template#L53-L56).

Ultimately the goal we've had with [aspendigital/octobercms](https://github.com/aspendigital/docker-octobercms) has been to mirror what's delivered by other installation methods (e.g. wizard,  commandline) which requires the modules to match those in the repo instead of relying on what composer finds on packagist.

Since we're already building a develop image for every commit, I think it'd make sense to leverage that and copy files into the workspace.

petehalverson:
I submitted a PR to @meysholdt's [fork with some recommendations.](https://github.com/meysholdt/october/pull/1)

Give it a go: https://gitpod.io/#https://github.com/petehalverson/october/tree/docker

Before mainlining this, I think there are a number of issues to consider, e.g.:

- Educating developers on:
  - Docker environment config
  - Making changes to composer.json
  - Dealing with proxies
  - Other general nuances of working within a container

I think it'd be important to get some feedback from others not so accustomed to working within a container and proxies. I've run into occasional hiccups that are hard to explain if you don't understand your dev environment is running behind a proxy. There are the occasional issues that crop up that I forget someone not accustomed to using containers may get frustrated by...

Finally, I think I'd rather remove the `.gitpod/` folder and move the custom Dockerfile along with the Apache customization to the live along side the base image . That way the only thing we need to add to the project repo is a `.gitpod.yml` file:

```yaml
image: aspendigital/octobercms:gitpod
ports:
  - port: 8080
    onOpen: open-preview
tasks:
- init: ...
```

Bringing some additional container support to October CMS by default is something I'd love to see!

petehalverson:
@meysholdt Is there any way to avoid the custom Apache config and serve the preview on port 80? Seems cumbersome to force projects based on official images to insert such a work around...

meysholdt:
hi @petehalverson!

thank you for the compliments and picking up may changes. I'll try to respond to your comments here and on https://github.com/meysholdt/october/pull/1.

> I created labs.octodock.com over a year ago...

Awesome! I tried to log in but was rejected, I assume my email needs to be whitelisted. Preview/Demo environments and hosting are certainly valuable. If it can produce a revenue stream that supports maintaining October CMS... perfect!

Surely you cloud integrate Theia, it's designed for such use case. However, it's just the IDE part and doesn't help you with backend and integrations. 

A low-hanging fruit could be to allow users to easily deploy from Gitpod to labs.octodock.com. This way you make Preview/Demo environments and hosting great, which Gitpod solved the IDE+workspace part.



> Before mainlining this, I think there are a number of issues to consider, e.g.:
> 
> Educating developers on:
> Docker environment config
> Making changes to composer.json
> Dealing with proxies
> Other general nuances of working within a container

I was hoping ver little extra education is needed when working with Gitpod. Main points:
* "you have an online Linux environemnt"
* "the environment is created by running the commands from the `Dockefile`".

And that's basically it. No need to know about docker cli, networking, volumes, etc.

Knowledge about `composer.json` I assume is valuable regardless of whether Gitpod is in the mix.

What knowledge do they need about proxies? Gitpod's proxy doesn't cache, so everything should behave as usual. 

> Finally, I think I'd remove the .gitpod/ folder and move the custom Dockerfile along with the Apache customization to the live along side the base image.

Sure, sounds perfect. I'm afraid we can't completely get rid of the custom Apache config, because we need to set `ServerName`, Port, `DocumentRoot`, and Linux-User. Port 80 may work at some point, but not yet today. See https://github.com/gitpod-io/gitpod/issues/581.

petehalverson:
I'm not expecting a developer will need to know docker cli, networking, volumes, etc. However, not being aware your environment could be frustrating.

For example, there are a number of core CMS functions that log IPs. Not being aware that you're behind a proxy could make you wonder why those are no longer accurate. Setting the CMS config `backendForceSecure` as `true` breaks the backend login while setting it as `false` just works because the proxy doesn't connect to the container internally over HTTPS.

These are corner cases, but I'm sure there are more that I'm not thinking of...

By mainlining Gitpod and directing OC developers to use it for contributions, you're also asking maintainers to use and understand the new workflow and be able to troubleshoot problems. They will be on the receiving end of issues that may arise from adopting your solution.

petehalverson:
I should mention, this is the primary reason I haven't made Octodock publicly available. My concern was that I'd be introducing more problems for maintainers by dictating a workflow.

Ultimately, I am a strong advocate for a 'container first' approach to October CMS. If adopting Gitpod moves us further that direction, I'm all for it.

.gitpod/apache/ports.conf
```diff
@@ -0,0 +1,2 @@
+# In Gitpod, we want tu run Apache on port 8080
```
alxy: Typo `tu`

bennothommo:

Issue #4330, opened 2019-05-11
Taglist form widget does not have preview context


Rike-cz:
- OctoberCMS Build: 454
- PHP Version: 7.2

### Description:
Taglist form widget behavior is the same in create/update context and in preview context. Select box can be dropdowned and new item selected in preview context.
![tablist](https://user-images.githubusercontent.com/4247582/57570569-b3b1ab00-7403-11e9-8b7b-c46853f20be4.png)

### Expected behavior:
No dropdown, only list of saved tags are expected.

LukeTowers:
Feel free to make a PR to add support for this (would include readOnly & disabled support too)

bennothommo:
Fixed in build 447 (https://github.com/octobercms/october/commit/acf45772551150402139f7cb9bfcd740fc0da5a1)

petehalverson:
This fails when the field is read only or disabled and value is an array (e.g. belongsToMany relationship).

PR #4349, opened 2019-05-26
Add taglist preview mode

Rike-cz:
#4330 solution.

### DIFF
--- a/modules/system/assets/ui/less/select.less
+++ b/modules/system/assets/ui/less/select.less
@@ -589,24 +589,3 @@
 .form-inline .select2-container--default {
     display: inline-block;
 }
-
-// Taglist preview mode
-//------------------------------------
-
-.taglist--preview {
-    overflow: hidden;
-    list-style-type: none;
-    padding-top: 0; padding-left: 0;
-    margin: 0;
-
-    .taglist__item {
-        color: @color-select-text;
-        background: @color-select-choice-bg;
-        border: 1px solid @color-select-choice-border;
-        border-radius: 4px;
-        cursor: default;
-        float: left;
-        margin: @padding-base-vertical 0 0 @padding-base-horizontal/2;
-        padding: 0 (@padding-base-vertical - 2);
-    }
-}
--- a/modules/system/assets/ui/less/select.less
+++ b/modules/system/assets/ui/less/select.less
@@ -589,3 +589,24 @@
 .form-inline .select2-container--default {
     display: inline-block;
 }
+
+// Taglist preview mode
+//------------------------------------
+
+.taglist {
+    overflow: hidden;
+    list-style-type: none;
+    padding-top: 0; padding-left: 0;
+    margin: 0;
+    
+    .taglist__item {
+        color: @color-select-text;
+        background: @color-select-choice-bg;
+        border: 1px solid @color-select-choice-border;
+        border-radius: 4px;
+        cursor: default;
+        float: left;
+        margin: @padding-base-vertical 0 0 @padding-base-horizontal/2;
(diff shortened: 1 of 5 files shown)


bennothommo:
@Rike-cz I've tested this and it works well. The one thing I would suggest is that the LESS code be moved to an asset file within the widget files, or to the backend LESS files somewhere. Where it is currently (within the Select2 styling) isn't very obvious.

w20k:
@bennothommo why? Those styles are related to OCMS select2 improvements/customizations within the core. I don't see any point of moving those. 
I've also tested it. Works great, thanks @Rike-cz!

w20k:
The only thing I'd really want to change, based on the BEM methodology. @Rike-cz could add something like `--preview` postfix`.tag--preview` to differentiate the state.

bennothommo:
@w20k Put it this way - if we upgrade Select2 one day and have to update the CSS/LESS along with it, are we going to remember that the taglist styling is at the bottom of that file?

w20k:
@bennothommo we need to update, if nothing in structure will changes, this file '@import "../vendor/select2/css/select2.less";'

Rike-cz:
I think taglist is derived from Select2, so that's why we have stylesheets in one file.

modules/system/assets/ui/less/select.less
```diff
@@ -588,4 +588,25 @@
 
 .form-inline .select2-container--default {
     display: inline-block;
+}
+
+// Taglist preview mode
+//------------------------------------
+
+.taglist {
```
w20k: @Rike-cz could you add `--preview` posfix? Thanks!

bennothommo:
@w20k Sorry, I didn't see that it links to a vendor LESS file, I presumed `modules/system/assets/ui/less/select.less` had all the Select2 code in it. I still think it might make sense in its own LESS file just purely for separation of concerns (in preview mode, the taglist isn't a Select2), but I'm not going to stand in the way if you guys disagree with me :)

LukeTowers:
I don't think it would be a bad idea to have the taglist specific CSS in its own file under its own directory. Most of the other formwidgets operate that way so I think it would be fine to do it that way for this too. In this case it's a small enough addition that it's not a huge deal either way, but I would say that the "proper" way of doing it is to have the taglist specific stuff in its own file under the taglist formwidget folder structure.

bennothommo:
@Rike-cz Last step, could you run `artisan october:util compile less` and provide the compiled `storm.css` file as part of your PR? Should be good to merge once done. :)

w20k:
Just minor note, why I don't like the idea of separation, either we should move everything from `select2.less` or leave as it is. Because right now it's just an override to `select2.less` styles.

Few examples:
_Dorpdown_
https://github.com/octobercms/october/blob/dd964fe3991abebfadd6c280dd24f98089da2201/modules/system/assets/ui/less/select.less#L218
_Single Select_
https://github.com/octobercms/october/blob/dd964fe3991abebfadd6c280dd24f98089da2201/modules/system/assets/ui/less/select.less#L251
_Multi Select_
https://github.com/octobercms/october/blob/dd964fe3991abebfadd6c280dd24f98089da2201/modules/system/assets/ui/less/select.less#L251
And so on.

bennothommo:
@w20k I think those ones are fine because, as you said, they are overrides to the original Select2 styling to theme them for October. The taglist preview however is *not* a Select2 dropdown, it's simply styled in a way that it looks like a Select2. :)

w20k:
Sorry, @Rike-cz did my best, but @bennothommo is right on this one :) My bad, haven't reviewed `Php` part.

PR #4358, opened 2019-05-30
Use temporaryUrls for protected files if the storage driver supports them

LukeTowers:
Requires https://github.com/octobercms/library/pull/406. This vastly improves performance of interacting with protected files in the backend that are located on remote storage services (S3, Rackspace) that support the use of temporary URLs.

Should this behaviour be configurable? @bennothommo @w20k @daftspunk

### DIFF
--- a/modules/backend/controllers/Files.php
+++ b/modules/backend/controllers/Files.php
@@ -2,6 +2,7 @@
 
 use View;
 use Cache;
+use Config;
 use Backend;
 use Response;
 use System\Models\File as FileModel;
@@ -65,18 +66,18 @@ class Files extends Controller
         $url = null;
         $disk = $file->getDisk();
         $adapter = $disk->getAdapter();
-        if (is_a($adapter, 'League\Flysystem\Cached\CachedAdapter')) {
+        if ($adapter instanceof \League\Flysystem\Cached\CachedAdapter) {
             $adapter = $adapter->getAdapter();
         }
 
-        if (is_a($adapter, 'League\Flysystem\AwsS3v3\AwsS3Adapter') ||
-            is_a($adapter, 'League\Flysystem\Rackspace\RackspaceAdapter') ||
+        if (($adapter instanceof \League\Flysystem\AwsS3v3\AwsS3Adapter) ||
+            ($adapter instanceof \League\Flysystem\Rackspace\RackspaceAdapter) ||
             method_exists($adapter, 'getTemporaryUrl')
         ) {
             if (empty($path)) {
                 $path = $file->getDiskPath();
             }
-            $expires = now()->addMinutes(60);
+            $expires = now()->addSeconds(Config::get('cms.storage.uploads.temporaryUrlTTL', 3600));
 
             $url = Cache::remember('backend.file:' . $path, $expires, function () use ($disk, $path, $expires) {
                 return $disk->temporaryUrl($path, $expires);
--- a/modules/backend/controllers/Files.php
+++ b/modules/backend/controllers/Files.php
@@ -1,6 +1,7 @@
 <?php namespace Backend\Controllers;
 
 use View;
+use Cache;
 use Backend;
 use Response;
 use System\Models\File as FileModel;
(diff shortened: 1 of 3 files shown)


modules/backend/controllers/Files.php
```diff
@@ -51,14 +52,54 @@ public function thumb($code = null, $width = 100, $height = 100, $mode = 'auto',
...
+     * @param string|null $path Optional, defaults to the getDiskPath() of the file
+     * @return string|null
+     */
+    protected static function getTemporaryUrl($file, $path = null)
+    {
+        // Get the disk and adapter used
+        $url = null;
+        $disk = $file->getDisk();
+        $adapter = $disk->getAdapter();
+        if (is_a($adapter, 'League\Flysystem\Cached\CachedAdapter')) {
```
bennothommo: @LukeTowers Any thoughts on switching these calls to `instanceof`? I find they're more readable, but it's a matter of preference.

modules/backend/controllers/Files.php
```diff
@@ -51,14 +52,54 @@ public function thumb($code = null, $width = 100, $height = 100, $mode = 'auto',
...
+            $adapter = $adapter->getAdapter();
+        }
+
+        if (is_a($adapter, 'League\Flysystem\AwsS3v3\AwsS3Adapter') ||
```
bennothommo: @LukeTowers Should this condition be changed to just check for the `getTemporaryUrl` method? The other two class checks are probably redundant.

bennothommo:
@LukeTowers Looks nifty. The only thing I would ask is what is supposed to happen if a temporary URL cannot be generated - I would think in instances where if one cannot be generated, it shouldn't be accessible for security reasons.

To answer your original question, I think it should be configurable as well - presumably the usage of temporary URLs in the first place and how long they are cached for should be config options.

modules/backend/controllers/Files.php
LukeTowers: If you can verify that works, sure. I was mirroring https://github.com/laravel/framework/blob/5.5/src/Illuminate/Filesystem/FilesystemAdapter.php#L471-L479 which has the explicit checks. @bennothommo

LukeTowers:
@bennothommo The only reasons one couldn't be generated that I can think of is either your credentials are no longer valid (in which case it'll fallback to the proxy method and fail too, returning a 404) or the file doesn't exist (in which case it'll still fallback to the proxy method and fail resulting in another 404.

Any suggestions for config key names / descriptions?

modules/backend/controllers/Files.php
LukeTowers: 6 of one, half a dozen of another. Doesn't really matter to me, `instanceof` is used elsewhere in October and Laravel so that's a valid reason to do that. I was mostly just experimenting with `is_a()` because I recently rediscovered it today.

modules/backend/controllers/Files.php
bennothommo: @LukeTowers Ah ok - it seems those checks are for using different methods though (`getAwsTemporaryUrl` and `getRackspaceTemporaryUrl`). If you've got everything feeding through to the one `getTemporaryUrl` method, then it would just be a case of checking for the existence of that method, I would think.

LukeTowers: I don't know if AWS & Rackspace adaptors have that method though, the laravel code doesn't call those methods on those adaptors.

bennothommo:
@LukeTowers Sounds good.

For the config values, maybe `cms.storage.uploads.allowTemporaryUrls` (boolean, default = `true`) and `cms.storage.uploads.temporaryUrlCacheTime` (integer, default = 60).

modules/backend/controllers/Files.php
LukeTowers: Looked into it more, they don't have those methods on the adaptor itself.

bennothommo: Yep, my mistake, I see how it all works now. Your implementation is good. :)

bennothommo:
@LukeTowers Found an issue, funnily enough, when trying to run `artisan october:env`.

```
[ErrorException]                                                                                                                     
  Declaration of System\Models\File::getPath() should be compatible with October\Rain\Database\Attach\File::getPath($fileName = NULL)
```

modules/system/models/File.php
```diff
@@ -42,13 +42,13 @@ public function getThumb($width, $height, $options = [])
     /**
      * {@inheritDoc}
      */
-    public function getPath()
+    public function getPath($fileName = null)
     {
         $url = '';
```
w20k: @LukeTowers, minor note: could we return value and not save it to a local variable `$url`, that's not even needed)

LukeTowers: Nah, I usually prefer to initialize a default value to return instead of having multiple return statements. Not really a big deal here, but I don't see a reason to change it.

PR #4360, opened 2019-05-31
Store limited list widget configuration in user preferences

### LINKED ISSUE #4357: Persist user's list settings even after logout from backend
- OctoberCMS Build: 455
- PHP Version: 7.2.17
- Plugins Installed: none

### Expected Behavior
When I customize the list settings of any backend list (like show/hide columns, customize the order they appear or sort a column), I expect this list settings to persist.

Bildschirmfoto von 2019-05-30 08-06-59

### Actal Behavior
If I log out from the backend and log back in, my list settings are reset to the defaults.

### Steps To Reproduce:
- fresh october cms install (installer or composer doesn't matter)
- go to a backend page with a list (e.g. Settings/Administrators)
- change any list setting (e.g. show/hide column)
- log out from the backend
- log back in
- visite the same backend page with list and see it in its default glory

### Did this ever work?
Yes. I was not sure if this was always like that and I just never logged out of the backend, but I checked with build 446 and there the list settings are persisted even when I log out.
So I assume that somewhere between 446 and 455 this bug was introduced.

bennothommo:
Fixes #4357.

This change stores the visible columns and item count options from the List configuration tool in the user preferences, thereby allowing the preferences to remain when the user logs out of October.


mjauvin:
Oh, I like this a lot @bennothommo !!!

mjauvin:
@bennothommo is there a way to "clear" this from the user preferences and restore "default" values?

bennothommo:
Thanks, @mjauvin. I'm at this moment implementing that very feature :P

LukeTowers:
@bennothommo why is it necessary to have a dependency on a `controller` property?

bennothommo:
@LukeTowers I wanted this to have feature parity with `Backend\Traits\SessionMaker` so that it could replace any session preference calls as necessary.

modules/backend/traits/PreferenceMaker.php
```diff
@@ -0,0 +1,136 @@
...
+     * @param string $key Unique key for the data store.
+     * @param mixed $value The value to store.
+     * @return void
+     */
+    public function putUserPreference(string $key, $value): void
+    {
+        $preferences = $this->getUserPreferences();
+        $preferences[$key] = $value;
+
+        $this->getStorage()::forUser()->set($this->getPreferenceKey(), $preferences);
```
LukeTowers: If the method is meant to be overrideable, should it be getPreferenceStorage()? Also, could it return an object instead of a class string?

bennothommo: Good points - it's now changed.

PR #4362, opened 2019-06-03
This fixes the issue with menus not being displayed in pages plugin

SebastiaanKloos:
@LukeTowers requested this pull request in [this](https://github.com/rainlab/pages-plugin/issues/396#issuecomment-498092719) conversation.

I don't know why, but this change is solving the issue of menus from the pages plugin not being displayed on my websites. Maybe it has something to do with my PHP version. I'm using PHP 7.3.4.

### DIFF
--- a/modules/cms/classes/Meta.php
+++ b/modules/cms/classes/Meta.php
@@ -42,7 +42,7 @@ class Meta extends CmsObject
             $this->content = $this->renderContent();
         });
         $this->bindEvent('model.afterFetch', function () {
-            $this->attributes = array_merge($this->parseContent(), $this->attributes);
+            $this->attributes = array_merge($this->attributes, $this->parseContent());
         });
     }
 


bennothommo:
@SebastiaanKloos I have tested this fix. While it does indeed fix the issue for static menus, it prevents the deletion of static pages from being stored in the database, as that too is a meta file (`meta/static-pages.yaml`). With this fix in place, deleted pages remain in the pages list.

I think a more suitable fix might involve just simply using the database YAML data for meta files if they are available, and completely ignoring the filesystem meta files.

Would you be willing to investigate this further for us?

bennothommo:
@SebastiaanKloos My mistake - the page issue I found seemed to be a caching issue. Your fix works as expected. I'll do a little more testing tonight and will merge if it's all good.

datune:
I tested this fix and can confirm it works. @LukeTowers Please merge this, thanks!

GinoPane:
I think that plugin is connected to the core too tight. We even have a setting in cms config specifically for the plugin, but actually it should belong to the plugin itself and be somewhere in admin.

LukeTowers:
@GinoPane the setting is not for the plugin, the plugin just uses the setting if it exists.

GinoPane:
@LukeTowers ah, ok, makes sense.

PR #4363, opened 2019-06-03
Change ExportModel download to use cache and streamed response

### LINKED ISSUE #4232: Temp folder is local only
- OctoberCMS Build: 447
- PHP Version: 7.3

### Description:
Exporting CSV file creates file in Temp folder but folder isnt shared between servers then sometimes it returns error of missing file. Same issue is with thumbnails for MediaLibrary #3275 

### Steps To Reproduce:
Use multiple servers for website and export CSV in backend or open MediaLibrary to see thumbs are missing sometimes.

Samuell1:
Fixes #4232 CSV export issue with multiple instances of server

### DIFF
--- a/modules/backend/models/ExportModel.php
+++ b/modules/backend/models/ExportModel.php
@@ -7,6 +7,8 @@ use Response;
 use League\Csv\Writer as CsvWriter;
 use ApplicationException;
 use SplTempFileObject;
+use Cache;
+use Symfony\Component\HttpFoundation\StreamedResponse;
 
 /**
  * Model used for exporting data
@@ -53,16 +55,17 @@ abstract class ExportModel extends Model
      */
     public function download($name, $outputName = null)
     {
-        if (!preg_match('/^oc[0-9a-z]*$/i', $name)) {
-            throw new ApplicationException(Lang::get('backend::lang.import_export.file_not_found_error'));
-        }
-
-        $csvPath = temp_path() . '/' . $name;
-        if (!file_exists($csvPath)) {
-            throw new ApplicationException(Lang::get('backend::lang.import_export.file_not_found_error'));
-        }
-
-        return Response::download($csvPath, $outputName)->deleteFileAfterSend(true);
+        $cachedExport = Cache::get('export::' . $name);
+        return new StreamedResponse(
+            function () use ($cachedExport) {
+                echo $cachedExport;
+            },
+            200,
+            [
+                'Content-Type' => 'text/csv',
+                'Content-Disposition' => 'attachment; filename="' . $outputName . '"',
+            ]
+        );
     }
 
     /**
@@ -135,13 +138,10 @@ abstract class ExportModel extends Model
         }
 
         /*
-         * Save for download
+         * Save in cache for download
          */
         $csvName = uniqid('oc');
-        $csvPath = temp_path().'/'.$csvName;
-        $output = $csv->__toString();
-
-        File::put($csvPath, $output);
+        Cache::put('export::'.$csvName, $csv->__toString(), 5);
 
         return $csvName;
     }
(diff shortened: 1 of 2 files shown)


modules/backend/models/ExportModel.php
```diff
@@ -53,16 +55,17 @@ public function export($columns, $options)
...
-        $csvPath = temp_path() . '/' . $name;
-        if (!file_exists($csvPath)) {
-            throw new ApplicationException(Lang::get('backend::lang.import_export.file_not_found_error'));
-        }
-
-        return Response::download($csvPath, $outputName)->deleteFileAfterSend(true);
+        $cachedExport = Cache::get('export::' . $name);
+        return new StreamedResponse(
+            function () use ($cachedExport) {
+                echo $cachedExport;
```
bennothommo: @Samuell1 Could this just use the `Cache::get` call above? You'll need to change the `use` parameter above to use `$name`.
```suggestion
                echo Cache::get('export::' . $name);
```

modules/backend/models/ExportModel.php
```diff
@@ -53,16 +55,17 @@ public function export($columns, $options)
...
+            },
+            200,
+            [
+                'Content-Type' => 'text/csv',
+                'Content-Disposition' => 'attachment; filename="' . $outputName . '"',
```
bennothommo: @Samuell1 You will need to add some handling to have a fallback value for `$outputName` if it is `null`, otherwise it will name the file as `null` when downloading, I believe. The old `Response::download()` call would use the filename of the source file if an `$outputName` was not specified.

modules/backend/models/ExportModel.php
```diff
@@ -135,13 +138,10 @@ protected function processExportData($columns, $results, $options)
...
         /*
-         * Save for download
+         * Save in cache for download
          */
         $csvName = uniqid('oc');
-        $csvPath = temp_path().'/'.$csvName;
-        $output = $csv->__toString();
-
-        File::put($csvPath, $output);
+        Cache::put('export::'.$csvName, $csv->__toString(), 5);
```
bennothommo: @Samuell1 Please add space between concatenation parts.

```suggestion
        Cache::put('export::' . $csvName, $csv->__toString(), 5);
```

bennothommo:
@Samuell1 You may want to look into adding a fix for this test too:
https://github.com/octobercms/october/blob/7590dfd30fb08242c6c8e111115659d38ac4aeb8/tests/unit/backend/models/ExportModelTest.php#L49

It is expecting the downloaded file to have a content type of `text/plain`, however, your changes return the content type of `text/csv`. I think both are valid so maybe the test case should be updated to check for either one.

modules/backend/models/ExportModel.php
Samuell1: @bennothommo Old Response::download() is using export.csv as $outputName, because its automatically passed from download controller.

bennothommo:
@Samuell1 thanks for making those changes. Could you take a look at the tests again though, still seems to be failing - Symfony is throwing an exception.

modules/backend/models/ExportModel.php
bennothommo: @Samuell1 No problem, if that's the case, that works.

Samuell1:
I think this needs to be just tested one more time to be 100% sure nothing is broken. I tested it before with blog plugin and it works fine. Anybody can test it with more rows in table?

w20k:
Might give have a minute today, later on. Will take a look @Samuell1.

LukeTowers:
Let's put this in 458 instead, I want to get 457 finalized and this seems like a risky change.

Samuell1:
@LukeTowers Hope it will be soon because after this change i have plans to implement chunking to allow exporting more items then few hundreds.

daftspunk:
There is a concern with writing such a large file through the `Cache` engine. I understand why you don't want this to use the file system since it will not carry across with load balancing / multiple servers

I should point out there is nothing stopping you from using this approach in your own projects without necessarily needing it in the core

PR #4364, opened 2019-06-03
Hide popup loading indicator if an error/flash is thrown from an AJAX handler

bennothommo:
Fixes https://github.com/rainlab/builder-plugin/issues/283

A popup widget using an AJAX handler will display a loading indicator whilst waiting for a response. If an error or a flash message is returned by the AJAX handler, the error is displayed correctly to the user, but the loading indicator and modal backdrop are not removed as the modal is technically not "open" at this stage.

This fix resolves this by tracking the loading state of the popup and removing the loading indicator and backdrop if an error occurs during load.

### DIFF
--- a/modules/system/assets/ui/js/popup.js
+++ b/modules/system/assets/ui/js/popup.js
@@ -86,10 +86,10 @@
                 },
                 error: function(jqXHR, textStatus, errorThrown) {
                     this.error(jqXHR, textStatus, errorThrown).done(function(){
-                        self.hide()
                         if (self.isLoading) {
-                            self.setLoading(false)
-                            setTimeout(function(){ self.setBackdrop(false) }, 100)
+                            self.hideLoading();
+                        } else {
+                            self.hide()
                         }
                         self.triggerEvent('popupError') // Deprecated
                         self.triggerEvent('error.oc.popup')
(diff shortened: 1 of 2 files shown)


modules/system/assets/ui/js/popup.js
```diff
@@ -86,6 +87,10 @@
                 error: function(jqXHR, textStatus, errorThrown) {
                     this.error(jqXHR, textStatus, errorThrown).done(function(){
                         self.hide()
+                        if (self.isLoading) {
```
w20k: @bennothommo I think you could just use `hideLoading()` here.

```
if (self.isLoading) {
   self.hideLoading()
}
```

bennothommo: @w20k Good point - changed! :)

w20k:
@ovillani could you test this PR?

bennothommo:
@w20k Could you confirm this works for you? If so, I'll merge it. :)

w20k:
@bennothommo will try 😄

w20k:
@bennothommo, later on, will do the testing!

w20k:
@bennothommo  - done. It works!

Screenshot:
Screenshot 2019-06-11 at 13 45 11

ovillani:
Great, thanks so much!

PR #4373, opened 2019-06-06
Add support for filters in pivot mode

mplodowski:
A little improvement to PR https://github.com/octobercms/october/pull/4241

### DIFF
--- a/modules/backend/behaviors/relationcontroller/partials/_manage_pivot.htm
+++ b/modules/backend/behaviors/relationcontroller/partials/_manage_pivot.htm
@@ -13,6 +13,9 @@
             <?php if ($relationSearchWidget): ?>
                 <?= $relationSearchWidget->render() ?>
             <?php endif ?>
+            <?php if ($relationManageFilterWidget): ?>
+                <?= $relationManageFilterWidget->render() ?>
+            <?php endif ?>
             <?= $relationManageWidget->render() ?>
         </div>
 


LukeTowers:
@alxy Could you test this?

alxy:
@LukeTowers May be able to test this out next week.

mplodowski:
I found two bugs at the moment, but I was unable to fix them.

The first seems to be related to original commit, when you try to search inside filter, it throws an error:

```
"A widget with class name 'relationProductsManageFilter' has not been bound to the controller" on line 598 of /modules/backend/Classes/Controller.php
```

The second is related to `number` and `numberrange` filter types. If the relation manager is opened inside popup window, you cannot set value of those inputs, it looks like they are disabled.

alxy:
@mplodowski Do you get the first behavior even if not in pivot mode? These issues would need to be resolved, before this can be merge, however I won't be able to do this anytime soon.

The second issue seems to be javascript related - I remember having simlar issues as well.

mplodowski:
@alxy Yes, the first behavior occurs even without pivot.

I will try to take a closer look into this soon.

alxy:
@mplodowski You would probably apply this to the numberfilter js code as well: https://github.com/octobercms/october/pull/4204/commits/2cf7fbef27c166ff5a50f965dcb24432b1b3f5b7

mplodowski:
@alxy I don't know what I should apply exactly? 

Can you help me to make it work?

alxy:
Sorry, the number filter has been fixed by me, at least for the normal case, where the relation controller is opened on a regular page. (In the initial implementation the filters opened behing the manage popup. If there are now multiple popups which overlay each other, things get more complex....)

For the other error I have no idea what is happening there....

PR #4377, opened 2019-06-09
Added 'selectable' column type

### LINKED ISSUE #2510: Define "array" list column type, similar to dropdown form field
Hello Guys,

I am using OctoberCMS and using one of its awesome plugins called as "Builder" http://octobercms.com/plugin/rainlab-builder and so far so good.

But i m stuck at one stage and confused how to achive this. This is my scenario below.

I have created one new field called as "status" in my one of the plugins in which i have key values like (0=>Inactive, 1=>Active) .. this works fine as i am able to insert or update values based on my status selection..

But the thing is, when i go to listing page, i have selected a "status" field but its showing me values 0 or 1 instead of Active or Inactive.

How can i show Active or Inactive instead of 0 and 1 using builder plugin or any other way.. I need to be able to search and sort as well ..

Can anyone guide me for the same ?

Thanks

mohsin:
The 'dropdown', 'radio', 'balloon-selector' and ‘checkboxlist’ type form fields do not have a list form field equivalent and so they show the internally stored key instead of the translated value. So, a new column type called 'selectable' is introduced here to overcome this limitation.

The workaround for this limitation was [mentioned on the OctoberCMS forum](https://octobercms.com/forum/post/dropdown-shows-its-value-than-key-name-in-list-controller#post-12382) but I found it unsatisfactory as it's a global model-level attribute conversion which doesn't solve the issue--using that technique breaks the fields at the form level as their update forms now gets the converted value and so the dropdown, balloon-selector, etc. fields show the incorrect value. Also, when writing RESTful APIs using OctoberCMS, the model runs the same attribute conversions to show the translated value than the key which isn't effective either--most RESTful APIs actually send the keys itself to the client device and the value's are filled in at the client-level as it's more network efficient that way, so that's another disadvantage of using the method, that was suggested in the forum. A backend list-level solution is what's needed and this new column type helps to solve that.

Fixes #2510.


LukeTowers:
@SaifurRahmanMohsin I'm assuming you copied the getOptionsFromModel from the Form widget? Perhaps we should create a new `\Backend\Traits\ModelOptions` trait to be used by the Form widget and the Lists widget and then put that code there.

SaifurRahmanMohsin:
That is correct but I removed a few parts that were specific to the FormController. Perhaps it might not be the best to make a common trait for this. You can check and decide on that.

SaifurRahmanMohsin:
Yes, that is probably a good idea.

LukeTowers:
@SaifurRahmanMohsin feel free to make that change on your branch so it shows up here.

SaifurRahmanMohsin:
There is just one issue with that... I remember editing some code that is not common to both. I’ll check and see if I can make a common trait, if not I’ll comment back here.

modules/backend/widgets/Lists.php
```diff
@@ -1698,6 +1710,84 @@ public function onToggleTreeNode()
...
+     * @param $field
+     * @param $fieldOptions
+     * @return mixed
+     */
+    protected function getOptionsFromModel($field, $fieldOptions, $fieldName)
+    {
+        /*
+         * Advanced usage, supplied options are callable
+         */
+        if (is_array($fieldOptions) && is_callable($fieldOptions)) {
```
bennothommo: @SaifurRahmanMohsin Could you just check if `$fieldOptions` is "callable" here?
```suggestion
        if (is_callable($fieldOptions)) {
```

modules/backend/widgets/Lists.php
```diff
@@ -1698,6 +1710,84 @@ public function onToggleTreeNode()
...
+            $fieldOptions = call_user_func($fieldOptions, $this, $field);
+        }
+
+        /*
+         * Refer to the model method or any of its behaviors
+         */
+        if (!is_array($fieldOptions) && !$fieldOptions) {
+
+            $methodName = 'get'.studly_case($fieldName).'Options';
```
bennothommo: @SaifurRahmanMohsin Please add spaces between concatenation portions.
```suggestion
            $methodName = 'get' . studly_case($fieldName) . 'Options';
```

modules/backend/widgets/Lists.php
SaifurRahmanMohsin: Yep, this works if you define the options field in the list fields config as a reference to a method in the model that provides the array of options.

LukeTowers: @bennothommo doesn't that have the potential to backfire (or at the very least introduce yet another way to load options), say if someone specified `options: global_function_here`?

bennothommo: @LukeTowers The same could happen with an array too. Perhaps this whole block should go.

mohsin: @bennothommo $fieldOptions is callable if you specify it in an array format as mentioned [here.](/octobercms/docs/pull/373/files)

bennothommo: @SaifurRahmanMohsin Thanks for that, you are correct. You can leave your code as is.

mohsin: Ok @bennothommo. Thanks anyway for mentioning this coz I realized that there is a 5th way of specifying the options that hasn't been documented.

SaifurRahmanMohsin:
@LukeTowers Hey, I've moved getOptionsFromModel to its own trait as you've suggested and made the List and Form widgets implement them. I've already tested it with the dropdown widget for form and list and it doesn't break anything but could you cross verify that it works fine at your end as well to ensure I didn't mess anything else up? Thanks

modules/backend/widgets/Lists.php
```diff
@@ -1154,6 +1155,18 @@ protected function evalCustomListType($type, $record, $column, $value)
...
     }
 
+    /**
+     * Process as selectable value types for 'dropdown', 'radio', 'checkboxlist', 'balloon-selector' form field types.
+     */
+    protected function evalSelectableTypeValue($record, $column, $value)
+    {
+        $fieldName = $column->columnName;
+        $fieldOptions = $column->config['options'] ?? null;
+        $fieldOptions = $this->getOptionsFromModel($record, $fieldOptions, $fieldName);
```
bennothommo: @SaifurRahmanMohsin You could maybe condense this down to one method call and remove the two lines above it?
```suggestion
        $fieldOptions = $this->getOptionsFromModel(
            $record,
            $column->config['options'] ?? null,
            $column->columnName
        );
```

modules/backend/widgets/Lists.php
```diff
@@ -1154,6 +1155,18 @@ protected function evalCustomListType($type, $record, $column, $value)
...
+        return $fieldOptions[$value];
```
bennothommo: @SaifurRahmanMohsin This should return null if the value does not exist.
```suggestion
        return $fieldOptions[$value] ?? null;
```

modules/backend/traits/ModelOptions.php
```diff
@@ -0,0 +1,92 @@
...
+                    list($model, $attribute) = $field->resolveModelAttribute($this->model, $field->fieldName);
+                }
+                catch (Exception $ex) {
+                    throw new ApplicationException(Lang::get('backend::lang.field.options_method_invalid_model', [
+                        'model' => get_class($model),
+                        'field' => $field->fieldName
+                    ]));
+                }
+            } else {
+                $methodName = 'get'.studly_case($fieldName).'Options';
```
bennothommo: @SaifurRahmanMohsin Please add spaces around the full stops.
```suggestion
                $methodName = 'get' . studly_case($fieldName) . 'Options';
```

SaifurRahmanMohsin:
@bennothommo I have verified your suggestions and they're all good and I've included them in the PR. Thank you.

bennothommo:
Thanks for that @SaifurRahmanMohsin. Just one last change, would you mind removing these 2 lines, as they won't be needed with my suggested change:
https://github.com/octobercms/october/blob/bdeda4d8a1a28ba25b912b1396bad41ec0f0c62b/modules/backend/widgets/Lists.php#L1163-L1164

modules/backend/traits/ModelOptions.php
```diff
@@ -0,0 +1,92 @@
+<?php namespace Backend\Traits;
+
+/**
+ * Model Options Trait
+ * Implements helper methods useful to extract the options from a given field
+ * that is used in the List and Form widgets.
+ */
+
```
LukeTowers: get rid of this empty line of whitespace please

modules/backend/traits/ModelOptions.php
```diff
@@ -0,0 +1,92 @@
...
+trait ModelOptions
+{
+    /**
+     * Looks at the model for defined options.
+     *
+     * @param $field
+     * @param $fieldOptions
+     * @return mixed
+     */
+    protected function getOptionsFromModel($field, $fieldOptions, $fieldName = '')
```
LukeTowers: Can you make these variable names more generic, less tied to fields specifically to reflect the fact that this method is used by multiple widgets?

SaifurRahmanMohsin: Ok, let me see what can be done.

mohsin: I'm trying to figure out what name should I give to make it more generic to widgets. I'm stuck between elements and attributes. The latter seems to be something specific to eloquent models in general and there's also a variable called attribute used in the code. Do you have any suggestion?

modules/backend/traits/ModelOptions.php
```diff
@@ -0,0 +1,92 @@
...
+                }
+                catch (Exception $ex) {
+                    throw new ApplicationException(Lang::get('backend::lang.field.options_method_invalid_model', [
+                        'model' => get_class($model),
+                        'field' => $field->fieldName
+                    ]));
+                }
+            } else {
+                $methodName = 'get' . studly_case($fieldName) . 'Options';
+                $model = $this -> model;
```
LukeTowers: woah, no whitespace between arrow and object please, should be: `$this->model`

modules/backend/traits/ModelOptions.php
```diff
@@ -0,0 +1,92 @@
...
+            }
+
+            if (
+                !$this->objectMethodExists($model, $methodName) &&
+                !$this->objectMethodExists($model, 'getDropdownOptions')
+            ) {
+                throw new ApplicationException(Lang::get('backend::lang.field.options_method_not_exists', [
+                    'model'  => get_class($model),
+                    'method' => $methodName,
+                    'field'  => $field->fieldName
```
LukeTowers: Does this line work in the Lists widget?

modules/backend/traits/ModelOptions.php
LukeTowers: `$fieldOptions` can just be `$options`, however the more I look at this the less I'm certain that it even works for the list type given `->resolveModelAttribute($this->model)` being called on a Model instance and `$field->fieldName` being used in multiple spots when the lists widget isn't sending a field instance it's sending a model instance. Please look into this further and make it more generic / better support both uses.

modules/backend/traits/ModelOptions.php
mohsin: I've made some modifications that fixes this to work on both widgets. Will post that in a bit.

SaifurRahmanMohsin:
@LukeTowers I have tested the code using 5 different possible ways of defining the field values for dropdown, radio, and balloon selector fields (4 of which are mentioned in the
OctoberCMS documentation) and everything works fine so I guess this is good to go.

modules/backend/traits/ModelOptions.php
```diff
@@ -0,0 +1,97 @@
...
+ * Model Options Trait
+ * Implements helper methods useful to extract the options from a given field
+ * that is used in the List and Form widgets.
+ */
+trait ModelOptions
+{
+    /**
+     * Looks at the model for defined options.
+     *
+     * @param $field
```
LukeTowers: There's still a conflict here in that Form is passing a `Backend\Classes\FormField` instance and Lists is passing a `October\Rain\Database\Model` instance - these two options are not identical to each other so they need to be treated differently in the method itself.

LukeTowers (changes requested):

modules/backend/traits/ModelOptions.php
mohsin: Yeah, I knew this but it worked fine so didn't want to change it. I'll see what can be done...

LukeTowers: It works fine for now but there are code branches that treat $field as a Field instead of a model which can cause issues if those branches are reached under the wrong conditions.

mohsin: @LukeTowers I do not think of a better way to make getOptionsFromModel method more generic as Lists and Forms use different classes for their fields i.e. Lists uses ListColumn and Form uses FormField. One solution would be to include the following code:
```
if ($this instanceof \Backend\Widgets\Lists) {
    $field = $this->getColumn($fieldName);
} else if ($this instanceof \Backend\Widgets\Form) {
    $field = $this->getField($fieldName);
}
```
at the top of the method to generalise the field variable. The other alternative was allowing mixed variables to be passed to the method and I've done exactly this in my last commit: the form sends a variable of type FormField and the list sends variable of type ListColumn. I've modified the internal code in the getOptionsFromModel so that it can handle both these field types ideally: the issue I faced was that ListColumn did not have a $value variable so I modified the code to use the method getValueFromData which is common to both types and has similar method signature. So I've generalised this as much as possible. Let me know if this is alright, or suggest me a better way of solving this.

SaifurRahmanMohsin:
@LukeTowers I've also made [a plugin](https://github.com/SaifurRahmanMohsin/oc-selectable-plugin) to test the selectable field code for the various ways of defining the values. You may use it to test if you'd like.

bennothommo:
Thanks for your work on this @SaifurRahmanMohsin. We're going to have to hold merging this until build 461 as we are trying to expedite build 460 being released, so we are just focusing on bug fixes for 460 now.

bennothommo:
@SaifurRahmanMohsin I'm happy to try and get this implemented in the next release (or the one after it, depending on when the Laravel 6 upgrades drop), if you can resolve the merge conflict above, and look into providing some unit tests that can test extracting the value from keys (ie. what would end up being shown in the List widget when using your methods).

Would you be interested in having a go at that?

bennothommo:
@SaifurRahmanMohsin are you still interested in finalising this PR? If so, would you mind reviewing [this comment](https://github.com/octobercms/october/pull/4377#issuecomment-592392130) and see if you can implement those final changes?

bennothommo:
@SaifurRahmanMohsin @LukeTowers I've been testing this out, as I have a project that will find some use for it, but it doesn't seem to support static field options from a `fields.yaml` file, which I feel is going to be the more common scenario for most field setups. It only supports callbacks, either directly to the model or to an external class.

It is possible this is the only way it'll work, treating the model as a single source of truth for the options, given that technically you could have more that one form that populates the model data, each potentially having their own set of options available for a field.

EDIT: I should clarify - @SaifurRahmanMohsin has made it so that the column config also passes a `options` parameter to the `getOptionsFromModel` method, however, this means you have to set the same set of options in two locations and ensure they are in sync.

SaifurRahmanMohsin:
@bennothommo What do you mean by two locations? I remember setting this to take it from the defined model the same way the form's dropdown, balloon-selector, etc. fields take their values so it's already a single source of truth for both forms and column definitions. Am I missing something? Hmm

mohsin:
@bennothommo What do you mean by two locations? I remember setting this to take it from the form/list model class the same way the form's dropdown, balloon-selector, etc. fields take their values so it's already a single source of truth for both form and column definitions. Am I missing something? Hmm

bennothommo:
@SaifurRahmanMohsin what I mean is, say I have the following field set up in `fields.yaml` for the form:

```
status:
    label: Status
    type: dropdown
    options:
        pending: Pending
        processing: Processing
        completed: Complete
```

and then I wanted to use the `selectable` type for a column for that field in `columns.yaml` for a list:

```
status:
    label: Status
    type: selectable
```

With that sort of setup, it won't know what the options are, because the options are defined in the `fields.yaml` config for the *form*, not the *list*. You've made an allowance for this by specifying the options in the `columns.yaml` as well, like so:

```
status:
    label: Status
    type: selectable
    options:
        pending: Pending
        processing: Processing
        completed: Complete
```

But that means that we now have two locations for the options that have to be kept in sync. It's probably not so bad for fields with only a small number of options, but there is every chance that people will have fields with many static options. The only way the `selectable` column will work around this is to define the options in the model through the callbacks (ie. `getDropdownOptions` or `getStatusOptions`), but I believe there will be a lot of people who will try the `selectable` type with their forms and lists set up the way I initially said, and won't realise that they have to convert the static options to use one of the callbacks instead.

mohsin:
A workaround maybe to read the form config and take those options but I'm not sure if that's the "best practice" coz now we have a list reading a form config.

bennothommo:
@SaifurRahmanMohsin That's correct - it wouldn't take into account any events that might manipulate the form config (ie. `backend.form.extendFields`). You'd have to pretty much run the entire initialization process for a form to ensure the correct options are available, which would be overkill.

bennothommo:
All that having been said, I'm not against the way it is set up now. It certainly is faster than having to write out a partial for these types of columns and customising the partial to display the "nicer-looking" value, which is what I've had to do every time I've set up columns like this.

mohsin:
@bennothommo I am trying to write the unit test for this as you requested but I'm unable to figure out why the system behaves weirdly. The goal here is for me to call getColumnValue and assert that with "Developer" which should be the right returned column value, thus proving it works fine. But I encounter a weird issue that doesn't let me write the test.

I made a video showing what I'm facing here:
https://youtu.be/iE9Rg0uI5V0

And here's a ready-made zip having my files to test the same:
https://drive.google.com/file/d/195QqOCHmgiUD0yAmdWuQ_oa-ZhuZIpnm/view?usp=sharing

If you can help me fix this or point me in the right direction, I'll be happy to include the test as well to this PR.

bennothommo:
@SaifurRahmanMohsin could you add what you have for the tests into the changes for this PR, and I'll be able to have a look into it.

mohsin:
Ok, here's what I changed:
1. Set role_id to 1 [here](https://github.com/octobercms/october/blob/d2b3f865f164c6662217962dc4f356506633b801/tests/fixtures/backend/models/UserFixture.php#L24).
2. Added the selectable test:
```php
public function testSelectableType()
{
    $user = new UserFixture;
    $list = $this->listWithSelectableFieldFixture($user);
    $selectableColumn = $list->getColumn('role_id');

    // Test if the type is set to selectable
    $this->assertEquals($selectableColumn->type, 'selectable');

    // Test if the value is processed correctly
    $this->assertEquals($list->getColumnValue($user, $selectableColumn), 'Developer');
}

protected function listWithSelectableFieldFixture($record)
{
    return new Lists(null, [
        'model' => $record,
        'columns' => [
            'id' => [
                'type' => 'text',
                'label' => 'ID'
            ],
            'role_id' => [
                'type' => 'selectable',
                'label' => 'Role',
                'options' => ['Client', 'Developer']
            ]
        ]
    ]);
}
```

Here I'm re-using the UserFixture since that's a model readily available in the tests framework to test this out. The role_id returns 1 and so the selection would be 'Developer'. However, I'm facing the weird bug which I mentioned in my last comment as the `$list->getColumnValue($user, $selectableColumn);` fails--basically **$selectableColumn** isn't the correct ListColumn item i.e. role_id on checking inside the method.

bennothommo:
@SaifurRahmanMohsin I'd say it's because `role_id` is not fillable, as it's filled by the `role` relation. I'm having a look into this now.

mohsin:
[The weird behavior](https://youtu.be/iE9Rg0uI5V0) occurs even if I custom define a field besides role. But sure, please do check and tell me if there's a solution. I'll be happy to check in the test case and we can have this closed soon.

bennothommo:
@SaifurRahmanMohsin I've deployed a unit test that works into this PR for you.

SaifurRahmanMohsin:
@bennothommo Excellent. The test passes on my machine as well. Thank you. Let me know if anything is left or this PR can be merged.

mohsin:
@bennothommo Excellent. The test passes on my machine as well. Thank you. Let me know if anything is left, otherwise you can go ahead and merge the PR.

bennothommo:
@SaifurRahmanMohsin I've done some changes to this PR to improve support for both form fields and list columns - would you mind giving that a test and letting me know if that works?

@LukeTowers I've also added in changes to the callback methods to match up the signatures with the docs. Let me know your thoughts, or if you want me to roll them back.

LukeTowers:
@bennothommo could you highlight the changes to the API you've made?

bennothommo:
@LukeTowers I've given it some thought and I'm going to roll it back. While I like a consistent API, this is too much of a breaking change.

daftspunk:
🔥🔥🔥 This is a crazy good feature @SaifurRahmanMohsin. Thank you! It is clear a lot of work went into building it. Although upon peer review, this implementation is an example where the code liability exceeds the underlying benefit; in other words, it adds too much technical debt to justify the feature. It is a great piece of work, just not the right implementation...

However, after some refactoring, I was able to get it to work. The final approach is to acknowledge that the list column is leveraging a form field's behaviour. So the workflow goes something like this

Lists Widget -> List Column -> type: selectable -> Form Field -> getOptions

The options logic then had to be bumped from the Form widget to the Form Field, which is a natural progression anyway. Previously there was just no reason to do it, now we have one

This feature should arrive in October CMS once the next major release comes out

Thanks again 🙏 All the best

PR #4378, opened 2019-06-10
fixed test hard coded theme url

panakour:


w20k:
Looks good and tested! Thanks @panakour!

Issue #4384, opened 2019-06-12
Block off front-end Service Worker from running in the Backend as a Security Threat and Reduce Lookups


ayumihamsaki:
I've just seen some small issues with the code, where we could probably fix:

The backend signin screen has two files:

`modules/backend/assets/js/auth/auth.js`

with the code:

```javascript
$(document).ready(function(){
    $(document.body).removeClass('preload')

    $('form input[type=text], form input[type=password]').first().focus()
})
```

and the file

`modules/backend/assets/js/auth/uninstall-sw.js`

with the code:

```javascript
// Only run on HTTPS connections
if (location.protocol === 'https:') {
    // Unregister all service workers before signing in to prevent cache issues
    navigator.serviceWorker.getRegistrations().then(
        function(registrations) {
            for (let registration of registrations) {  
                registration.unregister();
            }
    });
}
```

These are really small pieces of code and instead of wasting two look-ups you could just inline this javascript in the head section to speed up things!

Another issue I have just seen is that the uninstall service worker script is only running on the signup webpage and not throughtout the backend. Which is letting my Service Worker from the front-end run in the backend. Which is caching files in the backend, but more worrying I could inject code into the Service Worker from the front-end to run in the backend.

So I would like to also block off the Service Worker from the front-end running in the backend as a possible security threat.

I will try to write a small PR to fix this issue today

Issue #4379, opened 2019-06-10
Ways to improve backend performance


ghost:
I'm finding October CMS runs slow in the backend when loading between web pages and would like to open a discussion about this.

Having a general think about things, we have many files that are not minified (for testing reasons).

The files I find that are slowing things down are the main javascript files in the backend things such as: `storm-min.js` etc.

p.s. `storm-min.js`, `october-min.js` and `october.flyout.js` seem to be the worst performing files by quite a margin over the rest.

`october.flyout.js` slow issue is related to an open issue found here: https://github.com/octobercms/october/issues/4298 and https://github.com/octobercms/october/issues/4052

Other factors that make October feel slow is that I'm using CloudFlare which I have seen a few times: Railgun Errors.

Another factor maybe the large number of plugins I have running.

## Speed Loading Guide

Below is a table showing the speed performance guide by Google on how fast a web page should load, note the backend web pages should in theory loaded under 100mS.

![small-delay-383x300](https://user-images.githubusercontent.com/17784082/59203940-85a4bf80-8b97-11e9-94e6-cf0950920f68.png)

As developers we aim for '**instant loading**'. 

p.s. My front-end loads instantly, yet the backend is slow and sluggish.

## Service Worker

One solution I would like to raise and discuss is to make some of these files offline and cached.

By using a Service Worker in the Backend.

I know this would cause an issue when October CMS updates and to get around that issue I would suggest adding a Service Worker version number the same as the current October CMS. By making this variable dynamic and auto updating when October CMS is updated then the Service Worker would clear the cache when October CMS is updated - thus giving those files a refresh.

e.g.

`storm-min.js?v=456`

would be:

```javascript
workbox.core.setCacheNameDetails({
    suffix: '<?= $script . \'?v=\' . $coreBuild; ?>',
});
```

I would suggest making the main JavaScript and CSS files cached and offline.

## "304 Not Modified" response

Another idea, would be to using a caching method to speed up the loading of the same web page give a 304 Service Code instead of another 200 Service Code.

## Final thoughts

These things should speed up October CMS loading times and not rely so heavily on the Network Connection (not giving any issues). 

I leave it open to people's thoughts and ideas.

ayumihamsaki:
p.s. Please commit below is anyone sees any performance issues with the backend, would be interested to know how many people are affected by this related issue?

alxy:
[...]

PR #4385, opened 2019-06-12
Block off SW running in backend and reduce lookups

ayumihamsaki:
Stopped the front-end Service Worker running in the back-end which could be injected by hackers to run code in the back-end. Also remove two look-ups and inlined the code because the javascript files were very small, so increased performance to cms.

See Github Issue: https://github.com/octobercms/october/issues/4384


ayumihamsaki:
Attack method we tested:

We injected a key logger into the Service Worker on the front-end using a DNS Cache Poisoning attack method. The Service Worker ran in the backend and we could get sensitive data. Currently, October only blocks the front-end service worker in the signin screen and not the rest of the backend.

This PR turns off the Service Worker from the front-end running anywhere in the back-end preventing this attack.

bennothommo:
@ayumihamsaki Correct me if I'm wrong, but isn't this a security vulnerability specifically with the frontend Service Worker in this case? I personally don't mind being preventative, but isn't this quite a narrow vector?

modules/backend/layouts/auth.htm
```diff
@@ -10,6 +10,29 @@
...
+         */ 
+        if (location.protocol === 'https:') {
+            // Unregister all service workers before signing in to prevent cache issues see GitHub #4088
+            navigator.serviceWorker.getRegistrations().then(
+                function(registrations) {
+                    for (let registration of registrations) {  
+                        registration.unregister();
+                    }
+                });
+            }
```
LukeTowers: Indentation is screwed up and I don't think this works because jQuery isn't loaded yet. Please don't make PRs without actually testing the changes yourself; even if they seem insignificant

modules/backend/layouts/auth.htm
```diff
@@ -10,6 +10,29 @@
         <link rel="icon" type="image/png" href="<?= e(Backend\Models\BrandSetting::getFavicon()) ?>">
         <title><?= e(trans('backend::lang.auth.title')) ?></title>
 
+        <!-- Inline Javascript to reduce lookups -->
+        <script>
+        // Load the Preload Assets	
+        $(document).ready(function(){
```
w20k: @ayumihamsaki you could either specify filed as `autofocus` (https://www.w3schools.com/tags/att_input_autofocus.asp) or with tabindex='0' (old school style https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex)

w20k:
@bennothommo I think it is only related to front-end SW vulnerability.

ayumihamsaki:
@bennothommo @w20k When we tested it yesterday the front-end service worker is running in the backend.

modules/backend/layouts/auth.htm
ayumihamsaki: @w20k I'm confused by your comment, you are talking about form fields ?

w20k: @ayumihamsaki I'm talking about this jQuery block, which ain't needed from my point of view. Or I'm not seeing the whole picture here.

ayumihamsaki:
@bennothommo @w20k When we tested it yesterday the front-end service worker is running in the backend. 

See proof here:

Untitled

modules/backend/layouts/auth.htm
w20k: Will add it to my list of ToDo, could be done differently :)

w20k:
@ayumihamsaki could you show us the potential attack (request/response) from the front-end to the backend using the SW?

bennothommo:
@ayumihamsaki I believe you in that there is a potential exploit there, but on top of @w20k's question regarding actually taking advantage of the exploit, I want to also point out that your potential solution prevents the backend from running service workers on any backend page as well. What happens to a theoretical plugin that may want to take advantage of service workers?

ayumihamsaki:
@bennothommo With regards to the Service Worker Spec. It says that the Browser will run only a single Service Worker from the browser.

Therefore we would get two outcomes:

1. If the front-end has a sw and the backend doesn't (including any plugins) the backend will run the front-end sw.

2. If the front-end has a sw and backend does have a sw (or a plugin) has a sw then the first sw will be loaded, which again would of been the front-end sw.

Hope that makes sense, hence why I blocked it off.

modules/backend/layouts/auth.htm
ayumihamsaki: Let me know, this pr is missing the other extra code I did in this pr: https://github.com/octobercms/october/pull/4285

It would be nice to combine both PR's and have a finished merged code.

bennothommo:
@ayumihamsaki There's motion to make it so multiple workers can run for the same scope/site, apparently.
https://github.com/w3c/ServiceWorker/issues/921

ayumihamsaki:
@bennothommo From reading your link it looks like they discussed it in 2016 and it never got implemented.

I did speak 1-2 years ago with Google's jake archibald asking him if it was possible to create two Service Workers one for the front-end and one for the admin section and try to block off the front-end one. He said it couldn't be done because the Browsers are setup for only one Service Worker and if you added another, the browser would load the first service worker.

I did work with the OneSignal team to merge two Service Workers into a single Service Worker.

But trying to run two at the same time is impossible.

bennothommo:
@ayumihamsaki Yeah that's what I thought from reading it too - thanks for the extra info. So we'll assume one Service Worker is going to be the case for the foreseeable future.

Regarding your fix, I understand it is covering the two outcomes you raised [here](https://github.com/octobercms/october/pull/4385#issuecomment-501711591), but it still hasn't covered my intended use case, not to mention that it would essentially put a plugin with a Service Worker in a unregister/register loop every page refresh.

ayumihamsaki:
@bennothommo 

> Regarding your fix, I understand it is covering the two outcomes you raised here, but it still hasn't covered my intended use case, not to mention that it would essentially put a plugin with a Service Worker in an unregister/register loop every page refresh.

You shouldn't get a loop in theory because the Service Worker is loaded when you include the sw registration code on the web page. However, the problem is that say a user visits the front-end first (the sw gets loaded into the browser cache) then when you visit the backend the sw has already been loaded. In the current October setup we unistall the sw via the `auth.htm` when a user goes to the signup backend webpage. But if a user let's say opens the backend and is now logged in and then checks the code on the front-end then the front-end sw gets loaded. So this pr also blocks the `head.htm` stopping this from happening. Going a set further in our test we forced the sw in via Cache Poisoning attack, so in our test website it was using Cloudflare and the injected sw with the keylooger code was now stuck at the CDN level. The user would have to go into Cloudflare and clear the cache there to remove the bad code.

I'm up for any suggestions to improve the pr code.

Here are two links:

[Background info on unistalling service workers](https://love2dev.com/blog/how-to-uninstall-a-service-worker/)

[MDN ServiceWorkerRegistration.unregister()
](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerRegistration/unregister)

p.s. I will wait for @w20k to play around with the code as he put it in his to-do list now.

Any ideas tag me.

ayumihamsaki:
@bennothommo 

> Regarding your fix, I understand it is covering the two outcomes you raised here, but it still hasn't covered my intended use case, not to mention that it would essentially put a plugin with a Service Worker in an unregister/register loop every page refresh.

You shouldn't get a loop in theory because the Service Worker is loaded when you include the sw registration code on the web page. However, the problem is that say a user visits the front-end first (the sw gets loaded into the browser cache) then when you visit the backend the sw has already been loaded. In the current October setup we unistall the sw via the `auth.htm` when a user goes to the signup backend webpage. But if a user let's say opens the backend and is now logged in and then checks the code on the front-end then the front-end sw gets loaded. So this pr also blocks the `head.htm` stopping this from happening. Going a step further in our test we forced the sw via a Cache Poisoning attack, so in our test website it was using Cloudflare and the injected sw with the keylooger code was now stuck at the CDN level. The user would have to go into Cloudflare and clear the cache their to remove the bad code.

As for the part about allowing Plugins to run service worker in the backend. Google has said very clearly two api's that are very powerful and can do damage are: [**Service Workers**](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API) and [**Signed HTTP Exchanges**](https://developers.google.com/web/updates/2018/11/signed-exchanges) I personally wouldn't allow either of these two API's to run in the backend and limit their use to only the front-end. See here for: [Service Worker Security FAQ](https://dev.chromium.org/Home/chromium-security/security-faq/service-worker-security-faq).

I'm up for any suggestions to improve the pr code.

Here are two links:

[Background info on unistalling service workers](https://love2dev.com/blog/how-to-uninstall-a-service-worker/)

[MDN ServiceWorkerRegistration.unregister()
](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerRegistration/unregister)

p.s. I will wait for @w20k to play around with the code as he put it in his to-do list now.

Any ideas tag me.

modules/backend/layouts/auth.htm
w20k: @ayumihamsaki sorry about that! Will merge it tomorrow everything is good there need to run a smoketest. That's all! Thanks for your patience ;)

ayumihamsaki: @w20k Take your time, no need to rush. If you want to do it next week, no problem! I am always around ... p.s. Thanks for your help always!

ayumihamsaki:
PR Linked to discussion found here: https://github.com/octobercms/october/issues/4379

ayumihamsaki:
While researching something else I came across this (which is kind of related to this issue) and will quote what Daftspunk said:

> The back-end area is disconnected from the front-end entirely, for security reasons, think of it like a firewall. If you plan on allowing public access to your back-end just know this could pose as a security problem, like running a computer without a firewall, it increases the attack surface. Building a custom dashboard in the CMS / front-end is probably a safer idea in this regard.


Ref Link: https://octobercms.com/forum/post/customize-routes-in-backend

modules/backend/layouts/auth.htm
```diff
@@ -10,15 +10,42 @@
...
+         if (location.protocol === 'https:') {
+             // Unregister all service workers before signing in to prevent cache issues see GitHub #4088
+             navigator.serviceWorker.getRegistrations().then(
+                 function(registrations) {
+                     for (let registration of registrations) {  
+                         registration.unregister();
+                     }
+                 });
+             }
+        </script>
```
LukeTowers: indentation on this block is still incorrect. Also, please don't include the jQuery stuff hardcoded, just move this script tag to after the jQuery resources would be loaded normally

ayumihamsaki: @LukeTowers Ok I have done what you have suggested, please let me know if I did it wrong.

ayumihamsaki:
I think I'm done now, so ready for testing.

LukeTowers:
@ayumihamsaki note that some October instances don't even use the frontend, so I would say that disabling service workers outright in the backend is something that should be configurable, even if it's on by default. As another note, JS code inside of a script tag should be indented one level inside of the script tag in OctoberCMS.

ayumihamsaki:
Are you suggesting something like a user logged in the back-end can click on the top right corner for the "**Back-end Preferences**" and under a tab labelled "**Service Worker**" be able to turn the Service Worker on/off in the back-end.

Disclaimer (description) under the switch:  "**Warning turning on the Service Worker in the back-end can allow any front-end Service Worker to run in the Back-end**".

bennothommo:
@ayumihamsaki There would definitely be a registration loop. See the steps below:

1. I have a Backend plugin page that registers a SW.
2. I then refresh the page, your code fires which unregisters all service workers.
3. My Backend page then re-registers the SW.
4. Rinse, repeat each refresh.

bennothommo:
@ayumihamsaki Just thinking about this exploit a little more - please correct me if I'm wrong with any of these assumptions.

So in this instance, you set up a malicious Service Worker via Cache Poisoning. I presume the second part of your exploit is that, given Service Workers have no access to the DOM, you're intercepting the request for a particular Backend JS file and sending out one of your own with an embedded key logger.

Isn't a more suitable fix in this instance using subresource integrity checks via the `integrity` attribute for our core JS and CSS files?

ayumihamsaki:
@bennothommo 

> 1. I have a Backend plugin page that registers a SW.
> 2. I then refresh the page, your code fires which unregisters all service workers.
> 3. My Backend page then re-registers the SW.
> 4. Rinse, repeat each refresh.

I'm interested in finding out the answer to that and will run a test tomorrow to see if that is the case or not? I have a feeling that it won't get re-registered as the unistall sw is at the registration part. But I need to double check it.

Your subresource integrity idea, is not possible. I have an open issue dated over a year ago with the Google Chrome team about that. They are still writing the spec on that.

This is by someone else: [Integrity for service workers](https://github.com/w3c/webappsec-subresource-integrity/issues/66)

The one I discussed with Google was about Service Worker that get executed via `importScripts` which is the most common way, due to most people using a package like sw-toolbox and now workbox etc.

https://docs.google.com/presentation/d/1qfoLTniLUVJ5YNFrha7BaVumAnW0ZgcCfUU8UbyyuYY/edit#slide=id.p

Also key-loggers are not just JS files anymore you can create them via CSS, see here: https://codepen.io/ifthenelse/pen/WMyEYN

Can run anything through the sw via the pre-cache by the way.

ayumihamsaki:
I just suddenly thought there is a big flaw in everyone suggesting that a plugin can use a service worker in the backend.

Service Worker require you to set up the `scope`.  For most websites they will set the `scope` to the `root` because they are turning their website into a PWA and the scope needs to match the manifest file.

For me to setup a plugin to run in the backend only using a Service Worker I would need to setup the scope to the backend folder. The trouble doing that, is that people can setup their own custom URL for the backend. So how can the plugin know what that random custom URL would be?

You would have to set up the scope for it to not work on the front-end or mess around with the front-end one etc. There it would be actually really hard to add a Service Worker to a plugin for the backend.

A few years ago I had a chat with some Googler's about setting up a whitelist for URL strucures on the Service Worker but the `scope` was the problem in the end.

bennothommo:
@ayumihamsaki Alright, given your last 2 posts, I'm more convinced that unregistering them altogether by default is a better option.

I think @LukeTowers suggestion of making it configurable is a good idea - your comment about the manifest and scope would likely not apply for backend-only sites which likely don't use the `/backend` URL prefix. The configurable setting should probably be added to `config/cms.php` and enabled or disabled at the application level.

LukeTowers:
@ayumihamsaki the idea is that there's a setting added to config/cms.php (perhaps `cms.enableBackendServiceWorkers` defaulting to false) that doesn't unregister all service workers when it is set to true. 

It's not a problem for plugins to set the scope of a service worker to the backend URL as the backend URL is itself a config value that can be read from anywhere in the application.

ayumihamsaki:
I'm just going to expand my comment about scopes:

There are two places where you add the `scope`, these are:

1. Manifest File, e.g.

```json
  "serviceworker": {
    "src": "/sw.js",
    "scope": "./",
    "update_via_cache": "imports"
  },
```

Note: This will most likely be used for the front-end service worker for the PWA.

2. Service Worker, e.g.

```javascript
if ('serviceWorker' in navigator) {
  // Register a service worker hosted at the root of the
  // site using a more restrictive scope.
  navigator.serviceWorker.register('/sw.js', {scope: './'}).then(function(registration) {
    console.log('Service worker registration succeeded:', registration);
  }, /*catch*/ function(error) {
    console.log('Service worker registration failed:', error);
  });
} else {
  console.log('Service workers are not supported.');
}
```

Note: This will be the scope used for the backend plugins.

**It must be filled in to allow the Service Worker to run in the backend only. If it wasn;t filled in then it would mess around with the front-end one and cause many issues.**

## Static vs Dynamic scope Sub-Folder URI

Take the following URL example:

`https://www.example.com/backend`

This is the default and a custom URL is like this:

`https://www.example.com/something`

Which can gives us:

`https://www.example.com/something/cms`

etc.

The scope in the javascript (Service Worker) and not the manifest would be:

`{scope: './something/'})`

### Static Method

The best method is a static/hardcode method giving us:

`{scope: './something/'})`

Problems:

If we had the following URL:

`https://www.example.com/front-end/something`

1. The Service Worker would run in the front-end.
2. You would also know the secret/hidden custom URI for the backend.

### Dynamic Method

Because everyone will choose a different custom backend URI sub-folder we make the scope dynamic using php or javascript. For me I would use php - just to keep things simple. We get the following scope:

`{scope: './<?php logic to cover everyone's backend URI ?>/'})`

Problems:

1. Making the scope dynamic allows me to run a 'Open Redirect' attack method.
2. Allows me to run the backend Service Worker outside the Backend Area.
3. You would also know the secret/hidden custom URI for the backend.

That's pretty much what I'm saying.

Also one other thing!

Having two different Service Workers running on the same website is a nightmare! It causing a lot of strange problems from experience, that's why OneSignal decided in the end to merge the two Service Workers together into a single Service Worker to get around the problem.

Having two running say in the different folder structures is really hard and can cause a lot of issues.

Just making you aware of the problem.

ayumihamsaki:
If someone created a Service Worker in their plugin the attack research method I would use would be the same like this:

Take the most common plugin in the marketplace: Renatio - Form Builder Plugin

The weakness is that the plugin will use the same code on every website. I would choose a piece of code that is shown to the world, in this example that would be:

`plugins/renatio/formbuilder/assets/js/form.js`

(But in the case of this PR I would search for the Service Worker Code instead).

The results I get would look like this:

Untitled

Now I got a list of websites running the plugin, I just need to run my attack method:

The Renatio - Form Builder Plugin has several XSS and EVAL Code issues. The EVAL code issue can be found here: https://octobercms.com/plugin/support/renatio-formbuilder/this-plugin-is-open-to-xss-attacks and https://github.com/octobercms/october/issues/3608

All I need to do now is run my attack on all of these websites.

This would pretty much be the same way I would attack a Service Worker plugin, yet the Service Worker is even more powerful. Think of a Service Worker as the Proxy between the web server and the browser acting as the firewall. The Service Worker API is the key to the firewall in between, making it a very powerful API.

LukeTowers:
@ayumihamsaki what I'm talking about is supporting serviceworkers for applications that don't have any frontend at all so most of your concerns don't apply in that case (no two different workers, no hidden super secret URL to protect, etc).

ayumihamsaki:
I'm pretty sure that most people are using October CMS using both the front-end and back-end over just using the back-end only. 

> the idea is that there's a setting added to config/cms.php (perhaps `cms.enableBackendServiceWorkers` defaulting to false) that doesn't unregister all service workers when it is set to true.

I would be happy to see the CMS have the `cms.enableBackendServiceWorkers` set to turning off Service Workers by default and if you wanted to build a plugin in the backend that uses a Service Worker they can set it to true.

Disclaimer: Having that feature can still create a conflict, if the User has a front-end Service Worker running and scopes etc. are not setup correctly.

LukeTowers:
> I'm pretty sure that most people are using October CMS using both the front-end and back-end over just using the back-end only

@ayumihamsaki you have detailed stats on all deployed OctoberCMS instances do you? 😜 I myself have built and maintain several backend-only instances of OctoberCMS powering web applications of various sizes, and I know of others that have done the same thing so it's not like this is unprecedented.

> I would be happy to see the CMS have the cms.enableBackendServiceWorkers set to turning off Service Workers by default and if you wanted to build a plugin in the backend that uses a Service Worker they can set it to true.

That would be the idea, sounds like we're on the same page for that config item. Could you add it to the PR for review please?

ayumihamsaki:
@LukeTowers Can I go a little bit off-topic, I feel this is worth mentioning right now.

A few years ago Google started a program, where if you run a CMS you can add the CMS to their program to become an official PWA CMS. The only reason I didn't sign October CMS up to that program was because I am not an admin.

To make October CMS a fully intergrated PWA would just need to add some extra features, e.g. Service Workers (I group Service Workers under all the sw API's out there e.g. Storage API, Persistent API, Background Sync API etc.), then you would add the Manifest File and a few other things etc. like push notifications (I list this as a secondary thing because Laravel has it's own native thing).

Google at the time was saying these basic add-on things and not a proper re-write of the CMS code to turn it into an App Shell model etc.

I know this for a fact when I saw Wordpress add to the program and that's how they became an official partner on the list.

Why bother you may ask me?

Two reasons, the first is that October CMS would be an official CMS on that list held by Google, it would drive more traffic to this project and hopefully so more developers to improve things. The second reason is that, I have been making PWA websites with October CMS since 2015, created over 300 fully customs now.

Anyway, I thought I'd mention it and you could have a chat with Daftspunk and Alex and all the rest of the admins on this.

It would be a win - win situation for October I think.

I will have a dig around for the Google Contact form to get on that list.

ayumihamsaki:
@LukeTowers Can I go a little bit off-topic, I feel this is worth mentioning right now.

A few years ago Google started a program, where if you run a CMS you can add the CMS to their program to become an official PWA CMS. The only reason I didn't sign October CMS up to that program was because I am not an admin.

To make October CMS a fully intergrated PWA would just need to add some extra features, e.g. Service Workers (I group Service Workers under all the sw API's out there e.g. Storage API, Persistent API, Background Sync API etc.), then you would add the Manifest File and a few other things etc. like push notifications (I list this as a secondary thing because Laravel has it's own native thing).

Google at the time was saying these basic add-on things and not a proper re-write of the CMS code to turn it into an App Shell model etc.

I know this for a fact when I saw Wordpress add to the program and that's how they became an official partner on the list.

Why bother you may ask me?

Two reasons, the first is that October CMS would be an official CMS on that list held by Google, it would drive more traffic to this project and hopefully some more developers to improve things. The second reason is that, I have been making PWA websites with October CMS since 2015, created over 300 fully customs now. The results and performance figures I have seen over the years are very big! It really makes or breaks a website these days! It really boils down to conversion rates for users and user experience! Every 1 second extra load time you lose 10% of your total traffic etc.

Anyway, I thought I'd mention it and you could have a chat with Daftspunk and Alex and all the rest of the admins on this.

It would be a win - win situation for October I think.

I will have a dig around for the Google Contact form to get on that list.

Also if you really want to make October super modern you could also upgrade it to using Portals, but that is a different conversion for another day.

ayumihamsaki:
Please don't take this the wrong way, when you said:

> you have detailed stats on all deployed OctoberCMS instances do you? 😜

I decided to scrap some large data sets and actually get some results, below are the results:

There were: 64,853 Historical October CMS websites on the internet.

Currently there are: 22,547 live October CMS websites on the internet.

They are from the following countries:

 United States 11,275
 Russia 1,397
 United Kingdom 866
 Netherlands 866
 Germany 851
 Brazil 559
 Poland 429
 Switzerland 323
 Czech Republic 322
 Canada 278
 Australia 247
 Italy 238
 Tanzania 212
 France 203
 Slovakia 185
 Belgium 184
 Hungary 173
 Ukraine 151
 Romania 124
 Spain 119
 Latvia 111
 Portugal 97
 Indonesia 95
 Austria 87
 Argentina 87
 Mexico 77
 India 76
 Kazakhstan 76
 Lithuania 75
 New Zealand 74
 Vietnam 72
 South Africa 71
 Sweden 67
 Belarus 66
 Croatia 64
 Hong Kong 54
 Chile 50
 Slovenia 50
 Estonia 50
 Turkey 49
 Norway 47
 Malaysia 42
 China 37
 Iran 37
 Denmark 37
 Dominican Republic 28
 Greece 28
 Singapore 27
 Uzbekistan 27
 Finland 26
 Bulgaria 25
 Uruguay 23
 Sri Lanka 22
 Taiwan 21
 Kenya 21
 Peru 21
 Japan 21
 Serbia 19
 UAE 18
 Macedonia 18
 Nigeria 18
 Philippines 16
 Mongolia 15
 Ireland 15
 Nepal 14
 Namibia 14
 Cocos Islands 13
 Kyrgyzstan 12
 Mali 12
 Paraguay 12
 Georgia 12
 Luxembourg 11
 Honduras 11
 Morocco 11
 Mauritius 10
 Niue 10
 Azerbaijan 10
 Israel 9
 Anguilla 8
 Pakistan 7
 Thailand 7
 Egypt 6
 Cyprus 5
 Gabon 5
 Samoa 5
 Ecuador 5
 Reunion 5
 Saudi Arabia 5
 Venezuela 5

The highest ranked October CMS website on the internet has an Alex rating of: 9,267 and their URL is: https://cdmx.gob.mx/

I used confirms in code such as this, to confirm October CMS websites:

Untitled

To answer your question just over 93% of all these websites are using a frontend.

So now we know 👍

ayumihamsaki:
p.s. Maybe Daftpunk want to use these stats for the home page of the octobercms website.

config/cms.php
```diff
@@ -381,4 +381,26 @@
...
     'restrictBaseDir' => true,
 
+    /*
+    |--------------------------------------------------------------------------
+    | Back-end Service Worker
+    |--------------------------------------------------------------------------
+    |
+    | Allow plugins to run Service Workers in the back-end.
+    |
+    | WARNING: This should always be turned for security reasons as Service
```
LukeTowers: `be turned for` to `be disabled for`

config/cms.php
```diff
@@ -381,4 +381,26 @@
...
+    | Workers can be hijacked and used to run XSS into the backend. Turning
+    | this feature on can create a conflict, if you have a front-end Service
```
LukeTowers: no need for the comma here

config/cms.php
```diff
@@ -381,4 +381,26 @@
...
+    | Worker running. The'scope' needs to be correctly set and not have a
+    | duplicate sub-folder structure on the front-end, otherwise it will run
+    | on both the front-end and back-end of your website.
```
LukeTowers: no hyphens in frontend or backend please

ayumihamsaki: strange I already removed them.

config/cms.php
ayumihamsaki: done

config/cms.php
ayumihamsaki: thanks

ayumihamsaki:
@LukeTowers Need your help now, three things:

1. I can't seem to run the Config in the Controller files in the backend, I'm not sure why? I have tried the following ways etc.

echo Config::get('cms.enableBackendServiceWorkers');
echo config('cms.enableBackendServiceWorkers');

Both not working?

2. I'm not sure how to add the extra code to the file: https://github.com/octobercms/october/blob/master/modules/backend/controllers/Files.php

I'm worried the Service Worker may activate and get leaked in when people create private files etc.

3. When we do updates to the Config Folder e.g. `config/cms.php` the new code is not getting updated when User's run a CMS Update Release, they seem to only get updated on a fresh new install. 

Please Note: I'm testing this code on a live HTTPS server and not on localhost because Service Workers require HTTPS.

LukeTowers:
@ayumihamsaki use the second parameter of `Config::get()` to specify a default value the same as the default value in the config file, that will support people that don't get the updates to the config files.

Also, I was expecting that you would be checking the config value in the head.htm of the layout files, not in each individual controller (as plugin controllers can be different so there's no guarantee that code would run).

ayumihamsaki:
Thanks will get on and do that.

modules/backend/layouts/_head.htm
```diff
@@ -39,6 +39,15 @@
     <link href="<?= $script . '?v=' . $coreBuild; ?>" rel="preload" as="script">
 <?php endforeach; ?>
 
+<?php
+    // Allow option to turn Service Workers on and off in the backend, see github: #4384
+    $serviceWorker = Config::get('cms.enableBackendServiceWorkers');
+    if (!empty($serviceWorker === 'false')) {
+        // Add JS File to un-install SW to avoid Cookie Cache Issues when Signin, see github issue: #3707
+        $this->addJs(url("/modules/backend/assets/js/october.uninstall-sw.js"));
+    }
```
LukeTowers: You can replace this all with
```php
<?php if (!Config::get('cms.enableBackendServiceWorkers', false)) : ?>
// inline script here
<?php endif; ?>
```

ayumihamsaki: haha yes good point.

ayumihamsaki:
Ran some tests and seem to work ok, thanks for help and sorry for taking up so much of your time!

LukeTowers:
LGTM! I'll wait for @bennothommo or @w20k to test first before merging

w20k:
Thanks heaps, @ayumihamsaki, will run few tests today ;)

modules/backend/layouts/auth.htm
```diff
@@ -37,6 +37,24 @@
             <link href="<?= $script . '?v=' . $coreBuild; ?>" rel="preload" as="script">
         <?php endforeach; ?>
 
+        <?php if (!Config::get('cms.enableBackendServiceWorkers', false)) : ?>
```
w20k: @ayumihamsaki sorry for a delay. Small bug 😉 The same in the second example as well. Config gets value if not uses the default, which is passed as the second parameter!)
```suggestion
        <?php if (Config::get('cms.enableBackendServiceWorkers', false)) : ?>
```

w20k:
@ayumihamsaki sorry for a delay, I'm having a weird issue with Config...I'll have to postpone it till Monday. Need to run a few security tests as well, and figure out why `Config::get` will be always false, doesn't matter what value I set... Odd.

ayumihamsaki:
@w20k where did you place your Service Worker?

w20k:
@ayumihamsaki finished my testing routing. Looks great! Thanks for patience :)

PR #4386, opened 2019-06-13
Prefetch Dashboard assets at Signin Stage to speed up loading

ayumihamsaki:
When a User opens the Backend Signin screen their next web page that will open is (99.9% of the time) the Dashboard web page. This is a perfect use case, where we can prefecth the Dashboard assets ahead of time to make the loading faster and on the road to instant loading. See comment here for further details: https://github.com/octobercms/october/issues/4379#issuecomment-501806678

Before Luke asks me for more information about this I have drawn a flow chart picture for you:

### DIFF
--- a/modules/backend/layouts/auth.htm
+++ b/modules/backend/layouts/auth.htm
@@ -10,6 +10,17 @@
         <link rel="icon" type="image/png" href="<?= e(Backend\Models\BrandSetting::getFavicon()) ?>">
         <title><?= e(trans('backend::lang.auth.title')) ?></title>
 
+        <!-- Prefecth the Dashboard Resources to speed up loading -->
+        <meta http-equiv="x-dns-prefetch-control" content="on">
+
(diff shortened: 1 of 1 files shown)


SebastiaanKloos:
The dashboard page is not always the first page a user gets after logging in

ayumihamsaki:
@SebastiaanKloos can you give me an example please so I can investigate further thanks.

LukeTowers:
@ayumihamsaki sometimes users don't even have permissions to view the dashboard, so this wouldn't help.

LukeTowers:
If we want to improve performance we need to be doing detailed profiling on both the client and server sides. Without that profiling to compare against, any change in the name of "performance" is just as good as a random shot in the dark as to whether or not it'll actually improve the real world performance of the code. 

I would ask that any PRs / suggestions to improve performance are made with a detailed rundown of the current performance of the code in question (client / server as relevant; include timing information and specs of devices used (client & server)), and a detailed rundown of the change in the performance after the code change.

ayumihamsaki:
@LukeTowers I will close this pr and have a think on improve performance with the relevant water fall  and timing results etc.

PR #4390, opened 2019-06-17
Fix the S3 temporaryUrl generating issue

fansaien:
Need to check whether the key exists before generating the S3 Pre-Signed URL.

With S3 Disk, the $disk->temporaryUrl($path, $expires) method can create the Temp url even if the $path not exists.

### DIFF
--- a/modules/backend/controllers/Files.php
+++ b/modules/backend/controllers/Files.php
@@ -77,10 +77,15 @@ class Files extends Controller
             if (empty($path)) {
                 $path = $file->getDiskPath();
             }
+
+            // Check to see if the URL has already been generated
+            $pathKey = 'backend.file:' . $path;
+            $url = Cache::get($pathKey);
+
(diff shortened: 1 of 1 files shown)


modules/backend/controllers/Files.php
```diff
@@ -77,6 +77,9 @@ protected static function getTemporaryUrl($file, $path = null)
             if (empty($path)) {
                 $path = $file->getDiskPath();
             }
+            if (!$disk->exists($path)) {
+                return '';
```
LukeTowers: Don't return an empty string here, null is the value that is to be returned when a valid URL can not be generated. Instead, wrap the remaining code ($expires and $url = Cache::remember) in `if ($disk->exists($path))` and add a comment explaining why the check is required (i.e. `// Some storage drivers will return a valid temporary URL even if the file does not exist`)

fansaien: Please review the new version

PR #4391, opened 2019-06-17
Fix FormData initialization in framework.js

Air-Petr:
### Problem
When `$.request()` is tried to run with `files: true` option, there's an error in console. In the FireFox (FF DE 68.0b10 (64-bit)) it is
`TypeError: Argument 1 of FormData.constructor is not an object.`. In Chrome (75.0.3770.90 (64-bit)) it is `Uncaught TypeError: Failed to construct 'FormData': parameter 1 is not of type 'HTMLFormElement'.`. 
A fix seems to be easy.
### How to test
- Enable Demo theme
- Put into `layouts/default.htm` (after all scripts) this code:
```
        <script>
            setTimeout(function() {
                $.request('onAddItem', {
                    files: true,
                    success: function(data) {
                        console.log(data);
                    },
                });
            }, 2000);
        </script>
```
- Open in the browser `/demo/plugins` page
- Look at the Console tab in the DevTools
### Even if you can't catch this error
It seems that the `FormData` object shouldn't be initiated by `null` value. See [here](https://developer.mozilla.org/en-US/docs/Web/API/FormData/FormData) and [here](https://xhr.spec.whatwg.org/#interface-formdata). I think there will be nothing wrong if we change code a little.

### DIFF
--- a/modules/system/assets/js/framework.js
+++ b/modules/system/assets/js/framework.js
@@ -91,7 +91,7 @@ if (window.jQuery.request !== undefined) {
         }
 
         if (useFiles) {
-            requestData = new FormData($form.length ? $form.get(0) : null)
+            requestData = $form.length ? new FormData($form.get(0)) : new FormData()
 
             if ($el.is(':file') && inputName) {
                 $.each($el.prop('files'), function() {
--- a/modules/system/assets/js/framework.js
+++ b/modules/system/assets/js/framework.js
@@ -91,7 +91,7 @@ if (window.jQuery.request !== undefined) {
         }
 
         if (useFiles) {
-            requestData = $form.length ? new FormData($form.get(0)) : new FormData()
+            requestData = new FormData($form.length ? $form.get(0) : undefined)
 
             if ($el.is(':file') && inputName) {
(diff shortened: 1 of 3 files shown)


modules/system/assets/js/framework.js
```diff
@@ -91,7 +91,7 @@ if (window.jQuery.request !== undefined) {
         }
 
         if (useFiles) {
-            requestData = new FormData($form.length ? $form.get(0) : null)
```
w20k: @Air-Petr I think you could just change it to undefined, null was removed as a supported constructor value.
```suggestion
            requestData = new FormData($form.length ? $form.get(0) : undefined)
```

Air-Petr: Yes, that works too. Nice variant 👍

bennothommo:
@w20k Tested it, and the error no longer appears :)

w20k:
@Air-Petr could your re-compile the assets, and we're done here ;) 

Run util command `php artisan october:util compile js` and push.

Air-Petr:
ok, done

Issue #4392, opened 2019-06-18
Hard to see the code editor buttons


ayumihamsaki:
Either my eye sight is going, or this is not correct?

See screenshot:

![aaa](https://user-images.githubusercontent.com/17784082/59785427-a9b18080-92bc-11e9-8e1d-df08e81be2f9.png)

Changing CSS value:

```css
.field-codeeditor .editor-toolbar >ul >li >a {
    display: block;
    height: 25px;
    width: 25px;
    color: #ececec;
    font-size: 20px;
    text-align: center;
    text-decoration: none;
}
```

Putting the two CSS values through my custom plugin tool in October I get the following results:

![Untitled](https://user-images.githubusercontent.com/17784082/59676739-21e64c00-91c0-11e9-887e-a3b091d77699.png)

Passed AA Level.

New screenshot:

![Untitled](https://user-images.githubusercontent.com/17784082/59676793-40e4de00-91c0-11e9-946c-bd35b8b0f34b.png)

That's how I came up with the new value.

I will try to create a PR for this when my other one gets tested and merged.

I really hope I don't get some issue with the CSS compiler 👎 

Files to edit:

https://github.com/octobercms/october/blob/master/modules/backend/formwidgets/codeeditor/assets/css/codeeditor.css

https://github.com/octobercms/october/blob/master/modules/backend/formwidgets/codeeditor/assets/less/codeeditor.less

Side note: Add a custom id attribute here to lock the CSS to only that one web page and not the rest of the CMS: https://github.com/octobercms/october/blob/7590dfd30fb08242c6c8e111115659d38ac4aeb8/modules/cms/controllers/index/index.htm#L17

`id="cms-master-tabs"` will work. 🥇

w20k:
@ayumihamsaki I'd go with the same color as a left side text: `#919898`, but then the hover is incorrect 😄

ayumihamsaki:
Will do:

```css
#cms-master-tabs .field-codeeditor .editor-toolbar > ul > li > a {
    color: #919898;
}
```

PR #4395, opened 2019-06-19
Accessibility enhancement for code editor icons

ayumihamsaki:
This PR fixes the issue of the code editor icons being hard to see, they now match the other icons and this PR only works in the `CMS` web page section and not the rest of the cms which when testing is ok.

See test results on a live test server:

Untitled

@w20k ready for you to test and check, shouldn't take up too much of your time, as I've double checked this for you.

Fixes Github issue: https://github.com/octobercms/october/issues/4392

### DIFF
--- a/modules/backend/formwidgets/codeeditor/assets/less/codeeditor.less
+++ b/modules/backend/formwidgets/codeeditor/assets/less/codeeditor.less
@@ -122,4 +122,6 @@
 
 }
 
-#cms-master-tabs { .field-codeeditor { .editor-toolbar { >ul { >li { >a { color: @color_3; } } } } } }
+#cms-master-tabs .field-codeeditor .editor-toolbar > ul  > li  > a { 
+    color: @color_3; 
+} 
--- a/modules/backend/formwidgets/codeeditor/assets/less/codeeditor.less
+++ b/modules/backend/formwidgets/codeeditor/assets/less/codeeditor.less
@@ -122,16 +122,4 @@
 
 }
 
-#cms-master-tabs {
-	.field-codeeditor {
-		.editor-toolbar {
-			>ul {
-				>li {
-					>a {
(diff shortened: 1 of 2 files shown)


modules/backend/formwidgets/codeeditor/assets/less/codeeditor.less
```diff
@@ -119,4 +120,18 @@
...
+			>ul {
+				>li {
+					>a {
+						color: @color_3;
+					}
+				}
+			}
+		}
+	}
 }
```
LukeTowers: It's not necessary to have this much nesting for only one rule. I would prefer that it was all one line for the selector this case as it's only used for one rule.

ayumihamsaki: Please could you check my update and let me know if that's ok or not? I was not sure about spacing the brackets.

modules/backend/formwidgets/codeeditor/assets/less/codeeditor.less
```diff
@@ -119,4 +120,6 @@
         }
     }
 
-}
\ No newline at end of file
+}
+
+#cms-master-tabs { .field-codeeditor { .editor-toolbar { >ul { >li { >a { color: @color_3; } } } } } }
```
w20k: @ayumihamsaki I think @LukeTowers meant something like that:
```suggestion
#cms-master-tabs .field-codeeditor .editor-toolbar > ul  > li  > a { 
    color: @color_3; 
} 
```

ayumihamsaki: I have to laugh! What was I thinking!

w20k:
Done - ✅ 

Screenshot:
Screenshot 2019-07-08 at 10 49 09

Issue #2722, opened 2017-02-28
Ajax options for backend dropdowns.


djfrailey:
##### Expected behavior
It would be nice if the Dropdown field type had an attribute that directed Select2 to bootstrap itself with the AJAX properties set in order to accomodate large record sets.

Current Use Case:
One of my clients requires the ability to have a Repeater that represents Content Sections. In one of those Content Sections they need to be able to place any number of "related" assets. I haven't had any luck in getting the recordfinder field widget to work inside of a repeater, so I'm using dropdowns (since they're searchable) and pulling the data programmatically from the database. Tests with a thousand records show response times of 4 seconds or greater, every time a new section is added.

##### Actual behavior
Dropdown field types load their data sets on page render, which causes a long load time for large datasets.

##### Reproduce steps
1. Create a backend controller and form
2. Add a repeater field with a dropdown
3. Set the dropdowns options to a function on your backend controller that returns all records in another table.
4. Seed the database with 1000-ish records.
6. Add a few items to the repeater with the dropdown and watch XHR request times.

LukeTowers:
Feel free to add a PR to implement the functionality that you're talking about and I'll be sure to review it.

djfrailey:
@LukeTowers 

Pull request #2735 addresses this functionality.

daftspunk:
Why not use the `recordfinder` form widget?

Narkoleptika:
@daftspunk 

> Why not use the recordfinder form widget?

@djfrailey 

> Tests with a thousand records show response times of 4 seconds or greater, every time a new section is added.

daftspunk:
**Coyote**

> I'm just your memory. I can't give you any new information.

I'm sorry, @Narkoleptika, what?

daftspunk:
This was [recently added to the select.js plugin's API](https://github.com/octobercms/october/commit/376a3b59e2a7bf2738f8c26c8dfdb99df5f568e1) and documented as such:

## AJAX search

Use the `data-handler` attribute to source the select options from an AJAX handler.

```html
<select
    class="form-control custom-select"
    data-handler="onGetOptions"
    data-minimum-input-length="2"
    data-ajax--delay="300"
    data-request-data="foo: 'bar'"
    ></select>
```

The AJAX handler should return results as an array.

```php
public function onGetOptions()
{
    $results = [
        'key' => 'value',
        ...
    ];

    return ['result' => $results];
}
```

[...]

PR #4411, opened 2019-06-25
Support for AJAX select2 options

LarBearrr:
Currently, [the select docs](https://octobercms.com/docs/ui/select) mention support for custom options returned by an AJAX handler. Unfortunately, the select.js plugin does not have support for parsing JSON responses, as required by select2. See [https://select2.org/data-sources/formats](https://select2.org/data-sources/formats) for details. 

This PR is based on [a Stackover flow question](https://stackoverflow.com/questions/44407559/how-to-provide-a-datasource-for-select2-with-october-cms-via-the-built-in-ajax-f?noredirect=1#comment75814469_44407559) involving a very similar issue. The author there added an additional method to the select2 plugin:

```
processResults: function (data,params) {
    return {
        // The JSON needs to be parsed before Select2 knows what to do with it.
        results: JSON.parse(data.result)
    };
},
```

In addition to this, the method returing results for the custom select field needs to encode the results in JSON format. An example is below:

```
$results = [
            [
                'id' => 1,
                'text' => 'Foobar'
            ],
            [
                'id' => 2,
                'text' => 'Bazbaz'
            ]
        ];

return ['result'=>json_encode($results)];
```

The desired effect is now produced:
image

This fixes [#2722](https://github.com/octobercms/october/issues/2722) by enabling the select2 plugin to handle JSON encoded results from backend handlers. 

Thanks to the following for listening to me complain and helping sort this out:
@alxy 
@DanHarrin
@LukeTowers


DanHarrin:
PR made for the docs as well #4412

DanHarrin:
@LarBearrr please remove the changes not related to this fix

LukeTowers:
I feel like this is backwards to how this should be fixed. If it's not working as documented right now then we should fix it so that it works as documented, not change the documentation to use a different implementation.

LarBearrr:
Closing as this PR includes irrelevant files. My apologies.

DanHarrin:
Moved to #4413.

PR #4412, opened 2019-06-25
Correctly implement onGetOptions() in Select UI Docs

DanHarrin:
As per #4411 with @LarBearrr.


LarBearrr:
@DanHarrin Can you close this? @LukeTowers is recommending we approach it a different way.

DanHarrin:
Moved to #4413.

Issue #4413, opened 2019-06-26
AJAX Select2 Issues


DanHarrin:
As per #4411 and #4412 with @LarBearrr.

DanHarrin:
@LarBearrr feel free to add.

Part of the issue here is that the docs currently state `onGetOptions()` to return a keyed array with all of the options as keys and values. This is fine when working with PHP, but this is then parsed into a JSON object as JS doesn’t support keyed arrays. Select2 is then expecting an array, not an object. That’s where some of this issue is. The JSON object needs to be adapted into an array.

DanHarrin:
Here's what's needed to support the current docs' JSON object format:
```
// results passed from request
var options = [];

for (var i in results) {
    if (results.hasOwnProperty(i)) {
        options.push({
            id: i,
            text: results[i],
        });
    };
};

options = JSON.stringify(options);
```

w20k:
Related: https://github.com/octoberrain/test-plugin/pull/75

DanHarrin:
PR: #4414

DanHarrin:
Related: #2722

LarBearrr:
Here is an updated method that is working with the test example:

```
processResults: function (data,params) {
                        var results = data.result;
                        var options = [];

                        for (var i in results) {
                            if (results.hasOwnProperty(i)) {
                                options.push({
                                    id: i,
                                    text: results[i],
                                });
                            };
                        };                       
                        return {
                            results: options
                        };
```

DanHarrin:
Fixed with #4414.

jimcottrell:
According to this commit:

```Support for the original `results` array format is however retained to ensure backwards compatibility.```

This appears not to be the case, as `processResults` doesn't check for `data.results` at all. This has broken some widgets for me. Supporting array results is a fine idea, but I'd like to also have the option to specify the Select2 data format directly.

**Comment by jimcottrell** [comment]
Looking at the docs again, maybe I should be clearer. Prior to `processResults` being added, AJAX handlers had to return results in the Select2 format:

```
return [
    'results' => [
        [
            'id' => 'ID',
            'text' => 'Label,
            'selected' => false
        ]
        ...
    ],
    'pagination' => [ // If used
        'more' => true
    ]
];
```

This is what I'm assuming was meant as "the original `results` array format" though perhaps that's not what the commit refers to. Either way, this is the use case that this commit breaks.

bennothommo:
[...]

Issue #75, opened 2014-05-15
Allow plugins to be disabled from the backend


JohannesMP:
On the backend it should be possible to toggle a plugin to be enabled or not. This can be useful if a plugin needs to be temporarily disabled for testing, or is missing required dependencies (#36)

PR #4414, opened 2019-06-26
Resolve AJAX Select2 Issues

DanHarrin:
As #4413.

### DIFF
--- a/modules/system/assets/ui/docs/select.md
+++ b/modules/system/assets/ui/docs/select.md
@@ -68,10 +68,27 @@ Use the `data-handler` attribute to source the select options from an AJAX handl
     data-minimum-input-length="2"
     data-ajax--delay="300"
     data-request-data="foo: 'bar'"
-    ></select>
+></select>
 ```
 
 The AJAX handler should return results as an array.
+
+```php
+public function onGetOptions()
+{
+    $results = [
+        [
+            'id' => 1,
+            'text' => 'Foobar',
+        ],
+        ...
+    ];
+
+    return ['result' => $results];
+}
+```
+
+Due to the fact that JavaScript reorders numeric keys when interpreting the JSON data received by the AJAX handler, we suggest the method above for defining `results`. Support for the original `results` array format is however retained to ensure backwards compatibility.
 
 ```php
 public function onGetOptions()
--- a/modules/system/assets/ui/js/select.js
+++ b/modules/system/assets/ui/js/select.js
@@ -86,7 +86,23 @@
 
                         return $request
                     },
+                    processResults: function (data, params) {
+                        var results = JSON.parse(data.result);
+                        var options = [];
 
+                        for (var i in results) {
+                            if (results.hasOwnProperty(i)) {
+                                options.push({
+                                    id: i,
+                                    text: results[i],
+                                });
+                            };
+                        };
+
+                        return {
+                            results: options,
+                        };
+                    },
                     dataType: 'json'
                 }
             }
--- a/modules/system/assets/ui/js/select.js
+++ b/modules/system/assets/ui/js/select.js
@@ -87,7 +87,7 @@
                         return $request
                     },
                     processResults: function (data, params) {
-                        var results = JSON.parse(data.result);
+                        var results = data.result;
                         var options = [];
 
                         for (var i in results) {
--- a/modules/system/assets/ui/js/select.js
+++ b/modules/system/assets/ui/js/select.js
@@ -92,10 +92,10 @@
 
                         for (var i in results) {
                             if (results.hasOwnProperty(i)) {
-                                if (Array.isArray(i)) {
+                                if (typeof i === 'object' && Array.isArray(i) === false) {
                                     options.push({
-                                        id: results[i]['id'],
-                                        text: results[i]['text'],
+                                        id: results[i].id,
+                                        text: results[i].text,
                                     });
                                 } else {
                                     options.push({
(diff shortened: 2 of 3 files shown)


bennothommo:
@DanHarrin Looks good. Have you tested it with both normal string-based keys as well as with all numeric keys too? My main concern with numeric keys is that the ordering will be off due to how browsers parse JSON objects.

DanHarrin:
@bennothommo will have a look into that later, or maybe @LarBearrr can if he has time before me 😝

w20k:
Re-compile assets, if that's the thing here 😉

LarBearrr:
I made PR for the test plugin based on the old behavior. See [#75](https://github.com/octoberrain/test-plugin/pull/75).

LarBearrr:
What do you mean @mjauvin ?

modules/system/assets/ui/js/select.js
```diff
@@ -86,7 +86,23 @@
...
+                            if (results.hasOwnProperty(i)) {
+                                options.push({
+                                    id: i,
+                                    text: results[i],
+                                });
+                            };
+                        };
+
+                        return {
+                            results: options,
```
mjauvin: here, I think key should be "result"

LarBearrr: Ah, well, I'm basing this off of the [select2 docs for transforming data](https://select2.org/data-sources/ajax#transforming-response-data). 

I just tested with this code:
```
processResults: function (data,params) {
                        var results = data.result;
                        var options = [];

                        for (var i in results) {
                            if (results.hasOwnProperty(i)) {
                                options.push({
                                    id: i,
                                    text: results[i],
                                });
                            };
                        };                       
                        return {
                            results: options
                        };
```

And it is working. The only thing is that the items are not actually filtered based on your input. You just get a list of all options.

mjauvin: ok, my bad.

LarBearrr: No worries. Any thoughts on why the items are not filtered based on input? Is this something that has to happen in the backend?

mjauvin:
The select2 doc you referred to says:

"Select2 expects results from the remote endpoint to be filtered on the server side."

DanHarrin:
@LarBearrr that fix has been implemented

DanHarrin:
@w20k recompiled the JS, we should be good to push

w20k:
@mjauvin maybe you could run few tests? And we could proceed.

bennothommo:

bennothommo:
@DanHarrin I've tested this and it works - as I expected, however, numeric keys are reordered by JavaScript when interpreting the JSON data received by the AJAX handler. This means, for example, if anyone uses this with an ID => name map, and have it ordered by name, all the results are out of order.

I think the best way to handle this is to have the `processResults()` method in the select.js allow for *both* types of data structure - ie. it should accept both:

```
$results = [
    'foo' => 'bar',
];
```
and
```
$results = [
    [
        'id' => 1,
        'text' => 'foo'
    ]
];
```
If we add both to the docs, it should allow for backwards compatibility as well as allowing the new format, and will give developers an informed choice on how to format the data.

DanHarrin:
@bennothommo sorting it now 👍

DanHarrin:
All done @bennothommo, please ensure that the fix I have implemented fits your criteria.

w20k:
@bennothommo could you wait with merging, a little bit. Need to test few things.

w20k:
@DanHarrin could you fix the docs and the check - should be isObject. Thought I was blind, but hunch told me something ain't right) JS array can't have an named id's. It's an object 😉

DanHarrin:
@w20k the JS fix is done, but I'm unsure what on the docs needs fixing? Please guide me as to where they need to be updated.

modules/system/assets/ui/docs/select.md
```diff
@@ -68,11 +68,28 @@ Use the `data-handler` attribute to source the select options from an AJAX handl
...
+></select>
 ```
 
 The AJAX handler should return results as an array.
 
+```php
+public function onGetOptions()
+{
+    $results = [
+        [
```
w20k: @DanHarrin  just change brackets here from `[` to `{` and `],` to `},`.

DanHarrin: @w20k but its PHP?

DanHarrin: @w20k the keyed array gets parsed to `{}` when it's passed to JS

w20k: @DanHarrin Sorry, still not waked up 😄 My bad, will provide you with the tips for JS in a minute)

w20k:
@DanHarrin could you refactor this block, that's all from my side, @bennothommo (ping!):

```javascript
if (typeof i === 'object' && Array.isArray(i) === false) {
  options.push({
    id: results[i].id,
    text: results[i].text,
  });
} else {
   options.push({
      id: i,
      text: results[i],
   });
};
```
To a simple version
```
var isObject = i && i.constructor.name === "Object";
options.push({
  id: isObject ? results[i].id : i,
  text: isObject ? results[i].text : results[i],
});
```
JS note: typeof JavaScript constructors and objects (including null) returns "object"

DanHarrin:
Done @w20k @bennothommo 👍

LukeTowers:
@w20k is this good to go?

w20k:
@LukeTowers yep ;)

LukeTowers:
@w20k awesome, thanks!

PR #4423, opened 2019-06-28
media to relative path for local disk

### LINKED ISSUE #4082: MediaLibrary returns only fully qualified URLs
With one of latest upgrade came new feature based on link policy - MediaLibrary class returns [only fully qualified URLs](https://github.com/octobercms/october/blob/2dbc56aead2ef2355135349b83860457f48f6235/modules/system/classes/MediaLibrary.php#L544).
For my usecase is this solution unhandy, I prefer images with relative URLs, based on setting of `cms.storage.media.path`. I think this would be default solution and absolute paths would become optional.

### Proposal of solution
Add to config/cms.php option `cms.storage.media.forceAbsolutePath` to switch on current behavior. Implement this option in `MediaLibrary::getPathUrl`.
Or maybe @mjauvin or @w20k will come with something better.

panakour:
fixes #4082

### DIFF
--- a/modules/system/classes/MediaLibrary.php
+++ b/modules/system/classes/MediaLibrary.php
@@ -739,8 +739,9 @@ class MediaLibrary
      */
     protected function filterItemList(&$itemList, $filter)
     {
-        if (!$filter)
+        if (!$filter) {
             return;
+        }
 
         $result = [];
         foreach ($itemList as $item) {
--- a/modules/system/classes/MediaLibrary.php
+++ b/modules/system/classes/MediaLibrary.php
@@ -535,11 +535,11 @@ class MediaLibrary
     {
         $path = $this->validatePath($path);
 
-        if (Config::get('cms.storage.media.disk') === 'local') {
-            return $this->storagePath . $path;
-        }
-
         $fullPath = $this->storagePath.implode("/", array_map("rawurlencode", explode("/", $path)));
+
+        if (Config::get('cms.storage.media.disk') === 'local') {
+            return $fullPath;
+        }
 
         return Url::to($fullPath);
     }
--- a/modules/system/classes/MediaLibrary.php
+++ b/modules/system/classes/MediaLibrary.php
@@ -535,6 +535,10 @@ class MediaLibrary
     {
         $path = $this->validatePath($path);
 
+        if (Config::get('cms.storage.media.disk') === 'local') {
+            return $this->storagePath . $path;
+        }
+
         $fullPath = $this->storagePath.implode("/", array_map("rawurlencode", explode("/", $path)));
 
         return Url::to($fullPath);


modules/system/classes/MediaLibrary.php
```diff
@@ -535,6 +535,10 @@ public function getPathUrl($path)
     {
         $path = $this->validatePath($path);
 
+        if (Config::get('cms.storage.media.disk') === 'local') {
+            return $this->storagePath . $path;
+        }
+
         $fullPath = $this->storagePath.implode("/", array_map("rawurlencode", explode("/", $path)));
```
LukeTowers: Shouldn't you put the check after this line and return fullPath instead?

LukeTowers: Also @bennothommo can you explain why the code sniffer is saying that inline code structure isn't allowed? It shows up on line 742 in the filterItemList() method but this PR hasn't changed those lines.

bennothommo: @LukeTowers I can't limit the code sniffer to specific lines unfortunately, so it checks the whole file. I figured this would be fine as it would encourage a gradual cleanup of the code.

LukeTowers: Oops, @Samuell1 wrong PR for my comment about fixing this. @panakour can you fix that line?

panakour: @LukeTowers which line? I did not understand what is the problem

bennothommo: @panakour This line here:
https://github.com/octobercms/october/blob/e82a3698efef756cf8efae2b02f5352f3ebdc088/modules/system/classes/MediaLibrary.php#L742
You'll need to add the curly brackets around the `return` call in order to pass the code quality checks.

panakour: @bennothommo done

bennothommo: @panakour Thanks for that. Also, did you have any thoughts as to @LukeTowers's question here? https://github.com/octobercms/october/pull/4423#discussion_r298671428

panakour: I think it's better like i did it. It's unnecessary to execute the next line `        $fullPath = $this->storagePath.implode("/", array_map("rawurlencode", explode("/", $path)));
` if the storage disk is local.

LukeTowers: @panakour it's not unnecessary, that line exists for a reason - to URL encode any filenames that might have spaces or other illegal characters in them. Please make my suggested change so that the URL encoding happens for local files as well as remote files.

bennothommo:
@panakour As mentioned in https://github.com/octobercms/october/issues/4082#issuecomment-591042140, we'll need this to be configurable before it can be considered.

LukeTowers:
Fixed by https://github.com/octobercms/october/commit/ccb2c372bb1fb3980c1240ffb2f7aeef5e786dae. Media library will no longer force absolute URLs unless `cms.linkPolicy` is set to `force`.

PR #4424, opened 2019-06-28
Clean up index functionality in Repeater widget

### LINKED ISSUE #4402: Static page Repeater fields data doesn't saved properly after a field is deleted.
- OctoberCMS Build: ### 455
- PHP Version: 7.2
- Database Engine: MySQL
- Plugins Installed: Rainlab.StaticPages

### Description:
I'm using a simple repeater variable on static page:   
```
{repeater name="content_sections" prompt="Add another content section"}
<h3>
{text name="content_header" label="Content section" placeholder="Type in a heading and enter some content for it below"}{/text}
</h3>
<div>
{richeditor name="content_body" size="large"}{/richeditor}
</div>
{/repeater}
```
On static pages backend i create one static page and using repeater field i create 2 fields. Everything works ok, but if i delete one field and create a new one i get wrong data on save. As you can see on sample video, the problem starts when you delete one field. **If you delete one field you have to close the page and reopen it to be able to add more fields without a problem**. I test it on a clean install with latest build.    

I also noticed that after you delete the first field, the new field you create has filled fields and the data are the same with the first one. You can see it clearly on the video sample.   

Screenshot

### Steps To Reproduce:
1. Create one repeater variable on static page layout
2. Create one static page
3. Create 2 repeater fields and save page
4. Delete first repeater field
5. Create a new repeater field and save page
6. Wrong data saved on content file. Only one field data saved.

bennothommo:
- Remove `$indexCount` property, as it is no longer needed with this fix in place.
- Determine highest key number in data, and increment when adding another item in the repeater.
- Remove some old code and dependencies.

Fixes #4402.

### DIFF
--- a/modules/backend/formwidgets/Repeater.php
+++ b/modules/backend/formwidgets/Repeater.php
@@ -293,7 +293,6 @@ class Repeater extends FormWidgetBase
 
         $groupCode = post('_repeater_group');
 
-        // Gets the next available index number
         $index = $this->getNextIndex();
 
         $this->prepareVars();
@@ -322,6 +321,11 @@ class Repeater extends FormWidgetBase
         return $widget->onRefresh();
     }
 
+    /**
+     * Determines the next available index number for assigning to a new repeater item.
+     *
+     * @return int
+     */
     protected function getNextIndex()
     {
         if ($this->loaded === true) {
--- a/modules/backend/formwidgets/Repeater.php
+++ b/modules/backend/formwidgets/Repeater.php
@@ -3,7 +3,6 @@
 use Lang;
 use ApplicationException;
 use Backend\Classes\FormWidgetBase;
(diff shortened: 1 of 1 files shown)


LukeTowers:
Does this create an issue if someone adds three items, deletes the first, and then adds another? Wouldn't that go:
item1
item2
item3

then 
item2
item3

then 
item2
item3
item3
?

bennothommo:
@LukeTowers It accounts for that by checking the highest index number in the POST data. Eg:

I have three items:
item[0]
item[1]
item[2]

And I delete item[0] and item[1], then add a new item, it will still see the highest index is 2, so the next item will be item[3].

Alternatively, if I delete item[1] and item[2], it will see the highest index as 0, so the next item will be item[1].

LukeTowers:
@bennothommo awesome, missed the max() part. Thanks for thinking about that!

PR #4428, opened 2019-07-03
Fix error when $user is null

alxy:
I have no idea why the $user variable can ever be null in the first place, but for me it happens if I remove all main menu items (including the settings). The same check is already in place at the navigation manager: https://github.com/octobercms/october/blob/master/modules/backend/classes/NavigationManager.php#L509-L511
The change shouldn't break anything.

To reproduce the error, you can use this code:
```php
protected function removeMainMenuItems()
    {
        $navigationManager = NavigationManager::instance();
        foreach ($navigationManager->listMainMenuItems() as $mainMenuItem) {
            $navigationManager->removeMainMenuItem($mainMenuItem->owner, $mainMenuItem->code);
        }
    }
```

### DIFF
--- a/modules/system/classes/SettingsManager.php
+++ b/modules/system/classes/SettingsManager.php
@@ -365,6 +365,10 @@ class SettingsManager
      */
     protected function filterItemPermissions($user, array $items)
     {
+        if (!$user) {
+            return $items;
+        }
+        
         $items = array_filter($items, function ($item) use ($user) {
             if (!$item->permissions || !count($item->permissions)) {
                 return true;


bennothommo:
@alxy Just before we test this, would you mind removing the empty line here just to appease the Travis code quality check:
https://github.com/octobercms/october/blob/7c6631f2d418604aaa8167862a5cbeb1dd5abe6e/modules/system/classes/SettingsManager.php#L174

LukeTowers:
@alxy would you also be willing to add tests similar to what @daftspunk did in https://github.com/octobercms/october/commit/bb4c2da4043780b982273ebbeb983deae2845774?

alxy:
I have tracked this down to the issue that i tried to modify the navigation structure in the `boot()` method - and apparently `BackendAuth::getUser()` returns null there. Is there any other more appropriate place to put these kinds of code? (I seem to remember that in the past the $user was available in `boot()` by the way.)

LukeTowers:
BackendAuth::getUser() should be able to return the user in boot(), @bennothommo is this related to our controller middleware shenanigans from a while ago do you think?

bennothommo:
@LukeTowers it is a possibility, but that was functionality was restricted to controllers - it theoretically shouldn't have any bearing on plugins (I'm presuming @alxy meant the `boot()` method in a plugin)

LukeTowers:
@alxy are you able to test and verify that it worked previously (perhaps in Build 419) and make an issue for it with detailed reproduction instructions?

daftspunk:
Looks like nothing is happening here re: tests. The change is acceptable because it syncs up the pattern from file **modules\backend\classes\NavigationManager.php**

Thanks!

daftspunk:
BTW this is likely caused by the Laravel upgrade where the session was no longer available inside service providers, introduced in Laravel 5.5

PR #4433, opened 2019-07-05
Fix backend same filename check bug

wenlong-date:
**Summary:**
 In October backend cms module,when you have same filename in Pages and Partials, and you select the Partials's file, unfortunately, it can't be checked.

**Reproduce steps:**
 1. In October backend cms module, create a page name's index, create a partial name's index 
2. select the partial's index file

**Expected behavior:** The partial's index file will be selected.

**Actual behavior:** The page's index file will be selected and the partial's index file will not be selected

**Bug show:**
oc_file_checkbox_bug

### DIFF
--- a/modules/cms/widgets/templatelist/partials/_items.htm
+++ b/modules/cms/widgets/templatelist/partials/_items.htm
@@ -36,7 +36,7 @@
 
                     <input type="hidden" name="template[<?= e($item->fileName) ?>]" value="0" />
                     <div class="checkbox custom-checkbox nolabel">
-                        <?php $cbId = 'cb'.md5($this->itemType . '/' . $item->fileName) ?>
+                        <?php $cbId = 'cb' . md5($this->itemType . '/' . $item->fileName) ?>
                         <input
                             id="<?= $cbId ?>"
                             type="checkbox"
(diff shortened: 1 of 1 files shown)


modules/cms/widgets/templatelist/partials/_items.htm
```diff
@@ -36,7 +36,7 @@ <h4><a href="javascript:;"><?= e($item->title) ?></a></h4>
 
                     <input type="hidden" name="template[<?= e($item->fileName) ?>]" value="0" />
                     <div class="checkbox custom-checkbox nolabel">
-                        <?php $cbId = 'cb'.md5($item->fileName) ?>
+                        <?php $cbId = 'cb'.md5($this->itemType . '/' . $item->fileName) ?>
```
bennothommo: @wenlong-date Minor fix - would you mind adding some spacing around concatenated strings?
```suggestion
                        <?php $cbId = 'cb' . md5($this->itemType . '/' . $item->fileName) ?>
```

bennothommo:
@wenlong-date @LukeTowers confirmed the bug, and tested the fix, which does fix the issue. Great job @wenlong-date.

modules/cms/widgets/templatelist/partials/_items.htm
wenlong-date: surer! Thanks. I formatted code and had committed.

Issue #3415, opened 2018-02-26
Custom Form Widgets With AJAX Handlers Won't Work As Expected Inside Repeaters


theodark:
Adding a custom form widget, with AJAX handlers, inside a repeater should call that handler upon request. However the response throws an SystemException, even if the said custom widget uses the `$this->getEventHandler('onXXX')` method to access the handler.

In my case I was using the select2 to implement a custom "relation" style widget. Below is the partial that does the fetching request:

````html
<select class="form-control jq-select2" id="<?= $id ?>" name="<?= $name ?>">
</select>
<script lang="javascript" type="text/javascript">
	$(document).ready(function () {
		$('select.jq-select2').select2({
			'placeholder': '(Select Item)',
			'minimumInputLength': 3,
			'templateResult' : function (data) {
				if(!data.id) return data.text;
				console.log(data);
				var option = $(data.text);
				return option;
			},
			'ajax': {
				'delay': 500,
				'url': '',
				'dataType': 'json',
				'transport': function (params, success, failure) {
					$.request('<?= $this->getEventHandler('onFetchResultList') ?>',{
						'data' : { 'query' : params.data.term },
						'success' : success,
						'error' : failure
					});
				}
			}
		});
	});
</script>
````

And this ends up in a php function at the back end, in the style of:
````php
public function onFetchResultList() {
	return [
		"results" => $this->forgeQueryResults(strip_tags(input("query")))
	];
}
````
At rendering, I insert the widget inside a data repeater with name `contents_json` and I give it a name `magazine` - for example. In which case `$this->getEventHandler('onFetchResultList')` renders to `formContentsJsonForm1Magazine` when the partial is rendered in the form.

Upon ajax requests, October returns the following exception in the response
> "A widget with class name 'formContentsJsonForm1Magazine' has not been bound to the controller" on line 513 of /domains/example.domain/httpdocs/modules/backend/classes/Controller.php

I imagine that this should not happen under normal circumstances.
___

**For whom it may concern:** A quick workaround to this issue is to add another widget of the same kind at the top level form fields, then give it a name - f.e. `magazine_ajax` and set it to hidden, just to force the controller to load the widget class in the form. Then instead of using the `getEventHandler()` method in the partial, hardcode the name that the controller uses to identify that widget, in the javascript.  For example, the above `$.request` code of the partial would turn to

````javascript
	$.request('formMagazineAjax::onFetchResultList') ?>
````

##### October build
434

iotch:
[...]

PR #4434, opened 2019-07-06
Fix "widget not bound to controller" error for nested repeaters

### LINKED ISSUE #4344: Nested Repeater - A widget with class name has not been bound to the controller
- OctoberCMS Build: 455
- PHP Version: 7.1.9
- Database Engine: MySQL

### Description:
Nested Repeater does not work if it has **only one** repeater field. For example, I have a "Content" Repeater. This repeater has another "Blocks" Repeater inside it.
```
    content:
        label: Content
        type: repeater
        form:
            fields:
                blocks:
                    label: Blocks
                    type: repeater
                    form:
                        fields:
                            title:
                                label: Description
                                type: textarea

```
When I try to add Block I get this error:

> "A widget with class name 'formContentForm0Blocks' has not been bound to the controller" on line 598 of \modules\backend\classes\Controller.php

Here is a screenshot:
image

But if "Content" Repeater has more fields, I don't have this error. For example, I added "Name" field to "Content" Repeater:
```
    content:
        label: Content
        type: repeater
        form:
            fields:
                name:
                    label: Name
                    type: text
                blocks:
                    label: Blocks
                    type: repeater
                    form:
                        fields:
                            title:
                                label: Description
                                type: textarea
```
There is no error:
image

I also noticed that we will have the error if we use `radio` field type instead of `text` field type in the previous example. For example, in this case we will have the error:
```
    content:
        label: Content
        type: repeater
        form:
            fields:
                position:
                    label: Position
                    type: radio
                    options:
                        left: Left
                        center: Center
                        right: Right
                blocks:
                    label: Blocks
                    type: repeater
                    form:
                        fields:
                            title:
                                label: Description
                                type: textarea
```

### Steps To Reproduce:
Create a "content" column in your table. Add it to the `$jsonable` array in your model. Add my first example to the `fields.yaml` and try to add Content and Blocks items in the Repeater. You will get an error. After that add my second example to the `fields.yaml` and try to add Content and Blocks items in the Repeater. You will not get any errors.

bennothommo:
Since repeaters have no value by themselves, a repeater that contains only a child repeater was throwing a "widget not bound to controller" exception when adding a new item. This is due to the repeater `processItems()` method not creating a form widget for the inner repeater as it contained no value.

This fix specifically checks if a child repeater has called for a new item - if this is the case, and the current repeater has no value for this particular item, it will add a "stub" form widget to correctly bind the inner repeater widget.

Fixes #4344, ~~and quite possibly #3415 as well.~~

### DIFF
--- a/modules/backend/formwidgets/Repeater.php
+++ b/modules/backend/formwidgets/Repeater.php
@@ -115,7 +115,7 @@ class Repeater extends FormWidgetBase
             $this->loaded = true;
         }
 
-        $this->checkAjaxRequest();
+        $this->checkAddItemRequest();
         $this->processGroupMode();
 
         if (!self::$onAddItemCalled) {
@@ -364,17 +364,24 @@ class Repeater extends FormWidgetBase
     }
 
     /**
-     * Determines the repeater that has triggered an AJAX request.
+     * Determines the repeater that has triggered an AJAX request to add an item.
      *
      * @return void
      */
-    protected function checkAjaxRequest()
+    protected function checkAddItemRequest()
     {
         $handler = $this->getParentForm()
             ->getController()
             ->getAjaxHandler();
 
-        list($widgetName) = explode('::', $handler);
+        if ($handler === null || strpos($handler, '::') === false) {
+            return;
+        }
+
+        list($widgetName, $handlerName) = explode('::', $handler);
+        if ($handlerName !== 'onAddItem') {
+            return;
+        }
 
         if ($this->alias === $widgetName) {
             // This repeater has made the AJAX request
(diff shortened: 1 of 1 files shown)


bennothommo:
@LukeTowers I noticed you've added yourself as a reviewer for this task - did you want to review it and merge if its all good, to at least resolve #4344 and be able to proceed with build 458? It doesn't fix #3415, but I can tackle that in build 459.

Samuell1:
Can this fix same issue with forms in custom popups?

bennothommo:
@Samuell1 I don't think so - #3415 is a report for that error in a relation popup. I haven't been able to resolve that yet.

Samuell1:
@bennothommo But thats inside repeaters no? I mean custom created popups for create, update.

bennothommo:
@Samuell1 It definitely wouldn't fix those then - my fix is purely for repeaters only.

PR #4445, opened 2019-07-10
Fix text filter scope

alxy:
This tries to fix the behavior described here, which happens when more than one text filter scope is present: https://github.com/octobercms/october/pull/4241#issuecomment-510140574

The problem is that once the filter executes (i.e. new data is entered into the text input), it adds data to the `options.value` POST array. On regular lists, this array only contains the value of the field beeing edited/the filter beeing used. However, in the relation controller the POST data looks like this:
```
options[value][]: asd
options[value][]: have
```
The system will thus always use the first value in that array for each of the following text filters, meaning all fields will be filtered by the value of the first scope.

This PR adds the scope name to the POST data to make things expicit. I am, however, a bit concerned about side effects, as I do not understand why this comply array setup for the POST data is used in the first place.

### DIFF
--- a/modules/backend/widgets/Filter.php
+++ b/modules/backend/widgets/Filter.php
@@ -261,15 +261,7 @@ class Filter extends WidgetBase
                 break;
 
             case 'text':
-                $values = post('options.value');
-
-                if ($values !== null && $values !== '') {
-                    list($value) = $values;
-                }
-                else {
-                    $value = null;
-                }
-
+                $value = post('options.value.' . $scope->scopeName) ?: null;
                 $this->setScopeValue($scope, $value);
                 break;
         }
--- a/modules/backend/widgets/Filter.php
+++ b/modules/backend/widgets/Filter.php
@@ -116,7 +116,7 @@ class Filter extends WidgetBase
                     $after = $scope->value[0]->format('Y-m-d H:i:s');
                     $before = $scope->value[1]->format('Y-m-d H:i:s');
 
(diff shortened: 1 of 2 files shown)


LukeTowers:
@alxy could you perform the code cleanup tasks requested by Travis on the file? https://travis-ci.org/octobercms/october/jobs/556933466#L792

alxy:
@LukeTowers I can do this if it will be accepted. This is a very tedious task, especially since the output of travis is kind of raw. Can't we automate this? (Additionally, most of the things are not related to my specific PR but to the file in general. And I don't feel it helps if contributors should from now on clean up complete files even though they only changed minor things ;) For me this is really a game breaker.)

LukeTowers:
@alxy Thanks, I appreciate your contributions! This looks good to me, I'll just wait for @w20k to test it and confirm it fixes the issue and doesn't break existing non-relation manager filters and then it should be good to merge!

w20k:
@LukeTowers will do, will take me a few days or so. `Overcooked` at work on this week, not able to test/check PRs actively.

w20k:
Looks good so far, but need to test with other plugins, just to make sure.
Screenshot:
Screenshot 2019-07-12 at 16 13 56

alxy:
I would like to have @daftspunk s feedback on this one. Still unsure why it uses this complex array solution originally.

LukeTowers:
I looked at the original commit adding it and it's something that I let through. It's needlessly complex in that case for no reason, I just didn't catch that in my original review. As such I'm approving this change.

PR #4447, opened 2019-07-11
update cms zh-cn lang.php file

wenlong-date:
When i use OctoberCMS, i found the translations of `modules/cms/lang/zh-cn/lang.php` was out of date. So, i updated it .

### DIFF
--- a/modules/backend/lang/zh-cn/lang.php
+++ b/modules/backend/lang/zh-cn/lang.php
@@ -2,12 +2,14 @@
 
 return [
     'auth' => [
-        'title' => '管理'
+        'title' => '管理',
+        'invalid_login' => '无法匹配到您输入的相关内容，请检查后重试。'
     ],
     'field' => [
         'invalid_type' => '不合法的字段类型 :type.',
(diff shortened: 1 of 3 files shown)


wenlong-date:
Backend zh-cn lang.php file not finish translation at the moment . I will finish it before Jul 11  23:59:59.

wenlong-date:
@bennothommo  I have tried my best to translate the files and finished now.

modules/backend/lang/zh-cn/lang.php
```diff
@@ -501,7 +553,7 @@
         'label' => '媒体查找器',
         'default_prompt' => '点击 %s 按钮查找媒体项'
     ],
-   'media' => [
+    'media' => [
```
w20k: @bennothommo any ideas why CI approve space? Strange)

w20k: @wenlong-date could you remove space before 'media', doesn't seem to be correct with whitespace.

bennothommo: @w20k The extra space puts it back in correct alignment, from what I can tell, which is why it passes.

w20k: @bennothommo you're right, need a rest or two :) Thanks!

PR #4449, opened 2019-07-11
Code quality clean up

bennothommo:
I have done a clean-up of all our PHP files to match the spec in our [phpcs.xml](https://github.com/octobercms/october/blob/develop/phpcs.xml) definition. This was mainly due to several recent PRs being pinged by our CI tests due to code quality issues that were not introduced by the PRs themselves. With these changes, the code should be at a clean slate so the tests will only ping newly-introduced issues.

These changes are purely cosmetic and don't affect functionality, so it should be safe to include this in build 457.


LukeTowers:
@bennothommo please wait until the rest of the 457 issues / PRs are merged, I'd rather we have to fix the merge conflicts ourselves then make the other contributors do it in their PRs.

bennothommo:
@LukeTowers agreed.

w20k:
@bennothommo LGTM, though for me personally old style is more readable:

Old
```
       return new ArrayIterator($this->suppressTabs
            ? $this->getAllFields()
            : $this->getFields()
       );
```
New
```
        return new ArrayIterator(
            $this->suppressTabs ? $this->getAllFields() : $this->getFields()
        );
```

bennothommo:
@w20k I agree that ternary conditions are very subjective - PSR-1 and PSR-2 don't really have any rulings on how they should be formatted. The original way it was formatted failed in PHPCS simply because multi-line arguments should not begin on the same line as the calling function.

I generally believe that a ternary condition should always be one line, but when it's long, it should be formatted in a way that makes a clear separation between the condition and the results, so perhaps it could be formatted to the following:

```php
return new ArrayIterator(
    $this->suppressTabs
        ? $this->getAllFields()
        : $this->getFields()
);
```

LukeTowers:
@bennothommo I'm with you on how ternary conditions should be formatted. I'm fine with leaving it one line but I also don't care too much with it being moved to two lines. I'll leave that up to you to decide.

w20k:
Lost this one @bennothommo up to you to decide!

bennothommo:
@w20k @LukeTowers I'll change it to the multi-line definition as above - I think it makes most sense when compared to an `if` statement:
```
condition
    ? true
    : false
```
vs.
```
if (condition) {
    true
} else {
    false
}
```
We'll have to enforce it ourselves if we want to keep with this format, as PHPCS doesn't have a specific check for ternary conditions. While it's possible to code one up, I'd say it's quite low priority. ;)

LukeTowers:
@bennothommo I wouldn't worry about enforcing it too much. I'm probably going to stick to single line ternaries personally.

modules/system/console/ThemeSync.php
```diff
@@ -181,7 +181,9 @@ public function handle()
         }
 
         // Confirm with the user
-        if (!$this->confirmToProceed(sprintf('This will OVERWRITE the %s provided paths in "themes/%s" on the %s with content from the %s', count($validPaths), $themeName, $target, $source), function () { return true; })) {
+        if (!$this->confirmToProceed(sprintf('This will OVERWRITE the %s provided paths in "themes/%s" on the %s with content from the %s', count($validPaths), $themeName, $target, $source), function () {
+            return true;
+        })) {
```
LukeTowers: @bennothommo this is more clear being a one line closure. According to [the docs](https://laravel.com/api/5.5/Illuminate/Console/ConfirmableTrait.html#method_confirmToProceed) it supports a bool argument so this closure can just be replaced by `true`.

tests/UiTestCase.php
```diff
@@ -60,32 +67,39 @@ protected function getSweetConfirmation($expectedText = null, $clickOk = true)
 
     protected function waitForElementPresent($target, $timeout = 60)
     {
-        for ($second = 0; ; $second++) {
-            if ($second >= $timeout)
+        for ($second = 0;; $second++) {
```
LukeTowers: hmm, not really a fan of this. Is there a different control structure we can use? How about a while(true)? That'd be more clear that it's meant to evaluate infinitely. Same for `waitForElementNotPresent`

tests/unit/system/classes/CombineAssetsTest.php
```diff
@@ -49,7 +50,8 @@ public function testCombine()
         $this->assertNotNull($url);
         $this->assertRegExp('/\w+[-]\d+/i', $url); // Must contain hash-number
 
-        $url = $combiner->combine([
+        $url = $combiner->combine(
+            [
             'assets/js/script1.js',
             'assets/js/script2.js'
```
LukeTowers: Indentation on these two lines needs to be added

modules/system/console/ThemeSync.php
bennothommo: @LukeTowers I actually would prefer it being out of the `if` statement altogether, but I've changed it to `true` as per your recommendation, just to get the ball rolling.

tests/UiTestCase.php
bennothommo: @LukeTowers Changed to `while` loops.

tests/unit/system/classes/CombineAssetsTest.php
bennothommo: @LukeTowers done.

PR #4457, opened 2019-07-16
Fix layout of color picker for no color selected

anitaeisenhaber:
If the option. 
`allowEmpty: true`
is active the current design is a bit messy and doesn't act like the original spectrum color picker.

Before:
Screenshot 2019-07-16 at 16 58 09

After:
image

### DIFF
--- a/modules/backend/ServiceProvider.php
+++ b/modules/backend/ServiceProvider.php
@@ -76,6 +76,7 @@ class ServiceProvider extends ModuleServiceProvider
             $combiner->registerBundle('~/modules/backend/formwidgets/fileupload/assets/less/fileupload.less');
             $combiner->registerBundle('~/modules/backend/formwidgets/nestedform/assets/less/nestedform.less');
             $combiner->registerBundle('~/modules/backend/formwidgets/richeditor/assets/js/build-plugins.js');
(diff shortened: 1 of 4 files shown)


LukeTowers:
LGTM. Will wait for @w20k or @bennothommo to test and verify.

w20k:
@LukeTowers LGTM - tested

Screenshots:
Before:
Screenshot 2019-07-17 at 16 45 14
After:
Screenshot 2019-07-17 at 16 44 55

modules/backend/ServiceProvider.php
```diff
@@ -76,6 +76,7 @@ protected function registerAssetBundles()
             $combiner->registerBundle('~/modules/backend/formwidgets/fileupload/assets/less/fileupload.less');
             $combiner->registerBundle('~/modules/backend/formwidgets/nestedform/assets/less/nestedform.less');
             $combiner->registerBundle('~/modules/backend/formwidgets/richeditor/assets/js/build-plugins.js');
+            $combiner->registerBundle('~/modules/backend/formwidgets/colorpicker/assets/less/colorpicker.less');
```
LukeTowers: @w20k widget assets are recompiled when they're registered like this and then `php artisan october:util compile` is run.

PR #4459, opened 2019-07-17
Add chunk support for exporting and importing csv

Samuell1:
This should allow exporting extreme amount of rows without exhausting memory. If anyone can test this out it will help to find any bugs or if threshold is too big.

Merge after https://github.com/octobercms/october/pull/4363 PR is merged


LukeTowers:
@Samuell1 reason for close?

Samuell1:
@LukeTowers because it doesnt work when streamedresponse is called in other method, flush() needs to be callback for that otherwise it doesnt work.

PR #4461, opened 2019-07-17
Update jQuery to V3.4.1 and jQuery Migrate to V3.1.0

ayumihamsaki:
Connected to Github info page: https://github.com/octobercms/october/issues/3867

# Notes

## jQuery V3.4.1

Below is the following commits from their release:

- Build: Update Sinon from 2.3.7 to 7.3.1, other updates
- Tests: Fix the core-js polyfill inclusion method
- Build: Update the master version to 4.0.0-pre
- CSS: Don't automatically add "px" to properties with a few exceptions
- Build: Update Sizzle from 2.3.3 to 2.3.4
- Release: update AUTHORS.txt
- Build: Fix unresolved jQuery reference in finalPropName
- Build: Test on Node.js 12, stop testing on Node.js 6 & 11
- Event: Fix handling of multiple async focus events
- Event: Prevent leverageNative from registering duplicate dummy handlers
- Data: Separate data & css/effects camelCase implementations
- Event: Stop shimming focusin & focusout events
- Tests: Fix the new focusin/focusout test in IE
- Core: Remove deprecated jQuery APIs
- Tests: Restrict an event test fallback to TestSwarm
- Core: Drop support for IE <11, iOS <11, Firefox <65, Android Browser and PhantomJS
- Build: Fix AMD dependencies in curCSS
- Effect: Fix a unnecessary conditional statement in .stop()

This update should not create any issues to the current version of October CMS as it is a minor semantic update version number.

## jQuery Migrate V3.1.0

Below is the following commits from their release:

- IMPORTANT: Build: update dependencies to **_fix vulns_**
- Core: Warn on jQuery.isNumeric and jQuery.type
- Core: Get pre-4.0 compat by implementing removed functions
- CSS: Warn and fill access to jQuery.cssProps
- Core: Warn and fill jQuery.isArray

### JQMIGRATE: `jQuery.isNumeric()` is deprecated

**Cause**: This method was used by jQuery to determine if certain string arguments could be converted to numbers, but the name led people to apply their own interpretations to what the method means. As a result, it often doesn't meet the needs of specific cases. For example, a 25-character string of only digits is technically a valid number, but JavaScript cannot represent it accurately. The string `"0x251D"` is a valid hexadecimal number but may not be acceptable numeric input to a web form.

**Solution**: Use a test for being numeric that makes sense for the specific situation. For example, instead of `jQuery.isNumeric(string)`, use `isNan(parseFloat(string))` if a floating point number is expected, or `string.test(/^[0-9]{1,8}$/)` if a sequence of 1 to 8 digits is expected.

### JQMIGRATE: `jQuery.type()` is deprecated

**Cause**: This method returns a string that indicates the type of the argument, for example `"number"` or `"function"`. However, as the JavaScript language evolves this method has become problematic because new language constructs might require this function to either return a new string (potentially breaking existing code) or somehow map new constructs into existing strings (again, potentially breaking existing code). Examples of new recent JavaScript features include asynchronous functions, class constructors, `Symbol`s, or functions that act as iterators.

**Solution**: Review code that uses `jQuery.type()` and use a type check that is appropriate for the situation. For example. if the code expects a plain function, check for `typeof arg === "function"`.

### JQMIGRATE: `jQuery.cssProps` is deprecated

**Cause:** The `jQuery.cssProps` property is a public but undocumented object that allows CSS properties with one name to be mapped into another name. It was used for legacy browsers like IE8 that used non-standard names. This object is no longer used inside jQuery since all supported browsers now use the standard CSS property names.

**Solution:** Remove any uses of `jQuery.cssProps` in application code.

### JQMIGRATE: `jQuery.isArray` is deprecated; use `Array.isArray`
**Cause:** Older versions of JavaScript made it difficult to determine if a particular object was a true Array, so jQuery provided a cross-browser function to do the work. The browsers supported by jQuery 3.0 all provide a standard method for this purpose.

**Solution:** Replace any calls to `jQuery.isArray` with `Array.isArray`.

---

**_A quick search in October CMS code can not find any of these deprecated codes._**


LukeTowers:
@ayumihamsaki was there anywhere in the core JS files of October that use the newly deprecated code?

ayumihamsaki:
@LukeTowers 

> was there anywhere in the core JS files of October that use the newly deprecated code?

No, none I checked each one.

ayumihamsaki:
@LukeTowers 

> was there anywhere in the core JS files of October that use the newly deprecated code?

No none, I checked each one.

I think we are good to go on this one. But you may want @w20k just to do a quick double check if you like.

w20k:
@LukeTowers just wanted to clarify, we're going for this upgrade? Or it's on hold for now?

LukeTowers:
@w20k lets hold off for now, I want to get the preload issue dealt with on the frontend frameworks tag so that we can go stable pretty much immediately with 458

ayumihamsaki:
I thought the stable version was going to be v1.0.500 because the updates are coming out so fast. 😆

ayumihamsaki:
Need to add to this pr an updated FrameworkNode.php with the new Subresource Integrity for the loaded `jquery` and `jquery migrate` files.

As per comment here: https://github.com/octobercms/october/pull/4285#issuecomment-513784346

ayumihamsaki:
@LukeTowers @w20k 

- `$( handler )`
- `$( document ).ready( handler )`
- $`( "document" ).ready( handler )`
- `$( "img" ).ready( handler )`
- `$().ready( handler )`

As of jQuery 3.0, **only the first syntax is recommended**; the other syntaxes still work but are `deprecated`.

```js
$( document ).ready(function() {
  // Handler for .ready() called.
});
```

into

```js
$(function() {
  // Handler for .ready() called.
});
```

Just wondering if it would be a good time to update these, before jQuery v4 comes out and maybe they stop the backwards compatibility ?

The reason I write this comment is because the current version of jQuery Migrate says the following in dev-tools:

ffff

At some point jQuery Migrate may stop patching this.

w20k:
@ayumihamsaki it's not yet in the deprecated list (https://api.jquery.com/category/deprecated/). So, don't see a point, yet.

ayumihamsaki:
Also updated doc's https://github.com/octobercms/october/pull/4285#issuecomment-514580505

See here: https://octobercms.com/docs/markup/filter-theme#combiner-aliases

> Reference to the jQuery library (v3.4.0) used in the back-end. (JavaScript)

ayumihamsaki:
Hi, I'm just going through all my website jquery codes and updating any deprecated code. But it's a bit difficult because it's hard to tell if the error is coming from October or not? If October had no faults, then I can pin point the errors my side easier.

What I'm saying is, would it be possible to make a discussion about fixing the jquery deprecated code or not? Then I will open a github issue and go through and fix them.

ayumihamsaki:
Should `.resize();` not be debounced? Seems like October just uses resize without any limiting at all. Can't imagine that's a good idea.

w20k:
@ayumihamsaki it's better to have a `debounce` when using `resize()` function. It's due to a difference in timing which browsers use: IE, FF, and Chrome that time window is different. So, yeah, debounce helps a lot to make that difference as slim as possible.

LukeTowers:
@ayumihamsaki I noticed you made a lot of JS changes in your massive PR related to deprecations in jQuery, can you make those changes here instead please?

ayumihamsaki:
@LukeTowers I only did the `.ready()` removal in the big pr. I can't add to this pr because github has deleted it see here: https://github.com/octobercms/october/tree/jquery

But I can try and create a separate branch for the jquery ready.

---

@w20k totally makes sense what you are saying, that would explain why I am seeing some really annoying bugs in October.

Things like:

Main layout issue:

https://github.com/octobercms/october/blob/610e0c383cbf92f17def3ce64ddd59c1204f1eff/modules/backend/assets/js/october.layout.js#L78

Sidebar issue:

https://github.com/octobercms/october/blob/4e92686c1a7de662840fa160b7348003f5ba817d/modules/backend/assets/js/october.sidepaneltab.js#L92

etc.

This file is the worst out of all of October:

https://github.com/octobercms/october/blob/develop/modules/system/assets/ui/vendor/flot/jquery.flot.resize.js

ayumihamsaki:
- [ ]  jQuery bind
- [x] .ready()
- [ ] click
- [ ] resize (needs debounce)

ayumihamsaki:
May need some help with the debounce, shall I just fix the short-hand for now ?

ayumihamsaki:
List of files containing resize:

https://github.com/octobercms/october/blob/master/modules/system/assets/ui/vendor/flot/jquery.flot.resize.js

https://github.com/octobercms/october/blob/master/modules/backend/formwidgets/codeeditor/assets/js/codeeditor.js

https://github.com/octobercms/october/blob/master/modules/backend/assets/js/october.sidepaneltab.js

https://github.com/octobercms/october/blob/master/themes/demo/assets/vendor/bootstrap/js/modal.js

https://github.com/octobercms/october/blob/master/modules/backend/formwidgets/markdowneditor/assets/js/markdowneditor.js

https://github.com/octobercms/october/blob/master/modules/backend/assets/js/october.filelist.js

https://github.com/octobercms/october/blob/master/modules/system/assets/js/mailbrandsettings/mailbrandsettings.js

https://github.com/octobercms/october/blob/master/modules/backend/assets/js/october.layout.js

https://github.com/octobercms/october/blob/master/modules/system/assets/ui/js/toolbar.js

w20k:
@ayumihamsaki ping me on PR if you need JS/CSS help 😉

ayumihamsaki:
@w20k 

Basically I understand the:

https://github.com/octobercms/october/blob/d67f6c42bc2500a24574628dd93177eb912cfbbb/modules/backend/formwidgets/codeeditor/assets/js/codeeditor.js#L122

Do this:

```js
var debounced;

window.onresize = function() {
    clearTimeout(debounced);
    debounced = setTimeout(aceResize, 100);
};

function aceResize() {
    this.proxy(this.onResize);
}
```

I am stuck on the : 

https://github.com/octobercms/october/blob/d67f6c42bc2500a24574628dd93177eb912cfbbb/modules/backend/formwidgets/codeeditor/assets/js/codeeditor.js#L274

I'm also stuck on the: 

https://github.com/octobercms/october/blob/d67f6c42bc2500a24574628dd93177eb912cfbbb/modules/backend/formwidgets/codeeditor/assets/js/codeeditor.js#L372

Then there is very complex stuff I not mention yet.

w20k:
@ayumihamsaki if you really want to refactor resize to use a proper debounce option. I'd suggest creating a separate file `ocms-resize.js` (or something) and extend a jQuery object with a new function, which should be used for resize event. And don't forget only, mainly, you should replace only `$(window).resize()` event, not everything that has the method with the name `resize` in it.

Jquery extend with debounce example: https://stackoverflow.com/a/40534918/1053650

ayumihamsaki:
Thanks, that's a very cool repo.

LukeTowers:
@ayumihamsaki please move all the jQuery update related PRs (.bind(), .unbind(), .ready(), .click(), .resize()) into this PR

ayumihamsaki:
No problem will do that today, sorry about that, I thought it would be easier for w20k to test it thing separately.

ayumihamsaki:
Closed please go to this PR: https://github.com/octobercms/october/pull/4534

PR #4464, opened 2019-07-20
Add support for tab icons in theme customization

### LINKED ISSUE #4131: Tab Icons are not supported in Theme Data Model
- OctoberCMS Build: 447
- PHP Version: 7.1.17
- Database Engine: mysql
- Plugins Installed: n/a

### Description:
Extending Theme model with tab icon definitions don't display in page.

### Steps To Reproduce:

Either add the following to a plugin `Plugin.php`:

```php
public function boot()
{
     Event::listen('cms.theme.extendFormConfig', function ($themeCode, &$config) {
        $config['tabs'] = [
            'icons' => [
                'Global'  => 'icon-globe',
            ],
        ];
        array_set($config, 'tabs.fields.test_asset_var', [
            'label'    => 'Test Asset Var',
            'type'     => 'text',
            'assetVar' => 'testAssetVar'
            'span'     => 'left',
            'tab'      => 'Global',
        ]);
        return $config;
    });
}
```

or use the following:

* https://github.com/octoberrain/test-plugin/pull/62

Samuell1:
Resolves #4131


modules/backend/widgets/Form.php
```diff
@@ -711,6 +711,18 @@ public function addFields(array $fields, $addToArea = null)
...
+     * Add tab icons
+     *
+     * @param array $icons
+     * @return void
+     */
+    public function addTabIcons(array $icons)
+    {
+        $this->allTabs->primary->icons = $icons;
+        $this->allTabs->secondary->icons = $icons;
+        $this->allTabs->outside->icons = $icons;
```
bennothommo: @Samuell1 Will assigning the same icons array to all the tabs result in, for example, a primary tab called "Colours" and secondary tab called "Colours" having the same icon?

LukeTowers:
I'm not sure I like this, why does it not just work? Isn't the form initialized from the config just like a normal use of the Form widget? In that case, why doesn't this work properly? I'm more interested in answering those questions than adding specific exception handling for each individual thing that we may add support for to forms in the future.

Samuell1:
@LukeTowers because this is extending form not using already created form fields in theme

modules/backend/widgets/Form.php
Samuell1: Yes, but there is 0% chance that somebody wants to add same tabs names for theme customization.

LukeTowers:
@Samuell1 I see why, because it doesn't load the config properly: https://github.com/octobercms/october/blob/master/modules/cms/controllers/themeoptions/config_form.yaml#L9; it only uses extension to load the config. 

We can instead use `formExtendFieldsBefore($form)` and abuse the fact that `$this->config` is a public property on widgets (https://github.com/octobercms/october/blob/master/modules/backend/classes/WidgetBase.php#L26) to dynamically replace the config of the widget with the config defined by the theme. Let me know if you get stuck figuring out how to do that.

modules/backend/widgets/Form.php
bennothommo: @Samuell1 I wouldn't be so sure - for example, someone has a primary tab called Colours to edit the global colour scheme, and then a primary tab called Header, which has a secondary tab called Colours to handle just the colourization of the header element. Now both of those Colours tabs have the same icon.

Samuell1: @bennothommo Yeah, we solved it with @LukeTowers with more robust way

PR #4471, opened 2019-07-24
Add caching in ComponentBase

### LINKED ISSUE #2505: [feature_request] Cachable component
Hi OctoberCMS team,

I would like to have a feature that allows me to configure the caching behavior of a partial view in a component. To support that feature, I would like to add this class into the core of OctoberCMS (if you don't mind).

Please check my code at: https://gist.github.com/khoatran/f6fadc3a70b577910fc55408d66ed8a6 for the CachableComponent class and the example of how to use it.

Cachable components/widget actually is a very popular concept in other CMS. So, it's great if OctoberCMS can support this.

Best regards,

Samuell1:
Resolves #2505

### DIFF
--- a/modules/cms/classes/ComponentBase.php
+++ b/modules/cms/classes/ComponentBase.php
@@ -138,7 +138,8 @@ abstract class ComponentBase extends Extendable
     public function onRender()
     {
         if ($this->cacheTTL) {
-            return Cache::remember($this->getCacheKey(), $this->cacheTTL, function () {
+            $expires = is_integer($this->cacheTTL) ? now()->addSeconds($this->cacheTTL) : $this->cacheTTL;
+            return Cache::remember($this->getCacheKey(), $expires, function () {
                 return $this->renderPartial('@default');
             });
         }
--- a/modules/cms/classes/ComponentBase.php
+++ b/modules/cms/classes/ComponentBase.php
@@ -5,6 +5,7 @@ use Lang;
 use Config;
 use October\Rain\Extension\Extendable;
 use BadMethodCallException;
+use Cache;
 
 /**
  * Component base class
@@ -37,6 +38,11 @@ abstract class ComponentBase extends Extendable
      * @var boolean Determines whether the component is hidden from the back-end UI.
      */
     public $isHidden = false;
+
+    /**
+     * @var int Determines if component should cache default partial in onRender hook. Uses minutes.
+     */
+    public $cacheTTL = null;
 
     /**
      * @var string Icon of the plugin that defines the component.
@@ -131,6 +137,11 @@ abstract class ComponentBase extends Extendable
      */
     public function onRender()
     {
+        if ($this->cacheTTL) {
+            return Cache::remember($this->getCacheKey(), now()->addMinutes($this->cacheTTL), function () {
+                return $this->renderPartial('@default');
+            });
+        }
     }
 
     /**
@@ -325,5 +336,13 @@ abstract class ComponentBase extends Extendable
     public function __toString()
     {
         return $this->alias;
+    }
+
+    /**
+     * Returns the component's cache key
+     */
+    public function getCacheKey()
+    {
+        return 'component::' . $this->dirName;
     }
 }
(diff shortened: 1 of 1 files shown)


modules/cms/classes/ComponentBase.php
```diff
@@ -38,6 +39,11 @@ abstract class ComponentBase extends Extendable
      */
     public $isHidden = false;
 
+    /**
+     * @var int Determines if component should cache default partial in onRender hook. Uses minutes.
```
LukeTowers: This should be in seconds, that's what Laravel 5.9 is moving to so might as well do it here.

modules/cms/classes/ComponentBase.php
```diff
@@ -131,6 +137,11 @@ public function onRun()
      */
     public function onRender()
     {
+        if ($this->cacheTTL) {
+            return Cache::remember($this->getCacheKey(), now()->addMinutes($this->cacheTTL), function () {
```
LukeTowers: cacheTTL should also support being a datetime value in which case it should just be passed straight to Cache::remember

modules/cms/classes/ComponentBase.php
```diff
@@ -326,4 +337,12 @@ public function __toString()
     {
         return $this->alias;
     }
+
+    /**
+     * Returns the component's cache key
+     */
+    public function getCacheKey()
+    {
+        return 'component::' . $this->dirName;
```
LukeTowers: This should also have the name and alias of the component in here to be as unique as possible

modules/cms/classes/ComponentBase.php
```diff
@@ -40,7 +40,7 @@ abstract class ComponentBase extends Extendable
     public $isHidden = false;
 
     /**
-     * @var int Determines if component should cache default partial in onRender hook. Uses minutes.
+     * @var int Determines if component should cache default partial in onRender hook. Supports seconds or datetime.
```
LukeTowers: Typehint should also be updated to `integer|DateTime` please

modules/cms/classes/ComponentBase.php
```diff
@@ -138,7 +138,7 @@ public function onRun()
     public function onRender()
     {
         if ($this->cacheTTL) {
-            return Cache::remember($this->getCacheKey(), now()->addMinutes($this->cacheTTL), function () {
+            return Cache::remember($this->getCacheKey(), $this->cacheTTL, function () {
```
LukeTowers: You'll need to do it like the following to support the pre and post Laravel 6 upgrade:

```php
$expires = is_integer($this->cacheTTL) ? now()->addSeconds($this->cacheTTL) : $this->cacheTTL;
```

LukeTowers: and then use $expires in Cache::remember

modules/cms/classes/ComponentBase.php
```diff
@@ -326,4 +337,12 @@ public function __toString()
     {
         return $this->alias;
     }
+
+    /**
+     * Returns the component's cache key
+     */
+    public function getCacheKey()
+    {
+        return 'component::' . $this->dirName . $this->alias . $this->name;
```
LukeTowers:
```suggestion
        return 'component::' . $this->dirName . '-' . $this->page->getFilePath() . '-' . $this->alias;
```

bennothommo:
@nathan-van-der-werf @prhost Any particular reason for the thumbs down on this PR?

prhost:
> @nathan-van-der-werf @prhost Any particular reason for the thumbs down on this PR?

If an extra dependency is being created in a place it should not have, this problem can easily be solved by placing it in the component itself, or even using events to inject this cache system, outside of the developers' own custom, including the actual one. October documentation, that onRender doesn't call parent, then no immediate component would take effect, and if you have to change all or most of the onRender and call parent, it's easier to implement this snippet by hand when you need it.

The idea is cool, but I suggest @Samuell1 checking if there are any events that can inject this and create a plugin.

Samuell1:
Yeah, i was talking with @LukeTowers few days back that most of the components are dynamic that means they cant cache or they have custom onRender method to handle things and that means caching in components base is useless.

Its not my idea its from issue https://github.com/octobercms/october/issues/2505 @prhost

Issue #4419, opened 2019-06-27
Add Thai translation


anurat:
Hi,

I would like to contribute by adding a new language, namely Thai language, to October backend.
So far I am able to translate some of the text in the following:

update modules/backend/models/Preference.php
	function getLocaleOptions()
		add 'th' => [Lang::get('system::lang.locale.th'), 'flag-th'],

create modules/backend/lang/th/lang.php
	update text
create modules/cms/lang/th/lang.php
	update text
create modules/system/lang/th/lang.php
	'locale' add 'th' => 'ไทย',
	update text
create modules/system/lang/th/client.php
	update text
create modules/system/lang/th/validation.php
	update text

However there is an issue that prevents the text to show.
![alert cancel_button_text](https://user-images.githubusercontent.com/2355534/60237425-fc022b00-98d4-11e9-9cf2-1da5c03affd3.png)

I found that it is related to modules/system/lang/th/client.php that may not get loaded.
I'm wondering what I should do to fix the issue.

Best Regards,

LukeTowers:
@anurat after updating client.php you have to run `php artisan october:util compile lang` to recompile the JS language files.

anurat:
Thank you LukeTowers, it works!

LukeTowers:
No problem @anurat! I'll leave this open until you've made a PR to add the language and it's been merged

anurat:
Hi @LukeTowers ,

I've translated most of the text in lang and client files but some of them I couldn't find where they are used so I do not understand or not sure how I should translate them.

Here are the text I couldn't translate. If you could point out where they are used or explain what these mean, I'd really appreciate it.

modules/system/lang/th/lang.php

```php
    'combiner' => [
        'not_found' => "The combiner file ':name' is not found.",
    ],
    'mail_templates' => [
        // I know where to find this but not sure what it does
        'disable_auto_inline_css' => 'Disable automatic inline CSS',
    ],
```

modules/cms/lang/th/lang.php

```php
    'editor' => [
        // why is it commit instead of save/update?
        'commit_confirm' => 'Are you sure you want to commit your changes to this file to the filesystem? This will overwrite the existing file on the filesystem',
        'committing' => 'Committing',
        'commit_success' => 'The :type has been committed to the filesystem',
        'reset_success' => 'The :type has been reset to the filesystem version',
    ],
```

modules/backend/lang/th/lang.php

```php
    'field' => [
[...]

PR #4472, opened 2019-07-25
Add Thai translation

anurat:
Add Thai translation to backend, cms, system modules.
Most text are translated to Thai. 
Some of the text are left without translating.


anurat:
The pull request is related to this issue.
https://github.com/octobercms/october/issues/4419

modules/backend/lang/th/lang.php
```diff
@@ -0,0 +1,628 @@
...
+return [
+    'auth' => [
+        'title' => 'ส่วนการจัดการ',
+        'invalid_login' => 'รายละเอียดที่คุณเพิ่มเข้ามาไม่ตรงกับข้อมูลของเรา กรุณาตรวจสอบและลองอีกครั้ง',
+    ],
+    'field' => [
+        'invalid_type' => 'ประเภทของฟิลด์ไม่ถูกต้อง :type.',
+        'options_method_invalid_model' => "The attribute ':field' does not resolve to a valid model. Try specifying the options method for model class :model explicitly.",
+        'options_method_not_exists' => "The model class :model must define a method :method() returning options for the ':field' form field.",
+        'colors_method_not_exists' => "The model class :model must define a method :method() returning html color HEX codes for the ':field' form field.",
```
LukeTowers: Untranslated strings should be left out of the language files so that future updates to the default language (English) will propagate properly @anurat

modules/system/lang/th/validation.php
```diff
@@ -0,0 +1,121 @@
+<?php
+
+return [
```
LukeTowers: Since this file is entirely untranslated it should be removed.

anurat: Even though it is untranslated, if we left out the file, will there be any problem if the find is not found in th folder?

modules/backend/lang/th/lang.php
anurat: I see. Is there anyway to keep the untranslated fields so that I am able to update them later on?

LukeTowers: You could in theory comment them out with inline comments but I'm not really a fan of that. I would like it if there was an official plugin to review the language files and solicit missing translations from the community.

modules/system/lang/th/validation.php
LukeTowers: There shouldn't be, everything falls back to the default language (english), but feel free to test it.

anurat:
Thanks a lot!

w20k:
@anurat sorry for a delay with testing. 
Will test those soon, and ready to be merged!

w20k:
LGTM @LukeTowers, despite one thing:

@anurat I guess, you have to add translation inside of the `en` version as well.
https://github.com/octobercms/october/blob/bf809ba01d4b299a36b353c575d2c432ce6fc39a/modules/system/lang/en/lang.php#L8

Before changing to Thai:
Screenshot 2019-07-26 at 14 07 42

After changing to Thai:
Screenshot 2019-07-26 at 14 08 23


Screenshots:
Screenshot 2019-07-26 at 14 05 21
Screenshot 2019-07-26 at 14 06 07
Screenshot 2019-07-26 at 14 06 16
Screenshot 2019-07-26 at 14 06 23
Screenshot 2019-07-26 at 14 06 32
Screenshot 2019-07-26 at 14 06 37

anurat:
Thank you @w20k 
Will do shortly.

anurat:
By the way, should I go ahead and add Thai locale to all other languages as well?

w20k:
@anurat I've misled you a little bit. 
It's better to remove that block from your translations and just add support to `en\lang.php`. That's what ppl do mainly.

anurat:
Thank you @w20k @LukeTowers 
That was fun ^^

Issue #4454, opened 2019-07-14
Add full support for Click, Keyboard and Touch Navigation


ghost:
## Intro

Currently October CMS has good `mouse click` support with the frontend and backend.

However, October CMS has intermediate functionality with `touch` support.

It gets worse, October CMS has almost no `Keyboard Navigation` support.

W3c one of the main internet standards organizations has very specific rules for everything when it comes to this area.

## Example with Form Tabs

I recently did a very long [PR](https://github.com/octobercms/october/pull/4401) that allowed October CMS to use `Keyboard Navigation` support for all the Tabs in the frontend and backend.

Here we can see W3c rules for tabs:

![gay](https://user-images.githubusercontent.com/17784082/61186588-2be45900-a65f-11e9-855e-78257a4f2c8b.png)

About shows the Keyboard Navigation rules just for Tab fields.

## Rest of the Form and List Fields

I soon realized after coding this functionality that the rest of all the form and list fields don't have any support at all.

An example would be `radio buttons`, `checkboxes` etc.

I open this Github Issue with a `Priority: Low` label to add enhancement support for all form and list fields to give full support for Click, Keyboard and Touch Navigation.

List of October CMS current supported fields can be found here: https://octobercms.com/docs/ui/form

## Rest of October CMS

I suggest starting with the form and list fields first and then moving onto the main navigation areas such as the top menu and side panels etc.

### Help from everyone

You can help by two main ways, either creating a PR that adds the code for an item or you can update a lang file. I have started the ball rolling by adding the following section to the lang files:

```php
aria-label="<?= e(trans('backend::lang.aria-label.tabs')) ?>"
aria-label="<?= e(trans('backend::lang.aria-label.footer')) ?>"
```

Then add to the `lang` files:

```php
<?php
return [
    'aria-label' => [
        'tabs'   => 'Tabs',
        'footer' => 'Footer Section',
    ],
```

Previously I used Google Translator, which gives poor results, if you can speak a non-english language, you can help by updating the lang files.

### Doc's code examples

Also need to update the October documentation code examples.

For example, these are now wrong: https://octobercms.com/docs/ui/tab

Missing WAI-ARIA in all the HTML code examples.

ayumihamsaki:
## Radio buttons

Just playing around coding these, I leave my notes here for future ref:

![gay](https://user-images.githubusercontent.com/17784082/61187033-774d3600-a664-11e9-8e59-dabd7e0dfbb7.png)

[...]

PR #4478, opened 2019-07-26
Add Aria breadcrumbs

ayumihamsaki:
Related to here: https://github.com/octobercms/october/issues/4454


ayumihamsaki:
Finished.

modules/system/assets/ui/docs/breadcrumb.md
```diff
@@ -4,10 +4,12 @@ Display a breadcrumb on the page.
 
     <div style="padding: 20px">
         <div class="control-breadcrumb">
-            <ul>
-                <li><a href="#">Dash Board</a></li>
-                <li><a href="#">Blog Posts</a></li>
-                <li>Edit Post</li>
-            </ul>
+            <nav aria-label="Breadcrumb">
```
w20k: @ayumihamsaki why do you remove `control-breadcrumb` class?

ayumihamsaki: @w20k I didn't it on line 6 ?

w20k: Sorry, my bad, morning sleepy eyes :)

ayumihamsaki: haha, I thought I completely missed something!

w20k:
@ayumihamsaki is the current link not selected at all (focused?)?

Screenshot 2019-08-07 at 16 11 18

ayumihamsaki:
@w20k 

Over the years I have done a lot of research into that question!

1. According to the W3c spec on breadcrumbs you should NOT show a link to the current web page.

2. However, Google and Bing both have examples showing a link for the current web page. A few years ago I contacted both of them and told them in great detail their examples are wrong according to the spec.

3. Schema markup was created a few years later and they had mixed examples. At first they didn't have the current web page marked up and then the examples at present do have the current web page marked up.

4. Bad SEO practises going back 10 years ago it was a hack to link the current web page to increase the current web page Page Rank, something called: PageRank sculpting.

5. Going full circle and into 2019, WAI-ARIA spec with regards to breadcrumbs says to NOT markup the links using `role="link"` and just adding `aria-current="page"` to the current web page.

### So what does that all mean?

The reason I given a long answer is because the screenshot you showed is correct and not correct lol. For years I would have said yes don't add a link to the current web page according to the HTML spec set out by W3c and HTML Living Standard org.

But everything conflicts their standards and says to add a link to the current web page.

However, you have to ask a simple question and one I asked for many years!

If someone is on the current web page - why would they need to navigate to the same web page when they are on that web page? It doesn't make any sense.

So with all that info, my answer would be: No - I don't think you need to add the current link focused and the WAI-ARIA spec also says no. But all the other spec's conflict this.

ayumihamsaki:
Closed this and moved to here: https://github.com/octobercms/october/pull/4531

PR #4500, opened 2019-07-31
Add warning when debug mode is enabled in status widget

Samuell1:
This will add a warning in status widget to warn users that debug mode is enabled.

### DIFF
--- a/modules/backend/lang/en/lang.php
+++ b/modules/backend/lang/en/lang.php
@@ -120,6 +120,7 @@ return [
             'event_log' => 'Event log',
             'request_log' => 'Request log',
             'app_birthday' => 'Online since',
+            'debug_warning' => 'Debug mode is enabled! It\'s recommended to disable it on production.'
         ],
         'welcome' => [
             'widget_title_default' => 'Welcome',
--- a/modules/backend/lang/en/lang.php
+++ b/modules/backend/lang/en/lang.php
@@ -120,7 +120,7 @@ return [
             'event_log' => 'Event log',
             'request_log' => 'Request log',
             'app_birthday' => 'Online since',
-            'debug_warning' => 'Debug mode is enabled! It\'s recommended to disable it on production.'
+            'debug_warning' => 'Debug mode is enabled. This is not recommended for production installations.'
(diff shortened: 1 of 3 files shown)


modules/backend/lang/en/lang.php
```diff
@@ -120,6 +120,7 @@
             'event_log' => 'Event log',
             'request_log' => 'Request log',
             'app_birthday' => 'Online since',
+            'debug_warning' => 'Debug mode is enabled! It\'s recommended to disable it on production.'
```
bennothommo: @Samuell1 This might read better:
```suggestion
            'debug_warning' => 'Debug mode is enabled. This is not recommended for production installations.'
```

LukeTowers:
Screenshot?

Samuell1:
@LukeTowers Its here as first item in list
obrázok

bennothommo:
@LukeTowers Here it is on my local copy:

image

It's tested and confirmed working - not sure if we should include it in 458 or wait until 459 though.

bennothommo:
Blocked until development begins on build 459.

LukeTowers:
@Samuell1 @bennothommo what is the "No warnings to display" section used for? I feel like this is a warning that belongs in there but I'm not sure how it works.

Samuell1:
@LukeTowers thats system errors like missings dependencies or extensions etc.

bennothommo:
@LukeTowers it's for non-writable storage directories and missing PHP dependencies, I believe.

LukeTowers:
@Samuell1 @bennothommo what do you think about putting it in there anyways? For the user, I don't think they care about a distinction between a "system" warning and a "you shouldn't be doing this" warning, they're both warnings about the current configuration of the site.

Samuell1:
obrázok

@LukeTowers changed :)

bennothommo:
@LukeTowers sorry for not responding sooner - been out and about this weekend. I agree to have it in there, no real reason to separate them.

LukeTowers:
LGTM, just waiting on @daftspunk to tag 458

PR #4506, opened 2019-08-01
Content tabs do not work at all in the backend - this makes them now work

ayumihamsaki:
This is a complete re-write of the Content Tabs code. When I was testing something else, I discovered that these tabs do not work at all in the backend.

This pr fixes that issue and makes the content tabs work now in the backend forms.

---

One question though, there seems to be a strange line of code, found here: x

### DIFF
--- a/modules/system/assets/ui/less/tab.less
+++ b/modules/system/assets/ui/less/tab.less
@@ -398,16 +398,58 @@
             }
         }
     }
+}
 
-    //
-    // Content tabs
-    //
+//
+// Content tabs
+//
 
-    &.content-tabs {
-
-        > ul.nav-tabs {
+.control-tabs.content-tabs {
+    >ul.nav-tabs {
+        position: relative;
+        &:before {
+            content: ' ';
+            position: absolute;
+            bottom: 0;
+            left: 0;
+            width: 100%;
+            height: 1px;
+            background: #e3e5e7;
+        }
+        li {
+            margin-right: 0;
+            border-top: 1px solid #e3e5e7;
+            border-right: 1px solid #e3e5e7;
+            a {
+                padding: 10px;
+                margin: 5px;
+                border: none;
+                text-transform: uppercase;
+                background: transparent;
+                font-weight: 500;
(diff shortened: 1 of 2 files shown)


bennothommo:
@ayumihamsaki Please provide a screenshot of the affected functionality before and after your changes.

ayumihamsaki:
@bennothommo 

Before:

b4

After:

after

ayumihamsaki:
@w20k @bennothommo I need this tested and merged into October, because I need to calculate the `Content Tabs` height dimensions for another PR I'm working on.

daftspunk:
Content tabs do work and an example can be found on the Mail Templates page.

I have no idea what this rewrite is supposed to fix but it breaks the existing implementations of content tabs.

Before (working):
image

After (broken):
image

Samuell1:
I think he means content tabs in fields?

ayumihamasaki2019:
@Samuell1 👍 

@daftspunk You're testing the wrong thing! In your screenshots you are showing what you have called in the source code the "**outside tabs**" area. You need to test the content tabs as a **_secondary tab area_**.

Simple to do:

1. Create a YAML file with a secondary tabs.

2. Highlight it in dev-tools and change the name from `class="secondary-tabs"` to` class="content-tabs"` and you will see it is not working due to the CSS code in October using a single level `div > div` setup.

3. p.s. I think you're expecting the content tabs to look exactly the same, but they have been re-sized to allow the focus ring to select the whole tab area.

62254430-01c9bf80-b3f1-11e9-8650-d16f2183715b

Above example gif is showing a **secondary-content-tab** with **focus ring**. That's what this pr is fixing.

Hope that helps you understand it a bit better.

LukeTowers:
@ayumihamasaki2019 if you can create another PR with that explanation and the original "outside tabs" feature working correctly too then we can take another look at it.

ayumihamasaki2019:
@LukeTowers Need to wait for v460 to get rolled out and then re-test everything (due to way too many merges and reverts going on right now).

Issue #4507, opened 2019-08-01
Builder plugin has a display error in firefox


ayumihamsaki:
When running tests on some other pr's I came across this small error, this github issue just shows the error for future ref and will connect a pr for it.

The issue seems to be in firefox only - chrome works fine.

### Problem

![er1](https://user-images.githubusercontent.com/17784082/62283887-711fcd80-b44a-11e9-9b57-b4dbf3fd13a6.png)

### Solution

![er2](https://user-images.githubusercontent.com/17784082/62283905-79780880-b44a-11e9-8a47-8e020f866d75.png)

### Error

There is a single line of CSS code missing in the files - pr just adds the missing code back into october. May result in error with other plugins as well.

bennothommo:
Fixed in 8219f19e8d902670cddcad3dfc6ef28516418cc0.

PR #4508, opened 2019-08-01
Fix the top of the menu in builder and rich editors

ayumihamsaki:
This fixes the github issue found here: https://github.com/octobercms/october/issues/4507

### DIFF
--- a/modules/backend/assets/less/layout/fancylayout.less
+++ b/modules/backend/assets/less/layout/fancylayout.less
@@ -1,4 +1,3 @@
-@background_color_1: #bf5d07;
 
 body.breadcrumb-fancy .control-breadcrumb,
 .control-breadcrumb.breadcrumb-fancy {
@@ -72,25 +71,6 @@ body.breadcrumb-fancy .control-breadcrumb,
             }
         }
     }
-
-    .fancy-layout {
-        .control-tabs.master-tabs {
-            >div {
-                >div.tabs-container {
-                    background-color: @background_color_1;
-                }
-            }
-        }
-    }
-
-    .fancy-layout.control-tabs.master-tabs {
-        >div {
-            >div.tabs-container {
-                background-color: @background_color_1;
-            }
-        }
-    }	
-
     .control-tabs, &.control-tabs {
         &.master-tabs {
             overflow: hidden;
@@ -106,7 +86,7 @@ body.breadcrumb-fancy .control-breadcrumb,
             &.scroll-after:after { color: @color-fancy-master-tabs-active-text; }
 
             > div > div.tabs-container {
-                background: @color-fancy-master-tabs-bg;
+                background: #bf5d07;
                 padding-left: 20px;
                 padding-right: 20px;
 
--- a/modules/backend/assets/less/layout/fancylayout.less
+++ b/modules/backend/assets/less/layout/fancylayout.less
@@ -1,4 +1,3 @@
-
 body.breadcrumb-fancy .control-breadcrumb,
 .control-breadcrumb.breadcrumb-fancy {
     margin-bottom: 0;
--- a/modules/backend/assets/less/layout/fancylayout.less
+++ b/modules/backend/assets/less/layout/fancylayout.less
@@ -1,3 +1,5 @@
+@background_color_1: #bf5d07;
+
 body.breadcrumb-fancy .control-breadcrumb,
 .control-breadcrumb.breadcrumb-fancy {
     margin-bottom: 0;
@@ -70,6 +72,24 @@ body.breadcrumb-fancy .control-breadcrumb,
             }
         }
     }
+
+    .fancy-layout {
+        .control-tabs.master-tabs {
+            >div {
+                >div.tabs-container {
+                    background-color: @background_color_1;
+                }
+            }
+        }
+    }
+
+    .fancy-layout.control-tabs.master-tabs {
+        >div {
+            >div.tabs-container {
+                background-color: @background_color_1;
+            }
+        }
+    }	
 
     .control-tabs, &.control-tabs {
         &.master-tabs {
--- a/modules/backend/assets/less/layout/fancylayout.less
+++ b/modules/backend/assets/less/layout/fancylayout.less
@@ -85,7 +85,7 @@ body.breadcrumb-fancy .control-breadcrumb,
             &.scroll-after:after { color: @color-fancy-master-tabs-active-text; }
 
             > div > div.tabs-container {
-                background: #bf5d07;
+                background: @color-fancy-tabs-bg;
                 padding-left: 20px;
                 padding-right: 20px;
 
--- a/modules/backend/assets/css/october.css
+++ b/modules/backend/assets/css/october.css
@@ -885,6 +885,7 @@ body.breadcrumb-fancy .control-breadcrumb li:last-child:before,
 .fancy-layout .tab-collapse-icon:hover {text-decoration:none;opacity:1;filter:alpha(opacity=100)}
 .fancy-layout .tab-collapse-icon.primary {color:#475354;bottom:-25px;z-index:100;-webkit-transform:scale(1,-1);-moz-transform:scale(1,-1);-ms-transform:scale(1,-1);-o-transform:scale(1,-1);transform:scale(1,-1)}
 .fancy-layout .tab-collapse-icon.primary i {position:relative;display:block}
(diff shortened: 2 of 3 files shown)


bennothommo:
@ayumihamsaki It looks as though you edited `october.css` manually and then added the code to `fancylayout.less` after the fact. When running `php artisan october:util compile less`, the resulting CSS lines come out as:

```
.fancy-layout .fancy-layout .control-tabs.master-tabs >div >div.tabs-container {background-color:#bf5d07}
.fancy-layout .fancy-layout.control-tabs.master-tabs >div >div.tabs-container {background-color:#bf5d07}
```

Which is incorrect, as it's effectively looking for an element with class `fancy-layout` within an element with class `fancy-layout`.

Please ensure that you test your changes, and make any CSS changes directly in the LESS files and compile them using the proper CLI commands - otherwise, there will likely be issues when other contributors provide CSS changes.

modules/backend/assets/less/layout/fancylayout.less
```diff
@@ -1,3 +1,5 @@
+@background_color_1: #bf5d07;
```
bennothommo: @ayumihamsaki Please remove this colour, as we should be using the variables from `modules/backend/assets/less/core/variables.less`.

modules/backend/assets/less/layout/fancylayout.less
```diff
@@ -71,6 +73,24 @@ body.breadcrumb-fancy .control-breadcrumb,
         }
     }
 
+    .fancy-layout {
```
bennothommo: @ayumihamsaki Please remove this block and the one below, as this is not the correct location for these changes. You should instead insert your changes in line 109 below, and maybe use something like the `darken()` LESS function to apply a slight change to the colour but still use the original LESS variable.

ayumihamsaki: @bennothommo I don't think you're right with moving it to line 109 because line 94 has a `comma` not a `dot`, e.g.

```css
.control-tabs, &.control-tabs {
```

not

```css
.control-tabs. &.control-tabs {
```

bennothommo: @ayumihamsaki It is correct. LESS will render that two separate lines:

```
.fancy-layout .control-tabs
.fancy-layout.control-tabs
```

ayumihamsaki:
@bennothommo done.

bennothommo:
@ayumihamsaki Thanks, but I'm trying to avoid having a direct hex value in the LESS file, as it would make it easier for us to tweak the colour scheme in the future by simply editing the `variables.less` file.

Please change it to use the colour manipulation methods in LESS and provide it with the original variable.

ayumihamsaki:
@bennothommo Sorry my mis-understanding - have created a new variable and added it into the color scheme (as the hex value wasn't in that file for some strange reason?)

ayumihamsaki:
## Test results

### Google Chrome

Google Chrome has a over-the-top correction algo that auto fixes many coding issues. This is why you do not see this issue in Chrome.

### Firefox

Firefox doesn't doesn't have such an over-the-top correction algo and 99% of the time actually shows the error. The error seems to be the following logic order:

1. October.css has a `background:#d35400` on a code line.

2. October.css has a `background-color:#d35400` on another code line.

3. Builder plugin is trying to overwrite both of those two code lines with the correct code of: `background: #bf5d07;`

This is the reason why Firefox shows the error and not Chrome.

Correcting this error gives the following results:

### Before

aq1

### After

aw2

### Lastly

This fix will make all browsers display the exact same result, because instead of trying to inject the fix from Builder plugin, it comes direct from the `october.css` now.

bennothommo:
@ayumihamsaki I've tested a clean install of October on both Chrome and Firefox, and I'm not getting the issue you are reporting.

Chrome
Screenshot from 2019-08-02 23-53-16

Firefox
Screenshot from 2019-08-02 23-54-49

My best guess is that perhaps the sidebar orange strip that is in-built into the Builder plugin is not respecting the colour preferences that can be set in **Settings > Customise Back-end**. Given that on a fresh install, the sidebar strip and the strip behind the tabs use the same colour and height, I'm led to believe that it was intentional for the Builder plugin to have that top strip be continuous across the sidebar and the tab area.

ayumihamsaki:
@bennothommo not sure how you tested it, but I can show you proof as bright as day!

aaaaaaaaaaaa

100% Builder is correcting the color scheme here.

bennothommo:
@ayumihamsaki I'm saying that the Builder plugin is using the default tab background colour for the orange strip above the sidebar menu, so on a clean install, it is making the orange strip seem continuous. I'm suggesting that perhaps on your install, you have made a change in the **Customise back-end** screen which is applying a *different* colour to the tab background.

I've been able to replicate this by changing the secondary colour on that screen to blue - see below:

image

ayumihamsaki:
Mine is a fresh install on a local host it's 100% only there for testing.

### Chrome does this:

```css
.fancy-layout .control-tabs.master-tabs>div>div.tabs-container, .fancy-layout.control-tabs.master-tabs>div>div.tabs-container {
    background-color: #bf5d07;
}

.fancy-layout .control-tabs.master-tabs >div >div.tabs-container, .fancy-layout.control-tabs.master-tabs >div >div.tabs-container {
    /* background: #d35400; */
    padding-left: 20px;
    padding-right: 20px;
}
```

### Firefox does this:

```css
.fancy-layout .control-tabs.master-tabs>div>div.tabs-container, .fancy-layout.control-tabs.master-tabs>div>div.tabs-container {
    /* background-color: #bf5d07; */
}

.fancy-layout .control-tabs.master-tabs >div >div.tabs-container, .fancy-layout.control-tabs.master-tabs >div >div.tabs-container {
    background: #d35400;
    padding-left: 20px;
    padding-right: 20px;
}
```

In firefox `background` will over write `background-color`, I know this from previous research.

ayumihamsaki:
> I'm suggesting that perhaps on your install, you have made a change in the Customise back-end screen which is applying a different colour to the tab background.

No!

ssssssssss

Please read the code.

bennothommo:
@ayumihamsaki Please just humour me here, and try this:

- Go to **Settings** in the Backend.
- Click on **Customise back-end**
- Click on the **Colors** tab and then click the **Reset to Default** button on the bottom right.

Then go back to the Builder plugin and see if the issue still exists.

The reason I am suggesting this is because the block of code you have shown above with the `background-color` definition is provided by the `custom.less` file here: https://github.com/octobercms/october/blob/37eb5b0c3c535f7c0473fd9d0c02e4f405a154a6/modules/backend/models/brandsetting/custom.less#L124

Note that in your screenshot, the CSS is inline - it's not provided by a CSS file. That's because the `BrandSetting` model renders the CSS inline in the `<head>` tag whenever any changes are made to the Brand settings, which is part of the **Customise back-end** screen.

ayumihamsaki:
@bennothommo 

Sure no problem, I repeated the following steps:

- Go to Settings in the Backend.
- Click on Customise back-end
- Click on the Colors tab and then click the Reset to Default button on the bottom right.

Then go back to the Builder plugin and see if the issue still exists.

> Same issue.

---

You're right about the inline:

gggggggg

I leave it for you to decide if you want it continuous or not? Personally I like it (not continuous).

bennothommo:
I agree, but it's highly subjective. @LukeTowers @w20k - do either of you have any opinions on the above?

w20k:
Probably I'm blind as well but can't find anything related to the `tabs` inside `builder.css`.
https://github.com/rainlab/builder-plugin/blob/master/assets/css/builder.css

For me, FF and chrome (clean install via docker) - same colors by default.

ayumihamsaki:
@w20k I thought it was that css file at file, but it's being inlined via javascript I think.

bennothommo:
@ayumihamsaki I've fixed this in 8219f19e8d902670cddcad3dfc6ef28516418cc0. I've opted to have the strip be continuous as the default look (without branding) is for it to be a continuous orange script - it should now reflect whichever colour is chosen in the **Customise back-end** area.

@w20k The code was not in the Builder plugin - it is the custom CSS that is inserted into the `<head>` tag when someone customises the Backend colours.

ayumihamsaki:
@bennothommo 

That looks pretty good in your commit, having `background` instead of `background-color` will pass in firefox and using `@custom-dark-secondary`, I was afraid using a new color and then it might make that new color 20% darker. I think your commit is the best solution.

Thanks for your help.

bennothommo:
@ayumihamsaki Not a problem, thank you for the help as well.

Issue #4505, opened 2019-07-31
Setting page and sidebar


ghost:
I'm opening this github issue, to try and learn and better understand what the heck is going on with this: `sidenav-tree-root` class.

I need to find out what it's main purpose is in October?

It seems to be some stupid thing! So I am banging my head on the wall asking myself why is it there and what it's main purpose is?

I want to remove it totally out of October!

Here is what it's doing in the setting page:

![seee](https://user-images.githubusercontent.com/17784082/62225910-37e74f00-b3b1-11e9-8f16-d06bfc3ab91e.png)

If there is no good reason for having this class, I plan to do the following things:

1. Totally remove `sidenav-tree-root` class.

2. Fix the Settings sidebar `sidenav-tree` class.

By removing all the repeated and conflicting media queries:

End up having just this clean setup:

```css
@media only screen and (min-width: 400px) {
    .sidenav-tree {
        width: 300px;
    }
}
@media only screen and (max-width: 400px) {
    .sidenav-tree {
        width: 80vw;
    }
}
```

There is so much junk in this issue in the CSS and the LESS.

But I need to know what's the purpose of the `sidenav-tree-root` class before I remove it totally. For now I don't understand why it's there?

How to reproduce this stupid class:

- Open the settings page but make sure your top naivgation settings are set to: `Inline`

The other settings don't seem to try and inject this stupid class into October.

p.s. It's probably causing this issue as well: https://github.com/octobercms/october/issues/4052#issuecomment-453822232

----

This bit of code is stupid, I don't understand why someone has tried to merge the two classes into one: `sidenav-tree-root` class and this `sidenav-tree` class.

Here is two reasons why I think it's dumb:

1. 

![seee1](https://user-images.githubusercontent.com/17784082/62227263-75e57280-b3b3-11e9-8797-55c880ebfe5d.png)

2.

![seee2](https://user-images.githubusercontent.com/17784082/62227357-acbb8880-b3b3-11e9-929e-3d85da174743.png)

Here's a quick example of the sidebar all over the place:

![ezgif-2-d504a9fcdc92](https://user-images.githubusercontent.com/17784082/61581531-63e41400-ab17-11e9-8a02-57b372faba4c.gif)

Note: The above gif has been cleaned up by me and the code in October right now produces even worse results - just try it and see for yourself.

Lastly, here's some of the media query mess which is no good:

```css
@media (min-width: 1200px) {
    .sidenav-tree-root .sidenav-tree {
        width: 900px;
    }
[...]

Issue #3971, opened 2018-12-08
Would it be a good idea to add a Javascript dragbar to the backend in various places?


ghost:
I got one final idea for updating the code section in the October backend. I also would like to add it to the builder plugin.

I will show you my idea for the code editor section to add a Javascript dragbar, so users can adjust the size of the column to read the files more easily, or hide them a bit better.

## The Problem

Currently October allows you to either display it in full or hide it. There is no feature to allow you to fine tune the column size.

### Code Editor Section

The column displays the list of files, I would like to make the right side adjustable:

![untitled1](https://user-images.githubusercontent.com/17784082/49687508-c4c85a00-fafb-11e8-81dc-c237652dbcac.png)

and be able to do something like this:

![dragbar](https://user-images.githubusercontent.com/17784082/49687585-d1997d80-fafc-11e8-8fe1-361565e377f8.gif)

### Settings Section

![untitled13333333333](https://user-images.githubusercontent.com/17784082/49687748-6c935700-faff-11e8-911a-fa9b8e2ebbb2.png)

## Builder Plugin

I would like to add it also to here:

![untitled13333333333](https://user-images.githubusercontent.com/17784082/49687601-26d58f00-fafd-11e8-97d7-bcb3e4ae7a5a.png)

I would like to see what others think, to see if this is a good idea or not?

I personally think it would make using the backend a bit easier.

SebastiaanKloos:
That would be really nice. The adjustments should be saved in the session

ayumihamsaki:
Code Example: http://htmlpreview.github.com/?https://github.com/ayumihamsaki/october-drag-bar/blob/master/drag-bar.html

## Code Usage Options

**Global Scope:**
```javascript
$(selector).resizable(options);
```

### Options
The options parameter can be a map of any of these properties (default values are shown):

```javascript
var opt = {
    // optional selector for handle that starts dragging
    handleSelector: null,
    // resize the width
    resizeWidth: true,
    // resize the height
    resizeHeight: true,
    // the side that the width resizing is relative to
    resizeWidthFrom: 'right',
    // the side that the height resizing is relative to
    resizeHeightFrom: 'bottom',            
    // hook into start drag operation (event,$el,opt passed - return false to abort drag)    
    onDragStart: null,
    // hook into stop drag operation (event,$el,opt passed)
    onDragEnd: null,
    // hook into each drag operation (event,$el,opt passed)
    onDrag: null
    // disable touch-action on the $handle
[...]

PR #4512, opened 2019-08-01
Fix settings page sidebar

ayumihamsaki:
What have I done here?

Well I basically went hardcore and deleted a ton of complete crap out of the css and less files and added the correct code.

The issue, there was a bunch of conflicting media queries, classes that created a huge gap on the right hand side of the backend. You would either see it in a flash when loading or see the top navigation menu load on the left hand side of the screen and then get over-written to cover the rest of the screen. Someone tried to merge the sidebar and the top navigation code together.

Basically it was a total mess and totally unless code. So I deleted it. Then added a few lines of clean code and the result is this:

ezgif-2-087b529c55fc

Fixes this github issue: https://github.com/octobercms/october/issues/4505

Testing, normally doing something like this would break tihngs, but I'm so sure that the code I delete was total garbage, I'm sure the testers will not find errors or anything breaking - that's how bad the code was.


Samuell1:
It totally changes how it looks and worked before

ayumihamsaki:
I forgot to mention, if people want a bigger sized sidebar in the settings page, then adding a drag bar to extend this pr is my intention, see here: https://github.com/octobercms/october/issues/3971

Also on top of the dragbar, I would want to remove the display table inline rows and switch to using flexbox wrap to give a smooth feeling when a user adjusts the sidebar using a dragbar.

Samuell1:
But still it should be same size with or without dragbar as it was before, you still have empty right side of page why not to use that and in your gif there is logo getting smaller but doesnt hide on mobile resolutions.

ayumihamsaki:
As I said there are many issues with the current setup and it's all over the place! You suggest it should be the same size as before, but in truth it's not the same size because when you first open the settings page you get three columns then select something and get a single column. I don't know any professional website that has moving menu's all over the place, when a user selects something!

Here I am creating a more stable thing, where a user opens the settings page for the first time and gets one size and then opens the next thing in that section and gets the same sized sidebar.

Adding a dragbar to this pr the user can get a large sidebar when opening the settings page and clicking the next thing and get the same large sidebar.

Lastly, sorry but no! I am not going to add a dragbar to a moving all over the place sidebar. Having two things moving against each other would create an even worse mess!

Having a stable sidebar and adding a dragbar to a stable object allows a stable outcome!

It's a shame your gif is moving so fast and tries to cover up all the bugs, that I summarized above.

Samuell1:
Yes its bad but dont forget this is administration sidebar not a website menu, if there will be many items it will be a lot longer and it will take a lot of time to get where you want.

I am mostly showing with that gif that there are more items on bigger screen showed at once to faster navigate at first open and showing that on mobile there isnt any extremly tiny logo on right side.

ayumihamsaki:
I'm my idea is baised on two things, the first is your wonderful suggestion of adding a session storage to a dragbar and the second is a guess that the reason the rest of the settings sections have a sidebar set at 300px is not a width, but more of a min-width. I'm guessing it's setup that way to work on all web pages in the settings page using the min width of the sidebar.

So my idea is basically to have a screensize above 400px uses a min-width of 300px and then a dragbar can be used to increase the sidebar width - according to the user's spec. and then finally your idea of a session storage to auto save the sidebar width for future setting page loads. Then lastly have flexbox wrap to display the sidebar columns smoothly e.g. 1 column, 2 columns, 3 columns or even more for uses with 4k and upwards computer screens etc.

I understand what you are saying and it sounds like you want a 3 column layout for your personal websites. I don't mind putting this pr on hold until the dragbar pr is completed and then merging the two at the same time - thus giving you the increased sidebar width.

ayumihamsaki:
My idea is baised on two things, the first is your wonderful suggestion of adding a session storage to the dragbar and the second is a guess that the reason the rest of the settings sections have a sidebar set at 300px is not a width, but more of a min-width. I'm guessing it's setup that way to work on all web pages in the settings page using the min width of the sidebar.

So my idea is basically to have a screensize above 400px using a sidebar with a min-width of 300px and then a dragbar which can be used to increase the sidebar width - according to the user's spec. Then your idea of a session storage to auto save the sidebar width for future setting page loads. Then lastly have flexbox wrap to display the sidebar columns smoothly e.g. 1 column, 2 columns, 3 columns when the sidebar is increased manually in size as per the user spec.

This way it hopefully gives the user more control and a better stable sidebar.

I understand what you are saying and it sounds like you want a 3 column layout for your personal websites. I don't mind putting this pr on hold until the dragbar pr is completed and then merging the two at the same time - thus giving you the increased sidebar width.

ayumihamsaki:
@Samuell1 initial coding of the second part of this pr idea:

ezgif-2-3c1e12571716

daftspunk:
Most of the faults here are actually part of the design. Ghost/Ayumi: Please message me on Slack (spunky) and I can explain why this is not accepted.

PR #4513, opened 2019-08-03
Filterfields is now called before save and update of forms

### LINKED ISSUE #4144: Hidden Fields Not Saved

mohsin:
The filterfields method is called before the save or update of the form field's model is made so that any changes made by the user to the field using filterFields method is also reflected in the saved model data. This PR closes octobercms/october#4144.

### DIFF
--- a/modules/backend/behaviors/FormController.php
+++ b/modules/backend/behaviors/FormController.php
@@ -247,6 +247,10 @@ class FormController extends ControllerBehavior
         $this->controller->formBeforeSave($model);
         $this->controller->formBeforeCreate($model);
 
+        if (method_exists($model, 'filterFields')) {
+            $model->filterFields((object) $this->formWidget->getFields(), $this->context);
+        }
+
         $modelsToSave = $this->prepareModelsToSave($model, $this->formWidget->getSaveData());
         Db::transaction(function () use ($modelsToSave) {
             foreach ($modelsToSave as $modelToSave) {
@@ -313,6 +317,10 @@ class FormController extends ControllerBehavior
 
         $this->controller->formBeforeSave($model);
         $this->controller->formBeforeUpdate($model);
+
+        if (method_exists($model, 'filterFields')) {
+            $model->filterFields((object) $this->formWidget->getFields(), $this->context);
+        }
 
         $modelsToSave = $this->prepareModelsToSave($model, $this->formWidget->getSaveData());
         Db::transaction(function () use ($modelsToSave) {


bennothommo:
@SaifurRahmanMohsin whilst this change covers a model with the `filterFields` method, it does not cover models using the `model.form.filterFields` events. Would you be able to include that as well, perhaps in the way that the [Form Widget](https://github.com/octobercms/october/blob/develop/modules/backend/widgets/Form.php#L1204) handles it.

SaifurRahmanMohsin:
@bennothommo I just realised that there's more to this. The filterFields doesn't seem to change anything internally. When I called `$this->formWidget->getFields()` again it still shows the old data. This is problematic as the internal fields in the form need to be updated when filterFields gets called in order for it to update properly. I am looking into it to see if there's something else that can be done. I think the issue is that filterFields takes $this->formWidget->getFields as a value than a reference. If it were a reference then changes made in it would be reflected and would solve this particular issue.

bennothommo:
@SaifurRahmanMohsin That's rather surprising - I suppose that means that the intention of the event was to add or remove fields, not to modify them.

We could change it to a reference, but as you said, it would change the signature of the event. Given it is in widespread use at this point, I don't think that's safe. We could also change the `$allFields` property to a class of some sort, which would make it pass by reference in the events, but that would be a lot of work and again would result in a different behaviour for the event.

I'll have a think about this.

SaifurRahmanMohsin:
@bennothommo Any update on this? Have you thought of a good solution for this yet?

SaifurRahmanMohsin:
@bennothommo Hi, is there any update on this? If you offer the alternative solution I could make an additional PR to fix this. Or at-least we could change the API with the OctoberCMS Laravel 6.0 release and call this a breaking change so that people make the necessary changes. There has to be some reason why the method was designed that way.

bennothommo:
@SaifurRahmanMohsin you might need to refresh my memory here - what scenario is requiring the use of `filterFields`? I've been through the linked issues and PRs, and it appears that it's got to do with the saving of hidden fields, but I think @LukeTowers had suggested the use of the `beforeSave` event for handling those. Is there a specific use case that *has* to be done with `filterFields`?

PR #4514, opened 2019-08-03
Update __isset function to comply with the same checks as __get 

RickAcb:
The __isset function doesn't gives back the excepted result because the the __get function has different checks.

If you currently have a component that sets the variable you will be able to get this in the onEnd() function in the layout file. If you try to do isset upon the variable when calling it with the object operator it will return false.

Compoment:
```php
public function onRun()
{
    $this->page['var'] = 'Hello World';
}
```

Layout code:
```php
function onEnd() {
    var_dump(isset($this->var)); // FALSE
    var_dump(isset($this['var'])); // TRUE
    var_dump($this->var); // Hello World
    var_dump($this['var']); // Hello World
}
```

After the changes both isset methods will return TRUE.

### DIFF
--- a/modules/cms/classes/CodeBase.php
+++ b/modules/cms/classes/CodeBase.php
@@ -155,6 +155,14 @@ class CodeBase extends Extendable implements ArrayAccess
      */
     public function __isset($name)
     {
-        return isset($this->page->{$name});
+        if (isset($this->page->components[$name]) || isset($this->layout->components[$name])) {
+            return true;
+        }
+
+        if (isset($this->page->{$name})) {
+            return $value;
+        }
+
+        return array_key_exists($name, $this->controller->vars);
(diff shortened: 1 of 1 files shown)


modules/cms/classes/CodeBase.php
```diff
@@ -155,6 +155,14 @@ public function __set($name, $value)
      */
     public function __isset($name)
     {
-        return isset($this->page->{$name});
+        if (isset($this->page->components[$name]) || isset($this->layout->components[$name])) {
+            return true;
+        }
+
+        if (isset($this->page->{$name})) {
+            return $value;
```
bennothommo: @RickAcb I presume this should be returning `true` here, as `$value` is undefined.
```suggestion
            return true;
```

PR #4518, opened 2019-08-04
Finish the "october:util purge uploads" artisan command

bennothommo:
Continues and finishes the work done by @austinderrick in #4029.

Finishes the functionality for the `october:util purge uploads` command. This had to be done on a new branch as @austinderrick's original PR had diverged from the `develop` branch and couldn't be properly merged.


modules/system/console/OctoberUtil.php
```diff
@@ -294,7 +295,57 @@ protected function utilPurgeUploads()
             return;
         }
 
-        // @todo
+        $totalCount = 0;
+        $uploadsPath = Config::get('filesystems.disks.local.root', storage_path('app')) . '/uploads';
```
w20k: @bennothommo just an idea, maybe it would be useful to take `uploads` path from configs?

modules/system/console/OctoberUtil.php
```diff
@@ -294,7 +295,57 @@ protected function utilPurgeUploads()
...
+                        continue;
+                    }
+
+                    if (FileModel::where('disk_name', basename($file))->count()) {
+                        $this->warn('Skipped file in use: '. str_replace($uploadsPath, '', $file));
+                        continue;
+                    }
+
+                    // Skip .gitignore files
+                    if ($file === '.gitignore') {
```
w20k: @bennothommo is it really needed to store `.gitignore` the file inside of upload folder? :)

modules/system/console/OctoberUtil.php
bennothommo: @w20k Good idea, will change it around soon.

modules/system/console/OctoberUtil.php
bennothommo: @w20k Yes, I believe the `storage/app/uploads` folder is stored in Git with a `.gitignore` file on a normal install - if it removes this file, people using Git might start seeing their uploads tracked in Git as well.

bennothommo: See here: https://github.com/octobercms/october/tree/master/storage/app/uploads :)

w20k: You're right @bennothommo that's a bit odd, though I do remember it was, sort of, an old school way of doing things.

GitIgnore files everywhere:
Screenshot 2019-08-05 at 11 50 59

bennothommo: @w20k It's an old-school way, but it works - I imagine it ensures that the folders are there so people don't get any "missing folder" errors when uploading stuff. :)

modules/system/console/OctoberUtil.php
bennothommo: @LukeTowers not yet, still on the to-do list.

bennothommo:
@LukeTowers I'm gonna move this to 461. I have realised that this feature is only covering local filesystems currently, yet uploads can be stored on remote filesystems. Given this, I will need to re-do the command to also allow purging of uploads on remote filesystems.

LukeTowers:
Implemented in https://github.com/octobercms/october/commit/7fde924ec1d3c66791830e975c998815d0fe0b24

PR #4520, opened 2019-08-04
Add permission support for forms, lists and filters

### LINKED ISSUE #1837: Permission option for lists, forms and filters
I can set permissions for back-end pages. However I would like to add permission for lists columns, forms fields and filters elements. It would be nice a similar solution for this issue:

fields.yaml

```
fields:
    title:
        label: Title
        permission: author.plugin.access_view
```

columns.yaml

```
columns:
    name:
        label: Name
        permission: author.plugin.access_view
```

config_filter.yaml

```
scopes:
    category:
        label: autumn.tools::lang.categories.categories 
        permission: author.plugin.access_view
```

Samuell1:
Resolves #1837


bennothommo:
@Samuell1 I've added in some unit tests for the form permission check for you. I'll need to do the same for the lists and filters too.

LukeTowers:
The property should probably be `permissions` and require an array, just like menu items: https://octobercms.com/docs/plugin/registration#navigation-menus

bennothommo:
@LukeTowers I think the permissions check allows for either a single string or an array - I've changed it to `permissions` and allowed and checked for both types of definition in the tests.

(cc: @Samuell1)

modules/backend/widgets/Filter.php
```diff
@@ -561,6 +562,12 @@ public function addScopes(array $scopes)
         foreach ($scopes as $name => $config) {
             $scopeObj = $this->makeFilterScope($name, $config);
 
+            // Check if user has permission to show this filter
+            $permission = array_get($config, 'permission');
+            if ($permission && !BackendAuth::getUser()->hasAccess($permission)) {
```
LukeTowers:
```suggestion
            if ($permission && !BackendAuth::getUser()->hasAnyAccess($permission)) {
```
Should be `hasAnyAccess()` instead of `hasAccess()`, see https://github.com/octobercms/october/blob/master/modules/backend/classes/NavigationManager.php#L520

LukeTowers: Will need to be changed for all the permission checks (lists, forms, filters)

Samuell1: Can you then write it in single line if its single permission or then its needed to write it for everything as array?

LukeTowers: @Samuell1 permissions can still be defined as a string (`permissions: backend.manage_users` for example) as hasAnyAccess just calls hasAccess($permission, $all = false) which calls hasPermission() which takes an array or a string as input: https://github.com/octobercms/library/blob/master/src/Auth/Models/User.php#L460

LukeTowers: @Samuell1 the method called needs to be changed to `hasAnyAccess()` not just `hasAccess()`

LukeTowers:
@Samuell1 looks like I was wrong, hasAnyAccess typehints an array: https://travis-ci.org/octobercms/october/jobs/582823639#L387

just change to `$permissions = is_string($permissions) ? [$permissions] : $permissions;` then check `if (!empty($permissions)` instead of `if ($permissions`

bennothommo:
@Samuell1 I've fixed the unit tests for you.

@LukeTowers this should be good to review.

composer.json
```diff
@@ -51,6 +51,7 @@
     },
     "autoload-dev": {
         "classmap": [
+            "tests/concerns/InteractsWithAuthentication.php",
```
LukeTowers: Why does this have to be added to the composer file? Why can't it live with the rest of the test cases?

modules/backend/ServiceProvider.php
```diff
@@ -88,6 +90,14 @@ protected function registerAssetBundles()
         });
     }
 
+    /**
+     * Register database factories
+     */
+    protected function registerFactories()
+    {
+        App::make(Factory::class)->load(__DIR__ . '/factories');
```
LukeTowers: woah woah woah, why do we suddenly need factories?

LukeTowers:
This needs to use `hasAnyAccess()` or `hasAccess($permissions, false)` instead to match the NavigationManager: https://github.com/octobercms/october/blob/master/modules/backend/classes/NavigationManager.php#L542. In this PR's current form, all permissions defined are required in order to be permitted. The existing NavigationManager `permissions` definition uses an `OR` check instead of an `AND` check.

composer.json
bennothommo: @LukeTowers It's not a test case, it's a trait to enable emulating of an authenticated user. This was needed for the permissions tests. The reason for it being added there in Composer was because it is namespaced, but our tests folder in general is not currently namespaced.

modules/backend/ServiceProvider.php
bennothommo: @LukeTowers This is used for generating the authenticated user fixture for testing the permissions. I would also be keen to use it down the line for some other unit tests as well.

LukeTowers: Let's get @daftspunk's input on this

bennothommo: @LukeTowers As per @daftspunk's recommendations, this will be moved to the tests folder and only available for tests.

bennothommo: Further to the above, factories are now completely removed and replaced with a UserFixture class.

modules/backend/widgets/Filter.php
```diff
@@ -561,6 +562,12 @@ public function addScopes(array $scopes)
         foreach ($scopes as $name => $config) {
             $scopeObj = $this->makeFilterScope($name, $config);
 
+            // Check if user has permissions to show this filter
+            $permissions = array_get($config, 'permissions');
+            if ($permissions && !BackendAuth::getUser()->hasAccess($permissions, false)) {
```
LukeTowers: @Samuell1 what happens if $permissions is an empty array? Should this check be `!empty($permissions)` instead?

modules/backend/widgets/Lists.php
```diff
@@ -709,6 +710,10 @@ public function getColumns()
      */
     public function getColumn($column)
     {
+        if (!isset($this->allColumns[$column])) {
+            throw new ApplicationException('No definition for column ' . $column);
+        }
```
LukeTowers: Could you provide some more background on the inclusion of this?

Samuell1: I think that was added by @bennothommo

bennothommo: @LukeTowers I noticed that filters had an exception for a missing scope - see here:
https://github.com/octobercms/october/blob/33f5b3933f99c6dd6da93ab3bdd833d86dde317e/modules/backend/widgets/Filter.php#L890-L892
I copied the same idea to Lists to make it easy to test, however, it is arguably a breaking change. We could potentially make that method and the one in Lists return `null` if a scope/column does not exist instead.

LukeTowers: I'm fine with leaving it in, was just curious on the background. Not really a breaking change because it would have thrown an error previously when trying to access an index that didn't exist.

LukeTowers:
@Samuell1 looks great, can you make a PR to octobercms/docs documenting the use of this for filters, lists, and forms?

modules/backend/widgets/Form.php
```diff
@@ -688,6 +689,12 @@ public function addFields(array $fields, $addToArea = null)
             $fieldObj = $this->makeFormField($name, $config);
             $fieldTab = is_array($config) ? array_get($config, 'tab') : null;
 
+            // Check if user has permissions to show this field
+            $permissions = array_get($config, 'permissions');
+            if (!empty($permissions) && !BackendAuth::getUser()->hasAccess($permissions, false)) {
+                continue;
+            }
```
LukeTowers: @Samuell1 can you move this to be the first check of the loop? There's no need to make the fieldObj if we're not going to be using it.

modules/backend/widgets/Filter.php
```diff
@@ -561,6 +562,12 @@ public function addScopes(array $scopes)
         foreach ($scopes as $name => $config) {
             $scopeObj = $this->makeFilterScope($name, $config);
 
+            // Check if user has permissions to show this filter
+            $permissions = array_get($config, 'permissions');
+            if (!empty($permissions) && !BackendAuth::getUser()->hasAccess($permissions, false)) {
+                continue;
+            }
```
LukeTowers: @Samuell1 can you move this to be the first check of the loop? There's no need to make the $scopeObj if we're not going to use it.

LukeTowers:
@Samuell1 looking good, added some change requests to the docs PR and then it should be good to go.

Issue #3559, opened 2018-05-23
How can I remove permissions in external plugin without editing this plugin directly?


iboved:
How can I remove some permissions in external plugin without editing this plugin directly? For example, the external plugin has `registerPermissions` method with permissions in Plugin.php. I don't want to edit this plugin directly. I want to extend this plugin in my plugin in Plugin.php file and remove some permissions.
How can I remove permissions from external plugin in my Plugin.php?
Thank you in advance.

##### October build
October build: 434

LukeTowers:
Github issues really isn't the place for questions like this. Please use one of the many other avenues available instead of submitting new issues for your questions.

Forum: http://octobercms.com/forum
StackOverflow: http://stackoverflow.com/questions/tagged/octobercms
IRC: #october on Freenode or http://octobercms.com/chat
Slack: https://octobercms-slack.herokuapp.com/

As a side note, why do you want to remove permissions???

PR #4522, opened 2019-08-05
New method removePermission() for AuthManager

DanHarrin:
One of my plugins requires the programatic removal of a permission, so I added a method for `AuthManager` called `removePermission()`. It can be used like so:

```
BackendAuth::registerCallback(function ($manager) {
    $manager->deregisterPermission('October.Cms', 'cms.manage_content');
});
```


w20k:
Just a thought, might be useful - to add a method for User `Impeachment` (😄) to remove all User Permissions in one go :)

DanHarrin:
@w20k this is for hiding permissions in the list, not for removing permissions from a user :)

DanHarrin:
I've renamed this method to `deregisterPermission()` instead, to make it super clear 👍

DanHarrin:
Sorry, this doesn't actually work at the moment 😂 just debugging

DanHarrin:
It's actually working now @w20k haha 😂

LukeTowers:
@DanHarrin @w20k I prefer removePermission(), the method exists on the AuthManager class, not the User class so I think it's fine as removePermission().

DanHarrin:
@LukeTowers all done :+1:

modules/backend/classes/AuthManager.php
```diff
@@ -96,6 +96,26 @@ public function registerPermissions($owner, array $definitions)
         }
     }
 
+    /**
+     * Removes a single back-end permission
```
bennothommo: @DanHarrin Would you mind adding the parameters and return value to the docblock?
```suggestion
     * Removes a single back-end permission
     * @param string $owner Specifies the menu item's owner plugin or module in the format Vendor/Module.
     * @param string $code The code of the permission to remove
     * @return void
```

DanHarrin:
That's done too, @bennothommo 👍

LukeTowers:
Related: https://github.com/octobercms/october/issues/3559

modules/backend/classes/AuthManager.php
```diff
@@ -96,6 +96,29 @@ public function registerPermissions($owner, array $definitions)
...
+     * @param string $owner Specifies the menu item's owner plugin or module in the format Vendor/Module.
```
LukeTowers: Can you fix this docblock (and the one for registerPermissions) to say `Specifies the permissions' owner plugin or module in the format Author.Plugin`

DanHarrin:
Docblocks have been updated @LukeTowers 👍

bennothommo:
@DanHarrin @LukeTowers I've added some test cases for `AuthManager`, which should test this functionality along with registering permissions. I've also tested this in normal operation and it all works well.

LukeTowers:
@DanHarrin can you add this to the docs?

DanHarrin:
Sure thing @LukeTowers, all my other PRs need documenting as well so I'll get them all done when I find some time.

Issue #4524, opened 2019-08-05
WidgetManager’s removeReportWidget() Issues


DanHarrin:
- OctoberCMS Build: 457
- PHP Version: 7.3.1

I’ve noticed a couple of issues with `WidgetManager`’s `removeReportWidget()` method.

1. is that the class is missing a simple `use SystemException;`, so I get `Class 'Backend\Classes\SystemException' not found` error messages thrown at me when trying to call `removeReportWidget()`.
2. is that there’s a very odd issue with the detection of whether the widgets are loaded or not.

I am using the following code in my plugin’s `boot()` method to remove the system report widgets:
``` 
Event::listen('system.reportwidgets.extendItems', function($widgetManager) {
    $widgetManager->removeReportWidget('Backend\ReportWidgets\Welcome');
    $widgetManager->removeReportWidget('Cms\ReportWidgets\ActiveTheme');
    $widgetManager->removeReportWidget('System\ReportWidgets\Status');
}); 
```

By default, this doesn’t work. It throws an `Unable to remove a widget before widgets are loaded` exception. Here is the code for the `removeReportWidget()`:
```
 /**
 * Remove a registered ReportWidget.
 * @param string $className Widget class name.
 * @return void
 */
public function removeReportWidget($className)
{
    if (!$this->reportWidgets) {
        throw new SystemException('Unable to remove a widget before widgets are loaded.');
    }

    unset($this->reportWidgets[$className]);
}
```

As you can see, the exception is thrown when `$this->reportWidgets` is just an empty array. Fine, that should work. I now edit the method’s code to see what’s going on:
``` 
dd($this->reportWidgets);

  if (!$this->reportWidgets) {
    throw new SystemException('Unable to remove a widget before widgets are loaded.');
}

unset($this->reportWidgets[$className]);
```

This code is executed, and for some reason returns an array with the three system report widgets! This shouldn’t happen, as the conditional statement’s exception suggests that `$this->reportWidgets` has no items. Odd, right? I go further.
```
 if (!$this->reportWidgets) {
    dd($this->reportWidgets);

    throw new SystemException('Unable to remove a widget before widgets are loaded.');
}

unset($this->reportWidgets[$className]);
```

And the same three-item array is returned! I’m now very confused - how is the conditional passing *and* there be three items in `$this->reportWidgets`? It truly makes no sense to me. Finally, I comment that statement out completely, just leaving the `unset()`. Sure enough, no issues. The report widgets are deregistered successfully.

[...]

PR #4525, opened 2019-08-06
Fix issues with removeReportWidget()

DanHarrin:
As per #4524.


bennothommo:
@DanHarrin I don't think changing the visibility of the `$reportWidgets` property is a good idea - there would be a reason why it is protected. I would suggest instead making a new method that will return the array of report widget class names and use that instead. :)

DanHarrin:
That's just been done @bennothommo, sorry for the delay.

PR #4526, opened 2019-08-06
Use parseFile for caching and parsing

Samuell1:
This adds caching for backend parsing of configs when debug mode is disabled.

Tested with blog plugin on Post backend page:

**Before:**
obrázok

Green squares are yaml parsing calls
obrázok

**After:** 0 yaml parsed files

obrázok

### DIFF
--- a/modules/cms/classes/Theme.php
+++ b/modules/cms/classes/Theme.php
@@ -50,7 +50,6 @@ class Theme
 
     const ACTIVE_KEY = 'cms::theme.active';
     const EDIT_KEY = 'cms::theme.edit';
-    const CONFIG_KEY = 'cms::theme.config';
 
     /**
      * Loads the theme.
@@ -339,21 +338,7 @@ class Theme
             return $this->configCache = [];
         }
 
-        try {
-            if (Config::get('app.debug', false)) {
-                $config = Yaml::parseFile($path);
-            } else {
-                $cacheKey = self::CONFIG_KEY.'::'.$this->getDirName();
-                $config = Cache::rememberForever($cacheKey, function () use ($path) {
-                    return Yaml::parseFile($path);
-                });
-            }
-        }
-        catch (Exception $ex) {
-            // Cache failed
-            $config = Yaml::parseFile($path);
-        }
-
+        $config = Yaml::parseFile($path);
 
         /**
          * @event cms.theme.extendConfig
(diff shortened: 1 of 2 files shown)


prhost:
Maybe use something like `filemtime()` to check if you hear file changes, get worse or lose performance? there may be plugins that manipulate configMaker yaml via panel like Static Pages I guess, I haven't checked yet, but having a clever way to identify change and only use caching when no change would be nice (:

P.S. @ Samuell1 what program did you use to capture application performance? :D

Samuell1:
Static Pages edits yaml files in plugins? @prhost 

I am using https://blackfire.io

LukeTowers:
@prhost Anything that modifies config files is usually done by overriding the local class properties relating to that config, not changing the files themselves; so this change should be fine. 

If filemtime() has little to no impact, then I could see value in adding a `YAML::parseFile()` method that caches every parsed YAML file with a cache key of `yaml::/path/to/file-mtime`

@Samuell1 can you check the performance impact of adding filemtime() to a regular request and then also having it be called probably around 100 times in a single request (just for a theoretical maximum for how many times YAML::parseFile() would be called in a given request)

Samuell1:
@LukeTowers PR created on library. filemtime should be fine if you look at this benchmark 1000 files took only 0.02 (https://stackoverflow.com/a/19959695/4711464) but dont forget filemtime() caches too.

LukeTowers:
@Samuell1 could you do some performance testing on various backend pages (with the test plugin perhaps) to see how much of a performance impact the YAML caching has overall? Don't forget you'll need to replace any instances of `YAML::parse(File::getContents())`, although if File::getContents() is used for supporting remote disks (not sure if themes can be used in this way) then we'll have to make sure we're accounting for that.

Samuell1:
@LukeTowers If we implement this then we can remove that Theme config caching too, it will not only make backend faster but everything that uses Yaml parsing. Yes, i can do tests but next week.

LukeTowers:
@Samuell1 sounds great to me!

prhost:
I believe it will greatly improve performance since much of octobercms is yaml.  @LukeTowers is there any class that overrides the symfony yaml parser?  It would be interesting to implement a layer to make it default for everyone.

prhost:
Oops I saw it already exists. 😁

Samuell1:
Before caching yaml parser:

obrázok

After caching yaml parser:
obrázok


As you can see it calls less things and saves a lot of ram and time.

Samuell1:
@LukeTowers any news on this? Its just revert of caching for theme config.

PR #4531, opened 2019-08-08
Add WAI-Aria breadcrumbs to October CMS

ayumihamsaki:
Related to here: #4454


Samuell1:
I think stub files needs to be updated too with this changes no? (https://github.com/octobercms/library/tree/master/src/Scaffold/Console)

ayumihamsaki:
Yah I agree.

ayumihamsaki:
JS update code lines: https://github.com/octobercms/october/blob/4f141e0a577fad7d8a794eab04fe50c716d41190/modules/system/assets/ui/js/waiaria.js#L41-L53

(Ignore rest - I added it because the other pr's are not merged yet).

ayumihamsaki:
This is all finished now and ready for testing - enjoy!

After merged can expand to all the other places.

modules/system/assets/ui/js/waiaria.js
```diff
@@ -85,7 +87,7 @@ $(function() {
      */
     $('body').on('change', '.switch-field input[role="checkbox"]', function () {
```
bennothommo: @ayumihamsaki Will `event` need to be added to the function parameters here and on line 113 too?
```suggestion
    $('body').on('change', '.switch-field input[role="checkbox"]', function (event) {
```

ayumihamsaki: Done, thanks 👍

Issue #1956, opened 2016-04-23
Explore idea of remove author name from backend URLs


trsh:
When I use the command line "php artisan create:plugin darkweider.shop" I get kind of author-egocentric :)) file structure like darkweider/shop, what is I can live with. Probem is with URL's. 

URL's like /admin/darkweider/shop/orders does not make any sense at all (maybe for personal use only). Therefore I'm forced to rename what was tough a AuthorName to something that makes the URL more readable and acceptable for clients, like "/admin/e/shop/orders". This has consequences:  

1) Now author name is "e" what is not logical (it's not authors name) - especially not cool if I decide to publish the Plugin.
2) Unnecessarily longer URL. I think it's enough with the pluginName. 

So, as I think 90% DEV's will put something other in the AuthorName place, maybe it's worth to refactor? My proposal is:

a) Short term - authorName could be renamed to Package name, as it is more often the case... And make authorName for plugins deprecated. 
b) Longer term - remove authorName from plugin architecture (it stays can rest in Plugin.php).

----- UPDATED

In the struggle for PluginName uniqueness:

a) Don't bother - select, select2, e-shop, e-shop-extra is fine, always has been. AuthorName as additional info can help.
b) Let DEVS to customize the URL as they want And still refer to the correct controllers. Nowadays it's should not be rocket science. I see this a lot in other frameworks.

tsioukas:
The logic behind this architecture is to avoid conflict between plugins, not custom plugins but for external. Also if you see how backend routing works, with this feature can easily connect url request with plugins.

But I agree, isn’t pretty for customers to see this kind of url, and creates a mesh in admin logic when custom urls created. For me I use data as author name and the name of plugin for each main feature that I create. So I have a url like cms/data/pluginname/option.

Maybe we should go to an intermediate step. 

One of my ideas is to create a plugin that collects all installed plugins and create a re-route feature. So each admin can control backend SEO appearance without conflicts or changes in real plugin url (this also will solve compatibility problems with old plugins). 

I don’t know if this is possible but it;s a simple solution to have it both.

trsh:
I haven't seen the code. But mostly it's same thing. Server side receives URL, parses it, and then there is some underlying logic that finds the according Controller taking that data into account. What is need to done, is a possibility to inject some piece of logic, that can overrides/determinate those rules.

trsh:
[...]

PR #4538, opened 2019-08-11
Remove Author Name from Backend URLs (where possible)

DanHarrin:
This PR contains some changes I have made whilst starting work on a solution for #1956.

I used some of the information given by @LukeTowers [here](https://github.com/octobercms/october/issues/1956#issuecomment-356510337).

There are a number of issues that I'm a little stuck on, so if someone could give me a few pointers, that would be great. Currently, these changes...
- provide resolution for URLs that don't include the author's name.
- generate URLs without the author's name where possible (broken since 213e425).

Some issues with this PR include...
- AJAX requests made through backend controllers, like `onSave`, still redirect to the author's name URL version.
- DB requests are made to check if there are any other installed plugins with the same name. This is probably not the most optimised way of performing this check.

As I noted beforehand, these changes are in no way production-ready and **require heavy revision before they are production ready**.

### DIFF
--- a/modules/backend/classes/BackendController.php
+++ b/modules/backend/classes/BackendController.php
@@ -207,18 +207,23 @@ class BackendController extends ControllerBase
          * Look for a Plugin controller
          */
         if (count($params) >= 1) {
-            $pluginVersions = PluginVersion::applyEnabled()
-                ->where('code', ucfirst($params[0]) . '.' . ucfirst($params[1]))
-                ->orWhere('code', 'like', '%' . ucfirst($params[0]))
-                ->get();
+            $plugins = array();
 
-            $pluginVersion = count($pluginVersions) === 1 ? $pluginVersions->first() : null;
+            foreach (PluginManager::instance()->getPlugins() as $pluginCode => $plugin) {
+                if ($pluginCode === ucfirst($params[0]) . '.' . ucfirst($params[1])) {
+                    $plugins[] = $pluginCode;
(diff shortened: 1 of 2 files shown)


modules/backend/classes/BackendController.php
```diff
@@ -205,8 +206,26 @@ protected function getRequestedController($url)
         /*
          * Look for a Plugin controller
          */
-        if (count($params) >= 2) {
-            list($author, $plugin) = $params;
+        if (count($params) >= 1) {
+            $pluginVersions = PluginVersion::applyEnabled()
```
bennothommo: @DanHarrin Is there any reason for using `PluginVersion` over just looping through the results of `PluginManager::getPlugins()`? That should also contain the enabled plugin codes, and would save you an SQL query.

DanHarrin: just pushed changes using that method instead @bennothommo, URLs aren't rewriting now though

modules/backend/classes/BackendController.php
```diff
@@ -205,8 +206,31 @@ protected function getRequestedController($url)
         /*
          * Look for a Plugin controller
          */
-        if (count($params) >= 2) {
-            list($author, $plugin) = $params;
+        if (count($params) >= 1) {
+            $plugins = array();
```
Samuell1:
```suggestion
            $plugins = [];
```

PR #4539, opened 2019-08-11
Soft Component Support

danharrin:
This PR relates to issue #1431 and implements @LukeTowers' suggestions to @viamage's work on #3577.


bennothommo:
@DanHarrin It appears your new test failed in the unit tests:
https://travis-ci.org/octobercms/october/jobs/570575180#L417

Would you be able to double-check that the unit test is correctly set up?

DanHarrin:
@bennothommo I think I've fixed it?

w20k:
@DanHarrin you could run code quality checks on your local machine, as well ;)

DanHarrin:
Looks like they were fixed before you sent the message @w20k 😅

modules/cms/classes/ComponentManager.php
```diff
@@ -193,27 +193,29 @@ public function hasComponent($name)
...
+        if (!class_exists($className) && !$isSoftComponent) {
             throw new SystemException(sprintf(
                 'Component class not found "%s". Check the component plugin.',
                 $className
             ));
         }
 
-        $component = App::make($className, [$cmsObject, $properties]);
-        $component->name = $name;
+        if ($className && class_exists($className)) {
```
bennothommo: @DanHarrin The `$className` condition might be redundant here.
```suggestion
        if (class_exists($className)) {
```

modules/cms/classes/Controller.php
```diff
@@ -1436,31 +1436,52 @@ public function param($name, $default = null)
...
         $manager = ComponentManager::instance();
 
-        if ($addToLayout) {
-            if (!$componentObj = $manager->makeComponent($name, $this->layoutObj, $properties)) {
-                throw new CmsException(Lang::get('cms::lang.component.not_found', ['name'=>$name]));
-            }
-
-            $componentObj->alias = $alias;
-            $this->vars[$alias] = $this->layout->components[$alias] = $componentObj;
+        if ($isSoftComponent = $this->isSoftComponent($name)) {
```
bennothommo: @DanHarrin Could you take the variable assignment out of this condition and just check against the result? IMO, I find this sort of shortcut a bit of a code smell because it's easy to miss the single `=` when debugging.
```suggestion
        $isSoftComponent = $this->isSoftComponent($name);
        if ($isSoftComponent) {
```

modules/cms/classes/Controller.php
```diff
@@ -1436,31 +1436,52 @@ public function param($name, $default = null)
      * @param array  $properties  Component properties
      * @param bool   $addToLayout Add to layout, instead of page
      * @return ComponentBase Component object
```
bennothommo: @DanHarrin Would you be able to indicate in the docblock that this method can return `null` as well if a soft component is used but not found?

modules/cms/classes/Controller.php
```diff
@@ -1605,4 +1626,27 @@ protected function verifyCsrfToken()
...
+
+    /**
+     * Removes '@' from soft component name
+     * @param string $label
+     * @return string
+     */
+    protected function parseComponentLabel($label)
+    {
+        if ($this->isSoftComponent($label)) {
+            return str_replace('@', '', $label);
```
bennothommo: @DanHarrin Could this potentially strip legitimate `@` symbols?

modules/cms/classes/Controller.php
```diff
@@ -1605,4 +1626,27 @@ protected function verifyCsrfToken()
...
+    }
+
+    /**
+     * Checks if component name has @.
+     * @param string $label
+     * @return bool
+     */
+    protected function isSoftComponent($label)
+    {
+        return $label[0] === '@';
```
bennothommo: @DanHarrin Please change this to use `substr` instead to make the intention of this line more clear.
```suggestion
        return substr($label[0], 0, 1) === '@';
```

bennothommo (changes requested):

modules/cms/classes/Controller.php
DanHarrin: @bennothommo potentially, I've replaces `str_replace()` with `ltrim()`.

modules/cms/classes/Controller.php
DanHarrin: Done @bennothommo 👍

DanHarrin:
Fixed it all @bennothommo 👍

bennothommo (approved):

modules/cms/classes/Controller.php
bennothommo: @DanHarrin Docblocks should only have one `@return` tag. I would suggest the following instead: 
```
@return ComponentBase|null Component object. Will return `null` if a soft component is used but not found.
```

bennothommo:
Nice one @DanHarrin. Just one further change to the docblock and then all that's left is testing. :)

DanHarrin:
That docblock has been fixed @bennothommo 👍

bennothommo:
@DanHarrin I have tested this out. Whilst the mechanism itself works very well, it does prevent the page from being edited any further from within the CMS:

I tested it out by using the Vanilla theme from the marketplace, which has a requirement for the User plugin. I disabled the plugin, and then made the "account" and "session" components optional. That all worked fine. Then, when I tried to edit the "Account" page, the following showed:

image

This is correct, as the components do not exist when the User plugin is disabled... but, there's a few gotchas here:

- If I try to save the page, the properties in the optional component are lost. There is a warning that comes up when saving the page if this is the case, indicating that the component properties would be lost, so that's fine - but given it's optional, I think this behaviour could be changed just for the optional component.
- When I re-enabled the User plugin, it still said that the "@account" component was missing - because it no longer matches the correct component code for the "account" component as defined by the User plugin. There will need to be an allowance made for this check to ignore an @ symbol as the first character.
- Users in the CMS cannot currently add an @ symbol to make the component optional - it would have to be done through editing the file directly. A checkbox to make the component optional in the Inspector might be worth adding.

Would you be interested in having a look at these as well?

DanHarrin:
@bennothommo can you give me an idea as to where I could find where...
- the framework deletes an unknown component's properties
- the framework removes the `@` from the beginning of soft component names on save

Thanks!

bennothommo:
@DanHarrin This code block here handles the validation of the alias property with components:
https://github.com/octobercms/october/blob/develop/modules/cms/classes/ComponentHelpers.php#L24-L36
So I believe that is where you will need to edit to allow an @ character at the start of the component names.

bennothommo:
@DanHarrin As for the deletion, I haven't quite worked out where that is, but I believe all the component handling magic is in this file: https://github.com/octobercms/october/blob/develop/modules/cms/classes/CmsCompoundObject.php

DanHarrin:
As a status update, I just pushed support for a new `SoftComponent` class so that `UnknownComponent` isn't used in soft cases.
A soft component now

There currently is only one issue - when you try and save a soft component when it's missing, a space is prefixed onto its alias and its config properties are cleared. I've spent quite a few hours looking for where the components are passed back to the controller before they're saved to the file so that I can correct this, but haven't yet found anything. Any pointer as to where this could be found would be much appreciated.

I'm currently working on a checkbox for the component inspector to soften a component, as per @bennothommo's suggestion.

DanHarrin:
On second thoughts, should soft components even show on the list?

LukeTowers:
@DanHarrin I'm fine with them showing, what were you thinking in regards to not showing them?

DanHarrin:
@LukeTowers well I was questioning why the user would ever need to manipulate / view a component that isn't installed, it could clutter UI a little. It's not major and I'm fine with it either way, we'll see what others think.

bennothommo:
@DanHarrin I would keep them visible - it gives the end user a mechanism to remove the optional component if they wish. I would imagine theme developers might want to include optional components if people have certain plugins installed on their site along with the theme.

DanHarrin:
@bennothommo cool, they'll stay then.

Please could you give me any clues regarding previous comment?

> There currently is only one issue - when you try and save a soft component when it's missing, a space is prefixed onto its alias and its config properties are cleared. I've spent quite a few hours looking for where the components are passed back to the controller before they're saved to the file so that I can correct this, but haven't yet found anything. Any pointer as to where this could be found would be much appreciated.

Thanks!

bennothommo:
@DanHarrin I'm not sure where this happens - I think it is in the CmsCompoundObject class I linked earlier, but not sure at what point. I'm also in two minds about keeping the properties anyways - when a soft component is unavailable due to not being installed or disabled, there's technically no way to know which properties are valid, as there's no property configuration for the component being loaded.

DanHarrin:
Okay @bennothommo, what do you think @LukeTowers?

LukeTowers:
@DanHarrin @bennothommo I'm of the opinion that configuration should be retained. Just because the property validation isn't in place doesn't mean that all hell should break loose should invalid data creep in there, components should be robust enough to handle that case as you could just as easily manually edit the file and wreak havoc.

bennothommo:
@DanHarrin @LukeTowers very good points, and agreed.

bennothommo:
@DanHarrin I've found where the component settings are processed and the random space in front of missing components happens, in case you haven't found it yet. It happens here in the `upgradeSettings` method here:
https://github.com/octobercms/october/blob/develop/modules/cms/controllers/Index.php#L674

It appears that it is currently relying on POST data in order to fill the component settings - given that missing components do not generate an "Inspector" widget and therefore do not contribute any POST data, this would be why the component settings are also going missing after save.

I would imagine a suitable fix would be to detect if a soft component is missing, and perhaps filling the current property data direct from the model into the POST data or somewhere in the overall workflow of the [onSave](https://github.com/octobercms/october/blob/develop/modules/cms/controllers/Index.php#L161) event method.

DanHarrin:
This should now be fixed @bennothommo @LukeTowers 🙂

bennothommo:
Great job @DanHarrin - will give it another look over soon.

w20k:
Awesome job, @DanHarrin! will take look, hopefully soon)!

bennothommo:
@DanHarrin I've applied a couple of bug fixes for the aliasing of components (whether they are "soft" or not), and made it consistent with the displayed alias when editing a page. The fixes also allow for multiple instances of a soft component with different aliases.

@LukeTowers this is all tested, so if you're happy with it, it's good to merge.

bennothommo:
Hold off for the moment - still found some edge cases with the aliases.

bennothommo:
Thanks for your work on this @DanHarrin. We're going to have to hold merging this until build 461 as we are trying to expedite build 460 being released, so we are just focusing on bug fixes for 460 now.

DanHarrin:
Any movement on this? @LukeTowers

tests/fixtures/themes/test/pages/no-soft-component-class.htm
```diff
@@ -0,0 +1,5 @@
+url = "/no-soft-component-class"
+
+[@PeterPan\Nevernever\Land noComponentExist]
```
LukeTowers: Is there a test for a component that does exist being included as a soft component?

LukeTowers:
> Hold off for the moment - still found some edge cases with the aliases.

@bennothommo any updates on the testing of this?

bennothommo:
@LukeTowers ready for final review and testing. All edge cases I previously encountered are now gone, and I've added in the unit test you requested.

bennothommo:
Docs update here: https://github.com/octobercms/docs/pull/448

modules/cms/classes/ComponentManager.php
```diff
@@ -192,27 +192,30 @@ public function hasComponent($name)
      * @param array $properties The properties set by the Page or Layout.
      * @return ComponentBase The component object.
      */
-    public function makeComponent($name, $cmsObject = null, $properties = [])
+    public function makeComponent($name, $cmsObject = null, $properties = [], $isSoftComponent = false)
```
LukeTowers, line 195: Docblock needs to get updated

modules/cms/classes/Controller.php
```diff
@@ -1408,32 +1408,54 @@ public function param($name, $default = null)
      * @param string $alias       Alias to give the component
      * @param array  $properties  Component properties
      * @param bool   $addToLayout Add to layout, instead of page
-     * @return ComponentBase Component object
+     * @return ComponentBase|null Component object. Will return `null` if a soft component is used but not found.
+     * @throws CmsException
+     * @throws SystemException
```
LukeTowers, line 1413: Needs explanations for what states will trigger those exceptions

modules/cms/classes/Controller.php
```diff
@@ -1408,32 +1408,54 @@ public function param($name, $default = null)
...
 
-            $componentObj->alias = $alias;
-            $this->vars[$alias] = $this->page->components[$alias] = $componentObj;
+            $this->setComponentPropertiesFromParams($componentObj);
+            $componentObj->init();
+
+            return $componentObj;
+        /**
+         * We catch two exceptions below, as missing plugin can throw both and catch \Exception would be too global.
+         */
```
LukeTowers, line 1447: Just make this a single line comment `//` and add an extra line of whitespace between return and the comment

modules/cms/components/SoftComponent.php
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    /**
+     * @return array
+     */
+    public function componentDetails()
+    {
+        return [
+            'name'        => 'Soft component',
+            'description' => 'This component is missing but optional.'
```
LukeTowers, line 30: Can we add translation strings for this?

modules/cms/formwidgets/Components.php
```diff
@@ -42,7 +43,7 @@ protected function listComponents()
             try {
                 $componentObj = $manager->makeComponent($name, null, $properties);
 
-                $componentObj->alias = $alias;
+                $componentObj->alias = ((substr($name, 0, 1) === '@' && $alias !== $name) ? '@' : '') . $alias;
```
LukeTowers, line 46: Could you explain what this does exactly? Taking me too long to mentally parse

modules/cms/formwidgets/Components.php
```diff
@@ -57,9 +58,16 @@ protected function listComponents()
                 }
             }
             catch (Exception $ex) {
-                $componentObj = new UnknownComponent(null, $properties, $ex->getMessage());
-                $componentObj->alias = $alias;
-                $componentObj->pluginIcon = 'icon-bug';
+                if ($name[0] === '@') {
```
LukeTowers, line 61: Can you use starts_with() instead of all these different ways of looking at the start of the string?

LukeTowers (changes requested):

modules/cms/formwidgets/Components.php
bennothommo: This line determines if the component is a soft component (by being prefixed with an `@`) and if it is, and an alias is in use, it prefixes the alias with `@`. This is for the component property inspector - it allows people to add `@` to an alias to make any component a soft component, and ensures that the `@` is visible when the component properties are reloaded.

bennothommo:
Changes made @LukeTowers.

modules/cms/classes/ComponentManager.php
```diff
@@ -187,32 +187,40 @@ public function hasComponent($name)
...
     /**
      * Makes a component object with properties set.
+     *
      * @param string $name A component class name or code.
      * @param CmsObject $cmsObject The Cms object that spawned this component.
      * @param array $properties The properties set by the Page or Layout.
+     * @param bool $isSoftComponent Defines if this is a soft component.
+     *
      * @return ComponentBase The component object.
+     *
```
LukeTowers, line 197: I think we have the @throws immediately following the @return elsewhere in the codebase, we should do that here too.

bennothommo: I'd argue that spacing them out makes it easier to categorise and read. Thoughts?

LukeTowers: Sure, but I'm going off of what we already have in the codebase. We could consider a separate PR later to adjust that formatting across the codebase but for now we should stick with what's already being used.

PR #4543, opened 2019-08-12
Minor Spelling Errors and Inconsistencies

DanHarrin:
This PR fixes numerous "oportunity" spelling errors, and inconsistencies between the use of "post process" and "post-process" (of which the latter is more readable).


modules/cms/classes/AutoDatasource.php
```diff
@@ -100,7 +100,7 @@ public function sourceHasModel(string $source, Model $model)
 
             // Deleted paths are included as being handled by a datasource
             // The functionality built on this will need to make sure they
-            // include deleted records when actually performing sycning actions
+            // include deleted records when actually performing synching actions
```
LukeTowers: https://english.stackexchange.com/a/46473

DanHarrin: so you want that changed back? that source says that both British and Americans prefer `synching`...

LukeTowers: > For similar reasons and consistency, I'd recommend synced over synched and async over asynch.

Yes please

DanHarrin: Done @LukeTowers

PR #4546, opened 2019-08-14
Make the Code editor fullscreen button easier to see

ayumihamsaki:
I would like to request this PR for review, I find the fullscreen code editor button hard to see and would like to make it clearer - I strongly suggest making things with opacity is a bad practise and makes things hard to see!

This is a before screenshot:

image

This is an after screenshot (opacity removed):

image

I hope you agree the second image is less blurry and easier to see.


bennothommo:
@ayumihamsaki Apologies, but I disagree with this change at the moment. The reason it has the opacity is to give it a hover colour (and visual indication) when you hover over the icon. Your change would remove that.

You could try either darkening the colour of the icons (they are currently `#666`) or applying another colour for the hover effect.

LukeTowers:
I believe the opacity is also there to make it so that the buttons aren't covering up any code.

ayumihamsaki:
Would you be happy if I switched it the other way round.

Normal mode: opacity: 1

Hover mode: opacity: 0.7

What do you think?

LukeTowers:
The containing tab should retain the opacity, it's there for a reason

ayumihamsaki:
@LukeTowers 

I tested it and it definetly not doing what you are suggesting see here:

image

```css
.field-codeeditor .editor-toolbar >ul >li >a >i {opacity:0.7;filter:alpha(opacity=70);display:block}
.field-codeeditor .editor-toolbar >ul >li >a:focus >i {opacity:1;filter:alpha(opacity=100)}
```

They are both styling the `i` which is the font-awesome icons not the container.

e.g.

```html
<i class="icon-search"></i>
```

Doesn't make sense to me what you are saying.

---

Anyway, I will keep the opacity the same and try to find a darker color.

LukeTowers:
@ayumihamsaki fair enough, remove the opacity then

ayumihamsaki:
@LukeTowers @bennothommo ok thanks, this pr is ready then. Unless you need me to make any changes to it?

LukeTowers:
@ayumihamsaki 

> Apologies, but I disagree with this change at the moment. The reason it has the opacity is to give it a hover colour (and visual indication) when you hover over the icon. Your change would remove that.

Fix that please

ayumihamsaki:
@LukeTowers Thanks for quick reply, I will do this:

Normal mode: opacity: 1

Hover mode: opacity: 0.7

ayumihamsaki:
@bennothommo let me know if that works for you ok, or I need to change the color?

bennothommo:
@ayumihamsaki Yep, that's fine by me :)

w20k:
Let me guys test this out, I'm kind of on `disagreeing` edge with the change :)

w20k:
Sorry, @ayumihamsaki but disagree with this change, lowering the opacity on hover sound like a bad idea to me. More to it, from UX perspective it's, kind of, a good solution to make everything, despite your editor, less visible while you're using fullscreen mode.

ayumihamsaki:
@w20k 

I'm gonna add some logic here!

The fullscreen buttons have in normal `mode: opacity: 1` which gives them the most clarity. When someone hovers the mouse over the buttons the mouse has a styling of `cursor: pointer` which is actually covering most of the icon buttons area so actually the hover color is almost pointless - though you can see a small bit of the icons while you have the mouse pointer hovering over the icons.

No offence but I've waited 5 days for your decision and you haven't even given any real solution!

Come on, come on - I'm still waiting... lol

Would you be happy if I turned the buttons into the same as the default buttons in October then?

```css
.btn-default {
    color: #fff;
    border: none;
    background: #656d79;
}
.btn-default:hover, .btn-default:focus, .btn-default:active, .btn-default.active, .open .dropdown-toggle.btn-default {
    color: #fff;
    background: #1681ba;
    border-color: #1681ba;
}
```

w20k:
@ayumihamsaki I'm against the change. In a manner of changing opacity back and forth.

I'd even add opacity to a badge itself, but that's would be my own, project-based, design decision. Your suggestion is incorrect in a sense of changing the opacity on hover - `which is odd`. 

You could ask @LukeTowers & @bennothommo about removing opacity, again, that would be a solution I'd go for.
But by using a different class name, like: `btns__fullscreen` or something similar. 

PS: I had a fever for the past 4 days. Could I be excused, @ayumihamsaki? 😄

ayumihamsaki:
@w20k hope you feeling better now!

What about this then:

ezgif-2-341fd8da3065

That matches the same as a default btn in October.

bennothommo:
@ayumihamsaki @w20k Taking a step back - I agree with @w20k's sentiments. In all honesty, I think it should be *less visible* in full screen mode, not more. When someone goes full screen, it's because they are wanting less distraction and more space for code. I find that particular tab of buttons to stick out like a sore thumb in this view.

I'm more on @w20k's wavelength about trying to almost completely hide it (or fade it out a heap) until someone hovers over it, but at the same time I can respect that @ayumihamsaki is trying to make the contrast between the tab background and the buttons better. I think @ayumihamsaki's original contrast was good, except on hover, the icon should go full black as opposed to a lighter shade of grey.

ayumihamsaki:
@bennothommo Thanks for more feedback, so expanding on your comment we get the following now:

ezgif-2-5c6365d294db

Normal and hover mode: opacity: 1
Normal mode: same color
Hover mode: black color

- I think it's too much and too shrap with the color black maybe a slightly less tone?

ezgif-2-28ae13a8522a

Normal and hover mode: opacity: 1
Normal mode: same color
Hover mode: #191919 color (one tone lighter from black)

@w20k @LukeTowers Place your votes.

LukeTowers:
@ayumihamsaki full black is my vote

w20k:
@ayumihamsaki my vote is for a second (#191919). Pure black has some crispy edges.
@bennothommo, your vote?

ayumihamsaki:
p.s. my vote is exactly the same as w20k "is for a second (#191919). Pure black has some crispy edges." 

(sorry it's hard to see in the gif but on the cms it's more easier to spot).

bennothommo:
My vote is for #191919 - I tried the two out myself and solid black did look a bit "sharp".

ayumihamsaki:
@bennothommo @w20k @LukeTowers 

Ok, have updated the pr as per conversation. If someone wants to test it, thanks.

w20k:
@bennothommo @LukeTowers testing finished and ready to be merged if you have something to add plz do ;)

Screenshots:

_Before_
Screenshot 2019-09-03 at 09 57 28
Screenshot 2019-09-03 at 09 57 32

_After_
Screenshot 2019-09-03 at 09 56 02
Screenshot 2019-09-03 at 09 56 07

bennothommo:
@w20k Looks good to me.

ayumihamsaki:
p.s. Thank you to everyone for their feedback ❤️

PR #4549, opened 2019-08-15
Allow decompiled Backend JS assets

bennothommo:
This fixes one of my pet peeves of doing front-end asset work on the October backend.

Currently, when doing any work on the JS files, a developer has to run `php artisan october:util compile assets` or `php artisan october:util compile js` to be able to see the changes on the Backend. When making minor JS changes constantly, this can result in several minutes of wasted time waiting for the assets to compile.

This change will allow the individual assets that are compiled into the full compilation file to be loaded instead, allowing the developer to see their changes immediately. It introduces a new configuration variable, `cms.decompileBackendAssets`, that controls this functionality. By default, it is `false` and not tied to the `debug` value, requiring it to be explicitly enabled.

This change currently handles the JS files. I was going to also allow this for LESS files, but due to how LESS compiles, it's too difficult to seperate changes at this stage, so we'll still have to use `php artisan october:util compile less` for that. I'll add functionality for compiling LESS on the fly in another PR.

### Tasks
- [x] Allow JS file decompilation in the Backend.
- ~~Allow LESS file decompilation in the Backend.~~
- [x] Show warning in Status widget on Dashboard when `cms.decompileBackendAssets` is enabled.
- [x] Add unit tests, if required.
- [x] Add specific exceptions if decompilation fails (suggested by @nathan-van-der-werf)


modules/backend/helpers/Backend.php
```diff
@@ -156,4 +157,54 @@ public function dateTime($dateTime, $options = [])
...
+
+    /**
+     * Decompiles the compilation asset files
+     *
+     * This is used to load each individual asset file, as opposed to using the compilation assets. This is useful only
+     * for development, to allow developers to test changes without having to re-compile assets.
+     *
+     * @param string $file The compilation asset file to decompile
+     * @param boolean $skinAsset If true, will load decompiled assets from the "skins" directory.
+     * @return array
```
nathan-van-der-werf: Missing an @throws Exception

Also, it might be better to make specific exceptions for specific errors.

See for example: https://blog.nikolaposa.in.rs/2016/08/17/exceptional-behavior-best-practices/

It makes it possible to catch the correct "error".

bennothommo: @nathan-van-der-werf Good idea, will implement. :)

bennothommo: I've added `Backend\Helpers\Exception\DecompileException` - it's not unique to each individual exception, but it works with our error handler at least.

LukeTowers:
@bennothommo can you move this config item into a new config file, `develop.php`? I'd prefer to keep the cms config simpler (I'm thinking we might move other stuff out at some point too, like the restrictBaseDir and enableTwigStrictVariables)

bennothommo:
@LukeTowers sounds good - done :)

LukeTowers:
@bennothommo instead of decompiling, is there a way you can just pass this to the asset combiner with $this->addJs([]) to produce a "/combiner/asdlkfjpo34f3432" URL instead of the original compiled URL? This would also work for LESS files

bennothommo:
@LukeTowers I did try that, but I wasn't thrilled with the performance hit.

LukeTowers:
@bennothommo is a decompiler something that could be added as an Assetic filter?

bennothommo:
@LukeTowers In this case, I think it's fairly intrinsic to October and probably wouldn't suit going into Assetic. It's basically just pulling out the paths to the assets in the compilation JS files and allowing us to insert them individually into our scripts block. I can investigate further if you want.

LukeTowers:
That's fine then

DanHarrin:
This is a lifesaver @bennothommo!! Thanks for this!

PR #4550, opened 2019-08-15
Merge Order 7:  Add Keyboard navigation for WAI-ARIA Tooltips + fix security threat

ayumihamsaki:
This pr is made up of two parts:

1. I found a security threat, when you use: `target="_blank"` you must use rel="noopener noreferrer" to stop the threat from tabnapping. To understand this security issue see here: https://en.wikipedia.org/wiki/Tabnabbing

This threat is fixed on this line: https://github.com/ayumihamsaki/october/blob/940d34c4b416f3b9a1065281612db37f30799ea4/modules/backend/layouts/_mainmenu.htm#L54

2. Add keyboard navigation to tooltips, there are two tooltips in the top navigation section which I have added keyboard navigation, see screenshot of where I'm talking about:

image

Keyboard controls:

- `Tab` - select the button (displays the tooltip at the same time).
- `Shift + Tab` - select the previous button (displays the tooltip at the same time).
- `Escape` - Close the tooltip.

That's basically everything - I've tested this at the same time while coding it on a localhost test bed.

Oh one last thing:

Please add this pr to: https://github.com/octobercms/october/projects/9


LukeTowers:
@ayumihamsaki first of all, tabnabbing is not the term you're looking for:

> tabnabbing is different from most phishing attacks in that the user no longer remembers that a certain tab was the result of a link unrelated to the login page, because the fake login page is loaded in one of the long-lived open tabs in their browser

Tabnabbing is a phishing attack. What you're talking about is https://mathiasbynens.github.io/rel-noopener/, and is not a security issue under October's threat model because a malicious party does not have control over the frontend, you do.

ayumihamsaki:
@LukeTowers All a hacker needs to do is inject something into the cache on the frontend. A user from the backend clicks on the button to go to the frontend and then goes backwards to the backend again then the attack is complete.

LukeTowers:
Since that requires someone to take advantage of an unknown vulnerability on the frontend (probably a XSS attack) this is not a high priority security issue. I'm not saying we won't make the change, just that I disagree with your assessment of it. As a note: https://github.com/octobercms/october/blob/master/.github/CONTRIBUTING.md#reporting-security-issues

ayumihamsaki:
In this situation I don't see the point in wasting time emailing the octobercms website over a single line of code, which I spotted while doing the tooltips. There I just added it into this pr. It doesn't make sense.

LukeTowers:
@ayumihamsaki the point is about disclosure. It doesn't matter how easy it is to fix, if our Github issues list is disclosing vulnerabilities all over the place then that's not good for our users because now anyone can easily find out about it before a fix is released. Either don't say anything about it on a public issue unless we ask you why the change was made or send in an email to discuss it ahead of time please.

ayumihamsaki:
Ok understand 👍

ayumihamsaki:
@LukeTowers Sorry keep tagging you. With regards to you saying about **disclosing vulnerabilities**, would it be a good idea to add your comment into the `CONTRIBUTING.md` file, just so everyone knows about that and we are all on the same page.

LukeTowers:
I've added it to https://github.com/octobercms/october/blob/develop/.github/CONTRIBUTING.md#security-vulnerabilities

ayumihamsaki:
Thanks, if I do the same mistake again you have my blessing to shout at me, now I got no excuses!

daftspunk:
The tabnabbing fix has been implemented (b60adf29526a0b51f119ea7cf8bf93b6b915edf3)

Please do not bundle multiple changes into the same PR. The other changes have been discarded and will need to be resubmitted for consideration individually. Apologies in advance for any frustration this may cause

PR #4555, opened 2019-08-18
Update Arabic language

mabehiry:
Added some Arabic translations in system modules


modules/system/lang/ar/lang.php
```diff
@@ -48,4 +48,8 @@
     'settings' => [
         'menu_label' => 'الإعدادات',
     ],
+	'pagination' => [
```
bennothommo: @mabehiry Can you change the tab character to spaces please? :)

mabehiry: Ok, I did.
Thank you

Issue #4554, opened 2019-08-17
Backend render blocking gravatar image and plugin images


ghost:
### How to reproduce this issue

1. Don't add a custom avatar thumbnail - October will each time try to load an image from the website:

```
//www.gravatar.com
```

This gets even worse because October sends this website a Query String to get back a specific image size.

The other no-no is that October is trying to load the website with a `//` this means both the `HTTP` and `HTTPS` connections are being loaded. October should be loading `HTTPS` only external files.

That code can be found here: https://github.com/octobercms/october/blob/26173486d39cdf4312535ecf247e3d7a161713b4/modules/backend/models/User.php#L94-L117

2. If you have a lot of plugins all the images will render block the loading of the web page. This gets worse when the svg or png file sizes are big and not compressed.

### Screenshot of performance

![image](https://user-images.githubusercontent.com/17784082/63211952-311d4380-c0f6-11e9-8db1-5066c12e95e3.png)

In the above waterfall image we can see three problems.

1. The avator image - blocking the rendering.

2. The plugin images - blocking the rendering.

3. All the plugin images are being double loaded?

### Solutions

1. October should not need to load via an external connection to grab a default avatar thumbnail it should be saved in the October folder.

2. The avatar image should not need to have any query string's needed to work out the image size! The image size range is wide enough to be able to contain all it's image sizes without the need to set up `srcset`.

As per comment found here: https://github.com/octobercms/october/issues/4510#issuecomment-517287002

>Avatar account image sizes in the top navigation menu are now fixed to the following sizes: 45px, 60px and 78px.

> Want to mention something like min-size for account avatar's should be at least 78px in size and in a square format.

Therefore the range would be 45px - 78px meaning a default avatar image of 78px will cover all possible outcomes.

See a previous PR here: https://github.com/octobercms/october/pull/4511

So to fix points 1 and 2 October should load a default avatar image of 78px via a local connection. You can remove all that useless code and replace it with the following logic: Does the user's setup have an avatar? (yes or no) Yes ok use their image, no ok use the default image. As for sizes it's fixed with a `max-width` and `max-height` of 78px.

3. Loading all the plugin main icon images before the web page is rendered is causing me big headaches!

The correct solution would be to load a template at render time and then when the web page has loaded fill in the plugin images.

For example:

![image](https://user-images.githubusercontent.com/17784082/63212152-10ee8400-c0f8-11e9-9546-167f524b357e.png)

[...]

PR #4562, opened 2019-08-22
Lazyload top navigation icons and remove external loading of default avatar

ayumihamsaki:
Simple pr fixes the things I listed over here: https://github.com/octobercms/october/issues/4554

The pr fixes the following things:

1. For a default avatar image it now gets loaded via a data uri image on a local connection. Before the backend was subject to a possible huge delay in loading due to loading an external source and replied on no network congestion. Plus two network connections for HTTP and HTTPs, because of the old coding method of `//` where a browser had to waste time on deciding which is the best method.

2. All images are lazy loaded using the new Native lazyload API. Having 100 plugins or 1000 plugins on a web page will make the backend load in the exact same time.

3. All images have alt-tags, needed for screen readers and performance loading issues.

4. Added `role="img"` which works for screen readers, semantics etc.

5. Speed waterfall timing diagram - removes all this:

image


ayumihamsaki:
Not sure how to fix that conflict message - github is not really explaining much - plus my branch says it's up to date with develop branch?

ayumihamsaki:
Not sure how to fix that conflict message - github is not really explaining much - plus my branch says it's up to date with develop branch?

@bennothommo 

Hi, your PR https://github.com/octobercms/october/pull/4549 says it's merged but when I check the develop branch `modules/backend/layouts/_head.htm` it's not displaying in github?

image

and

image

You would think the develop branch should be showing your merged pr from 6 days ago? But it's not and I have a feeling that's what is causing the conflict here!

w20k:
LGTM, needs testing.

ayumihamsaki:
@w20k recomplied.

w20k:
Looks good to me! 
Tested - ✅ 
Ping-ping @LukeTowers!

Screenshots:
_Deafult_
Screenshot 2019-09-03 at 14 19 49
Screenshot 2019-09-03 at 14 21 10

_Custom Image_
Screenshot 2019-09-03 at 14 22 20

ayumihamsaki:
Wait one second don't merge yet please.

ayumihamsaki:
Ok now ready to merge! 👍

modules/backend/layouts/_mainmenu.htm
```diff
@@ -59,7 +59,7 @@
                 <a
                     href="javascript:;" onclick="$.oc.layout.toggleAccountMenu(this)"
                     title="<?= e(trans('backend::lang.account.signed_in_as', ['full_name' => $this->user->full_name])) ?>">
-                    <img src="<?= $this->user->getAvatarThumb(90, ['mode' => 'crop', 'extension' => 'png']) ?>" class="account-avatar" />
+                    <img role="img" loading="lazy" src="<?= $this->user->getAvatarThumb(90, ['mode' => 'crop', 'extension' => 'png']) ?>" alt="<?= e(trans($this->user->full_name)) ?>" class="account-avatar" width="78" height="78" />
```
LukeTowers: @ayumihamsaki why is `$this->user->full_name` passed through `trans()`? A user's name is not going to be a translation key.

modules/backend/layouts/_head.htm
```diff
@@ -50,6 +55,10 @@
     <link href="<?= $script . '?v=' . $coreBuild; ?>" rel="preload" as="script" importance="high">
 <?php endforeach; ?>
 
+<?php foreach ($fonts as $font) : ?>
+    <link href="<?= $font; ?>" rel="preload" as="font" type="font/woff2" crossorigin importance="high">
```
LukeTowers: Why is the font type specified as woff2 here hardcode when the font itself is woff? Also, why is it hardcoded instead of depending on the file extension?

modules/backend/models/User.php
```diff
@@ -103,17 +103,12 @@ public function getAvatarThumb($size = 25, $options = null)
...
-        $default = array_get($options, 'default', 'mm');
-
         if ($this->avatar) {
             return $this->avatar->getThumb($size, $size, $options);
         }
 
-        return '//www.gravatar.com/avatar/' .
-            md5(strtolower(trim($this->email))) .
-            '?s='. $size .
-            '&d='. urlencode($default);
```
LukeTowers: We can't just get rid of the gravatar entirely, we need to detect when one exists and cache it locally.

modules/backend/layouts/_mainmenu.htm
ayumihamsaki: My bad, didn't know that.

modules/backend/models/User.php
ayumihamsaki: @LukeTowers can we load it from local instead then? I could save it to file location:

```html
https://github.com/octobercms/october/tree/develop/modules/backend/assets/images
```

modules/backend/layouts/_head.htm
ayumihamsaki: Done.

modules/backend/models/User.php
LukeTowers: @ayumihamsaki the data URI isn't the problem, the problem is that people that use gravatars will start seeing mystery man instead of their gravatar. What we should do is add an `afterSave()` method to the user model to attempt to pull in a local copy of the user's gravatar if they don't already have an avatar set. That way existing users just need to go to their profile page and save once to get their gravatar added back.

ayumihamsaki: I understand what you mean, I just don't know how to code it.

modules/backend/layouts/_head.htm
```diff
@@ -50,6 +50,8 @@
     <link href="<?= $script . '?v=' . $coreBuild; ?>" rel="preload" as="script" importance="high">
 <?php endforeach; ?>
 
+    <link href="<?php Url::asset('modules/system/assets/ui/font/fontawesome-webfont.woff?v=1.0.1'); ?>" rel="preload" as="font" type="font/woff" crossorigin importance="high">
```
LukeTowers: this shouldn't be indented here, the other ones are indented because they're in a PHP code block.

modules/backend/layouts/_mainmenu.htm
```diff
@@ -24,7 +24,7 @@
                         <a href="<?= $item->url ?>">
                             <span class="nav-icon">
                                 <?php if ($item->iconSvg): ?>
-                                    <img class="svg-icon" src="<?= Url::asset($item->iconSvg) ?>">
+                                    <img role="img" loading="lazy" class="svg-icon" src="<?= Url::asset($item->iconSvg) ?>" alt="<?= e(trans($item->label)) ?>" width="30" height="30">
```
LukeTowers: why is role="img" required? Also I'm not sure if the alt text should be just the label repeated as screen readers will then be repeating the label of the menu item twice. I'm not sure we should have an alt text for these icons TBH as menu items are required to have labels set AFAIK. If we do decide to have alt text, then it should be in a string like "$label menu item icon" and / or allowed to be overridden by the devs.

modules/backend/layouts/auth.htm
```diff
@@ -37,6 +37,8 @@
             <link href="<?= $script . '?v=' . $coreBuild; ?>" rel="preload" as="script" importance="high">
         <?php endforeach; ?>
 
+            <link href="<?php Url::asset('modules/system/assets/ui/font/fontawesome-webfont.woff?v=1.0.1'); ?>" rel="preload" as="font" type="font/woff" crossorigin importance="high">
```
LukeTowers: Indentation

modules/backend/models/User.php
LukeTowers: Something like the following:

```php
use October\Rain\Network\Http;

public function afterSave()
{
    if ($this->avatar()->exists()) {
       return;
    }

    $avatarRequest = Http::get('http://www.gravatar.com/avatar/' . md5(strtolower(trim($this->email))) . '?s=' . $size . '&d=404');

    if ($avatarRequest->code === 200) {
         $avatar = new System\Models\File();
         $avatar->fromData($avatarRequest->body, $this->email . '-gravatar.jpg');
         $avatar->save();
         $this->avatar()->attach($avatar);
    }
}
``` 
testing required of course but that should get most of the job done. Btw `http://` is intentional, some servers have broken open SSL installations that cause issues when requesting HTTPS URLs from the server so it's probably better to just request HTTP for this.

Samuell1: And if i dont want to download file per every admin? Because server will just have 6 same images if they dont use gravatar

w20k:
@LukeTowers what do you think if the use of `gravatar` could be a part of user configuration or config (and by default return base64 image, if disabled. And true by default.

modules/backend/models/User.php
ayumihamsaki: > Btw http:// is intentional, some servers have broken open SSL installations that cause issues when requesting HTTPS URLs from the server so it's probably better to just request HTTP for this.

Doesn't work like that, you got Mixed Content API stopping that, then if the certificate is missing or broken the borwser will block it and then lastly if the website doesn't have the redirect in .HTaccess you wouldn't even get the file access.

Again why can't I do this:

https://github.com/octobercms/october/tree/develop/modules/backend/assets/images/avatar-default.jpg

Then you got a stable connection and don't reply on a bunch of what if's for the external connection.

ayumihamsaki: @Samuell1 

Is there a piece of code you can link to, which is doing that, or show me in a screenshot, gif, or video. I didn't even know October does that.

modules/backend/layouts/_mainmenu.htm
ayumihamsaki: I removed the role="img" my bad. As for the rest, I'm going to run some tests with a screen reader and see what is going on and will comment back.

modules/backend/models/User.php
ayumihamsaki: @LukeTowers e.g.

```php
<?php

public function afterSave()
{
    if ($this->avatar()->exists()) {
       return;
    }

    $avatarRequest =  Url::asset('modules/backend/assets/images/gravatar.jpg');

    if ($avatarRequest->code === 200) {
         $avatar = new System\Models\File();
         $avatar->fromData($avatarRequest->body, $this->email . '-gravatar.jpg');
         $avatar->save();
         $this->avatar()->attach($avatar);
    }
}
```

ayumihamsaki:
@w20k @LukeTowers 

October avatar seems to be a really poor experience right now, take the following example:

1. In my test localhost, I have opened the backend and is displaying the default avatar.

2. I add a custom image and click save, I get a confirmation message - but the avatar is not updated, see here:

image

I think October should update the avatar when a user clicks 'Save' button. It works ok with 'Save and close' button.

modules/backend/layouts/_mainmenu.htm
ayumihamsaki: @LukeTowers see screenshot as proof:

image

So adding alt-tags are needed for a screen reader and they are not displayed twice!

> If we do decide to have alt text, then it should be in a string like "$label menu item icon" and / or allowed to be overridden by the devs.

I think you don't need to do that.

Right now the alt-tag is displaying:

```html
<img loading="lazy" class="svg-icon" src="http://localhost/modules/backend/assets/images/dashboard-icon.svg" alt="Dashboard" width="30" height="30">
```

I plan to create another PR for the top navigation with regards to adding keyboard navigation and screen reader support. So I will be adding extra code that will turn that section into a special container, basically the screen reader will be reading it like this:

"Main menu, contains eight items,.. item one of eight dahsboard,.. etc."

So I think it's best to have the alt-tag changed to this:

```html
alt="<?= e($item->label) ?>"
```

Let me know if you are happy with all that and your thoughts?

ayumihamsaki: image

Removed the alt-tag.

modules/backend/models/User.php
LukeTowers: > And if i dont want to download file per every admin? Because server will just have 6 same images if they dont use gravatar. This should as option in admin settings like "Use Gravatar image"

@Samuell1 gravatar will only be used if the user has a gravatar set up for themselves, same as it works currently. 

@ayumihamsaki 

> Doesn't work like that, you got Mixed Content API stopping that, then if the certificate is missing or broken the browser will block both HTTPS and HTTP and then lastly if the website doesn't have the redirect in .HTaccess you wouldn't even get the file access for the HTTP.

It does work like that, the browser is not the one requesting the HTTP URL, as far as the browser is concerned a gravatar image magically appears from the server running October, the browser never talks to Gravatar. Thus, no mixed content warning.

Again why can't I do this:

https://github.com/octobercms/october/tree/develop/modules/backend/assets/images/avatar-default.jpg

Because we're not going to drop support for gravatars out right. Displaying a static image without enabling people to still use their gravatars will break gravatars.

LukeTowers:
@ayumihamsaki 

> Also could we re-design this avatar thing, I really hate October dumping things on the right side - likewise with other people's plugins doing the same thing. It makes no sense - when we got a huge amount of space below for example:

No. That space can be used by plugins extending the user and makes a lot of sense in certain situations: 
Screen Shot 2019-09-04 at 9 26 24 AM

> I add a custom image and click save, I get a confirmation message - but the avatar is not updated, see here:

That can be easily fixed by just returning `Redirect::refresh()` on update.

@w20k 

> what do you think if the use of gravatar could be a part of user configuration or config (and by default return base64 image, if disabled. And true by default.

I don't mind it except that it needs to be a database / branding setting, and we still need to support gravatar in the way that I mentioned.

modules/backend/layouts/_mainmenu.htm
ayumihamsaki: @LukeTowers 

Ok done some research and here is the following corrections and results:

### Menu item links

image

Solution: Removed the alt-tag, the result is:

```html
Name: "Dashboard"

    aria-labelledby: Not specified
    aria-label: Not specified
    Contents: "Dashboard" <---
    title: Not specified
```

### Preview the website link

image

Solution: There was no alt-tag, but the result was blank and I've added an `aria-label` attribute that now gives the result for the screen reader, the result is:

```html
Name: "Preview the website"

    aria-labelledby: Not specified
    aria-label: "Preview the website" <---
    Contents: Not specified
    title: Not specified
```

### Avatar link

image

Solution: I've removed the `alt-tag` and I've added an `aria-label` attribute that now gives the result for the screen reader, the result is:


```html
Name: "Signed in as Admin Person" 

    aria-labelledby: Not specified
    aria-label: "Signed in as Admin Person" <---
    Contents: Not specified
    title: Not specified
```

Hope that is fine with you, now passes my screen reader tests.

ayumihamsaki: @LukeTowers Have marked this as un-resolved. Please can you give me a suggestion for a better description of this sub-menu for the avatar, see previous comment (sorry I was writing it while I was grabbing each test - maybe you saw an old cached version in github).

modules/backend/models/User.php
ayumihamsaki: @LukeTowers I'm still confused, going to add your code anyway. If @w20k comment gives us the option to turn gravatar support on/off then that's all good.

ayumihamsaki:
@w20k @LukeTowers Gonna carry on with this tomorrow, my head spinning today, think I got the flu?

Things left to do:

1. Luke to confirm if this commit is ok: https://github.com/octobercms/october/pull/4562/commits/807fa363933d18a68ff0578c3aac9359193b02fb#diff-c9b7cd5fc13b6d1a8cf3c154b91aefc8R63

As per comment: https://github.com/octobercms/october/pull/4562#discussion_r320834188

2. What code line/file will this go into?

> That can be easily fixed by just returning Redirect::refresh() on update.

3. What will be the name of the config property and which config file do you want to use?

> what do you think if the use of gravatar could be a part of user configuration or config (and by default return base64 image, if disabled. And true by default.

> I don't mind it except that it needs to be a database / branding setting, and we still need to support gravatar in the way that I mentioned.

Also are you thinking of just creating a wrapper around the: https://github.com/octobercms/october/blob/8be0b60f23c98b238649ba20c018f703fbd43b57/modules/backend/models/User.php#L125-L132

daftspunk:
Once again, there is simply too much going on in this PR. Please limit changes to one per pull request so we can properly review each change on an individual basis. 

This has been [partially] implemented in b37641e94d4349e6015cdda508eade14b4a882dd

PR #4567, opened 2019-08-25
Include file extension in combiner

alekseyp:
This PR let's asset combiner to generate final URLs that includes file extension.

It will help services that depent on file extension (like CloudFlare) to recognize what kind of file it is.
It will also help CDNs to cache those files.

Looks like no docs needed be updated.


LukeTowers:
Reverted by https://github.com/octobercms/october/commit/7e98e199a46040dbf22b8211d33115ffae780e32

PR #4569, opened 2019-08-26
Remove logging of ApplicationException

### LINKED ISSUE #4568: ApplicationException is logged unexpectedly
- OctoberCMS Build: 458
- PHP Version: 7.2.17

#### Expected behavior
`ApplicationException` should not be logged as stated in [its comment](https://github.com/octobercms/library/blob/master/src/Exception/ApplicationException.php#L5).
Also, `ApplicationException` is in [the list](https://github.com/octobercms/library/blob/master/src/Foundation/Exception/Handler.php#L23) of exceptions not to be logged.

#### Actual behavior
`ApplicationException` is logged.

### Steps To Reproduce:
Create a page that throws an `ApplicationException` like below, and check the log file.

```
url = "/test"

==
function onStart() {
	throw new \October\Rain\Exception\ApplicationException('application exception');
}

==
hello
```

pikanji:
This fix resolves #4568 .

Even though logging is suppressed by [this dontReport list](https://github.com/octobercms/library/blob/master/src/Foundation/Exception/Handler.php#L23), `ApplicationException` is logged by [ErrorHandler](https://github.com/octobercms/october/blob/master/modules/system/classes/ErrorHandler.php#L45).
`ErrorHandler::beforeHandleError` is the one that logging `ApplicationException`.
It is called by `ErrorHandler::handleException`, which is called in the listener of `exception.beforeRender` event registered by [ServiceProvider::registerErrorHandler](https://github.com/octobercms/october/blob/master/modules/system/ServiceProvider.php#L262)

### DIFF
--- a/modules/system/classes/ErrorHandler.php
+++ b/modules/system/classes/ErrorHandler.php
@@ -35,18 +35,6 @@ class ErrorHandler extends ErrorHandlerBase
     // }
 
     /**
-     * We are about to display an error page to the user,
-     * if it is an ApplicationException, this event should be logged.
-     * @return void
-     */
-    public function beforeHandleError($exception)
-    {
-        if ($exception instanceof ApplicationException) {
-            Log::error($exception);
-        }
-    }
-
-    /**
      * Looks up an error page using the CMS route "/error". If the route does not
      * exist, this function will use the error view found in the Cms module.
      * @return mixed Error page contents.


bennothommo:
@LukeTowers my two cents - given that it's mentioned twice that it is not meant to be logged, and it's only be added ad-hoc to the ErrorHandler to be logged, I think it's probably a safe bet it is not *meant* to be logged.

daftspunk:
This looks like it might have been a typo, it should read `SystemException`. Overall October's platform makes inconsistent use of these exceptions likely because of the general confusion around their meaning.

To clarify the difference:

- *SystemException* occurs when something unhealthy happened within the system. It couldn't write to the disk. The database has fallen over. An unhappy path has occurred due to a failure in the running system itself, either by configuration or environment or by a logic error from the programmer. In a perfect world, a System Exception should never happen.

- *ApplicationException* occurs when the user has caused something to go wrong. They've filled out a form incorrectly. They didn't follow a procedure properly. These are treated as basic validation errors usually deriving from an expected path in the software where the application is not being used correctly. In a normal operating environment, Application Exceptions are likely to occur frequently.

It would appear October in general needs an overall audit to ensure the correct use of these exception types is applied.

It would also make sense that `SystemException` is logged, but `ApplicationException` probably shouldn't be logged by default.

daftspunk:
I'm leaving this open because it contains actionable information but the current code itself should not get merged. It ApplicationException should be renamed to SystemException.

LukeTowers:
@daftspunk do you mean rename the exception class itself? I use it all the time in my project's for its intended purpose, telling the user they've done something wrong.

daftspunk:
I mean don't log messages that tell the user they've done something wrong, only log system errors.

Change code to this, but maybe look at the impact of doing so first:
```
    /**
     * We are about to display an error page to the user,
     * if it is an SystemException, this event should be logged.
     * @return void
     */
    public function beforeHandleError($exception)
    {
        if ($exception instanceof SystemException) {
            Log::error($exception);
        }
    }
```

LukeTowers:
Did a quick look through the core, all SystemException uses right now are valid, they're mostly used for cases where either the system trips on something that it can't recover from or as a "developer validation" tool (i.e. perhaps it could recover from it, but the developer should really just follow the error message to fix whatever it is they're doing wrong).

PR #4574, opened 2019-08-26
Update jQuery mousewheel from v3.1.9 to v 3.2.0

ayumihamsaki:
This update fixes all the deprecated jQuery code and adds several features like rate limiting for the mousewheel etc.

Link to this issue: https://github.com/octobercms/october/issues/3867


ayumihamsaki:
@bennothommo Can you please change the label on this, I would like to put a block on this pr and update it again when jQuery have fixed the passive event listener.

daftspunk:
New lib looks almost identical except for the addition of `normalizeOffset` option. Should be safe to merge in its current state.

LukeTowers:
@daftspunk it wasn't merged yet because of the following:

> [update] jQuery team have re-opened my issue and may now consider adding a switch to turn on passive mode for event listeners. In October CMS having the event listeners turned on reduced jank issues and increase performance when I ran some tests. So it would be nice to update this pr in the near future and have the event listeners using passive mode - github issue can be found her: jquery/jquery-mousewheel#202

PR #4576, opened 2019-08-27
Initial unit tests for front-end framework

bennothommo:
## What this does

This is a proof-of-concept for unit tests for our front-end JavaScript libraries and frameworks. This current implementation uses an NPM pipeline to test the `framework.js` file that is made available to the themes.

The pipeline leverages [jsdom](https://github.com/jsdom/jsdom) to create a virtual DOM to test the scripts and HTML in a fake environment, and [mocha](https://github.com/mochajs/mocha/) and [chai](https://github.com/chaijs/chai) to run and assert the test cases.

image

## What's included

At this stage, there's just a single test case file for `framework.js` which can be found in `tests/js/cases/system/framework.test.js`. In addition, a helper module has been set up to quickly create a fake DOM, found in `tests/js/helpers/fakeDom.js`.

## Future applicability

Ideal candidates for future application of this testing framework would be our form widgets that are controlled primarily through JavaScript, such as the repeater and datatable widgets.

## How to test this out

Before starting, you will need Node.js and NPM installed - https://nodejs.org/en/ for more information on this.

1. Checkout the `wip/js-unit-tests` branch.
2. Run `npm i` to install the necessary dependencies as defined in `package.json`.
3. Run `npm run test` to run the test cases. The test cases can be found in `tests/js/cases`.

## Other changes

One minor change was made to the real `framework.js` - changing a redirect from `window.location.href` to `window.location.assign`, so that it could be mocked in the tests (JS properties cannot be easily mocked). This should have no effect on the functionality.


modules/system/assets/js/framework.js
```diff
@@ -267,7 +267,7 @@ if (window.jQuery.request !== undefined) {
              * Custom function, redirect the browser to another location
              */
             handleRedirectResponse: function(url) {
-                window.location.href = url
+                window.location.assign(url)
```
w20k: @bennothommo you're losing IE 11 compatibility with this feature :) 
https://developer.mozilla.org/en-US/docs/Web/API/Location/assign#Browser_compatibility

If you want to test `location.href`, might use it as: 
```
global.window = Object.create(window);
const url = "http://dummy.com";
Object.defineProperty(window, 'location', {
  value: {
    href: url
  }
});
expect(window.location.href).toEqual(url); 
```
Or

```
delete global.window.location;
global.window = Object.create(window);
global.window.location = {
  port: '123',
  protocol: 'http:',
  hostname: 'localhost',
};
```

bennothommo: @w20k I thought that said yes for Internet Explorer? :P

w20k: @bennothommo I think it's edge only not IE11  😄If I'm correct.

bennothommo: @w20k do you have a copy of IE11 to test with? I've only got Edge on my Windows laptop :(

w20k: @bennothommo on Parallels I think have it, not the latest version, but will test this out!

w20k: @LukeTowers I would like to raise one concern, should we support or `say` that we still support IE 11?
I've tried testing it on my parallels and here is how it looks...
Can we officially drop IE11?

_Would say it not usable at all with my IE11 version_
Screenshot 2019-09-03 at 21 48 16
Screenshot 2019-09-03 at 21 48 40
Screenshot 2019-09-03 at 21 48 48

LukeTowers: @w20k afaik it should be working fine on IE 11, ping me again in a couple days if I haven't gotten around to running it on a real computer.

w20k: Will do!

w20k: @LukeTowers ping-ping, IE 11 test run :)

bennothommo: I've been able to confirm that `location.assign()` does exist on IE11.

w20k: Cool @bennothommo, will review your PR one more time, it's good for merging!

w20k:
@bennothommo could you update Travis? I think we could merge this one after a quick review :)

bennothommo:
@w20k All sync'd up :)

w20k:
LGTM, @bennothommo. Awesome work!

PR #4583, opened 2019-08-30
Add translations for plugin author to system module

LarBearrr:
There are some inconsistencies with references to translations when the CMS module is disabled. For example, the install plugins page references the `by_author` translation in the CMS module which will result in the value not being translated if the CMS module is not installed. 
image

This PR adds a `by_author` key to the system modules lang files where there was a value available in the CMS module lang files. The translation key has also been updated in the `modules/system/controllers/updates/_install_plugins.htm` partial to use the new key.

image


w20k:
@LarBearrr could you copy every string there is? :) 

Made a list for you 😉 
- [ ]  `ca/lang.php`
- [ ]  `ru/lang.php`
- [ ]  `en/lang.php`
- [ ]  `cs/lang.php`
- [ ]  `fa/lang.php`
- [ ]  `sk/lang.php`
- [ ]  `et/lang.php`
- [ ]  `th/lang.php`
- [ ]  `uk/lang.php`
- [ ]  `tr/lang.php`
- [ ]  `vn/lang.php`
- [ ]  `pt-br/lang.php`
- [ ]  `zh-cn/lang.php`
- [ ]  `fi/lang.php`
- [ ]  `hu/lang.php`
- [ ]  `kr/lang.php`
- [ ]  `nl/lang.php`

LarBearrr:
@w20k How do you mean "every string?" Just the one in question, or _every_ CMS string?

w20k:
@LarBearrr sorry about that :) Just double-check if you have copied everything and I didn't miss something in my list. Thanks!

LarBearrr:
I can't edit your list, but it appears there is no `th` in the system module at all.

w20k:
@LarBearrr wait a sec, I'm losing my mind 😄 Could you correct me if I'm wrong))
Haven't you just copied those lines (`theme.by_author`) from the `cms/lang/{lang}/lang.php` to `system/lang/{lang}/lang.php`?

`th` is here:
https://github.com/octobercms/october/blob/master/modules/cms/lang/th/lang.php#L27

LarBearrr:
@w20k Are you on slack right now?

w20k:
@bennothommo @LukeTowers if you have a minute, could you peer review! Thanks ;) 

Tested, and double-checked keys (list ^).

bennothommo:
@w20k @LarBearrr Looks fine to me - I've noticed that `cms::lang.theme.by_author` is also used here: https://github.com/octobercms/october/blob/master/modules/system/controllers/updates/_install_themes.htm#L50 - not sure if that's going to need changing too?

LarBearrr:
Well, I don’t think so in that case because you’re installing a theme which is only used by the CMS module, so it would have to be installed. Let me know though.

bennothommo:
@LarBearrr as far as I can tell, that tab in the Updates section is still available even if the CMS module is disabled.

LukeTowers:
@bennothommo I think it's fine for now, really that section should be handled by the CMS module period but that's a PR for another day.

Issue #3856, opened 2018-10-11
Improve Tab UX for large numbers of tabs and smaller screen sizes


ayumihamsaki:
Using the code from this webpage: https://octobercms.com/docs/ui/tab

I will show the following screenshots and errors:

## Master tabs

![1](https://user-images.githubusercontent.com/17784082/46809999-a5b18400-cd67-11e8-90ea-e2bb50977280.png)

The CSS is NOT broken, however the User Experience seems rather poor to scroll the tabs.

## Primary tabs

![2](https://user-images.githubusercontent.com/17784082/46810182-fcb75900-cd67-11e8-8893-894eddfdb9d1.png)

The CSS is BROKEN (See the border-bottom CSS is not working correctly), also the User Experience seems rather poor to scroll the tabs.

## Secondary tabs

![3](https://user-images.githubusercontent.com/17784082/46810370-70f1fc80-cd68-11e8-8b15-cc5c53cb1a16.png)

The CSS is NOT broken (although I am not a fan of having the text up against the column), however the User Experience seems rather poor to scroll the tabs.

## Content tabs

![4](https://user-images.githubusercontent.com/17784082/46810543-d0e8a300-cd68-11e8-8811-a2ac66f7e23b.png)

The CSS is BROKEN (See the border-bottom CSS is not working correctly), also the User Experience seems rather poor to scroll the tabs.

### Conclusion

The Primary and Content Tabs have broken CSS.

All the Tab Designs seem to have poor UX.

Could we not have a proper tab solution something like this: https://codyhouse.co/demo/responsive-tabbed-navigation/index.html

## Desktop Version:

![5](https://user-images.githubusercontent.com/17784082/46810738-3e94cf00-cd69-11e8-8615-4aa477757ca6.png)

## Tablet Version

![6](https://user-images.githubusercontent.com/17784082/46810754-46547380-cd69-11e8-8631-c82779776316.png)

## Phablet Version

![7](https://user-images.githubusercontent.com/17784082/46810766-4d7b8180-cd69-11e8-806e-6818cc03b396.png)

## Mobile Version

![8](https://user-images.githubusercontent.com/17784082/46810781-54a28f80-cd69-11e8-88ce-06e551edbb74.png)

LukeTowers:
@ayumihamsaki I like the suggestion. Submit a PR and I would be happy to review it.

ayumihamsaki:
@LukeTowers 

[...]

PR #4587, opened 2019-09-01
Fix the Primary tabs overflow error

ayumihamsaki:
I first reported this way back here: https://github.com/octobercms/october/issues/3856

This is the error:

image

This is the fixed version:

image


daftspunk:
Nice!

daftspunk:
Note this had to be rolled back because it caused the draggable area to always overflow. We will need to find a better solution.

PR #4578, opened 2019-08-28
Allow checkbox's to have checked state

ayumihamsaki:
This fixes the coversation found here: https://github.com/octobercms/october/issues/4573

Please don't close that issue when this pr is merged as I haven't even started to do the work on checkbox's and checkbox lists.

Please merge this asap as I think this carries a 'high priority' label.

Also going to update the doc's, pr link here: x

## Test results:

```yaml
fields:
    show_content1:
        label: 'Display content'
        type: checkbox
        default: true
    show_content2:
        label: 'Display content'
        type: checkbox
        default: false
    show_content3:
        label: 'Display content'
        type: checkbox
        default: 0
    show_content4:
        label: 'Display content'
        type: checkbox
        default: 1
    show_content5:
        label: 'Display content'
        type: checkbox
        default: null
```

Outputs:

image

Issue #4573, opened 2019-08-26
Fix Checkbox and Checkbox list fields and add the missing example code in the doc's


ghost:
According to the W3c there are two types of checkboxes these are:

### 1. (Two State) Checkbox

Here we have a single checkbox example:

![ezgif-2-6fb0fa627d42](https://user-images.githubusercontent.com/17784082/63714262-59651a80-c839-11e9-9af2-818ca3f0166b.gif)

The doc's can be found here: https://octobercms.com/docs/backend/forms#field-checkbox

A working example can be found here: https://octobercms.com/docs/ui/checkbox

### 2. (Tri State) Checkbox List

Here we have a checkbox list example:

![ezgif-2-fe59d9a75021](https://user-images.githubusercontent.com/17784082/63714641-148db380-c83a-11e9-9a4a-04fa7765a02a.gif)

The doc's can be found here: https://octobercms.com/docs/backend/forms#field-checkboxlist

A working example can be found here: **_DOES NOT EXIST!_**

### Problems with October's Checkbox List

The problem is that October CMS for it's Checkbox List is actually using a (Two State) and not a (Tri State) method I tricked you in the above example!

Let's now show you a proper (Tri State) method:

![ezgif-2-b2c6c24e7347](https://user-images.githubusercontent.com/17784082/63714952-b0b7ba80-c83a-11e9-901b-af61b0571781.gif)

Did you spot the difference?

Let me explain, let's say there are 50 checkbox's in the Checkbox list how long would it take your Users/Clients before they finished selecting all the checkbox's in the checkbox list? Even more important, how long would it take before they started to get annoyed?

A (Tri State) method allows your Users/Clients to be able to select all the checkbox's, de-select all the checkbox's and allow their custom selection.

### Solutions

October CMS needs to do three things:

1. Add a working Checkbox list example to their doc's.

2. Change the Checkbox list from a (Two State) into a (Tri State) method.

3. Add WAI-ARIA to the checkbox and Checkbox list fields.

36864:
Just a side note, checkbox list already allows selecting/deselecting all entries by setting `quickselect` to true. `quickselect` is true by default on lists larger than 10 elements. It's not as fancy as your example, but it gets the job done:

![image](https://user-images.githubusercontent.com/7127588/63763932-9ecb2b80-c8bd-11e9-9cbc-c495d240865a.png)

[...]

PR #4401, opened 2019-06-21
Add Tab WAI-ARIA Support

ayumihamsaki:
Adds support for WAI-ARIA to October's Tabs, I not tested this yet, but then code is pretty simple and I think I can't get this wrong (I hope).

Relates to github issue: https://github.com/octobercms/october/issues/4400

## Summary of features this PR adds (for release notes)

1. Added WAI-ARIA to the layout templates files.
2. Added Focus-ring polyfill to allow better keyboard navigation and screen reader support.
3. Added WAI-ARIA to tabs and created keyboard arrow controls for tabs.
4. Added `event.key` polyfill to allow removal of deprecated `event.keyCode`, `event.charCode` and `event.which`.
5. Fixed all the whitespace, tabbing and indent issues with all the lang files.
6. Keyboard navigation now allows the following key actions: `Tab`, `Up Arrow`, `Down Arrow`, `Left Arrow`, `Right Arrow`, `Home` and `End` buttons.
7. Added keyboard navigation for tab panels.

PR #4588, opened 2019-09-01
Merge Order 8: Add Keyboard navigation forcheckbox and checkbox lists

ayumihamsaki:
Linked to github issue: https://github.com/octobercms/october/issues/4454 and https://github.com/octobercms/october/issues/4573

Here is a short youtube video showing it in action (when a user clicks or presses a key the dom will update), see here: https://youtu.be/xO60cdkN7Eo

Keyboard Navigation according to W3c is as follows:

image

I have added some extra controls because I find that setup a bit confusing, so you can either press '`Enter`' or '`Space bar`' to select a checkbox value.

### A cleaner focus ring

I don't like how browsers work with checkboxes and focus rings. To me the way they process checkboxes when a user switches between keyboard controls and mouse clicking is no good.

For example (normal browser) a user selects a checkbox using keyboard controls via the '`Tab`' key, which displays a focus ring. A user then clicks the same checkbox with a mouse and the focus ring is still there. Doesn't look nice at all.

My version is different, a user selects a checkbox using keyboard controls via the '`Tab`' key, which displays a focus ring. When a user clicks the same checkbox with a mouse the focus ring is hidden. Select it with 'Tab' and the focus ring comes back. This is a more tidy method.

### Testing this PR

You need to install two other PR's to add the required code before you can test this PR.

The following PR's should be install before this PR: https://github.com/octobercms/october/pull/4401 and https://github.com/octobercms/october/pull/4578

---

Lastly, this is for admins.

Please can you add this PR to the '`Accessibility Enhancements`' projects list, e.g. https://github.com/octobercms/october/projects/9

Thanks.


modules/system/assets/ui/js/waiaria.js
```diff
@@ -0,0 +1,324 @@
...
+            'aria-checked': 'true',
+            'tabindex': '0',
+            'checked': true
+        });
+    });
+
+    /*
+     * Slider switch buttons
+     */
+    $('body').on('change', '.switch-field input[role="checkbox"]', function () {
```
w20k: @ayumihamsaki you lost `event` var here.

modules/system/assets/ui/js/waiaria.js
```diff
@@ -0,0 +1,324 @@
...
+            });
+        }
+    });
+    
+    /*
+     * Turn off custom tab focus ring
+     * Important - 'keydown' selects the form and 'keyup' selects the form field.
+     * 'keydown' must be used.       
+     */
+    $('body').on('click keydown', '.form-widget', function() {
```
w20k: Here as well ;)

modules/system/assets/ui/js/waiaria.js
ayumihamsaki: @w20k thanks 👍

daftspunk:
This should have been addressed in 8471c7305488a71413aaf6e68efdbf57618b3d8d

Checkboxes now have keyboard support

ayumihamasaki2019:
@daftspunk 8471c73 is for checkbox - this pr is for checkbox lists (two different widgets)

PR #4598, opened 2019-09-05
Fail CSRF token checks if the session expires.

### LINKED ISSUE #4595: Error if session expired
- OctoberCMS Build: 458
- PHP Version: 7.2
- Database Engine: mysql

### Description:
If the login page opened a long time, and session expired, there is error 
ErrorException: hash_equals(): Expected known_string to be a string, null given in /modules/backend/classes/Controller.php:795

bennothommo:
Fixes #4595.

If the session expires, we cannot do a valid CSRF token check, so this accounts for that possibility. I'm pretty certain this fix should work, but I've made it a PR in case there's some fringe case I'm not aware of.

### DIFF
--- a/modules/backend/classes/Controller.php
+++ b/modules/backend/classes/Controller.php
@@ -786,7 +786,7 @@ class Controller extends ControllerBase
 
         $token = Request::input('_token') ?: Request::header('X-CSRF-TOKEN');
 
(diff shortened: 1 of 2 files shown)


LukeTowers:
LGTM!

PR #4599, opened 2019-09-05
Update the checkbox list widget

ayumihamsaki:
Related to this github issue: https://github.com/octobercms/october/issues/4573

### Old design

image

### New design (normal)

image

and on mobile devices (below 435px)

image

### New design (nested)

image

I hope you can see that the theme is then the same throughout the new design.


ayumihamsaki:
Changed buttons to default now look like this:

image

PR #4600, opened 2019-09-05
 Add a working checkbox list doc code example to october website 

ayumihamsaki:
Relates to this girhub issue: https://github.com/octobercms/october/issues/4573

Currently there is no checkbox list code example on the october cms website. This PR adds a working and tested code example.

See screenshot results from test plugin:

image

- Have added this code to the test plugin as well, see commit: ---


bennothommo:
@ayumihamsaki Hmmm, I'm not personally a fan of the buttons look, just purely because it becomes the focal point of the field. The focal point should be the checkboxes as they're the most important bit, hence why I think it was designed previously as simple text links.

ayumihamsaki:
@bennothommo any suggestion, because the old method completely doesn't work for touch screens and fails the tap-target rule. Happy to change it as long as people using a touch screen can us the features.

ayumihamsaki:
> Maybe the selection buttons should be in the secondary colour ( same as the Save and Close button )

Credit: seanthepottingshed 

New screenshot:

image

ayumihamsaki:
@bennothommo Have updated the code and first comment, please re-review, thanks.

PR #4604, opened 2019-09-06
Speed up the loading of the plugin assets by adding preload to them

ayumihamsaki:
Can be related to this issue: https://github.com/octobercms/october/issues/4379

When a user adds some assets in the formwidget folder in a plugin the assets are loaded in the `<head>` section.

Because of this there is no reason not to preload those assets at the same time to reduce the loading time.

### Before

image

### After

image


daftspunk:
Not sure about this. This only makes sense if we're loading the JS at the bottom of the file, which we're not. The change essentially makes the page load slower because there is more markup on every page.

PR #4606, opened 2019-09-06
Add missing Checkboxlist before and after hints

ayumihamsaki:
I don't understand why but Daftspunk has added a checkbox list in the form examples with the HTML code:

```html
<p class="help-block before-field">What cars would you like in your garage?</p>
```

Yet there is no matching YAML code?

Very strange!

This PR adds the missing code and allows a user to add a `before` and `after` hint to their Checklist widgets. No styling has been added as the code is already in October.

Here is an example screenshot:

image

YAML API code is as follows:

```json
hintBefore: 'hint displayed at the top'
hintAfter: 'hint displayed at the bottom'
```


daftspunk:
`hintAfter` and `hintBefore` are not valid configuration items. You may be thinking of `commentAbove` and `comment`.

ayumihamasaki2019:
@daftspunk 

See the October website link here: https://octobercms.com/docs/ui/form

Open the '**Complete example**' source code and you will see your example code at the bottom under **_Checkbox List_** section, the code line:

```html
<p class="help-block before-field">What cars would you like in your garage?</p>
```

See you wrote `class="before-field"` maybe the example code is wrong on the October website? But the strange thing is that it works in the backend with the styling.

> hintAfter and hintBefore are not valid configuration items. You may be thinking of commentAbove and comment.

Yah I think you are right, however it's not mentioned in the doc's found here: https://octobercms.com/docs/backend/forms#field-checkbox

bennothommo:
@ayumihamasaki2019 It is mentioned in the [Field options](https://octobercms.com/docs/backend/forms#form-field-options) section, which is meant to apply to all field types.

Issue #3839, opened 2018-10-03
Cloudflare issue after updating to latest octobercms 443 build.


zakir1929:
I updated my octobercms from 437 to 443 build.
and now the header menu is not showing up in the way it was showing before and flash messages for add/edit/delete of any record is not showing up in backend only.

##### Expected behavior
![image](https://user-images.githubusercontent.com/16069667/46395039-abc0b880-c708-11e8-8062-8087b5b130fb.png)

##### Actual behavior
![image](https://user-images.githubusercontent.com/16069667/46395077-cabf4a80-c708-11e8-9b2f-64e2a082d5df.png)

and in browser console I can see this
![image](https://user-images.githubusercontent.com/16069667/46395101-e4f92880-c708-11e8-92fc-0d40d85d2683.png)

##### Reproduce steps
just update to latest 443 build and have cloudflare installed on it.

##### October build
443

LukeTowers:
Turn off CloudFlare performance settings on your backend: https://github.com/octobercms/october/issues/3602#issuecomment-399692351

LukeTowers:
Actually, I'll try adding onto https://github.com/octobercms/october/commit/f068b13851150c6d12f4f2949497e21b980b7f15

LukeTowers:
@zakir1929 https://github.com/octobercms/october/commit/29541394a3fc773f498c2d6c142302ca2cb927d2 should fix it in theory.

zakir1929:
Thankyou @LukeTowers  this problem has been solved :) .

but there is a new problem now. 
The editor in whole website is not showing up
![image](https://user-images.githubusercontent.com/16069667/46490029-80cb8700-c824-11e8-8a35-fc63c3c90d28.png)

LukeTowers:
@zakir1929 it sounds like CloudFlare just isn't playing nice. I would recommend you just disable the performance on the entire backend now.

LukeTowers:
@zakir1929 we will use https://github.com/heathdutton/cloudflare/pull/4 & https://github.com/octobercms/october/commit/e246427463ac5edcc6b7115b33838a4367a4f9a9 to fix this issue going forward.

PR #4611, opened 2019-09-10
Fix CloudFlare's Rocket Loader break the script loading

serkanbektas:
Cloudflare Rocket Loader break the script loading, so many plugins (as RTE, Media Manager) are nor working. **data-cfasync** option set **false** applied for backend plugins.

It also used on #4092 , #3841, and #3839

I have expanded it to other plugins.

### DIFF
--- a/modules/backend/layouts/_head.htm
+++ b/modules/backend/layouts/_head.htm
@@ -30,7 +30,7 @@
 ?>
 
 <?php foreach ($styles as $style) : ?>
-    <link href="<?= $style . '?v=' . $coreBuild; ?>" rel="stylesheet" importance="high">
+    <link data-cfasync="false" href="<?= $style . '?v=' . $coreBuild; ?>" rel="stylesheet" importance="high">
     <link href="<?= $style . '?v=' . $coreBuild; ?>" rel="preload" as="style" importance="high">
 <?php endforeach; ?>
 
@@ -49,11 +49,11 @@
              // Unregister all service workers before signing in to prevent cache issues, see github issue: #3707
              navigator.serviceWorker.getRegistrations().then(
                  function(registrations) {
-                     for (let registration of registrations) {  
+                     for (let registration of registrations) {
                          registration.unregister();
                      }
                  });
-         }		
+         }
     </script>
 <?php endif; ?>
 
--- a/modules/system/traits/AssetMaker.php
+++ b/modules/system/traits/AssetMaker.php
@@ -3,6 +3,7 @@
 use Url;
 use Html;
 use File;
+use Event;
 use System\Models\Parameter;
 use System\Models\PluginVersion;
 use System\Classes\CombineAssets;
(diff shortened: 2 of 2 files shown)


w20k:
@LukeTowers haven't we had the same issue already?

serkanbektas:
> @LukeTowers haven't we had the same issue already?

I put the related pull request links. #4092 is not enough for the solution. File Manager, Media Manager, Rich Text Editor is effected by that. I added cache false for all of the backend javascript files which not need to be there. Backend assets are sent in AJAX framework responses so that's need to be cached.

LukeTowers:
@w20k I want to just make the default be cache="false" so we don't have to add it to every single backend script.

LukeTowers:
@serkanbektas ping me if I don't get that done by tomorrow.

bennothommo:
@serkanbektas @LukeTowers would it be easier to just change the following line:
https://github.com/octobercms/october/blob/e2f23eb18492219e58c96bcbfd152f92d1f49a17/modules/system/traits/AssetMaker.php#L132
to use the following:
```php
        if (!isset($attributes['cache']) || $attributes['cache'] === 'false') {
```
There's no real harm in adding the `data-cfasync` attribute to every JS file, and let people override it if required. Anyone not using Cloudflare will ignore the attribute.

LukeTowers:
@serkanbektas can you implement @bennothommo's suggestion please

bennothommo:
@LukeTowers I've had a re-think about this.

I propose instead of deploying a blanket fix for CloudFlare, we instead roll out a couple of events to AssetMaker that allow modification of the attributes for asset files - eg. `system.assets.modifyAttributes($name, $type, $attributes)` - and handle this via a plugin.

The main reasoning is that CloudFlare is not the only CDN (albeit it's one of the most well known), and other CDN providers might have different attributes that achieve the same thing. I personally haven't yet needed to deploy with CloudFlare, and I'm thinking although it's harmless having the attribute in there, I'd still prefer it was optional.

LukeTowers:
@bennothommo I'm not opposed to that, although we'd want to get https://github.com/heathdutton/cloudflare to implement it as soon as we add the event so it can be ready to go.

LukeTowers:
@heathdutton do you have any input on the above?

heathdutton:
I haven't been maintaining said plugin, but I'm open to PRs. I say go for it.

LukeTowers:
@serkanbektas please implement @bennothommo's suggestion of firing an event (`system.assets.modifyAttributes($name, $type, $attributes)`) instead and add a PR to https://github.com/heathdutton/cloudflare to listen to that event and do the modification there.

serkanbektas:
Thanks to all for help.

I modified my PR as @bennothommo said and opened a PR at @heathdutton’s Cloudflare’s plugin repo to listen new event.

serkanbektas:
This is another PR for event listening https://github.com/heathdutton/cloudflare/pull/3

LukeTowers:
Replacing this with https://github.com/octobercms/october/commit/e246427463ac5edcc6b7115b33838a4367a4f9a9

LukeTowers:
@serkanbektas @heathdutton @bennothommo we will use https://github.com/heathdutton/cloudflare/pull/4 & https://github.com/octobercms/october/commit/e246427463ac5edcc6b7115b33838a4367a4f9a9 to fix this issue going forward.

PR #4615, opened 2019-09-11
Use composer to set build version

Samuell1:
This allows to set build depending on installed version in your composer.


LukeTowers:
If the version number is only used for display (and not for update checking) then use the full version number. If not, then use the same format as a regular update would.

Side note, we need to look for the `october/system` package, not `october/cms` as both `cms` and `backend` have the potential of not being present on a given installation; technically speaking the only required packages are `october/rain` and `october/system`.

modules/system/classes/UpdateManager.php
```diff
@@ -365,21 +365,29 @@ public function uninstall()
     }
 
     /**
-     * Asks the gateway for the lastest build number and stores it.
+     * Asks the composer/gateway for the lastest build number and stores it.
```
bennothommo: @Samuell1 Please change the comment to the following:
```suggestion
     * Asks the gateway (or Composer) for the latest build number and stores it in the database.
```

modules/system/classes/UpdateManager.php
```diff
@@ -365,21 +365,29 @@ public function uninstall()
...
      * @return void
      */
     public function setBuildNumberManually()
     {
-        $postData = [];
+        // Use installed.json to detect latest version
+        $composerInstalled = base_path('vendor/composer/installed.json');
```
bennothommo: @Samuell1 Any reason for using this file over using `composer.lock` in the root folder?

modules/system/classes/UpdateManager.php
```diff
@@ -365,21 +365,29 @@ public function uninstall()
...
+        if (file_exists($composerInstalled)) {
+            $packages = json_decode(file_get_contents($composerInstalled), true);
```
bennothommo: @Samuell1 I'm concerned we will run into memory issues reading this file entirely, as Composer dependencies could be huge depending on the number of plugins used and various other factors. Given that this command is run in a command-line environment, would it be safe enough to just run `composer show october/system` through `shell_exec` and get the version that way?

bennothommo (changes requested):

modules/system/classes/UpdateManager.php
prhost: @bennothommo would depend on having composer installed, not all october environments or projects use composer, including the october installer itself does not use composer.

bennothommo: @prhost the normal "set build" functionality is underneath this code block, and still works for non-Composer based installations.

prhost: I meant that running the `composer show october/system` command via shell will depend on whether OS has composer bin installed.

bennothommo: @prhost True, but I think it's reasonable enough to assume that the binary is installed if the `vendor/composer` folder exists. We could fall back to reading the installed.json file if it doesn't.

Samuell1: installed trough ocms installer and vendor with installed.json is there

prhost: I would particularly vote to create a metadata or constant with the october version in package `october/system` and commit and keep it up to date when releasing a version.

What do you think? Because there are tools I for example step into my project to eliminate unwanted waste and files, making the project lighter when I create an artifact or release and go into production. Reading this into a json file is creating an unnecessary dependency.

Samuell1: I dont know but installed,json is internal composer file that is needed for composer to work idk if any tool can remove it.

bennothommo: @Samuell1 thanks for checking the installer. Would you mind testing the memory usage of just decoding the JSON file as you had it before, with a whole heap of dependencies installed? If the memory doesn't become too high, I'm fine with leaving it as is.

@prhost If, for example, you delete the `installed.json` file, the system will simply use the old method of pinging the October server for the build number. There's no explicit dependency on the `installed.json` file - it's just a shortcut for determining the build number if it's available.

Samuell1: @bennothommo yes

Its not only shortcut but exact version of your ocms installed locally. I seen many times issue that version installed locally wasnt same as set in administration.

prhost: This is exactly why the build version should be bundled with some file in the repository. and @Samuell1 does work october without installed.json, what doesn't work is composer, but in production environment you don't need composer, so I'm talking about dependency.

A `define("OCTOBER_VERSION", 420);` like laravel does would already solve all that.

But this is all my opinion, @bennothommo decides whatever he wants (:

Samuell1: You cant run ocms without vendor folder.

bennothommo: While I'm not opposed to a constant defining the version, it does mean one extra manual step for us when pushing out a new version of October. I would prefer if it was automated in some way.

Samuell1: Just tested that json_decode it spikes to 18MB of ram :D

prhost: > You cant run ocms without vendor folder.

Not what I meant hehe. october does not depend on composer, it depends on its vendor dependencies, the installed.json file is not a file that october or vendor depends on, deletes and your application will work normally.

But I agree with @bennothommo , updating constant is yet another process when the release is released. But the form that is currently not the correct one, as the query to webservice will always return the last version released, which may not be the current correspondent.

Perhaps the opinions of others could help.

Samuell1: It depends how it can be most simple from both sides implemented.

Samuell1: Any ideas @bennothommo what to use to prevent big memory usage from parsing that json?

w20k: Use non-memory based parsers 😄 `json_decode` -> get everything inside of memory and proceed.
Like: https://github.com/halaxa/json-machine or https://github.com/salsify/jsonstreamingparser

Samuell1: @w20k does exists something in ocms like that without needing to add another package?

w20k: The only way you could do it is a raw string parsing (regex or by char). 
I don't think there is an option out of the box in PHP (might be wrong, brain exploded from a different language I'd to use in past few weeks).

LukeTowers: @Samuell1 18MB should be fine if that's all that's happening in the request, most environments won't have anything less than 128MB allotted.

Samuell1: @LukeTowers that size depends on size of that json, but that 18 is just once and it should be fine, but maybe ben knows better solution.

bennothommo: @Samuell1 Yeah I'm happy with 18MB. Even if it was 4 times the size, we're looking at maybe 72MB of memory usage, which is not too bad. If it becomes common, we'll look at some form of streaming or another method of getting the version.

bennothommo (approved):

bennothommo:
@Samuell1 I've just ran a test of this on Composer, and I've noticed that saving the build as `1.0.<build>` is making October think it is completely out-of-date - my guess is the Updates & Plugins area is interpreting `1.0.x` as version 1.

See here:
Screenshot from 2019-09-26 21-52-50

I quickly made an edit to just save the last number in the version as the build, as it would receive it from the October server, and it seems to work correctly:

Screenshot from 2019-09-26 21-53-50

So either, you will need to save just the last number of the version as the build, or you'll need to update the Update checker to interpret the full version correctly. I'll post the change I made as a review.

modules/system/classes/UpdateManager.php
```diff
@@ -365,21 +365,29 @@ public function uninstall()
...
+            $packageIndex = array_search('october/system', array_column($packages, 'name'));
+            $build = str_replace('v', '', array_get($packages[$packageIndex], 'version', 420));
```
bennothommo: @Samuell1 Suggested change to get just the final part of the version as the build:
```suggestion
            $version = str_replace('v', '', array_get($packages[$packageIndex], 'version', '1.0.420'));
            $build = substr($version, strrpos($version, '.') + 1);
```

LukeTowers:
@bennothommo it might be worthwhile to fix the UpdateManager to process the entire version number correctly, just in case we ever do decide to get on the semantic versioning wagon.

bennothommo:
@Samuell1 Did you want to have a crack at @LukeTowers suggested change, or would you like me to?

Samuell1:
@bennothommo Yes you can but we need mostly change from @daftspunk to change *ping* api to return full ocms version.

bennothommo:
@Samuell1 @LukeTowers I've taken another look at this, and there's another edge case I've found. This change does not handle working on branches in Composer.

For example, if you use the Laravel 6 branch, the version in `installed.json` is recorded as `dev-wip/laravel-6`, which doesn't correlate to a build - same as if someone uses `dev-develop` to get the latest code from GitHub. While it might be prudent to, in these cases, retrieve the build number from the server, it'll almost certainly be the wrong build number.

I'm almost wondering if @prhost's suggestion to have the build number stored as a file (perhaps in the `system` module) and deprecating/removing all other methods of retrieving the build number is perhaps the best way forward. Sure - it'll be an extra step for the maintainers, but it allows us to have correct build numbers no matter the installation type.

LukeTowers:
@bennothommo hmm, I'm not really a fan of that since it has the same issue as just using composer in the first place, inaccurate reporting of branches.

`composer show october/backend` shows the currently installed commit hash via the source property, could we utilize that?

Samuell1:
@bennothommo isnt that because laravel branch doesnt have tagged version in github and ocms system?

bennothommo:
@LukeTowers hmm, the hash probably isn't useful in this scenario, since those too don't correlate to a build. Realistically, the only reason we need the build at all is for doing updates via the October server (and for troubleshooting).

@Samuell1 that's correct - I'm just making it known that we can't associate a build number from a branch.

Samuell1:
@bennothommo i think this cant be solved with using this weird versioning system because everything should be tagged as version even branches like prelease or alpha/beta testing.

prhost:
Example how Laravel does: https://tecadmin.net/check-laravel-version/

LukeTowers:
@prhost lol, they're not even automating it: https://github.com/laravel/framework/commit/b2457b1ddb9c59396313eae4e948bdc5fa5251db. Good to know it's just as painful to figure out what version of Laravel is being used as it is for October.

Samuell1:
@LukeTowers thats cool then we know how to do it now :D

LukeTowers:
@Samuell1 that wasn't a vote for their method, it's pretty much as bad as our own right now 😛

prhost:
what do you think of this library? https://github.com/antonioribeiro/version would it be too much to use something similar to that?

LukeTowers:
@prhost I'm looking for something that can be more automated / reliant on the actual code present itself rather than a manually updated file. One of ideas that has been tossed around before is computing a hash for the october modules directory structure and mapping those hashes to version numbers on the marketplace side. That would be ideal in my books so long as it's not too performance intensive since the version display functionality would be able to detect modified installations of October.

prhost:
I think the safest alternative would be to version this, I believe you have a sequence of tasks to perform before launching a new build, put this task of setting the build version in the constant.

bennothommo:
I reckon we make working out a solution to this a little easier by assuming that someone working on a branch has *no* build number (and implicitly disabling `enableCoreUpdates`). As I mentioned before, the build is only used for querying updates directly from the October servers, and it's very, very unlikely that someone working on a branch is going to want updates from the server.

So, the only people who need build numbers are people who use the wizard installer, or people who use the standard Composer method.

LukeTowers:
I just did a quick benchmark on this, and recursively hashing every single file contained under `~/modules` takes approximately a second using the following code:

```php
use App;
use October\Rain\Halcyon\Datasource\FileDatasource;

$eta=-hrtime(true);

$files = new FileDatasource(base_path('modules'), App::make('files'));

$paths = $files->getAvailablePaths();

foreach ($paths as $path => &$hash) {
  $hash = md5($path . file_get_contents(base_path('modules/'.$path)));
}

$eta+=hrtime(true);

// Convert to microseconds
echo $eta/1e+6;
```

I think for the level of accuracy this would give us it's well worth the 1s processing cost especially when it only has to be calculated on demand, whenever the system is trying to verify it's currently installed version. Thoughts on this @bennothommo @prhost @Samuell1? 

Additionally we could extend this code a bit to do some pretty fancy stuff when it comes to detecting modifications to core files (or even plugin files). We could even utilize this to generate hashes for every previous version of October to have a tinker snippet that could be run in the console to identify the version of October when coming across a version older than whatever version we'd include this logic in.

prhost:
@LukeTowers It sounds like a good idea, but if any changes to modules occur, whether on purpose or unintentionally as a self-indentation from IDE, one more space, would that be a problem?

LukeTowers:
@prhost that's one of the features of it, if anything is changed in the modules folder then you now have a "modified" version of October and we can just tell you to fix your install before spending time trying to debug why an installation is broken even though it was modified in an unsupported manner.

bennothommo:
@LukeTowers I'm quite suprised it was that fast - I had thought about that piece of code prior and assumed it would take at the very least a number of seconds.

I like the idea - perhaps the only suggestion I would make is strip out all newlines when doing the MD5 sum, as I would imagine that's the *only* thing that an OS might impose on files.

Samuell1:
@LukeTowers Now i was thinking but this only hashes paths of files no? What if you change only content of file.

LukeTowers:
@Samuell1 this hashes paths and content. 

@bennothommo do different OS's actually change file contents to force a different newline characters or is it just the defaults that are used when creating new files?

bennothommo:
@LukeTowers not usually, but I've seen it happen. It's more of a precautionary suggestion.

LukeTowers:
That might hurt the performance quite a bit @bennothommo but you're welcome to give it a try.

Samuell1:
@LukeTowers oh i see it right now, maybe you can use hash_file method https://www.php.net/manual/en/function.hash-file.php and there is benchmark even with algoritms maybe something will be faster and better

LukeTowers:
@Samuell1 tested it out and performance between the existing code and `hash_file()` instead was pretty much exactly the same regardless of the hashing algorithm used. About a third of the second on average is spent just scanning the directory for paths.

Samuell1:
Any news on this? @LukeTowers

bennothommo:
@Samuell1 @LukeTowers I've done up a proof-of-concept of an October CMS manifest, which would allow one to generate an update manifest and then compare their installation against the manifest to determine the build.

https://github.com/bennothommo/october-manifest

I've done it outside of the October architecture for now, as I would imagine such a facility would need to be run on the October CMS marketplace servers, but if it's worth pursuing, I'll implement the functionality into October.

LukeTowers:
@bennothommo that looks incredible! Any reason why it doesn't include the CMS module?

bennothommo:
@LukeTowers it was so people who didn't have the CMS module installed could still do a comparison.

LukeTowers:
@bennothommo Are we able to dynamically require modules based on what's present in cms.loadModules?

bennothommo:
@LukeTowers I'll see what I can do :).

LukeTowers:
Closing in favour of #5087

PR #4626, opened 2019-09-18
Polish locale update

Magiczne:
Hi!
I've updated polish localization files to include newest updates.
I've corrected some mistakes too and removed old unused strings.


Magiczne:
Actually why? Spacing aligment is now the same as in e.g. english localization file, shouldn't it be unified?

LukeTowers:
@Magiczne if they were to be unified I'd prefer they were unified by changing all of the arrays in October to have aligned values, but that's not going to happen so there's no point in making broad changes like that to just some files. As to why formatting changes like that aren't allowed it's because it clutters the history of the file and makes it unnecessarily difficult for me to review what's actually being changed about the code.

PR #4635, opened 2019-09-25
update zh-cn translations of backend

everyx:
improve translate and add some missing translation items.


modules/backend/lang/zh-cn/lang.php
```diff
@@ -553,7 +553,7 @@
         'label' => '媒体查找器',
         'default_prompt' => '点击 %s 按钮查找媒体项'
     ],
-    'media' => [
+   'media' => [
```
LukeTowers:
```suggestion
    'media' => [
```

PR #4651, opened 2019-10-02
Add event before sending backend user invite email

rejuvenatedigital:
Adds ability to extend data passed to backend user invite email

### DIFF
--- a/modules/backend/models/User.php
+++ b/modules/backend/models/User.php
@@ -151,6 +151,8 @@ class User extends UserBase
             'password' => $this->getOriginalHashValue('password'),
             'link' => Backend::url('backend'),
         ];
+        
+        Event::fire('backend.user.beforeSendInvite', [$this, &$data]);
 
         Mail::send('backend::mail.invite', $data, function ($message) {
             $message->to($this->email, $this->full_name);


rejuvenatedigital:
> There's already a couple events that you can use: https://octobercms.com/docs/api/mailer/beforesend, https://octobercms.com/docs/api/mailer/preparesend

Hi Luke, I've tried to use both of these but they won't let me add to the $data array, am I doing something wrong here? 

`Event::listen('mailer.beforeSend', function ((string|array) $view, (array) $data, (\Closure|string) $callback) {
    $data['test'] = 'testing';
    return true;
});`

LukeTowers:
You would need to request the $data array by reference (`&$data`), and if that still doesn't work then we need to send it by reference too, so you'd make a PR to the library repo to change those events to send that variable by reference.

There's also `View::share('variable_name', $varData);` that can be used to share a variable with all email templates, so you could potentially listen to after create (note, example hasn't been tested)

```
\Backend\Models\User::extend(function ($model) {
    $model->bindEvent('model.afterCreate', function () use ($model) {
        if ($model->send_invitation) {
            View::share('my_var', $data);
        }
    });
});
```

PR #4652, opened 2019-10-02
Allow partial overrides in subfolders

### LINKED ISSUE #4347: Overriding partials in subfolders
Given the following component structure:

```
components/mycomponent
├── default.htm
├── subfolder
│   └── default.htm
```

If I want to override the `subfolder/default.htm` partial in my theme, I would add it like so:

```
mytheme
├── partials
│   ├── mycomponent
│   │   └── subfolder
│   │       └── default.htm
```

Unfortunately, this currently does not work. The override is ignored.

I tracked this issue down to this line:

https://github.com/octobercms/october/blame/master/modules/cms/classes/Controller.php#L1003

It seems like subfolder overrides are explicitly excluded. This line has been there from the very early days. Removing the if check does not seem to break anything obvious, but subfolder overrides begin to work without any futher changes.

So I'm wondering, does any one have any idea why it's there? What's wrong with including partials from subfolders?

tobias-kuendig:
This PR is my proposal to fix #4347. 

Changed behaviour:

* Component partials in subfolders can now be overridden
* Including partials from outside the base path results in a ApplicationException now (if `cms.restrictBaseDir` is true)

Adding the local file check to the `File::validatePath` method turned out to be not a good solution. Passing the absolute path to this method would break the "max nesting" check. My solution is to introduce a new `File::validateIsLocalFile` method that does this check depending on the `cms.restrictBaseDir` setting.

### DIFF
--- a/modules/cms/helpers/File.php
+++ b/modules/cms/helpers/File.php
@@ -1,5 +1,8 @@
 <?php namespace Cms\Helpers;
 
+use October\Rain\Filesystem\Filesystem;
+use Config;
+
 /**
  * Defines some file-system helpers for the CMS system.
  *
@@ -64,6 +67,26 @@ class File
             if (!self::validateName($segment)) {
                 return false;
             }
+        }
+
+        return true;
+    }
+
+    /**
+     * Validates a CMS object path is inside the application's base directory.
+     * @param string $filePath Specifies a path to validate
+     * @return boolean Returns true if the file path is local. Otherwise returns false.
+     */
+    public static function validateIsLocalFile($filePath)
+    {
+        $restrictBaseDir = Config::get('cms.restrictBaseDir', true);
+
+        if ($restrictBaseDir && !app(Filesystem::class)->isLocalPath($filePath)) {
+            return false;
+        }
+
+        if (!$restrictBaseDir && realpath($filePath) === false) {
+            return false;
         }
 
         return true;
--- a/modules/cms/helpers/File.php
+++ b/modules/cms/helpers/File.php
@@ -1,7 +1,7 @@
 <?php namespace Cms\Helpers;
 
-use October\Rain\Filesystem\Filesystem;
 use Config;
+use File as Filesystem;
 
 /**
  * Defines some file-system helpers for the CMS system.
@@ -81,7 +81,7 @@ class File
     {
         $restrictBaseDir = Config::get('cms.restrictBaseDir', true);
 
-        if ($restrictBaseDir && !app(Filesystem::class)->isLocalPath($filePath)) {
+        if ($restrictBaseDir && !Filesystem::isLocalPath($filePath)) {
             return false;
         }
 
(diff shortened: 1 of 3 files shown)


modules/cms/helpers/File.php
```diff
@@ -68,4 +71,24 @@ public static function validatePath($filePath, $maxNesting = 2)
...
+    /**
+     * Validates a CMS object path is inside the application's base directory.
+     * @param string $filePath Specifies a path to validate
+     * @return boolean Returns true if the file path is local. Otherwise returns false.
+     */
+    public static function validateIsLocalFile($filePath)
+    {
+        $restrictBaseDir = Config::get('cms.restrictBaseDir', true);
+
+        if ($restrictBaseDir && !app(Filesystem::class)->isLocalPath($filePath)) {
```
bennothommo: @tobias-kuendig You can use the `File` facade instead of directly interacting with the application provider.
```suggestion
        if ($restrictBaseDir && File::isLocalPath($filePath)) {
```

tobias-kuendig: Good point!

modules/cms/helpers/File.php
```diff
@@ -1,7 +1,7 @@
 <?php namespace Cms\Helpers;
 
-use October\Rain\Filesystem\Filesystem;
 use Config;
+use File as Filesystem;
```
LukeTowers: Use it as File please, everywhere else does.

tobias-kuendig:
The class itself is called File already!

modules/cms/helpers/File.php
tobias-kuendig: The class itself is called File already...

LukeTowers: So it is, turns out I'm not that great at code review at 2am. Who would have thought 🤷‍♂

tobias-kuendig:
I just noticed that the `{% if partial('') %}` function throws an exception now if the partial does not exist. It looks like it wasn't checked before if the partial exists when validating the path.

tobias-kuendig:
Anything I can do to get this PR viewed at?

LukeTowers:
@bennothommo @w20k is all that's left here just for me to review it?

bennothommo:
@LukeTowers it might still need testing. I haven't tested it myself yet.

daftspunk:
@tobias-kuendig Thanks for this contribution, it was difficult to review since it was conflating two separate issues (tightening of security + loosening of override rules)

I have made some adjustments that make the code easier to understand, whilst maintaining the parity of logic. Please review 7902cfa58a09eff0165a9962f562d1bced2d8045

Thanks!

tobias-kuendig:
:ballot_box_with_check: sub-folder overrides work
:ballot_box_with_check: Partials from outside the project cannot be included
:ballot_box_with_check: `partial()` twig helper does not throw exception if the requested partial does not exist

Looks all good to me! Thank you for the review and the improvements!

PR #4653, opened 2019-10-02
Added getConfig helper method to get config values

SaifurRahmanMohsin:
Added getConfig to make it easier for developers to fetch the config data from a list column while overriding the list items through extension. This also makes the class more compatible with [FormField](https://github.com/octobercms/october/blob/master/modules/backend/classes/FormField.php) which already has the same helper function.

### DIFF
--- a/modules/backend/classes/ListColumn.php
+++ b/modules/backend/classes/ListColumn.php
@@ -220,6 +220,17 @@ class ListColumn
     }
 
     /**
+     * Returns a raw config item value.
+     * @param  string $value
+     * @param  string $default
+     * @return mixed
+     */
+    public function getConfig($value, $default = null)
+    {
+        return array_get($this->config, $value, $default);
+    }
+
+    /**
      * Returns this columns value from a supplied data set, which can be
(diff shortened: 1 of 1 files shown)


LukeTowers:
@SaifurRahmanMohsin can you provide an example of how this method would get used?

SaifurRahmanMohsin:
I'm using this for a very specific purpose i.e. to implement wriite-able dropdowns in my list controller and my code looks something like this:

```
    \Acme\User\Controllers\MyController::extendListColumns(function ($list, $model) {
        $list->addColumns([
            'my_special_column' => [
                'label' => 'My Special Column',
                'type'  => 'partial',
                'path'  => '$/acme/user/assets/partials/my_column.htm'
            ]
        ]);
    });

// In my_column.htm
    <?php
        $columnOptions = $column->getConfig('options', []);
    ?>
// Dropdown render code goes here...
```

But you can use this method for a variety of purposes. It's pretty much as useful as it's FormField equivalent as it solves the same purpose but for columns type fields.

LukeTowers:
LGTM

PR #4655, opened 2019-10-03
Null Error with $user

AakbariP:
System has error when $user is null


bennothommo:
@AakbariP the `master` branch is updated when a new stable version of October is released. If you need the fix for the issue when `$user` is null, please use the edge updates. You can enable them in `config/cms.php`.

Issue #4185, opened 2019-03-07
 Wrong parameter value in php section of page template if it's value equals '0'


gaabora:
- OctoberCMS Build: any
- PHP Version: any

### Description:
I'm getting wrong parameter value in php section of page template if it's value equals '0'.
$this->param('paramname') returns null while 'paramname' is actually '0'

### Steps To Reproduce:
navigate to /test/0 using page template below
```
title = "test"
url = "/test/:slug"
layout = "default"
==
<?php
function onStart() {
	dump($this->param('slug')); // returns null
}
?>
==
HI! {{ this.param.slug }} // but here is ok, prints HI! 0
```

gaabora:
it definitely cause by isset function in getParameter() of Cms\Classes\Router
```
public` function getParameter($name, $default = null)
    {
        if (isset($this->parameters[$name]) && !empty($this->parameters[$name])) {
            return $this->parameters[$name];
        }

        return $default;
    }
```
so how about
`if (array_key_exists($name, $this->parameters)) { ...`?

bennothommo:
Fixed by #4657.

PR #4657, opened 2019-10-03
fix issue 4185

gaabora:
This fixes wrong parameter value in php section of page template if it's value equals '0'
https://github.com/octobercms/october/issues/4185

### DIFF
--- a/modules/cms/classes/Router.php
+++ b/modules/cms/classes/Router.php
@@ -304,7 +304,7 @@ class Router
      */
     public function getParameter($name, $default = null)
     {
-        if (isset($this->parameters[$name]) && !empty($this->parameters[$name])) {
+        if (array_key_exists($name, $this->parameters)) {
             return $this->parameters[$name];
         }
 
--- a/modules/cms/classes/Router.php
+++ b/modules/cms/classes/Router.php
@@ -304,7 +304,7 @@ class Router
      */
     public function getParameter($name, $default = null)
     {
-        if (isset($name, $this->parameters) && ($this->parameters[$name] === '0' || !empty($this->parameters[$name]))) {
+        if (isset($this->parameters[$name]) && ($this->parameters[$name] === '0' || !empty($this->parameters[$name]))) {
             return $this->parameters[$name];
         }
 
--- a/modules/cms/classes/Router.php
+++ b/modules/cms/classes/Router.php
@@ -304,7 +304,7 @@ class Router
      */
     public function getParameter($name, $default = null)
     {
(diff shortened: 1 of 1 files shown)


bennothommo:

bennothommo:
@gaabora It might be better to keep the original line there, but instead of doing an `!empty` check, just simply check to see that the parameter is not an empty string, `false` or `null`. This should help to minimise any unintended side effects with this change.

LukeTowers:
@bennothommo github actions bot still in develop?

LukeTowers:
@bennothommo to expand on your suggestion, what are "empty" values that are actually acceptable? We could just explicitly allow those values before falling back to the !empty() check. I think I would prefer that tbh.

bennothommo:
@LukeTowers TBH I'm not sure. From what I've seen, the only two values I think would be valid that are "empty" values are 0, and an empty string, but I can't say with any certainty whether others are in use, ie. an empty array.

LukeTowers:
@bennothommo how about 0 as a string? `"0"` Not sure 100% but I think empty() considers that to be falsey.

bennothommo:
@LukeTowers you're right, it does. So I would assume that needs to be valid too.

gaabora:
So should I change it to
`if (array_key_exists($name, $this->parameters) && ($this->parameters[$name] !=='' || $this->parameters[$name] !== null)) {`

bennothommo:
@gaabora I'd try the following:
```php
if (isset($this->parameters[$name]) && !in_array($this->parameters[$name], ['', 0, '0'])) {
```

gaabora:
> 
> 
> @gaabora I'd try the following:
> 
> ```
> if (isset($this->parameters[$name]) && (in_array($this->parameters[$name], ['', 0, '0']) || !empty($this->parameters[$name]))) {
> ```
`in_array` needs 3rd parameter to be `true` for strict check, otherwise there's no point,
also strict comparing with `or` condition will work faster than `in_array`

bennothommo:
@gaabora It would only be microseconds difference, so I'm happy with `in_array` because it's more readable. You're correct though in that we'd need the strict check, so feel free to add that in too.

LukeTowers:
@gaabora Isn't `0` by itself an acceptable value?

gaabora:
> @gaabora Isn't `0` by itself an acceptable value?

Did you mean integer `0` value? Then there's no way in this function for $this->parameters[$name] to be === `0`, only `'0'`, cause it produces from string parsing.

PR #4661, opened 2019-10-04
Add support for non-numeric menu counter values

LarryBarker:
This PR makes a small change to the `NavigationManager` class where it checks the menu item's counter value. This allows for non-numeric values to be used as counter labels (e.g. New, Beta, etc.). This is useful when you want to display a value other than a number, possibly to highlight a new feature or change to your plugin. 

image

**TODO**
- [ ] Document changes


Samuell1:
I think **counterClass** is not needed because you can attach different class to data-menu-id like this:

```
.counter[data-menu-id="shop/product"] {
    background: #27b061 !important;
}
```

LarBearrr:
@Samuell1 Fair point, especially considering whatever class was to be applied would have to be added anyway, so why not just target the element directly. My only thought here was providing a cleaner "interface" for applying the class.

I'm curious what @LukeTowers thinks about the class. Would it be worth while to add some default classes for the counter color, and provide a list of the available colors in the documention? Such that:

`counterColor`:

- red
- green
- orange
- yellow
- blue

Samuell1:
@LarBearrr yeah thats true too. 

Maybe better will be to separate them like to add "badge" option? Because if you dont skip counter and allow string values main value will be broken because it counts all values together.

LarBearrr:
I thought about the counter vs. badge nomenclature, and it seems like adding an additional configuration is unnecessary. I could see just renaming `counter` to `badge` but this would be a breaking change. 

The way it works right now, if the main menu item has a text value, it does not display a sum of the side menu item counters; it works as if the counter was set to `false`. In my screen shot, you can see that the counters are still working when they are numeric. Luke may want to clarify on this, but I believe if the main menu counter is set to anything at all, it does not display a sum of the side menu items.

Samuell1:
Yes but if you dont set main counter to false and other counter to string values what will happen?

LarBearrr:
Oh, you're saying like if the main menu counter is not set to either false or a string, meaning it should show a sum of the side menu items, and say one of the side menu items is a string, what will this do?

It looks like there will need to be a check in the `listSideMenuItems` to continue if the value is not  numeric: https://github.com/octobercms/october/blob/6b97277ffd13cddc2c06d32e860c8a672f4728db/modules/backend/classes/NavigationManager.php#L366-L373

Samuell1:
Yeah just check if its numeric should be fine.

LukeTowers:
@LarBearrr I'd rather have the option be `badge` to separate it explicitly and have `badge` overrule `counter`.

LarBearrr:
@LukeTowers Are you thinking to overload the counter value if badge is set? Or implement additional logic in the menu partials to render the badge if present?

LarBearrr:
Here are a couple screens after adding the `badge` option to the `NavigationManager`:

Side menu item with badge and multiple numeric counters summed in main menu:
image
```
'test' => [
                'label'    => 'Playground',
                'url'      => Backend::url('october/test/people'),
                'icon'     => 'icon-child',
                'order'    => 200,
                'permissions' => ['october.test.access_plugin'],

                'sideMenu' => [
                    'people'    => [
                        'label' => 'People',
                        'icon'  => 'icon-database',
                        'badge' => 'HOT',
                        'url'   => Backend::url('october/test/people'),
                    ],
                    'posts'     => [
                        'label'     => 'Posts',
                        'icon'      => 'icon-database',
                        'counter'   => 2,
                        'url'       => Backend::url('october/test/posts'),
                    ],
                    'users'     => [
                        'label' => 'Users',
                        'counter' => 4,
                        'icon'  => 'icon-database',
                        'url'   => Backend::url('october/test/users'),
                    ],
```

Main menu item with badge overriding numeric counter values from side menu items:
image
```
'test' => [
                'label'    => 'Playground',
                'url'      => Backend::url('october/test/people'),
                'icon'     => 'icon-child',
                'order'    => 200,
                'badge'    => 'NEW',
                'permissions' => ['october.test.access_plugin'],

                'sideMenu' => [
                    'people'    => [
                        'label' => 'People',
                        'icon'  => 'icon-database',
                        'badge' => 'HOT',
                        'url'   => Backend::url('october/test/people'),
                    ],
                    'posts'     => [
                        'label'     => 'Posts',
                        'icon'      => 'icon-database',
                        'counter'   => 2,
                        'url'       => Backend::url('october/test/posts'),
                    ],
                    'users'     => [
                        'label' => 'Users',
                        'counter' => 4,
                        'icon'  => 'icon-database',
                        'url'   => Backend::url('october/test/users'),
                    ],
```

modules/backend/classes/MainMenuItem.php
```diff
@@ -44,6 +44,11 @@ class MainMenuItem
      */
     public $counterLabel;
 
+    /**
+     * @var null|mixed
```
LukeTowers:
```suggestion
     * @var null|string
```

modules/backend/classes/NavigationManager.php
```diff
@@ -349,7 +351,8 @@ public function listMainMenuItems()
         }
 
         foreach ($this->items as $item) {
-            if ($item->counter === false) {
+            if ($item->badge) {
+                $item->counter = $item->badge;
```
LukeTowers:
```suggestion
                $item->counter = (string) $item->badge;
```

modules/backend/classes/NavigationManager.php
```diff
@@ -349,7 +351,8 @@ public function listMainMenuItems()
         }
 
         foreach ($this->items as $item) {
-            if ($item->counter === false) {
+            if ($item->badge) {
+                $item->counter = (string) $item->badge;
                 continue;
             }
```
LukeTowers:
```suggestion
            if ($item->badge) {
                $item->counter = (string) $item->badge;
                continue;
            }
            if ($item->counter === false) {
                continue;
            }
```

modules/backend/classes/SideMenuItem.php
```diff
@@ -47,6 +47,11 @@ class SideMenuItem
      */
     public $counterLabel;
 
+    /**
+     * @var null|mixed
```
LukeTowers:
```suggestion
     * @var null|string
```

LukeTowers:
Just need the changes documented in a PR to the docs repo @LarryBarker

modules/backend/classes/SideMenuItem.php
LarryBarker: Good call :)

PR #4663, opened 2019-10-04
Fix #4251: improved email client support for branded mails

### LINKED ISSUE #4251: Branded emails: incorrect word-break in Outlook clients
- OctoberCMS Build: 451
- PHP Version: 7.1
- Database Engine: n.a. (5.7.19)
- Plugins Installed: n.a.

### Description:
All emails using the default branded css do seem to have word-break issues in Outlook clients. Words are broken at random points due to the "word-break: all" css setting.

See the following link for example renders: [https://litmus.com/checklist/emails/public/15ff7f9](https://litmus.com/checklist/emails/public/15ff7f9)

### Steps To Reproduce:
1. Create a mail-template containing large sentences.
2. Send out a (test) mail
3. Open mail in any Outlook client.

### Possible solution:
The corresponding CSS is located in:
`modules/system/models/mailbrandsetting/custom.less`

> body {
>     background-color: @body-bg;
>     color: @text-color;
>     height: 100%;
>     hyphens: auto;
>     line-height: 1.4;
>     margin: 0;
>     -moz-hyphens: auto;
>     **-ms-word-break: break-all;**
>     width: 100% !important;
>     -webkit-hyphens: auto;
>     -webkit-text-size-adjust: none;
>     **word-break: break-all;**
>     word-break: break-word;
> }
> 

Changing "break-all" to "break-word" fixes the problem. I'm unsure however why "word-break: break-all;" was used in the first place.

vevers:
Fixed issue #4251 where word-break issues in Outlook clients (9.1% market share) were causing major readability issues.

- Updated word-break style settings.
- Added .break-all class, which can be used to make sure content does not expand beyond its container.
- Updated panel and promotion panel to use break-all class by default (in case long strings like coupon codes are used).
- Tested in all major email clients for compatibility using Litmus.

**Before**: https://litmus.com/checklist/emails/public/24587be
**After**: https://litmus.com/checklist/emails/public/d80507b

Note: this does not fix the Android 6.0 layout issue.

### DIFF
--- a/modules/system/models/mailbrandsetting/custom.less
+++ b/modules/system/models/mailbrandsetting/custom.less
@@ -87,7 +87,7 @@ img {
     max-width: 100%;
 }
 
-.break-all, .break-all *{
+.break-all, .break-all * {
     -ms-word-break: break-all;
     word-break: break-word;
     word-break: break-all;
--- a/modules/system/models/mailbrandsetting/custom.less
+++ b/modules/system/models/mailbrandsetting/custom.less
@@ -74,8 +74,6 @@ code {
     color: @text-color;
     font-size: 16px;
     line-height: 1.5em;
-    -ms-word-break: break-all;
-    word-break: break-word;
     word-break: break-all;
 }
 
@@ -88,8 +86,6 @@ img {
 }
 
 .break-all, .break-all * {
-    -ms-word-break: break-all;
-    word-break: break-word;
     word-break: break-all;
 }
 
--- a/modules/system/models/mailbrandsetting/custom.less
+++ b/modules/system/models/mailbrandsetting/custom.less
@@ -13,11 +13,10 @@ body {
     line-height: 1.4;
     margin: 0;
     -moz-hyphens: auto;
-    -ms-word-break: break-all;
+    -ms-word-break: break-word;
     width: 100% !important;
     -webkit-hyphens: auto;
     -webkit-text-size-adjust: none;
-    word-break: break-all;
     word-break: break-word;
 }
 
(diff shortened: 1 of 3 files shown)


modules/system/models/mailbrandsetting/custom.less
```diff
@@ -85,6 +87,12 @@ img {
     max-width: 100%;
 }
 
+.break-all, .break-all *{
```
LukeTowers:
```suggestion
.break-all, .break-all * {
```

modules/system/models/mailbrandsetting/custom.less
```diff
@@ -75,6 +74,9 @@ code {
     color: @text-color;
     font-size: 16px;
     line-height: 1.5em;
+    -ms-word-break: break-all;
+    word-break: break-word;
```
bennothommo: @vevers Is this necessary with the `break-all` rule directly underneath?

modules/system/models/mailbrandsetting/custom.less
```diff
@@ -85,6 +87,12 @@ img {
     max-width: 100%;
 }
 
+.break-all, .break-all * {
+    -ms-word-break: break-all;
+    word-break: break-word;
```
bennothommo: @vevers Same as above.

modules/system/models/mailbrandsetting/custom.less
vevers: I kept the original structure of using a combination of `-ms-word-break: break-all; word-break: break-all; word-break: break-word;` as used in the body rule intact, assuming it was required for some weird edge case or specific email clients. 

Just tried simplifying the rules as suggested, tested it again in Litmus and didn't see any problems.

bennothommo:
Looks great @vevers. Tested and works well.

PR #4666, opened 2019-10-05
Prevent flash of unstyled tabs on page load

tobias-kuendig:
While working on the lazy tabs implementation I tried to fix another problem that was bugging me for far too long: The fact that tabs are only styled correctly once they are initialized via JS.

This PR proposes a simple fix for this problem: The required (styled) span tags are added to the server rendered markup. The JS logic is extended to only add the span tags if they are not present already (this is to prevent backwards compatibilty).

### Before this change

tabs before

### After this change

tabs after


bennothommo:
@tobias-kuendig I've tested this with the Test plugin and it works as advertised. However, the `storm-min.js` diff is showing some weird changes that were not a part of your changes to the `tab.js` file. Did you run `php artisan october:util compile assets`?

LukeTowers:
@bennothommo I believe those are from https://github.com/octobercms/october/commit/c627e8a0ddc4e5b59f890c67145c8de892c0dca5, I don't think the assets were recompiled against that commit. Feel free to analyze the changes in to make sure they're just including those changes.

bennothommo:
@LukeTowers yep that was it. Cheers for letting me know.

Issue #4216, opened 2019-03-26
Media finder permission error message


gergo85:
### Steps To Reproduce:
1. Create an administrator and disabled the "Upload and manage media contents - images, videos, sounds, documents" permission.
2. Log in as the created administrator.
3. Click on any "mediafinder" form widget and will get the following error:
![error](https://user-images.githubusercontent.com/2959112/54995643-803bce00-4fc7-11e9-868f-018f202ba782.JPG)

LukeTowers:
That's up to the developers to implement to dynamically remove fields that their users shouldn't have access to. While we could in theory hook into the backend.forms.extendFields event to dynamically remove all mediafinder formwidgets when the user doesn't have the appropriate permission to use them, that could lead to a headache when a developer has specified that field as required but the user can't actually fulfill that requirement.

Thoughts?

gergo85:
You are right, but it would be better to display a user-friendly error message instead of default one.

LukeTowers:
Where would you display it though? Feel free to make a PR with your suggestions.

LukeTowers:
Fixed in https://github.com/octobercms/october/commit/85fadbfef32b08a30b82d7a12a5b2a70e3c63cfc

PR #4669, opened 2019-10-07
Check user permission for Mediafinder

gergo85:
Issue: #4216


mjauvin:
Looks like a decent fix to me.

LukeTowers:
@mjauvin were you able to test it?

mjauvin:
@LukeTowers Yes, tested and works fine.

I think this is a good approach as it only makes the field "readonly" when user does not have permission to edit.

LukeTowers:
Fixed in https://github.com/octobercms/october/commit/85fadbfef32b08a30b82d7a12a5b2a70e3c63cfc

Issue #4670, opened 2019-10-07
Make relation form type's select portable


multiwebinc:
I am creating a plugin and followed [the documentation](https://octobercms.com/docs/backend/forms#widget-relation) to create my model's `fields.yaml` file with the following `relation` type:

```
fields:
    cars:
        label: 'Available cars'
        span: auto
        type: relation
        select: concat(year, ' ', make , ' ' , model)
```

However since I'm using sqlite, I get an error:

> SQLSTATE[HY000]: General error: 1 no such function: concat (SQL: select "id", concat(year, ' ', make , ' ' , model) AS selection from "example_cars_cars" order by "name" asc)

So I therefore need to change it to:

    select: year || ' ' || make || ' ' || model

Since I think that a concatenation function in this context would be very common, what about defining one so that if sqlite is being used, it will transform `concat(year, ' ', make , ' ' , model)` into `year || ' ' || make || ' ' || model`, so that switching between database types (e.g. moving from dev to production) wouldn't be as much of an issue.

daftspunk:
Unless I am missing something obvious... support for this has existed for a long time inside DbDongle. I'm not sure why it doesn't work in your particular case. If it is still a problem, feel free to ping me here, after submiting a PR to the [test plugin](https://github.com/octoberrain/test-plugin) to demonstrate the defect.

Output of DbDongle, which is currently applied to the `select` value:

```
>>> (new October\Rain\Database\Dongle('sqlite'))->parseConcat("concat(year, ' ', make , ' ' , model)");
=> "year || ' ' || make || ' ' || model"
```

PR #4671, opened 2019-10-07
Make relation form type's select portable

multiwebinc:
Proposed fix for https://github.com/octobercms/october/issues/4670. I made it a switch to make it extendable to other database drivers.

### DIFF
--- a/modules/backend/formwidgets/Relation.php
+++ b/modules/backend/formwidgets/Relation.php
@@ -172,7 +172,7 @@ class Relation extends FormWidgetBase
      */
     protected function getPortableSqlSelect($sql)
     {
-        switch(config("database.connections.".config('database.default').".driver"))
+        switch (config("database.connections.".config('database.default').".driver"))
         {
             case 'sqlite':
                 $sql = preg_replace_callback(
--- a/modules/backend/formwidgets/Relation.php
+++ b/modules/backend/formwidgets/Relation.php
@@ -172,8 +172,7 @@ class Relation extends FormWidgetBase
      */
     protected function getPortableSqlSelect($sql)
     {
-        switch (config("database.connections.".config('database.default').".driver"))
-        {
+        switch (config("database.connections.".config('database.default').".driver")) {
             case 'sqlite':
                 $sql = preg_replace_callback(
                     '|concat\s*\((.*)\)|',
--- a/modules/backend/formwidgets/Relation.php
+++ b/modules/backend/formwidgets/Relation.php
@@ -26,7 +26,9 @@ class Relation extends FormWidgetBase
     public $nameFrom = 'name';
 
     /**
-     * @var string Custom SQL column selection to use for the name reference
+     * @var mixed Custom SQL column selection to use for the name reference
+     * If array, values will be concatenated together. If string, it will be
+     * be used directly in the query.
      */
     public $sqlSelect;
 
@@ -39,11 +41,6 @@ class Relation extends FormWidgetBase
      * @var string Use a custom scope method for the list query.
      */
     public $scope;
-
-    /**
-     * @var string Define the order of the list query.
-     */
-    public $order;
 
     //
     // Object properties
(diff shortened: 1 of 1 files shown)


bennothommo:
@multiwebinc Whilst this is good work, my main concern is that a regular expression match may not catch every permutation of a `concat()` query that people could do, and thus it would not cover all bases.

A possible different implementation of this would be instead providing a `selectConcat` property for the relation form widget, which could be provided by the developer as an array. We could then take the array and populate a controlled concatenation statement to the query based on the database type.

What are your thoughts on that?

multiwebinc:
I think that's a great idea.

multiwebinc:
I got started coding this and I'm wondering how would you want the yaml to be. There needs to be a way to distinguish between column names and strings.

If I have the following yaml:

        selectConcat:
            - year      # column
            - ', '      # string
            - make      # column
            - ' '       # string
            - model     # column

And in PHP, they're all just strings with no quotation marks around them, so when I do:

    return 'CONCAT(' . implode(', ', $values) . ')';

The result is

    "CONCAT(year, , , make,  , model)"

which is obviously wrong. The only way I can see to get the quotation marks in there is to use **double** quotation marks in yaml, which seems super ugly and error-prone to me:

        selectConcat:
            - year
            - "', '"
            - make
            - "' '"
            - model

Can anyone think of a better way?

bennothommo:
@multiwebinc I'm fine with double-quoted strings, especially if they are intending to use different separators between fields. Otherwise, perhaps another option called "concatSeparator" which defines a single string to use between all fields in the "selectConcat" array.

multiwebinc:
I think it might be better to just use the `select` property instead of introducing a new one. If it's an array, concatenate. If not, use it direct like it is now.

p.s. Sorry for the many commits here. I often manage to mangle things up on github.

bennothommo:
@multiwebinc That's fine - we squash it when we merge anyway (with credit given to you, of course).

LukeTowers:
Hmm, I'm almost wondering if we should extend the base grammar to allow for a separate `selectConcat` method that can be defined on each driver's grammar class instead. Then it's available everywhere.

multiwebinc:
@LukeTowers I wouldn't know how to begin doing that, to be honest.

LukeTowers:
Related https://github.com/laravel/framework/issues/5676

LukeTowers:
@bennothommo what do you think about taking a stab at implementing it in the Builder / Grammar classes?

bennothommo:
@LukeTowers sounds like fun. :)

prhost:
I do not want to be boring but I do not see this code snippet as good, too many conditions to try to solve a situation that occurs infrequently, not all that use sqlite, that is, a little more processing for nothing.

I would come up with the idea of creating a new optional formWidget parameter, which does the similar functionality of the `select` parameter but uses Yaml tags to assemble the result. Symfony did something like this in its Yaml configs.

Example:
```
customSelect: "%year% %make% %model%"
```

LukeTowers:
@prhost the whole point of the `select` property is to have the database do the concatenation so you can perform searches, filtering, and sorting operations on that concatenated result. Having YAML do it would move that into the application layer making those operations undoable.

prhost:
> @prhost the whole point of the `select` property is to have the database do the concatenation so you can perform searches, filtering, and sorting operations on that concatenated result. Having YAML do it would move that into the application layer making those operations undoable.

Exactly, my idea is precisely not to mess with `select` but create another optional parameter to set the value, the right is pure sql in `select`, the developer must adapt to the format of your driver query.

In a few years, this `switch()` will be great trying to predict all that are divergences between drivers.

LukeTowers:
@prhost the suggestion is to add to the grammar, so it's not a switch in the October application code, it's a grammar statement in the Laravel internals so that developers don't need to worry about what db structure they're targeting.

multiwebinc:
@bennothommo Excellent work. I'm closing this PR.

bennothommo:
@multiwebinc Are you sure this should be closed? While the `selectConcat` does provide the functionality you are looking for, it still needs to be linked to the YAML configuration value, I would think.

multiwebinc:
@bennothommo Ok, I'll wait for https://github.com/octobercms/library/pull/434 to be merged first to make sure there aren't any major changes.

LukeTowers:
@multiwebinc https://github.com/octobercms/library/pull/434 has been merged, are you still interested in getting this PR merged?

multiwebinc:
@LukeTowers I still think it would be good to merge it after modifications to take advantage of https://github.com/octobercms/library/pull/434. If anyone could do that, that would be great.

daftspunk:
Moved to #4670

PR #4672, opened 2019-10-08
Make RecordFinder's SearchWidget work when RecordFinder is in a Repeater

MartynasJanu:
RecordFinder's SearchWidget doesn't work if the RecordFinder is inside a Repeater. The problem is that the widgets are not initialized and not bound to controller.

This change preserves the relevant POST keys from the sent from the Repeater and adds them to the SearchWidget form as hidden fields. When the SearchWidget's onSubmit is activated, those keys are sent and the Controller will initialize all related widgets, including the Repeater, it's child RecordFinder and it's child SearchWidget. Works with nested Repeaters too.

There might be a better way to achieve this but after reverse-engineer the widget code, I only came up with this. Suggestions are welcome, but I will be using this for my project, so wanted to also share.

### DIFF
--- a/modules/backend/formwidgets/RecordFinder.php
+++ b/modules/backend/formwidgets/RecordFinder.php
@@ -299,7 +299,67 @@ class RecordFinder extends FormWidgetBase
             $this->searchWidget->setActiveTerm(null);
         }
 
+        $this->setParentFormPostKeysToPreserve();
         return $this->makePartial('recordfinder_form');
+    }
+
+    /**
+     * If elements from the parent Repeated don't get initialized, this RecordFinder is not initialize,
+     * which means that the searchWidget is not initialized and will not work from a repeated. If values
+     * and keys can be preserved from the parent RecordFinder form when calling the searchWidget, the
+     * RecordFinder will get initialized and will work
+     * @return void
+     */
+    protected function setParentFormPostKeysToPreserve()
+    {
+        $arrayName = $this->formField->arrayName;
+        $arrayNameLen = strlen($arrayName);
+        $preservePosts = [];
+        $idPrefix = $this->formField->idPrefix;
+
+        $flatPost = $this->flattenPost(post());
+        foreach ($flatPost as $key => $value) {
+            // preserve parent form
+            if (substr($key, 0, $arrayNameLen) === $arrayName) {
+                $preservePosts[$key] = $value;
+            }
+
+            // preserve repeater's _loaded var
+            if (substr($key, -7) === '_loaded' &&
+                strpos($idPrefix, substr($key, 0, -7)) !== false
+            ) {
+                $preservePosts[$key] = $value;
+            }
+        }
+
+        $this->vars['preservePosts'] = $preservePosts;
+    }
+
+    /**
+     * Credit: BryanErwin
+     * @param array $postData
+     * @param string $prefix
+     * @return string
+     */
+    protected function flattenPost(array $postData, string $prefix = '')
+    {
+        $result = [];
+        foreach ($postData as $key => $value) {
+            if (is_array($value)) {
+                if ($prefix === '') {
+                    $result = $result + $this->flattenPost($value, $prefix . $key);
+                } else {
+                    $result = $result + $this->flattenPost($value, $prefix . '[' . $key . ']');
+                }
+            } else {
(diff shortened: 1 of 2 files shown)


LukeTowers:

MartynasJanu:
> This pull request has been made to the wrong branch. Please review [the contributing guidelines](https://github.com/octobercms/october/blob/master/.github/CONTRIBUTING.md#pull-requests) as all PRs need to be made to the develop branch.
> 
> I'll fix it for you this time, but please ensure you make any future PRs to the develop branch, not the master branch.

Thanks, my bad.

modules/backend/formwidgets/recordfinder/partials/_recordfinder_form.htm
```diff
@@ -1,5 +1,13 @@
 <div id="<?= $this->getId('popup') ?>" class="recordfinder-popup">
     <?= Form::open() ?>
+    <?php
+
+    foreach ($preservePosts as $key => $value) {
+        ?>
+            <input type="hidden" name="<?= $key ?>" value="<?= $value ?>" />
+        <?php
+    }
+    ?>
```
LukeTowers: This should be 
```php
<?php foreach ($preservePosts as $key => $value) { ?>
    <input type="hidden" name="<?= $key ?>" value="<?= $value ?>" />
<?php } ?>
```

modules/backend/formwidgets/RecordFinder.php
```diff
@@ -299,9 +299,69 @@ public function onFindRecord()
...
     }
 
+    /**
+     * If elements from the parent Repeated don't get initialized, this RecordFinder is not initialize,
+     * which means that the searchWidget is not initialized and will not work from a repeated. If values
+     * and keys can be preserved from the parent RecordFinder form when calling the searchWidget, the
+     * RecordFinder will get initialized and will work
+     * @return void
+     */
+    private function setParentFormPostKeysToPreserve(): void
```
LukeTowers: Private methods are not used in OctoberCMS as protected is better for extensibility most of the time. We also still support PHP 7.0 so `: void` return types can't be used either.

LukeTowers:
Related https://github.com/octobercms/october/issues/3415, https://github.com/octoberrain/test-plugin/pull/78. I appreciate the contribution but this looks like it'll have to be handled in the Repeater widget

MartynasJanu:
> Related #3415, [octoberrain/test-plugin#78](https://github.com/octoberrain/test-plugin/pull/78). I appreciate the contribution but this looks like it'll have to be handled in the Repeater widget

I see. Makes sense and would be more elegant, however I didn't see how the Repeater could affect what the custom widget sends from it's form, so I recon this might be a huge overhaul to how the Repeater and/or widgets work. I hope I'm wrong though and this will get fixed soon and simply

LukeTowers:
@MartynasJanu could you send me an example of the before / after of what fields get sent for the search request?

MartynasJanu:
@LukeTowers 

Original:

```
_session_key: XX0i0F63DH2JsHOFC1Wy0CtwWjUMGGOpFwb3WZW0
_token: hHw4TNlwCt1Cg9iTWRmDQh41FJ9tiksepdkekhgP
formStudentsForm0StudentSearch[term]: x
recordfinder_flag: 1
```

With my changes:
```
_session_key: XX0i0F63DH2JsHOFC1Wy0CtwWjUMGGOpFwb3WZW0
_token: hHw4TNlwCt1Cg9iTWRmDQh41FJ9tiksepdkekhgP
Session[students][0][student]: 
Session[students][0][attended]:   
formStudents_loaded: 1
formStudentsForm0StudentSearch[term]: x
recordfinder_flag: 1
```

MartynasJanu:
@LukeTowers The Repeater widget could probably be changed to process it's children, but then the problem I see is that it wouldn't know the index for the child widgets. Like in my example `students` Repeater could initialize `student` RecordFinder, but it couldn't initialize `Session[students][0][student]` or any other index. Maybe another title/id could be used for widgets that doesn't involve the index and the indexes only for the POST values?...

LukeTowers:
@MartynasJanu I think I've come up with a potential solution, will ping you in the PR in a moment

MartynasJanu:
Fixed in  #4674

PR #4674, opened 2019-10-09
Fix #3415, #4672

### LINKED ISSUE #3415: Custom Form Widgets With AJAX Handlers Won't Work As Expected Inside Repeaters

LukeTowers:
Handle child FormWidgets making AJAX requests outside of the repeater's form container. Note that this won't pass on the state of the repeater item as it exists on the webpage because that information won't be sent to the server in an AJAX request sent from outside of the repeater's form container unless the FormWidget sending the orphaned request specifically includes that data in their request, in which case the regular handling will kick in and initialize the widget properly. A discussion should be had whether this fix makes sense to be done in the Repeater FormWidget or if we should force FormWidgets to have the responsibility of initializing their parent repeaters if they're going to be making orphaned AJAX requests.

Should fix #3415, #4672.
Related: https://github.com/octoberrain/test-plugin/pull/78

### DIFF
--- a/modules/backend/formwidgets/Repeater.php
+++ b/modules/backend/formwidgets/Repeater.php
@@ -220,6 +220,21 @@ class Repeater extends FormWidgetBase
             ? post($this->formField->getName())
             : $this->getLoadValue();
 
+        // Detect when a child widget is trying to run an AJAX handler
+        // outside of the form element that contains all the repeater
+        // fields that would normally be used to identify that case
+        $handler = $this->controller->getAjaxHandler();
+        if (!$this->loaded && starts_with($handler, $this->alias . 'Form')) {
+            // Attempt to get the index of the repeater
+            $handler = str_after($handler, $this->alias . 'Form');
+            preg_match("~^(\d+)~", $handler, $matches);
+
+            if (isset($matches[1])) {
+                $index = $matches[1];
+                $this->makeItemFormWidget($index);
+            }
+        }
+
         if (!$this->childAddItemCalled && $currentValue === null) {
             $this->formWidgets = [];
             return;
@@ -269,7 +284,7 @@ class Repeater extends FormWidgetBase
         $config = $this->makeConfig($configDefinition);
         $config->model = $this->model;
         $config->data = $this->getValueFromIndex($index);
-        $config->alias = $this->alias . 'Form'.$index;
+        $config->alias = $this->alias . 'Form' . $index;
         $config->arrayName = $this->getFieldName().'['.$index.']';
         $config->isNested = true;
         if (self::$onAddItemCalled || $this->minItems > 0) {


LukeTowers:
@MartynasJanu This is my proposed solution from the Repeater. I'd like your thoughts (and @bennothommo's thoughts) on handling this from the Repeater instead of forcing the FormWidgets that want to make orphaned AJAX requests to support being in a repeater (orphaned AJAX request means that the AJAX request is sent from an element that is outside of the repeater form, thus does not include any of the fields necessary for the repeater to properly initialize, #4672 would be an example of handling it from the RecordFinder's point of view).

bennothommo:
@LukeTowers I think I tried a fix like that a couple of months back - it works for a single layer of repeater, but as soon as you start talking about nested repeaters, the same issues occur. I think these lines partly form this fix: https://github.com/octobercms/october/blob/355448573479d8fd65dce13f356acaa0a4e8c425/modules/backend/formwidgets/Repeater.php#L243-L246

modules/backend/formwidgets/Repeater.php
```diff
@@ -220,6 +220,21 @@ protected function processItems()
             ? post($this->formField->getName())
             : $this->getLoadValue();
 
+        // Detect when a child widget is trying to run an AJAX handler
+        // outside of the form element that contains all the repeater
+        // fields that would normally be used to identify that case
+        $handler = $this->controller->getAjaxHandler();
+        if (!$this->loaded && starts_with($handler, $this->alias . 'Form')) {
```
LukeTowers: @bennothommo the starts_with() should handle the child repeaters, although I haven't tested.

LukeTowers: @bennothommo tested it, it works perfectly in the child repeater too!

LukeTowers: Perhaps we can revisit the child repeater approach to see if we can simplify it.

bennothommo: @LukeTowers True, but the regex below would only pick up one index, it wouldn't pick up the other indexes. Unless I'm misreading it.

MartynasJanu: @LukeTowers looks good from first glances. Tested with nested Repeaters too. I guess now the question remains, what could this break in some cases? If nothing, I believe this is suitable

bennothommo: @LukeTowers awesome. While I'm sure we can simplify it in places, I'm also hesitant to make too big a change to them after the last fiasco. :P

MartynasJanu: @LukeTowers Since the index is passed on via the headers as the handler name and the Repeater simply extracts the index from it, without making any alterations, it makes it fool-proof in my opinion. I thought maybe if the Repeater had a number of RecordFinders, there could be problems but after testing I realized the key is just about stripping part of a string, so there should be no harm done in any case. My analytical 0.02

PR #4677, opened 2019-10-09
Enable to change theme partial override for component

Rike-cz:
#4128


bennothommo:
@Rike-cz would you mind explaining the use case for this? My understanding from looking at your proposed change is that if the component does not have a default partial, then this change will allow someone to have a default partial for the component in the theme. Is this correct?

Rike-cz:
> @Rike-cz My understanding from looking at your proposed change is that if the component does not have a default partial, then this change will allow someone to have a default partial for the component in the theme. Is this correct?
My idea is explained [here](https://github.com/octobercms/october/pull/4128#issue-251727430). What I want is to have default partial for the component when component is used as snippet. But it works for only-components too. It looks like big change, that does not pass some tests :-(

config/cms.php
```diff
@@ -380,5 +380,19 @@
...
+     |--------------------------------------------------------------------------
+     |
+     | Enable using default theme partial for Component fallback where exists.
+     |
+     | If is enabled, especially each snippet not need to have own theme
+     | override, but uses default one.
+     |
+     */
+    
+    'enableComponentThemeDefaultPartialFallback' => false,
```
bennothommo: @Rike-cz This config name is too large. Maybe something shorter?
```suggestion
    'defaultAliasPartialFallback' => false,
```

config/cms.php
```diff
@@ -380,5 +380,19 @@
     */
 
     'restrictBaseDir' => true,
+    
+    /*
+     |--------------------------------------------------------------------------
+     | Component default partial fallback
```
bennothommo:
```suggestion
     | Allow default alias fallback for component partials
```

config/cms.php
```diff
@@ -380,5 +380,19 @@
...
+     |--------------------------------------------------------------------------
+     |
+     | Enable using default theme partial for Component fallback where exists.
+     |
+     | If is enabled, especially each snippet not need to have own theme
+     | override, but uses default one.
```
bennothommo:
```suggestion
     | If set to true, the default alias for a component will also be used to find
     | theme-overridden partials for components.
     |
     | eg. a component with a default alias of "blogPost" but with an assigned
     | alias of "blogPost1" will be checked to see if a partial exists for "blogPost1"
     | in the theme, before checking for a partial for "blogPost".
```

modules/cms/classes/ComponentPartial.php
```diff
@@ -106,6 +109,19 @@ public static function loadOverrideCached($theme, $component, $fileName)
         if ($partial === null) {
             $partial = Partial::loadCached($theme, $component->alias . '/' . $fileName);
         }
+        
+        if ($partial === null && Config::get('cms.enableComponentThemeDefaultPartialFallback')) {
```
bennothommo:
```suggestion
        if ($partial === null && Config::get('cms.defaultAliasPartialFallback')) {
```

modules/cms/classes/ComponentPartial.php
```diff
@@ -106,6 +109,19 @@ public static function loadOverrideCached($theme, $component, $fileName)
...
+            $classes = array_flip(ComponentManager::instance()->listComponents());
+            $default_alias = array_get($classes, Str::normalizeClassName($component->name));
```
bennothommo: @Rike-cz you can just use `array_search` here instead. Also, as a convention, we use camelCase variables.
```suggestion
            $defaultAlias = array_search(ComponentManager::instance()->listComponents(), Str::normalizeClassName($component->name));
```

bennothommo (changes requested):

Rike-cz (commented):
I think it's done.

LukeTowers:
@Rike-cz what's the rationale for having this behind a config value?

Rike-cz:
@LukeTowers Passing tests & backward compatibility. 
This PR changes previous behavior and does not pass "partial override" tests because does not fail when expected due to fallback.

LukeTowers:
@Rike-cz is this a breaking change then? Or do the tests just need to get updated?

Rike-cz:
@LukeTowers I think so. That's why I used config switch. But maybe @bennothommo has different statement?
When I tried it first, I have problem with passing [these tests](https://travis-ci.org/octobercms/october/jobs/492655828?utm_medium=notification&utm_source=github_status).

bennothommo:
@Rike-cz it is not failing the tests any more, as far as I can see. I think I'm on board with what @LukeTowers is saying - this doesn't need to be behind a config. The original behaviour is still intact - if an override uses the exact alias, it is used instead of your fallback.

bennothommo:
@LukeTowers @Rike-cz I've refactored the code a bit and removed the config, and the tests are still failing - my previous success was simply because I hadn't enabled the config yet.

It is *kinda* a change in behaviour, in that there is the potential for some sites who had created a partial override with the filename of a default component alias - these sites may encounter the override being used in more places than before. However, I feel this is a very unlikely scenario, so I'm happy to update the tests to reflect the new behaviour if you wish.

Rike-cz:
@Rike-cz I haven't problem with removing config and updating tests. But this decision is only for maintainers.

LukeTowers:
@bennothommo can we get @daftspunk's input on this first just to verify that's acceptable? It could be a relatively common case given that most of the time the default alias is used rather than an instance specific one, but it's true that in those cases there's likely to be only one instance in the first place.

LukeTowers:
@Rike-cz is this replaced by https://github.com/octobercms/october/pull/5139?

Rike-cz:
@LukeTowers Could be. I commented one line in code, where my fallback could be better used. Waiting for contributor`s reply.

daftspunk:
Hi @Rike-cz, we want to discourage overriding component partials this way. Default markup was only ever intended to be a sample use case that should be copied over to the theme. So the solution here would be to fork the component to your theme, reference then it from the snippet partial. Hope this helps

PR #4680, opened 2019-10-10
Fix #4679: Unable to set customview path for relation list

### LINKED ISSUE #4679: Unable to set customview path for relation list
When you render a relation in the backend, you should be able to use a custom view for that relation table - cause that's usally possible when rendering list widgets.

zoomyboy:
This addresses the issue that it should be possible to set a custom view path for a relation. The relation list is rendered inside of the form of the parent and generally has another "meaning" than the normal table which is rendered by default. So setting a custom view path for that relation list is probably a good option. 

I set this to null by default because no path meansis's just not added to the ViewPath of the list widget:

```
Backend\Widgets\Lists:213

        $this->recordsPerPage = $this->getUserPreference('per_page', $this->recordsPerPage);

        if ($this->showPagination == 'auto') {
            $this->showPagination = $this->recordsPerPage && $this->recordsPerPage > 0;
        }

        if ($this->customViewPath) {
            $this->addViewPath($this->customViewPath);
        }

        $this->validateModel();
        $this->validateTree();
```

If the view Path is null, the list just behaves as before - meaning the config value isn't set at all.

### DIFF
--- a/modules/backend/behaviors/RelationController.php
+++ b/modules/backend/behaviors/RelationController.php
@@ -666,6 +666,7 @@ class RelationController extends ControllerBehavior
             $config->recordsPerPage = $this->getConfig('view[recordsPerPage]');
             $config->showCheckboxes = $this->getConfig('view[showCheckboxes]', !$this->readOnly);
             $config->recordUrl = $this->getConfig('view[recordUrl]', null);
+            $config->customViewPath = $this->getConfig('view[customViewPath]', null);
 
             $defaultOnClick = sprintf(
                 "$.oc.relationBehavior.clickViewListRecord(':%s', '%s', '%s')",


LukeTowers:
@zoomyboy can you also make a PR to the docs repo (octobercms/docs) documenting this?

LukeTowers:
Fixes #4679

zoomyboy:
https://github.com/octobercms/docs/pull/409

daftspunk:
Thanks @zoomyboy!

mjauvin:
@daftspunk @LukeTowers aszbanas-extremecoding Is that documented somewhere ?

LukeTowers:
@mjauvin here: https://github.com/octobercms/docs/pull/409

mjauvin:
I mean, shouldn't that be in october documentation?

--
Marc

LukeTowers:
@mjauvin @daftspunk might have to update the docs site for it to show up

mjauvin:
@LukeTowers never mind, it's already in the docs, I thought that was a new option in the relation config file:

https://octobercms.com/docs/backend/relations#configuring-relation

PR #4683, opened 2019-10-11
Support readonly and disabled fields when value is an array

petehalverson:
Issue #4330

### DIFF
--- a/modules/backend/formwidgets/taglist/partials/_taglist.htm
+++ b/modules/backend/formwidgets/taglist/partials/_taglist.htm
@@ -1,18 +1,32 @@
 <?php
     $selectedValues = is_array($selectedValues) ? $selectedValues : [];
     $availableOptions = $useKey ? $fieldOptions : array_unique(array_merge($selectedValues, $fieldOptions));
+    $displayOnlyOptions = [];
+
+    foreach ($availableOptions as $key => $option) {
+        if (!strlen($option)) {
+            continue;
+        }
+        if (($useKey && in_array($key, $selectedValues)) || (!$useKey && in_array($option, $selectedValues))) {
+            $displayOnlyOptions[] = $option;
+        }
+    }
 ?>
 <!-- Tag List -->
 <?php if ($this->previewMode || $field->readOnly || $field->disabled): ?>
     <ul class="form-control taglist--preview" <?= $field->readOnly || $field->disabled ? 'disabled="disabled"' : ''; ?>>
-        <?php foreach ($availableOptions as $key => $option): ?>
-            <?php if (!strlen($option)) continue ?>
-            <?php if (($useKey && in_array($key, $selectedValues)) || (!$useKey && in_array($option, $selectedValues))): ?>
-                <li class="taglist__item"><?= e(trans($option)) ?></li>
-            <?php endif ?>
+        <?php foreach ($displayOnlyOptions as $option): ?>
+            <li class="taglist__item"><?= e(trans($option)) ?></li>
         <?php endforeach ?>
     </ul>
-    <?php if ($field->readOnly): ?>
+    <?php if ($field->readOnly && is_array($field->value)): ?>
+        <?php foreach ($displayOnlyOptions as $option): ?>
+        <input
+            type="hidden"
+            name="<?= $field->getName() ?>[]"
+            value="<?= $option ?>">
+        <?php endforeach ?>
+    <?php else: ?>
         <input
             type="hidden"
             name="<?= $field->getName() ?>"


petehalverson:
[Octodock control test](https://1a179b3d9129.octodock.com:8443/backend/october/test/posts) using [test plugin fork](https://github.com/petehalverson/test-plugin/commit/97b4b09f34065b6941e6764549a5c1b0272496d7) with some preloaded tags

[Confirm fix](https://dc68f32d2217.octodock.com:8443/backend/october/test/posts). Uses PR https://github.com/octobercms/october/pull/4683 + test plugin

LukeTowers:
@petehalverson I'd feel more comfortable if readyOnly save data was handled on the server side, something like

```
function getSaveValue()
{
     if ($this->readOnly) { return $this->getLoadValue(); }
}
```

petehalverson:
Agreed. @jimcottrell suggested something similar. I thought it was odd forms posts values for readonly fields as hidden input fields.

I didn't know if there was history to why, so I stuck with the status quo to keep the PR simple.

LukeTowers:
I'm not sure of the history either, but I do know that support for disabled, previewMode, and readOnly has never really been fully fleshed out for all fields / widgets so my guess would be any existing cases of readOnly being present as a hidden input is a workaround because a full proper solution to that was never developed.

bennothommo:
I've confirmed the fixes through @petehalverson's Octodock instances, and my own copy too for good measure. It's good to merge once @LukeTowers' comments above are considered.

LukeTowers:
@bennothommo were you able to test out my proposed server side fix instead?

bennothommo:
Whoops @LukeTowers, only just saw your comment. Will try and test it out in the next day or 2.

jimcottrell:
Thinking about this a bit, I don't know if there actually should be a server-side change for `readOnly`. The HTML spec doesn't actually have a `readonly` attribute on `select` elements, which is why [the Select2 project disagrees about how to handle it](https://github.com/select2/select2/issues/3387), so we don't get to just pass that through on the tag list.

For elements that do support `readonly`, the expectation is that there's still an input field (which distinguishes it from October's `previewMode` to me), but no interaction is possible with it. The value is still submitted with the form (unlike `disabled`), so it seems like it should still be on the widget to provide the correct value, so there's no reason to override it server-side.

That said, I don't care for the need for hidden inputs with the array/not-array distinction. I'd rather just instantiate Select2 and disable input with one of the workarounds and let the data flow normally.

LukeTowers:
@jimcottrell I suggest a server side change as a solution to another problem, which is that readOnly doesn't actually mean readOnly in truth because the browser can still send whatever values back it wants to. In my mind, a dev setting readOnly on a field means that there should be no way that field is able to change the data that it is presenting, which should be handled on the server side. Thoughts?

jimcottrell:
@LukeTower That was pretty much my initial thought too, but in thinking about it, I started feeling that `readOnly` was set up in October with the intent to mirror the HTML definitions and is an attribute of the UI form field, i.e. "the field doesn't take user input" but doesn't actually mean "the field data cannot change" (a programmatic JS field update could still take effect).

It starts getting confusing with `select` elements (with no `readonly` attribute) and custom widgets and such, so I would understand the clarity of the server-side `readOnly`  field change prevention, but then shouldn't it just be replaced in practice by `previewMode`? At that point, there's no real distinction between the two.

bennothommo:
@LukeTowers I'm leaning towards agreeing with @jimcottrell on this one. Our docs state `readOnly` "prevents the field from being modified", whereas `disabled` "prevents the field from being modified and excludes it from the saved data." - however, **readOnly** should really say "prevents the field from being modified *by the user*".

There is an implication when using the HTML5 `readonly` attribute that the field can actually be modified by an external system - ie. JavaScript. You can use JavaScript to modify the value of a read-only field, and its new value will still be sent through the POST data just fine. It's only when you use the `disabled` attribute that the field is fully excluded and thus non-modifiable.

LukeTowers:
@bennothommo @jimcottrell fair enough. I think there was a comment in an issue somewhere that I defined each mode (disabled, readOnly, previewMode) and what exactly the differences between them were. I feel like we need to either find that and revise it as necessary or make a new one that can be agreed upon and then add that to the docs so that we can base any future changes off of those definitions.

bennothommo:
@LukeTowers Here's what I reckon we define the attributes as:

- **disabled:** prevents the field from being modified entirely and prevents any values in the field from being saved. Options: true or false.
- **hidden:** excludes the field from being displayed in the form and prevents any values in the field from being saved. Options: true or false.
- **readOnly:** prevents the field from being modified by the user. The field value can still be modified by events or JavaScript. Options: true or false.

If we stick with these definitions, then the only curve-ball thrown is in cases like the TagList where the read-only version of the widget is rendered as HTML. In these cases, we would have to render the value as hidden fields like @petehalverson has included in this PR to allow the field to be manipulated via JavaScript. If we decide to exclude JavaScript, then we can use your suggestion of using the `getLoadValue()`, as long as its fired before the events fire.

LukeTowers:
@bennothommo under those definitions, what's the use case for ever using `hidden: true`? Is it just for when you want the data to be accessible to JS?

bennothommo:
@LukeTowers sorry for the delay in responding.

I don't think I've ever really saw a need for the `hidden` attribute, I've always thought it didn't serve much of a point - a field that needs to be hidden except under certain scopes will use form scopes, a field that always needs to be hidden should probably be removed from the config completely.

I can't remember if it was you, or @daftspunk, who said it's useful for plugins that wish to disable fields externally through the `extendFields` event.

bennothommo:
@LukeTowers just pinging you on the above. You happy with those definitions? I haven't yet tried running the `getLoadValue` readOnly override - can't recall if this is being done in another PR.

LukeTowers:
@bennothommo we need to get @daftspunk's input on this

LukeTowers:
@bennothommo what's the status on this? @jimcottrell @petehalverson is this still relevant?

bennothommo:
@LukeTowers it was waiting for @daftspunk to comment on. I'm still interesting in changing the behavior to work according to my [comments here](https://github.com/octobercms/october/pull/4683#issuecomment-554944853).

LukeTowers:
@bennothommo after taking another quick look I feel like hidden and readOnly should be swapped:

- disabled: prevents the field from being modified entirely and prevents any values in the field from being saved. Options: true or false.
- hidden: excludes the field from being displayed in the form. The field value can still be modified by events or JavaScript. Options: true or false.
- readOnly: prevents any values in the field from being saved. Options: true or false.

bennothommo:
@LukeTowers I can agree with those definitions :)

daftspunk:
Looks good @petehalverson . Replicated and tested fine on my end. Thanks

PR #4685, opened 2019-10-11
Adding two additional robots META fields to CMS.Pages

FlusherDock1:
In the current realities of the web, the presence of robots meta tags is just as mandatory as title and description. Quickly hiding pages from search, or creating pages that are hidden from search initially, but open to users, is already a common practice.

At the moment, there is no other option for the user of the admin panel to hide the page from search robots, except to hide from everyone, or install third-party plugins.

With this PR, I want to start a discussion on the need for these two fields in CMS.Pages.
If this PR is approved, I will add translations for all other localizations, and create a PR in RainLab.Pages


m49n:
good idea

bennothommo:
@FlusherDock1 Could this simply be two checkboxes, one asking if they wish to "Allow search engines to index this page?", and another to "Allow search engines to follow links?". We can determine the correct `robots` meta value based on the combination of checkbox values. 

Additionally, there is another useful robots value called `noarchive`, that allows developers to prevent a page from having an archive copy stored of it. I believe that the Wayback Machine (webpage archive) system respects this meta value as well. This could be "Allow search engines to archive this page?".

FlusherDock1:
> Could this simply be two checkboxes, one asking if they wish to "Allow search engines to index this page?", and another to "Allow search engines to follow links?". We can determine the correct `robots` meta value based on the combination of checkbox values.

It can be whatever community likes to see it :) 

From users of Russian telegram chat i already see some examples in form of radio buttons.
IMAGE 2019-10-11 12:43:43

But your example with checkboxes is more suitable, because on the other tab there are also checkboxes.

>Additionally, there is another useful robots value called noarchive, that allows developers to prevent a page from having an archive copy stored of it. I believe that the Wayback Machine (webpage archive) system respects this meta value as well. This could be "Allow search engines to archive this page?".

That sounds good, already preparing another commit with this checkbox...

36864:
+1 for checkboxes. It's more compact and the phrasing is clear enough for non-technical users to understand.

36864:
You could try using `switch`es instead.

FlusherDock1:
> You could try using `switch`es instead.

What's the difference between switch and checkbox, if the problem with output through IF statement?

36864:
Brainfart, I use switches for a similar situation but forgot I have custom logic in place to explicitly set a value on the field when the switch is in the off position. (which also would work for checkboxes, of course).
Maybe that's a PR waiting to happen.

Samuell1:
@FlusherDock1 You can use custom attribute that will use switches like @36864 suggested and return value from multiple checkboxes.

Example:
```
public function getMetaRobotsAttribute()
{
 // code here
}
```

Then you can just call it like this.page.meta_robots

modules/cms/classes/Page.php
```diff
@@ -28,6 +28,10 @@ class Page extends CmsCompoundObject
         'is_hidden',
         'meta_title',
         'meta_description',
+        'meta_title',
```
LukeTowers:
```suggestion

```
`meta_title` is already in this array

LukeTowers:
@FlusherDock1 could you post a screenshot?

FlusherDock1:
@LukeTowers sure.
Снимок экрана 2019-10-17 в 10 21 19

FlusherDock1:
@LukeTowers done.

modules/cms/classes/Page.php
```diff
@@ -112,11 +115,25 @@ public static function getNameList()
...
+    /**
+     * Helper that returns formatted robots tags.
+     * @return string
+     */
+    public function getMetaRobotsAttribute()
+    {
+        $options = [];
+        $options[] = $this->meta_robots_index ? 'noindex' : 'index';
+        $options[] = $this->meta_robots_follow ? 'nofollow' : 'follow';
+        $options[] = $this->meta_robots_archive ? 'noarchive' : 'archive';
```
LukeTowers: These properties should be prefixed with `no`: `$this->meta_robots_noindex` to reflect how they're actually used.

FlusherDock1:
@LukeTowers done.

modules/cms/classes/Page.php
```diff
@@ -112,11 +115,25 @@ public static function getNameList()
...
+        $options[] = $this->meta_robots_noarchive ? 'noarchive' : 'archive';
+        return implode(',', $options);
+    }
+
     /**
      * Helper that makes a URL for a page in the active theme.
      * @param mixed $page Specifies the Cms Page file name.
      * @param array $params Route parameters to consider in the URL.
      * @return string
+     * @throws CmsException
```
LukeTowers: Why have these been added?

FlusherDock1: PHPStorm does this by himself. 

I'll clear that in the next commit.

FlusherDock1:
I translated everything through Google Translate. I think the errors will be found later anyway...

36864:
This is a bit off-topic, and regards only your localization effort.

On the one hand, having machine-translated localization strings can be better than just having unlocalized strings. On the other hand, it makes it more difficult for translators to figure out what needs to be translated.

I recently did a pass on one translation file because I found an untranslated string, and noticed that a lot of translation keys were either missing or untranslated, so I was able to focus on fixing those while mostly ignoring the rest of the file.

If every new addition is machine-translated, it's more difficult for a new contributor to determine what needs a manual pass. As translations can sometimes be subjective (one word can have multiple translations, and it's important to maintain consistency in those choices), in some cases where the machine translation is actually decent, it's also more difficult to determine which choices previous translators made, and which translation is the "correct" one used throughout.

As I understand it, there's currently no policy on localization for new translation keys in PRs, but maybe there should be. I'll write up an issue about it for further discussion.

FlusherDock1:
@36864 You're absolutely right. But what to do in such a situation? Finding a native speaker of each language in the October community is not an easy task, and thus any innovation will be on the shelf for a very long time waiting for its translation

LukeTowers:
@FlusherDock1 @36864 our usual policy is to only accept human-translated strings for new strings and merge new functionality in without full locale coverage rather than have machine translated coverage.

This allows users of locales that haven't been added yet to see the new functionality and provide their own human translations for the keys without the potential for confusion that can arise from the machine translated keys.

FlusherDock1:
> @FlusherDock1 @36864 our usual policy is to only accept human-translated strings for new strings and merge new functionality in without full locale coverage rather than have machine translated coverage.
> 
> This allows users of locales that haven't been added yet to see the new functionality and provide their own human translations for the keys without the potential for confusion that can arise from the machine translated keys.

Ok, fair point, i understand that. What i need to do?

w20k:
@FlusherDock1 leave only the English version and any language that you are fluent with ;)

FlusherDock1:
@w20k done.

LukeTowers:
@w20k is this ready to merge? We won't merge it right away, we'll wait for 460 to be finished first but just checking in on the status.

w20k:
@LukeTowers not yet, haven't tested front-end, yet.

w20k:
@LukeTowers works fine ;) The only thing is missing - a docs section for `dummies`)
Screenshots:
Screenshot 2019-10-24 at 16 22 24
Screenshot 2019-10-24 at 16 24 57

FlusherDock1:
Why it's not merged on 461?

LukeTowers:
@FlusherDock1 because 461 was pushed out for PHP 7.4 support. Adding it to 462.

LukeTowers:
@FlusherDock1 After another look at this with @daftspunk we've decided to revert it. This would be a perfect candidate for inclusion through a plugin. We strive to keep the core footprint tiny, and accepting these three fields would open the floodgates to accepting every SEO field imaginable. Please feel free to submit this as a separate plugin to the marketplace!

FlusherDock1:
Thanks for the clarification

Issue #4686, opened 2019-10-11
Invalid error message.


nameewgeniy:
- OctoberCMS Build: 458
- PHP Version: 7.3
- Database Engine: PostgreSQL

### Description:
Validation error message for rule exists must be negative in Russian. The rule exists generates an error if the value in the database is absent, but an error message in Russian reports that such a value already exists.

### Steps To Reproduce:
![image](https://user-images.githubusercontent.com/10694033/66647966-ff76b380-ec32-11e9-9d35-25e36179075d.png)

Now: ["Выбранное значение для :attribute уже существует."](https://github.com/octobercms/october/blob/c1fe12f732278df8a706fccf62baae8cd918ca30/modules/system/lang/ru/validation.php#L43)
Probably should be: "Выбранное значение для :attribute отсутствует."

w20k:
@nameewgeniy you could provide a PR for `ru/validation.php` to fix this translation issue.

LukeTowers:
Fixed by #4687

PR #4687, opened 2019-10-11
Update validation.php

nameewgeniy:
Validation error message for rule exists must be negative in Russian. The rule exists generates an error if the value in the database is absent, but an error message in Russian reports that such a value already exists.


w20k:
Thanks, @nameewgeniy !

LukeTowers:
Fixes #4686

PR #4689, opened 2019-10-11
Update portuguese translation

36864:
Added missing translation keys from lang/en and translated a few keys that were added since the last translation update.


modules/backend/lang/pt-pt/lang.php
```diff
@@ -407,21 +465,25 @@
         'hint' => 'Este registo mostra a lista de acessos dos administradores. Os registros são mantidos por um período de :days dias.',
         'menu_label' => 'Registo de Acesso',
         'menu_description' => 'Veja a lista de acessos à administração.',
+        'id' => 'ID',
         'created_at' => 'Data & Hora',
+        'type' => 'Tipo'
```
bennothommo: @36864 Please add a comma at the end. This line will cause a parse error currently.
```suggestion
        'type' => 'Tipo',
```

36864: That's what I get for editing this in the browser.
Went ahead and just replaced every instance of `'\n` with `',\n`

Issue #4561, opened 2019-08-22
Add other model's relation events


Rike-cz:
By now there are only attach/detach events for [BelongsToMany relation](https://octobercms.com/docs/api/model/relation/afterattach). Adding other events will make using OctoberCMS more comfortable. Good example there: https://github.com/chelout/laravel-relationship-events. 
Please add this feature request [to bounty](https://www.bountysource.com/teams/october/issues).

LukeTowers:
Please only use GitHub Issues for feature requests if you are using it as a pre-Pull Request collaboration tool. If you aren't planning on implementing it yourself, then it would be better suited being asked for on one of these channels:

Forum: http://octobercms.com/forum
IRC: #october on Freenode or http://octobercms.com/chat
Slack: https://octobercms-slack.herokuapp.com/

The support demand for October is increasing, while the available time the founders have to give remains the same. Feature requests are by default the lowest priority for us, as we don't have time to implement everyone's suggestions by ourselves. However there are ways you can help:

- Submit a pull request that introduces this feature and we'll review it. Please note that the amount of effort making the PR (code quality, rationale for inclusion in the project, explanation of changes) is directly proportional to the likelihood that we will review and merge it.
- [Make a financial contribution to the project](http://octobercms.com/fundraising) so we can allocate more time towards making improvements like these that benefit everyone

Rike-cz:
### More details
Just now we have this 4 relation events for BelongsToMany relation:
- model.relation.beforeAttach (string $relationName, array $attachedIdList, array $insertData),
- model.relation.afterAttach (string $relationName, array $attachedIdList, array $insertData),
- model.relation.beforeDetach (string $relationName, array $attachedIdList),
- model.relation.afterDetach (string $relationName, array $attachedIdList).

But other events and relations are not covered. This is table of full covered relations and their events:
#### hasOneOrMany
Events | Params
--- | ---
| model.relation.beforeCreate | string $relationName, Model $related |
| model.relation.afterCreate | string $relationName, Model $related |
| model.relation.beforeUpdate | string $relationName, Model $related |
| model.relation.afterUpdate | string $relationName, Model $related |
| model.relation.beforeSave | string $relationName, Model $related |
| model.relation.afterSave | string $relationName, Model $related |

#### belongsTo
Events | Params
--- | ---
| model.relation.beforeUpdate | string $relationName, Model $related |
| model.relation.afterUpdate | string $relationName, Model $related |
[...]

PR #4691, opened 2019-10-12
Attempt proposal of relation controller's events

Rike-cz:
Someone very clever said me "try it" :-) So I am trying to make proposal of relation controller's events handling as in form controller behavior they are.
If you will find this proposal useful, we can make it.

### DIFF
--- a/modules/backend/behaviors/RelationController.php
+++ b/modules/backend/behaviors/RelationController.php
@@ -1096,7 +1096,11 @@ class RelationController extends ControllerBehavior
 
             $modelsToSave = $this->prepareModelsToSave($newModel, $saveData);
             foreach ($modelsToSave as $modelToSave) {
+                if ($this->controller->methodExists('relationBeforeSave')) $this->controller->relationBeforeSave($modelToSave, $this->relationName, $this->relationObject);
+                if ($this->controller->methodExists('relationBeforeCreate')) $this->controller->relationBeforeCreate($modelToSave, $this->relationName, $this->relationObject);
                 $modelToSave->save(null, $this->manageWidget->getSessionKey());
+                if ($this->controller->methodExists('relationAfterSave')) $this->controller->relationAfterSave($modelToSave, $this->relationName, $this->relationObject);
+                if ($this->controller->methodExists('relationAfterCreate')) $this->controller->relationAfterCreate($modelToSave, $this->relationName, $this->relationObject);
             }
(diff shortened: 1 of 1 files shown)


bennothommo:
@Rike-cz given there's a fair few repeated callback calls, could you simplify it by creating a new method called `fireCallback()` that accepts the callback name and related arguments? Perhaps using this sort of method:

```php
protected function fireCallback(string $callback, ...$arguments)
{
    if ($this->controller->methodExists($callback)) {
        call_user_func_array([$this->controller, $callback], $arguments);
    }
}
```

Then you can simply use the following to fire a callback:

```php
$this->fireCallback('relationBeforeSave', $modelToSave, $this->relationName, $this->relationObject);
```

Thoughts?

Rike-cz:
@bennothommo No problem, if this concept of events is in general good and acceptable feature for all maintainers.

LukeTowers:
@Rike-cz shouldn't these events be implemented at the model level, not the controller level?

Rike-cz:
@LukeTowers Absolutely, but as in form controller behavior, you can use both approaches.

LukeTowers:
@Rike-cz I would prefer for them to exist as model events first, and then if we need to we can look at implementing them as relationcontroller methods

bennothommo:
@Rike-cz are you still interested in finishing this? If so, you will need to implement this based on @LukeTowers [suggestion](https://github.com/octobercms/october/pull/4691#issuecomment-542302260).

Rike-cz:
@bennothommo This PR is waiting for @LukeTowers deal with this [Bounty](https://www.bountysource.com/issues/79185728-add-other-model-s-relation-events).

bennothommo:
@Rike-cz @LukeTowers what is the status with this PR? It appears to be for a bountied issue that @Rike-cz placed the bounty for, but Rike indicated that they're waiting on @LukeTowers for something...?

LukeTowers:
I'm not sure what @Rike-cz is waiting on me for?

Rike-cz:
> @Rike-cz shouldn't these events be implemented at the model level, not the controller level?

Trying to deal with it with @mjauvin #4561. When we finish that, this PR can be closed as redundant (but I think some events in RelationController can me handy).

LukeTowers:
Closing in favour of https://github.com/octobercms/library/pull/537

Issue #4694, opened 2019-10-15
Improve responsiveness of backend elements


ghost:
Going to show two quick examples but I feel the problem extends to all the other backend elements.

### Example 1 - Buttons

I'm pretty suprised with even basic buttons in the backend are not responsive at all, see example:

![ezgif-6-7dd4903067da](https://user-images.githubusercontent.com/55155131/66845933-df176380-ef68-11e9-9aca-85c81aa3fa3a.gif)

Here we can see the search box cover the buttons. In a responsive world the buttons should be moved down a row automatically.

### Example 2 - Tables

This follows carries on from the github issue: https://github.com/octobercms/october/issues/4618

Not sure why it was closed? But there is still no solution to this issue.

In fact it gets worse when you start investigating this issue with tables and soon realize that many plugins use **partial files with the list widget** and all these tables are completely hanging out of the screen when testing October V460 (that's not been released yet).

Right now the lists widget seems to use a very old school method with css media queries and will turn on the responsive table when the screen is a certain width. Maybe the orginal design thought this would be good for mobile phones?

However, when testing tablets we have large table data hanging out of the screen and no way to scroll the data back in to view with the lists widget.

On Mobile phone devices it gets even worse and half way through the screen widths we have a css media query removing the left sidebar and then there is a huge block of screen width where you can't scroll the table at all.

Most developers will say well I use October CMS with desktop computers so who cares? Well when you create a list widget with lots of data and a few columns. I reduce the screen size down from 8K to 1920 x 1080 and the list widget data is hanging out of the screen. I only have 6 columns to see this issue. Meaning I can use the list widget for 4K and 8K computer screens and desktop screens with lower resolutuion - the data starts hanging out of the screen view and no way to pull the table data into view.

This is no good at all right now with the list widget and worse when you use a list widget in a partial file to create a secondary layer.

### Modern day solution

1. Buttons - Should not have a overflow: hidden function and get covered up.

2. Tables - Should be responsive when the data goes out of the screen view and not use a css media query to try and second guess things!

The list widget gets worse when a developer adds more columns in the table, because the list widget table will hang out of the screen at a higher screen resolution causing a bigger gap of no functionability at all.

The list widget has major issues when you start creating a second layer of data in a partial or more deeper layers. The responsive feature just goes out of the window at that point.

Basically the current setup makes a lot of features totally useless in most screen sizes.

### Out of the box

I can adjust my plugins, but these kind of features should be out of the box and applied to evereything in the backend as standard.

[...]

PR #4697, opened 2019-10-17
Show cursor as grab when buttons are draggable

ayumihamasaki2019:
Currently when you resize the screen and hover over the buttons they mouse cursor is pointer. This PR changes the mouse cursor to grab to indicate that the buttons are draggable.

As per github issue: https://github.com/octobercms/october/issues/4694

Before:

b4

After:

af

When the buttons are not draggable the mouse curor is pointer.

This just makes it easier to understand when the drag function is activated.

### DIFF
--- a/modules/system/assets/ui/less/toolbar.less
+++ b/modules/system/assets/ui/less/toolbar.less
@@ -271,3 +271,12 @@ html.mobile {
         }
     }
 }
+
+.toolbar-item.toolbar-primary.scroll-before {
+    button {
+        cursor: grab !important;
+    }
+    a {
+        cursor: grab !important;
+    }
+}
--- a/modules/system/assets/ui/storm.css
+++ b/modules/system/assets/ui/storm.css
@@ -3127,6 +3127,7 @@ html.mobile [data-control=toolbar].is-native-drag {overflow:auto;-webkit-overflo
 .compact-toolbar div.control-toolbar input.form-control.icon.search {background-position:right -78px}
 #layout-side-panel div.control-toolbar div.loading-indicator-container.size-input-text .loading-indicator,
 .compact-toolbar div.control-toolbar div.loading-indicator-container.size-input-text .loading-indicator {top:6px}
+.toolbar-item.toolbar-primary.scroll-before button, .toolbar-item.toolbar-primary.scroll-before a {cursor:grab !important}
 div.scoreboard {position:relative;padding:0}
 div.scoreboard:after,
 div.scoreboard:before {display:none;position:absolute;top:50%;margin-top:-7px;height:9px;font-size:10px;color:#bbb}


FlusherDock1:
Cursor shouldn't be with drag icon just because you hovering buttons. It can be misunderstood that buttons not clickable but draggable. I think cursor should be with drag icon when you click and hold on buttons.

ayumihamasaki2019:
@FlusherDock1 How to know if the buttons are draggable when the mouse cursor doesn't tell you - if you don't click them? The whole point of the mouse `cursor:drag` is to indicate that the element is draggable. It's common sense, that a button is clickable (that's it's purpose). But it's not obvious that the elements are draggable.

bennothommo:
@ayumihamasaki2019 I somewhat agree with @FlusherDock1's point-of-view, insofar that the primary function of the buttons is to be clicked on, so having it show as a drag icon by default is hiding the primary intention of the buttons. It made sense for the table headers purely because they're not necessarily clickable.

Additionally, it's likely that in most cases, the only people who will need to drag the toolbar buttons will be people on mobile devices, which means there is no need for a mouse icon.

ayumihamasaki2019:
@bennothommo A couple of things I disagree with you on (sorry):

>It made sense for the table headers purely because they're not necessarily clickable.

You're wrong - the table headers are clickable and when you click them the columns are sorted in order.

> Additionally, it's likely that in most cases, the only people who will need to drag the toolbar buttons will be people on mobile devices, which means there is no need for a mouse icon.

That's a poor assumption! Let's say I have a long list of buttons, when the draggable function would show up on a larger screen size. In my case sometimes I have a list of horizontal buttons in my plugins and so that would appear on a desktop device, thus a mouse cursor would be a factor.

bennothommo:
@ayumihamasaki2019 

> @bennothommo A couple of things I disagree with you on (sorry):

Haha, I welcome discussion, so by all means, disagree away. :)

> You're wrong - the table headers are clickable and when you click them the columns are sorted in order.

This is true, but in that case, I would've made the same argument as I am making now with the buttons. I must've missed this one when that change was being reviewed.

> That's a poor assumption! Let's say I have a long list of buttons, then the draggable function would show up on a larger screen size. In an example, let's say I have a list of horizontal buttons in one of my plugins, it would then appear on a desktop device, thus a mouse cursor would be a factor. Actually I do have some plugins that have 6 - 10 buttons in a row so that conflicts with your statement.

Personally, if you need more than 4-5 full-size buttons in a toolbar, I think your whole navigation could probably be refactored, either into a dropdown menu of actions, or made more contextual with certain elements in your page.

ayumihamasaki2019:
@bennothommo

> Personally, if you need more than 4-5 full-size buttons in a toolbar, I think your whole navigation could probably be refactored, either into a dropdown menu of actions, or made more contextual with certain elements in your page.

Yah I agree, personally I'm not a big fan of the current method, I think it poor design, built from an age before flexbox.

I much perfer to see the buttons use flexbox wrap method and when the screen size gets small to make the search box reduced even more. I think it's time to make all the buttons 100% width.

For example:

Above 320px screen width use flexbox wrap method.

Below 320px screen width make all buttons 100% width for each row and the search box on the first row.

----

What I'm saying is happy to keep the discussion going on the github issue. Just not a fan of the current setup!

Samuell1:
Its just to save space on smaller displays, you dont want to have every button on every new row i cant imagine how much it takes space from smaller screens. If you prefer it like that you can manually override it trough plugin or backend css i did same thing with scoreboard :D

LukeTowers:
Going to close this because the primary function of the buttons is to be clicked on. If you're on a device that has a mouse pointer and also has too little screen space to show the buttons then that tells me that you have too many buttons and should refactor it to use dropdowns or other UI elements.

I would however be fine with a PR to make the "scrollable container" the buttons are in have the grab icon so long as the buttons themselves retain the regular pointer.

PR #4698, opened 2019-10-18
Fix: Not reflecting "defaultMask" when uploading file@CMS

kmarty:
When uploading file through "CMS"->"Files"->"Add" -> "Upload file(s)", uploaded file doesn't have set file permissions according to "cms.defaultMask.file" from Config.

This patch fixes it so "defaultMask" can be different from umask and file has correctly set its permissions (usefull when required permissions are other than "644") as well as it has files/directories directly created in CMS section.

### DIFF
--- a/modules/cms/widgets/AssetList.php
+++ b/modules/cms/widgets/AssetList.php
@@ -17,7 +17,6 @@ use RecursiveIteratorIterator;
 use RecursiveDirectoryIterator;
 use DirectoryIterator;
 use Exception;
-use Config;
 
 /**
  * CMS asset list widget.
--- a/modules/cms/widgets/AssetList.php
+++ b/modules/cms/widgets/AssetList.php
@@ -665,7 +665,9 @@ class AssetList extends WidgetBase
              */
             $uploadedFile = $uploadedFile->move($this->getCurrentPath(), $uploadedFile->getClientOriginalName());
 
-            @chmod($uploadedFile, octdec(Config::get('cms.defaultMask.file')));
+            if (Config::get('cms.defaultMask.file')) {
+                @chmod($uploadedFile->getRealPath(), octdec(Config::get('cms.defaultMask.file')));
+            }
 
             $response = Response::make('success');
         }
--- a/modules/cms/widgets/AssetList.php
+++ b/modules/cms/widgets/AssetList.php
@@ -17,6 +17,7 @@ use RecursiveIteratorIterator;
 use RecursiveDirectoryIterator;
 use DirectoryIterator;
 use Exception;
+use Config;
 
 /**
  * CMS asset list widget.
@@ -663,6 +664,8 @@ class AssetList extends WidgetBase
              * Accept the uploaded file
              */
             $uploadedFile->move($this->getCurrentPath(), $uploadedFile->getClientOriginalName());
+
+            @chmod($this->getCurrentPath().'/'.$uploadedFile->getClientOriginalName(), octdec(Config::get('cms.defaultMask.file')));
 
             $response = Response::make('success');
         }
--- a/modules/cms/widgets/AssetList.php
+++ b/modules/cms/widgets/AssetList.php
@@ -663,9 +663,9 @@ class AssetList extends WidgetBase
             /*
              * Accept the uploaded file
              */
-            $uploadedFile->move($this->getCurrentPath(), $uploadedFile->getClientOriginalName());
+            $uploadedFile = $uploadedFile->move($this->getCurrentPath(), $uploadedFile->getClientOriginalName());
 
-            @chmod($this->getCurrentPath().'/'.$uploadedFile->getClientOriginalName(), octdec(Config::get('cms.defaultMask.file')));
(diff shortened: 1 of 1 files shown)


bennothommo:

kmarty:
OK, thank you. Next time to `develop` branch.
Just one question: Do I have to pay attention to "All checks have failed"? Because it "fails" because some:
```
ReflectionException: Class October\Rain\Tests\fixtures\Database\SampleClass does not exist

/home/travis/build/octobercms/october/vendor/october/rain/tests/Database/UpdaterTest.php:15
```
Which has nothing with me, I suppose.

modules/cms/widgets/AssetList.php
```diff
@@ -664,6 +665,8 @@ protected function checkUploadPostback()
              */
             $uploadedFile->move($this->getCurrentPath(), $uploadedFile->getClientOriginalName());
 
+            @chmod($this->getCurrentPath().'/'.$uploadedFile->getClientOriginalName(), octdec(Config::get('cms.defaultMask.file')));
```
LukeTowers:
```suggestion
            @chmod($this->getCurrentPath() . '/' . $uploadedFile->getClientOriginalName(), octdec(Config::get('cms.defaultMask.file')));
```

LukeTowers: @kmarty why not just use $uploadedFile->getPath()?

LukeTowers:
@kmarty we've switched to using GitHub actions instead of Travis for CI on the develop branch, so don't worry about that. When you next push a commit it will run the GitHub actions tests

modules/cms/widgets/AssetList.php
kmarty: Perhaps because of my ignorance :-/.
But anyway, _$this->getCurrentPath()_ contains destination path whilst _$uploadedFile->getPath()_ contains source path (tmp path).
Just guess, but after _$uploadedFile->move()_ the file is already not in _$uploadedFile->getPath()_.

LukeTowers: Looked into it further and `->move()` returns the new file reference, so you can do `$uploadedFile = $uploadedFile->move()` and then `$uploadedFile->getPath()` should work

kmarty: OK, I said nothing, then :-)

modules/cms/widgets/AssetList.php
```diff
@@ -662,7 +663,9 @@ protected function checkUploadPostback()
             /*
              * Accept the uploaded file
              */
-            $uploadedFile->move($this->getCurrentPath(), $uploadedFile->getClientOriginalName());
+            $uploadedFile = $uploadedFile->move($this->getCurrentPath(), $uploadedFile->getClientOriginalName());
+
+            @chmod($uploadedFile, octdec(Config::get('cms.defaultMask.file')));
```
LukeTowers: @kmarty did you test this? Shouldn't it be getPath() not the object by itself?

LukeTowers: Also should there be some sort of check to see if there's actually a valid permission value set in the config since the default is null?

kmarty: > @kmarty did you test this? Shouldn't it be getPath() not the object by itself?
Definitely not `getPath()` in this case.
Frankly, I wasn't sure whether I have to use `getRealPath()` or `__toString()` (probably cleaner(?)), but in real test, simply `$uploadedFile` works too.

LukeTowers: Let's go with getRealPath() if it works, that'll be more explicit about what's going on.

modules/cms/widgets/AssetList.php
```diff
@@ -662,7 +663,11 @@ protected function checkUploadPostback()
             /*
              * Accept the uploaded file
              */
-            $uploadedFile->move($this->getCurrentPath(), $uploadedFile->getClientOriginalName());
+            $uploadedFile = $uploadedFile->move($this->getCurrentPath(), $uploadedFile->getClientOriginalName());
+
+            if (Config::get('cms.defaultMask.file')) {
+                @chmod($uploadedFile->getRealPath(), octdec(Config::get('cms.defaultMask.file')));
```
bennothommo: @kmarty Just a minor change - we shouldn't suppress the error here.
```suggestion
                chmod($uploadedFile->getRealPath(), octdec(Config::get('cms.defaultMask.file')));
```

modules/cms/widgets/AssetList.php
```diff
@@ -662,7 +663,11 @@ protected function checkUploadPostback()
...
+            }
```
LukeTowers:
```suggestion
            File::chmod($uploadedFile->getRealPath());
```

modules/cms/widgets/AssetList.php
```diff
@@ -17,6 +17,7 @@
 use RecursiveDirectoryIterator;
 use DirectoryIterator;
 use Exception;
+use Config;
```
LukeTowers:
```suggestion
```

PR #4700, opened 2019-10-19
Preset buttons for editor toolbar settings

### LINKED ISSUE #4187: UX: Toolbar buttons comment is confusing
image

This comment suggests that a user has to add the toolbar buttons including the `[` and `]`. Doing this won't work and will fail without a visible error.

Further, this comment suggests the displayed list is the current/default configuration of the toolbar which it is not.

Since this form tab is mostly empty the options list could be moved to a separate partial.

Thoughts?

tobias-kuendig:
This is my proposal to fix #4187.

I have added three buttons below the toolbar configuration textarea that can be used to insert a preset into the textarea.

toolbar

The PR also includes a fix to a problem where the "divider" element in the toolbar was not wide enough on larger screen sizes. This lead to display errors where more toolbar rows would appear than defined in the config.

### Full (before bugfix)

grafik

### Full (after bugfix)

grafik

### Default

grafik

### Minimal

grafik


w20k:
It's tested, but needs to have an additional smoke test before merging with the latest `release`.

modules/backend/models/EditorSetting.php
```diff
@@ -76,6 +76,20 @@ class EditorSetting extends Model
         'oc-cell-thick-border' => 'Thick Border',
     ];
 
+    /**
+     * Editor toolbar presets for Froala.
+     */
+    protected $editorToolbarPresets = [
+        'default' => 'paragraphFormat, paragraphStyle, quote, bold, italic, align, formatOL, formatUL, insertTable, 
+                      insertLink, insertImage, insertVideo, insertAudio, insertFile, insertHR, html',
+        'minimal' => 'bold, italic, underline, |, insertLink, insertImage, |, html',
```
LukeTowers, line 85:
```suggestion
        'minimal' => 'paragraphFormat, bold, italic, underline, |, insertLink, insertImage, |, html',
```

LukeTowers:
@tobias-kuendig could you also manually add in the CSS changes to the compiled file since we can't recompile it without the DRM'd source?

tobias-kuendig:
Done!

LukeTowers:
Thanks @tobias-kuendig, awesome contribution as always!

PR #4703, opened 2019-10-19
Prevent Cross-Site Tracing (XST) enhacement idea

ayumihamasaki2019:
See github issue: https://github.com/octobercms/october/issues/4702

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -56,4 +56,10 @@
     RewriteCond %{REQUEST_FILENAME} !-f
     RewriteRule ^ index.php [L]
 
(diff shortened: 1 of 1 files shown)


bennothommo:
Closing this, as after further investigation, no modern installation of Apache or Nginx is going to be affected. Apache has disabled the TRACE method by default since 2013 at the very latest, and Nginx has basically never allowed TRACE unless it's been manually added. This means that it's extremely likely that anyone who has TRACE enabled *wants* it to be enabled.

PR #4712, opened 2019-10-23
Restore support for Select2 data format in custom select control

jimcottrell:
Referencing the conversation in #4413, this PR will restore the ability to use the Select2 data format in AJAX handlers while retaining the recently added support for a simple associative array of results.

It also updates the documentation with a more extensive example of the data format and (hopefully) a little more clarity.

### DIFF
--- a/modules/system/assets/ui/js/select.js
+++ b/modules/system/assets/ui/js/select.js
@@ -90,6 +90,7 @@
                         var results = data.result || data.results,
                             options = []
 
+                        delete(data.result)
                         if (results[0] && results[0].id) { // Already in Select2 format
                             options = results
                         }
--- a/modules/system/assets/ui/js/select.js
+++ b/modules/system/assets/ui/js/select.js
@@ -87,23 +87,25 @@
                         return $request
                     },
                     processResults: function (data, params) {
-                        var results = data.result;
-                        var options = [];
+                        var results = data.result || data.results,
+                            options = []
 
-                        for (var i in results) {
-                            if (results.hasOwnProperty(i)) {
-                                var isObject = i != null && i.constructor.name === 'Object';
-                                
-                                options.push({
-                                    id: isObject ? results[i].id : i,
-                                    text: isObject ? results[i].text : results[i],
-                                });
-                            };
-                        };
+                        if (results[0] && results[0].id) { // Already in Select2 format
+                            options = results
+                        }
+                        else {
+                            for (var i in results) {
+                                if (results.hasOwnProperty(i)) {
+                                    options.push({
+                                        id: i,
+                                        text: results[i],
+                                    })
+                                }
+                            }
+                        }
 
-                        return {
-                            results: options,
-                        };
+                        data.results = options
+                        return data
                     },
                     dataType: 'json'
                 }
--- a/modules/system/assets/ui/js/select.js
+++ b/modules/system/assets/ui/js/select.js
@@ -91,10 +91,10 @@
                             options = []
 
                         delete(data.result)
-                        if (results[0] && results[0].id) { // Already in Select2 format
+                        if (results[0] && typeof(results[0]) === 'object') { // Pass through Select2 format
                             options = results
                         }
-                        else {
+                        else { // Key-value map
                             for (var i in results) {
                                 if (results.hasOwnProperty(i)) {
                                     options.push({
--- a/modules/system/assets/ui/docs/select.md
+++ b/modules/system/assets/ui/docs/select.md
@@ -71,24 +71,47 @@ Use the `data-handler` attribute to source the select options from an AJAX handl
 ></select>
 ```
 
-The AJAX handler should return results as an array.
+The AJAX handler should return results in the [Select2 data format](https://select2.org/data-sources/formats).
 
 ```php
 public function onGetOptions()
 {
-    $results = [
-        [
-            'id' => 1,
-            'text' => 'Foobar',
+    return [
+        'results' => [
+            [
+                'id' => 1,
+                'text' => 'Foo',
+                'disabled' => true
+            ],
+            [
+                'id' => 2,
+                'text' => 'Bar',
+                'selected' => true
+            ],
+            [
+                'text' => 'Group',
+                'children' => [
+                    [
+                        'id' => 3,
+                        'text' => 'Child 1'
+                    ],
+                    [
+                        'id' => 4,
+                        'text' => 'Child 2'
+                    ]
+                    ...
+                ]
+            ]
+            ...
         ],
-        ...
+        'pagination' => [
+            'more' => true
+        ]
     ];
-
-    return ['result' => $results];
 }
 ```
 
-Due to the fact that JavaScript reorders numeric keys when interpreting the JSON data received by the AJAX handler, we suggest the method above for defining `results`. Support for the original `results` array format is however retained to ensure backwards compatibility.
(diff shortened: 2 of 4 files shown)


LukeTowers:
Are we able to add unit tests for each of the three supported methods?

modules/system/assets/ui/js/select.js
```diff
@@ -87,23 +87,25 @@
                         return $request
                     },
                     processResults: function (data, params) {
-                        var results = data.result;
-                        var options = [];
-
-                        for (var i in results) {
-                            if (results.hasOwnProperty(i)) {
-                                var isObject = i != null && i.constructor.name === 'Object';
```
bennothommo: @jimcottrell We'll need to maintain the Object handling, in case some people are working with that format.

modules/system/assets/ui/js/select.js
```diff
@@ -87,23 +87,25 @@
...
+                                    options.push({
+                                        id: i,
+                                        text: results[i],
+                                    })
+                                }
+                            }
+                        }
+
+                        data.results = options
+                        return data
```
bennothommo: @jimcottrell Would you be able to keep the original `return` call as is, and add any extra parameters (such as `pagination`) as required? It might seem risky sending the entire data object as is to the Select2.

modules/system/assets/ui/js/select.js
jimcottrell: I'm not clear what you're referring to.

modules/system/assets/ui/js/select.js
jimcottrell: Can you clarify what the risk would be?

modules/system/assets/ui/js/select.js
jimcottrell: Are you referring to the Select2 format as the "Object handling"? That format is just passed through to Select2.

bennothommo: Sorry @jimcottrell, I see what you are doing now. Don't worry about my comment.

modules/system/assets/ui/js/select.js
bennothommo: @jimcottrell I suppose I'm just being overly paranoid and considering that, potentially in the future, Select2 may introduce other applicable properties to their data format that may have unintended issues with our implementation. I just figured if we only looked for and passed through the pertinent keys (`results` and `pagination`) and discard the rest, we just keep the data that's expected.

jimcottrell: I can see what you mean, although actually I took the other side of that, because I wrote it this way so that if the Select2 format did add any other functionality requiring other keys, they'd still be passed through without requiring changes to `processResults`.

I can change it if you think it's important, but from my perspective, the data came from an AJAX handler specifically written to send search results to Select2, and Select2 is a library we included to begin with, so the security needs should be minimal.

bennothommo:
Just to note, here's the data types we can expect from the AJAX handler in the PHP side:

Simple key-value pair:

```
[
    'key' => 'value',
    'key2' => 'value2',
]
```

Results array:

```
[
    [
        'id' => 'key',
        'text' => 'value',
    ],
    [
        'id' => 'key2',
        'text' => 'value2',
    ],
]
```

Select2 format:

```
[
    'results' => [
        [
            'id' => 'key',
            'text' => 'value',
        ],
        [
            'id' => 'key2',
            'text' => 'value2'
        ]
    ],
    'pagination' => [
        'more' => true
    ]
]
```

As far as I can tell, there hasn't been much handling for option groups previously, but I wonder if it's also worth tackling this as well. I would think the following formats would make sense:

Results array:

```
[
    [
        'id' => 'key',
        'text' => 'value',
    ],
    [
        'text' => 'groupName',
        'children' => [
            [
                'id' => 'child1',
                'text' => 'value1',
            ],
            [
                'id' => 'child2',
                'text' => 'value2',
            ]
        ]
    ],
]
```

Select2 format:

```
[
    'results' => [
        [
            'id' => 'key',
            'text' => 'value',
        ],
            [
            'text' => 'groupName',
            'children' => [
                [
                    'id' => 'child1',
                    'text' => 'value1',
                ],
                [
                    'id' => 'child2',
                    'text' => 'value2',
                ]
            ]
        ]
    ],
    'pagination' => [
        'more' => true
    ]
]
```

jimcottrell:
I'm confused. The whole point of using the [Select2 format](https://select2.org/data-sources/formats) is that it already supports everything necessary, including option groups, and it already looks exactly like what you're proposing. My documentation change added a group to the example so that it would be clear that it could be done.

The key-value mapping at a `result` key was originally documented, but it only just started working with the recent addition of `processResults` (which of course broke the Select2 format I've been using, hence this PR), so it's now official even if it's likely no one's ever used it. The other format documented was a very limited subset of the Select2 format on the `result` key. That's supported simply by supporting the full Select2 format at either `results` or `result`.

That said, I don't know where this "results array" format comes from. It's just Select2 format without a `result` or `results` key. It wouldn't have been able to be passed directly to Select2 prior to `processResults`, and the addition of `processResults` didn't add support for it, so why is it entering the consideration of something that has to be supported? What am I missing?

bennothommo:
@jimcottrell Just proceed as is. Please make sure you provide unit tests, if possible.

jimcottrell:
I've added unit tests and a simpler documentation example. Please let me know if you feel anything is missing.

jimcottrell:
Any word on this? I hadn't thought it would be particularly controversial to restore this functionality.

LukeTowers:
@jimcottrell we've forked OctoberCMS. Feel free to resubmit your PR here: https://github.com/OctoberCMF/october. Read more about it here: https://github.com/OctoberCMF/october/issues/4

bennothommo:
@jimcottrell It definitely isn't a controversial change - we're just under the pump with a couple of other fixes at the moment, so we haven't had a chance to test this out and confirm it all works.

modules/system/assets/ui/js/select.js
bennothommo: @jimcottrell You make good points - we'll leave it as is for now.

bennothommo:
@jimcottrell Finally got a chance to test this - it works well, except for one case I found.

If the first item in the `results` array is a group item, the new `processResults` functionality fails to detect it as a Select2 data format, as it has no `id` key.

For example:
```
[
    [
        'text' => 'groupName',
        'children' => [
            [
                'id' => 'child1',
                'text' => 'value1',
            ],
            [
                'id' => 'child2',
                'text' => 'value2',
            ]
        ]
    ],
    [
        'id' => 'key',
        'text' => 'value',
    ]
]
```

Will fail to be correctly attributed as Select2-formatted data.

I will leave a comment on the line I think should change to support this - once you've made that change, if you could do a quick adjustment to unit tests to cover this type of data, then it's good to merge.

modules/system/assets/ui/js/select.js
```diff
@@ -87,23 +87,26 @@
...
-                        };
-
-                        return {
-                            results: options,
-                        };
+                        var results = data.result || data.results,
+                            options = []
+
+                        delete(data.result)
+                        if (results[0] && results[0].id) { // Already in Select2 format
```
bennothommo: @jimcottrell Would you mind changing the line to the following:
```suggestion
                        if (results[0] && ((results[0].id && results[0].text) || (results[0].text && results[0].children))) { // Already in Select2 format
```
This should correctly detect both types of Select2 data.

jimcottrell:
That's a good catch. I prefer a little simpler check, so see how this works for you.

bennothommo:
@jimcottrell Perfect! Thanks heaps :)

PR #4720, opened 2019-10-26
Implement XSRF checking for AJAX handlers

LukeTowers:
Refs #4699
Refs #4701

WIP, testing needed

### DIFF
--- a/modules/cms/classes/Controller.php
+++ b/modules/cms/classes/Controller.php
@@ -136,11 +136,29 @@ class Controller
      * Finds and serves the requested page.
      * If the page cannot be found, returns the page with the URL /404.
      * If the /404 page doesn't exist, returns the system 404 page.
+     * * If the parameter is omitted, the current URL used.
+     *
      * @param string $url Specifies the requested page URL.
-     * If the parameter is omitted, the current URL used.
-     * @return string Returns the processed page content.
+     * @return Response Returns the processed page content.
      */
     public function run($url = '/')
+    {
+        $response = $this->runInternal($url);
+
+        if (Config::get('cms.enableCsrfProtection') && $response instanceof \Symfony\Component\HttpFoundation\Response) {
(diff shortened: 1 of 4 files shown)


Samuell1:
Now question is if this is not breaking GDRP because data are saved in cookies.

modules/cms/classes/Controller.php
```diff
@@ -256,7 +266,13 @@ public function run($url = '/')
             return $result;
         }
 
-        return Response::make($result, $this->statusCode);
+        $response = Response::make($result, $this->statusCode);
+
+        if (Config::get('cms.enableCsrfProtection')) {
+            $this->addCsrfCookie($response);
```
bennothommo: @daftspunk Should we perhaps call this method `addXsrfCookie` to make it more clear what it is?

modules/cms/classes/Controller.php
```diff
@@ -1583,6 +1592,32 @@ protected function setComponentPropertiesFromParams($component, $parameters = []
     // Security
     //
 
+    /**
+     * Adds anti-CSRF cookie.
+     * Adds a cookie with a token for CSRF checks to the response.
+     * @return void
```
bennothommo: @daftspunk This would be returning a `Response` object, not void.

modules/system/assets/js/framework.js
```diff
@@ -68,6 +68,11 @@ if (window.jQuery.request !== undefined) {
             requestHeaders['X-OCTOBER-REQUEST-FLASH'] = 1
         }
 
+        var csrfToken = getCSRFToken()
```
bennothommo: @daftspunk as above, this should probably be `getXSRFToken` to make it clear it is separate to any CSRF tokens passed from input.

bennothommo:
@Samuell1 I think it would be fine, it's not personal data and it's not used for identifying any user.

PR #4723, opened 2019-10-28
Refactor relation saving in multi-relation widget

### LINKED ISSUE #4246: beforeSave event is fired twice in relationManager
- OctoberCMS Build: 451 latest dev @ [aeeb791](https://github.com/octobercms/october/commit/aeeb791899b4bbbddf114dfb24d01581022ec695)
- PHP Version: 7.2.15
- Database Engine: 10.3.13-MariaDB Homebrew
- Plugins Installed: October.Test

### Description:
When creating an item in a relation manager, the `beforeSave` event is called twice.

### Steps To Reproduce:
0. Install latest October 
0. Clone [PR Test plugin](https://github.com/CptMeatball/test-plugin/tree/before-save-called-twice)
1. Navigate to Playgrounds -> Posts
2. Create a new comment
3. Check Log

bennothommo:
Prevents the relationModel from saving until when the relation is established with `$this->relationObject->add`, preventing multiple event calls for the model.

Fixes #4246.

### DIFF
--- a/modules/backend/behaviors/RelationController.php
+++ b/modules/backend/behaviors/RelationController.php
@@ -1096,7 +1096,9 @@ class RelationController extends ControllerBehavior
 
             $modelsToSave = $this->prepareModelsToSave($newModel, $saveData);
             foreach ($modelsToSave as $modelToSave) {
-                $modelToSave->save(null, $this->manageWidget->getSessionKey());
+                if (get_class($modelToSave) !== get_class($newModel)) {
+                    $modelToSave->save(null, $this->manageWidget->getSessionKey());
+                }
             }
 
             $this->relationObject->add($newModel, $sessionKey);


modules/backend/behaviors/RelationController.php
```diff
@@ -1096,7 +1096,9 @@ public function onRelationManageCreate()
 
             $modelsToSave = $this->prepareModelsToSave($newModel, $saveData);
             foreach ($modelsToSave as $modelToSave) {
-                $modelToSave->save(null, $this->manageWidget->getSessionKey());
+                if (get_class($modelToSave) !== get_class($newModel)) {
```
LukeTowers: Does $newModel always get saved in `$this->relationObject->add($newModel, $sessionKey)`? And does the add method work if the newModel object doesn't exist yet?

bennothommo: @LukeTowers As far as I can tell, yes, the model is always saved in `$this->relationObject->save` - unless it's a `belongsTo(Many)` relation. Does the relation manager support that type of relation? The save occurs whether deferred binding is enabled or not.

This change is only for the "create" portion of the Relation manager, so `$newModel` is always effectively a new instance of the model that the relation manager is tied to. If you edit or delete the relation, this happens in different methods, and don't seem to exhibit the same doubling-up of the save events.

bennothommo:
@daftspunk When you have a moment, would you be able to let me know if I'm going down the wrong path here? I might not be considering something that you've already previously considered.

daftspunk:
Ah yes, don't do this. This will indeed break instances of some relationship types since the model will now no longer save; the `save()` method is not called as part of `add()` for some relationship types. I will take a look at this soon to see if there is an alternative solution that doesn't break anything.

bennothommo:
@daftspunk Thanks heaps. From what I could tell, it was the "belongsTo(Many)" types that don't call `save()` as part of their `add()` methods, but I wasn't sure whether they were in widespread usage for Relation widget.

daftspunk:
After giving this some thought, we could probably use an `isDirty()` check in the HasOneOrMany relation trait and this would prevent the save from happening twice; if it's not dirty, don't worry about saving anything.

daftspunk:
@bennothommo Here you go, I believe this should solve it: https://github.com/octobercms/library/commit/2133e2fd5f59150344ebb3d10943b5b62da12ad3

bennothommo:
@daftspunk Thanks a lot - confirmed your fix working. Closing this PR in favour of merging your fix.

PR #4725, opened 2019-10-28
Fix empty Richeditor class lists from breaking widget

### LINKED ISSUE #4579: Empty Image Markup Classes cause not uploading image from Froala editor
- OctoberCMS Build: 455
- PHP Version: 7.3.3
- Database Engine: 10.3.8-MariaDB-1:10.3.8+maria~xenial-log
- Plugins Installed: N/A

### Description:
If there are empty image markup classes, image upload from Froala WYSIWYG editor does not work, it throws JS error in console:

```
build-min.js?v455:formatted:13176 Uncaught TypeError: Cannot read property 'title' of null
    at FE.html (build-min.js?v455:formatted:13176)
    at _content (build-min.js?v455:formatted:8494)
    at _build (build-min.js?v455:formatted:8549)
    at Object.buildList (build-min.js?v455:formatted:8569)
    at U (build-min.js?v455:formatted:11894)
    at n (build-min.js?v455:formatted:11655)
    at HTMLImageElement.ne (build-min.js?v455:formatted:12505)
    at k (build-min.js?v455:formatted:11948)
    at HTMLImageElement.n (build-min.js?v455:formatted:12136)
    at HTMLImageElement.dispatch (jquery.min.js?v=455:2)
```
js

The problem is that this part of JS code is directly from [Froala](https://github.com/froala/wysiwyg-editor/blob/v2.9.3/js/plugins/image.min.js), so do not know if it is really OctoberCMS issue, but if I add some dummy class to list, there is not error in console and image uploads succesfully. 

Settings are stored in database in table `system_settings` under record `backend_editor_settings` as

```
"html_style_image": [
  {
    "class_label": "Foo",
    "class_name": "bar"
  }
]
```
which is then converted to JS object:

```
{
  bar: "Foo"
}
```

But if there are no classes it looks like:

```
"html_style_image": [
  []
]
```
which is then converted to JS object (first picture line 13172):

```
{
  "": null
}
```
Possible fix in OctoberCMS could be, that this strange object should be converted to empty JS object `{}`, which I tried in isolated function in console.

### Steps To Reproduce:
1. Delete all image markup classes (Settings -> Editor Settings -> Markup Classes)

markup-classes

2. Try to upload any image from Froala WYSIWYG editor (image will not be uploaded really, it is grayed out and error is thrown in console)

froala

bennothommo:
Fixes #4579.

As reported, a JavaScript error occurred if someone empties any of the class lists from the Settings -> Editor Settings -> Markup Classes section, or has entries that are missing a label or class definition. This fixes this issue, and also allows users to have NO classes for any of the class lists if they so desire - in this case, the corresponding Froala action can be clicked, but simply shows no selections.

### DIFF
--- a/modules/backend/formwidgets/RichEditor.php
+++ b/modules/backend/formwidgets/RichEditor.php
@@ -1,10 +1,12 @@
 <?php namespace Backend\FormWidgets;
 
 use App;
+use Config;
 use File;
 use Event;
 use Lang;
 use Request;
+use Backend;
 use BackendAuth;
 use Backend\Classes\FormWidgetBase;
 use Backend\Models\EditorSetting;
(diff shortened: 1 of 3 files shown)


w20k:
LGTM, @bennothommo!
@emzet maybe you could test this out?

PR #4726, opened 2019-10-29
Translate korean

tbear18:
I translated English because most Korean looked broken like "\uad75\uac8c". If this is not a broken character, cancel the Pull Request. Also, if there is something wrong, please write comments. Do I have to make my Branch a developer?


bennothommo:
@tbear18 Thanks for your efforts, but that particular language file is provided by a third-party vendor (Froala). If we update Froala, that language file will need to be overwritten anyway. I would suggest perhaps providing these language updates to the Froala project instead: https://github.com/froala/wysiwyg-editor.

PR #4733, opened 2019-11-02
Add HttpOnly Cookie Protection

ayumiTemp:
October doesn't have HttpOnly Cookie Protection.

### DIFF
--- a/config/session.php
+++ b/config/session.php
@@ -152,6 +152,19 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | HTTP Access Only
+    |--------------------------------------------------------------------------
+    |
+    | Setting this value to true will prevent JavaScript from accessing the
+    | value of the cookie and the cookie will only be accessible through
+    | the HTTP protocol. You are free to modify this option if needed.
+    |
+    */
+    
+    'http_only' => true,
+
+    /*
+    |--------------------------------------------------------------------------
     | Same-Site Cookies
     |--------------------------------------------------------------------------
     |


bennothommo:
Hello @g54qeqtg452 aka. Ayumi. Your assertion is incorrect - October has HTTP-only cookies enabled by default. Please see the `StartSession` middleware in Laravel 5.5, which October uses in its default middleware: https://github.com/laravel/framework/blob/5.5/src/Illuminate/Session/Middleware/StartSession.php#L179. Note that if the `http_only` config variable is *not* provided, it null-coalesces to `true`. The only reason to add this config variable would be to *disable* HTTP-only cookies.

With this in mind, do you still want this config value included?

g54qeqtg452:
@bennothommo 

I agree with your comment, I didn't realise October had that Middleware setup. 

> The only reason to add this config variable would be to disable HTTP-only cookies.

Hard, I can't argue that. Though I just expected October to have that setting and default have it to true.

Not sure what @daftspunk as done to this commit?

@bennothommo Need to ask you something!

I'm running tests on October right now with regards to Cookies. Google Chrome, Firefox, Opera and Safari are in the middle of changing the security settings for SameSite Cookies.

I now have a deadline of Feb 2020.

The following rules apply:

1. October Session cookie must be set to `SameSite LAX or STRICT`.

2. Third party cookies in October must be set to `SameSite None and Strict to TRUE`.

If we don't do this October gets blocked in Chrome V80+

---

> I would like to create a pull request and set SameSite to LAX as default.

---

This will then fix this upcoming issue. People can choose to make their October SameSite Strict if they want.

My websites all have SameSite Strict for their October Session cookie and no problems at all.

Let me know your thoughts.

I'm in middle of testing everything with some special dev tags open in Google Canary V80.

g54qeqtg452:
@bennothommo Oh forgot to say, I found a security issue with the settings section sidebar in the backend.

When you collapse things in the sidebar in the settings section it creates an insecure cookies called: `sidenav_treegroupStatus` and `sidenav_treesearch`

These need to be fixed as they will get blocked in browsers and also allow bad things to happen when a user goes from a backend to the frontend to a backend.

It would be nice to have those two cookies set to:

HttpOnly - true
Secure - true
SameSame - Strict

To give complete protection to users.

g54qeqtg452:
@bennothommo Oh forgot to say, I found a security issue with the settings section sidebar in the backend.

When you collapse things in the sidebar in the settings section it creates an insecure cookies called: `sidenav_treegroupStatus` and `sidenav_treesearch`

Cookies created over here: https://github.com/octobercms/october/blob/master/modules/backend/assets/js/october.sidenav-tree.js

Better to create them using Laravel instead, javascript only creates insecure cookies (which will end up getting blocked by all browsers soon).

These need to be fixed asap and also allow bad things to happen when a user goes from a backend to the frontend to a backend.

It would be nice to have those two cookies set to:

HttpOnly - true
Secure - true
SameSame - Strict

To give complete protection to users.

See proof:

image

**Note**: My october_session cookie on the same website in the screenshot has:

HttpOnly - true
Secure - true
SameSame - Strict

LukeTowers:
@g54qeqtg452 I've warned you several times on multiple occasions to report security issues through our established security policy. Stop doing it in GitHub issues. We're more than happy to work with reporters on issues they consider security related but we will do so through our established processes.

g54qeqtg452:
@LukeTowers @daftspunk I emailed both of you with regards to my last two comments here. Please let me know in email reply. Then I happy to help get this issue sorted out...

g54qeqtg452:
Also forgot to say, this code line is wrong and out-dated now:

https://github.com/octobercms/october/blob/c1fe12f732278df8a706fccf62baae8cd918ca30/config/session.php#L166

It should be changed to:

```
Supported: "none", "lax", "strict" 
```

Adding the extra info of:

1. October Session cookie must be set to SameSite LAX or STRICT.
2. Third party cookies in October must be set to SameSite None and Strict to TRUE.

I feel would help users understand better to setup the cookie settings, so I was thinking of doing this:

```
    /*
    |--------------------------------------------------------------------------
    | Same-Site Cookies
    |--------------------------------------------------------------------------
    |
    | This option determines how your cookies behave when cross-site requests
    | take place, and can be used to mitigate CSRF attacks. By default, we
    | do not enable this as other CSRF protection services are in place.
    |
    | In the strict mode, the cookie is not sent with any cross-site usage
    | even if the user follows a link to another website. Lax cookies are
    | only sent with a top-level get request.
    |
    | October session cookie must be set to SameSite LAX or STRICT.
    | Third party cookies in October must be set to SameSite None and Strict to TRUE.
    |
    | Supported: "none", "lax", "strict"
    |
    */
    'same_site' => strict,
```

(I've written this here as I have sent the emails to you both a few minutes ago).

bennothommo:
@g54qeqtg452 Can you provide us a link to the Chrome and Firefox repos that dictate this upcoming requirement of SameSite cookies?

g54qeqtg452:
@bennothommo I gave you the link already - but happy to give it again with extra info:

https://www.chromium.org/updates/same-site
https://groups.google.com/forum/#!msg/mozilla.dev.platform/nx2uP0CzA9k/BNVPWDHsAQAJ
https://groups.google.com/a/chromium.org/forum/#!msg/blink-dev/AknSSyQTGYs/8lMmI5DwEAAJ
https://blog.chromium.org/2019/10/developers-get-ready-for-new.html

Work started on Chrome 76 and Firefox 69. The completed rollout is in Chrome v80 and Firefox v--- (I need check my notes for the exact version number for Firefox, I read it a few days ago).

Work on other browsers such as Safari and Opera are also going on and you can see a list of browser issues here: https://www.chromium.org/updates/same-site/incompatible-clients

---

To answer your comment more directly. It all depends on two main issues I feel.

1. Cookies on-site and not being used on another website needs to be set as follows:

HttpOnly - true
Secure - true
SameSame - Strict or Lax

> My thinking is to change the default settings in the config to try and avoid browser blocking. I would like to see Strict being set and if need be users can change it down to LAX.

2. Cookies that are used as third party, need to be set as follows:

HttpOnly - true
Secure - true
SameSame - none

> My thinking is that plugins that are embed on websites or pull in third-party scripts would need to be changed to allow them to work, in the same way as CORS setup. So some plugins may need adjusting their code if they use cookies and depending on if their data is being used for embeds etc.

Hope that helps, feel free to ask me more if needed.

Issue #4718, opened 2019-10-25
RelationController remove single relation UI


madmicrowave:
- OctoberCMS Build: 458
- PHP Version: >=7.1
- Database Engine: >= MySql 5.7

So I have following problem so relation type belonsTo and use RelationController view, with toolbar buttons "create|add|remove" after click on remove relation is detached on db properly but ajax request still show that user data.

### Steps To Reproduce:
1. Setup bolongsTo relation ans build simple form
2. Link relation any type  create|add
3. Click remove, form still show attached relation data, if i reload page user data is not filled, double click on remove data is also removed

LukeTowers:
What? I don't understand what the issue is. Could you provide a step by step of what you're trying to do along with screenshots / gifs?

MaksimsGee:
So basically has following relation: Order -> belongsTo|hasOne -> User

And happens following:
1. attach relation
2. delete|remove relation(and it actually removed from first click)
3. form don`t refresh relation after update and respond form with filled fields

![our-zzzzz-1](https://user-images.githubusercontent.com/11761923/67851989-49f19e80-fb14-11e9-9549-276819ef980e.gif)

LukeTowers:
Could be related to #3470 and #3476

LukeTowers:
@paishin is this the same issue as #3476?

DanHarrin:
I'm having this same issue, I'll have a look and see if I can PR a fix.

w20k:
@DanHarrin could you double-check if PR:https://github.com/octobercms/october/pull/3476 fix your issue?

DanHarrin:
@w20k thanks for that, it does fix the issue on "unlink", but it's still present on "delete".

w20k:
@DanHarrin if you'd have time to make a PR for the. Ping me. Will review and test ;)

DanHarrin:
All done @w20k 😄

w20k:
Omg, your crazyy fast 😄 Thanks, @DanHarrin!

bennothommo:
Confirmed bug. Fixed by #4741, thanks to @DanHarrin.

PR #4741, opened 2019-11-06
Refresh Relation Manager on Unlink and Delete

### LINKED ISSUE #3470: BelongsTo (HasMany) Relation Unlink and Display Issues
##### Reproduce steps
https://stackoverflow.com/questions/49492510/octobercms-belongsto-hasmany-relation-issues

##### October build
434

Following the description from linked Stackoverflow issue and by @LukeTowers suggestion I tried to reproduce the issue using  [test plugin](https://github.com/daftspunk/oc-test-plugin).

I slightly changed `Posts` `Status` field and got this:

image

So, the layout issue was not reproduced (probably it is connected with specific styles on RainLab/BlogPost page), but the unlinking issue is still there. Here's the demo video: https://youtu.be/Zz9oOCf3UN4

This is the only change that I made to oc-test-pugin in plugins/october/test/models/post/fields.yaml:

from 

```
status:
        label: Status
        type: partial
        path: $/october/test/controllers/posts/_status.htm
        span: auto
        commentAbove: Generic model filtered by relation conditions.
        context: [create, update, relation]
```

to

```
status:
        label: Status
        type: relation
        span: auto
        commentAbove: Generic model filtered by relation conditions.
        context: [create, update, relation]
```

DanHarrin:
This fixes #3470 and #4718 and by refreshing the relation manager's form when the user unlinks or deletes the related record.

### DIFF
--- a/modules/backend/behaviors/RelationController.php
+++ b/modules/backend/behaviors/RelationController.php
@@ -1183,6 +1183,7 @@ class RelationController extends ControllerBehavior
                 $relatedModel->delete();
             }
 
+            // Reinitialise the form with a blank model
             $this->initRelation($this->model);
 
             $this->viewWidget->setFormValues([]);
@@ -1280,6 +1281,7 @@ class RelationController extends ControllerBehavior
                 $this->relationObject->dissociate();
                 $this->relationObject->getParent()->save();
 
+                // If the relation manager isn't using deferred binding, reinitialise the form with a blank model
                 if (is_null($sessionKey)) {
                     $this->model->refresh();
                     $this->initRelation($this->model);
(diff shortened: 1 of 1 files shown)


LukeTowers:
Please add a comment to both sections explaining the purpose of the added code

DanHarrin:
Done @LukeTowers

bennothommo:
Will give this a test tonight :)

bennothommo:
@DanHarrin @LukeTowers 1 month later, I've finally tested it and confirmed it working.

Test scenario:
- Using the Test plugin, go to "Posts" and then click a post in the list.
- In Alternative Views at the top, click "Relation Controller" to change the Status field to a relation controller widget.
- Link a status to the Post.
- Click Unlink. Before the fix, the linked status would still remain after unlinking, until the page was refreshed. With this fix, the linked status is removed.

I ran the same test with the delete action by modifying the available buttons for that Status relation controller widget. It too works with the fix in the place.

Issue #3952, opened 2018-12-01
[UX] Add "Clear filters" button to filter


vosco88:
When you have couple of filter scopes with something selected it is a bit annoying to have to click through all of them to clear them. 
My suggestion would be to add a button to the filter toolbar that would clear all filters. 

I was trying to look on how to approach this as I would like to code it but probably would need some guidance and help. As I am not sure how to approach clearing all.

w20k:
Hi @vosco88  could you add a screenshot of this issue? Would be a bit easier to understand. It might be fixed in the dev branch already 😄

LukeTowers:
@vosco do you have a mockup of how you would like envision it looking? Ideally two mockups, one with one filter and one with a lot of filters

vosco88:
@LukeTowers I was envisoning it as something like this quick mockup 
![image](https://user-images.githubusercontent.com/14372933/49332908-fb5f1b80-f5b5-11e8-8f21-ba89a5be3697.png)

I do not think I would specify it differently with different amount of filters, but basic idea is you click the button  and it clears all applied filters from the list. 

@w20k it is not an issue per say, more a feture request :)

LukeTowers:
@vosco88 I would probably want it at the end, or as a button outside of the filter container

bennothommo:
Closing as it has been over a month since any activity on this occurred and we are trying to figure out what issues are still relevant. If this is still something that you would like to see through to fruition please respond and we can get the ball rolling.

LukeTowers:
@Teranode or @alxy is this something you'd be interested in looking into?

w20k:
Could be also great addition to add `clear` (cross) button to search input :)

Samuell1:
@w20k clear button on search input is already there

Eoler:
> clear button on search input is already there

Not on Number/Text filter scopes.

w20k:
@robinbonnes do it :) Added it to my todo list, but it's way at the bottom!

msimkunas:
@robinbonnes Are you still planning to make that PR sometime?

robinbonnes:
> @robinbonnes Are you still planning to make that PR sometime?

I've supplied a screenshot and asked to reopen the PR: https://github.com/octobercms/october/pull/4744

I'm using the following myself in a plugin, to add the clear filter button globally for all lists that have list filters:

Add `/filters/_scope_button.htm` in your plugin:
```
<!-- Button scope -->
<a
[...]

PR #4744, opened 2019-11-06
Added new 'clear'-scope as requested in #3952

robinbonnes:
…/october/issues/3952.

Format in config_filter.yaml:

```
    clear:
        label: Clear filters
        type: clear

```

### DIFF
--- a/modules/backend/widgets/Filter.php
+++ b/modules/backend/widgets/Filter.php
@@ -264,9 +264,9 @@ class Filter extends WidgetBase
                 $value = post('options.value.' . $scope->scopeName) ?: null;
                 $this->setScopeValue($scope, $value);
                 break;
-				
+
             case 'clear':
-                foreach($this->getScopes() as $scope) {
+                foreach ($this->getScopes() as $scope) {
                     $this->setScopeValue($scope, null);
                 }
 
--- a/modules/backend/widgets/Filter.php
+++ b/modules/backend/widgets/Filter.php
@@ -264,6 +264,14 @@ class Filter extends WidgetBase
                 $value = post('options.value.' . $scope->scopeName) ?: null;
                 $this->setScopeValue($scope, $value);
                 break;
+				
+            case 'clear':
+                foreach($this->getScopes() as $scope) {
+                    $this->setScopeValue($scope, null);
+                }
+
+                return ['.control-filter' => $this->render()];
+                break;
         }
 
         /*
--- a/modules/backend/widgets/Filter.php
+++ b/modules/backend/widgets/Filter.php
@@ -270,8 +270,7 @@ class Filter extends WidgetBase
                     $this->setScopeValue($scope, null);
                 }
 
-                return ['.control-filter' => $this->render()];
-                break;
+                return ['.control-filter' => preg_replace('/^<[^>]+>|<\/[^>]+>$/', '', $this->render())];
         }
 
         /*
--- a/modules/backend/widgets/Filter.php
+++ b/modules/backend/widgets/Filter.php
@@ -187,6 +187,7 @@ class Filter extends WidgetBase
         }
 
         $scope = $this->getScope($scope);
+        $updateScopePartial = false;
 
         switch ($scope->type) {
             case 'group':
@@ -267,10 +268,21 @@ class Filter extends WidgetBase
 
             case 'clear':
                 foreach ($this->getScopes() as $scope) {
-                    $this->setScopeValue($scope, null);
+                    $value = null;
+
+                    switch ($scope->type) {
+                        case 'checkbox':
+                            $value = false;
+                            break;
+                        case 'switch':
+                            $value = '0';
+                            break;
+                    }
+
+                    $this->setScopeValue($scope, $value);
                 }
 
-                return ['.control-filter' => preg_replace('/^<[^>]+>|<\/[^>]+>$/', '', $this->render())];
+                $updateScopePartial = true;
         }
 
         /*
@@ -279,6 +291,11 @@ class Filter extends WidgetBase
         $params = func_get_args();
 
         $result = $this->fireEvent('filter.update', [$params]);
+
+        if ($updateScopePartial) {
+            $this->prepareVars();
+            $result[] = ['.control-filter' => $this->makePartial('filter_scopes')]; 
+        }
 
         if ($result && is_array($result)) {
             return call_user_func_array('array_merge', $result);
--- a/modules/backend/widgets/Filter.php
+++ b/modules/backend/widgets/Filter.php
@@ -267,10 +267,10 @@ class Filter extends WidgetBase
                 break;
 
             case 'clear':
-                foreach ($this->getScopes() as $scope) {
+                foreach ($this->getScopes() as $s) {
                     $value = null;
 
-                    switch ($scope->type) {
+                    switch ($s->type) {
                         case 'checkbox':
                             $value = false;
                             break;
@@ -279,7 +279,7 @@ class Filter extends WidgetBase
                             break;
                     }
 
-                    $this->setScopeValue($scope, $value);
+                    $this->setScopeValue($s, $value);
                 }
 
                 $updateScopePartial = true;
--- a/modules/backend/widgets/Filter.php
+++ b/modules/backend/widgets/Filter.php
@@ -294,7 +294,7 @@ class Filter extends WidgetBase
 
         if ($updateScopePartial) {
             $this->prepareVars();
-            $result[] = ['.control-filter' => $this->makePartial('filter_scopes')]; 
+            $result[] = ['.control-filter' => $this->makePartial('filter_scopes')];
         }
 
         if ($result && is_array($result)) {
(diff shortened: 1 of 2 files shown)


bennothommo (changes requested):
@robinbonnes Please review and fix the code style issues reported here:
https://github.com/octobercms/october/pull/4744/checks?check_run_id=291252747#step:6:8

robinbonnes:
@bennothommo fixed ✔️

Samuell1:
I think this will be better as option in config_list.yaml something like `allowResetFilter: true`.

robinbonnes:
Thats also an idea, however on this way you can specify where to place the clear filter button in the filter list (start, end) and what the button text should be (the label).

Now it does reset all filters to `null`, so it totally clears the filters. I think resetting them back to default is better?

LukeTowers:
@robinbonnes can you post a screenshot?

Samuell1:
@robinbonnes Yeah, it should reset to default because if something cant be null then you get DB errors.

modules/backend/widgets/Filter.php
```diff
@@ -264,6 +264,14 @@ public function onFilterUpdate()
                 $value = post('options.value.' . $scope->scopeName) ?: null;
                 $this->setScopeValue($scope, $value);
                 break;
+
+            case 'clear':
+                foreach ($this->getScopes() as $scope) {
+                    $this->setScopeValue($scope, null);
+                }
+
+                return ['.control-filter' => $this->render()];
```
bennothommo: @robinbonnes ~~Would you mind putting the array key and value on a new line - it makes it more readable as an array.~~ Never mind, see discussion below.

modules/backend/widgets/Filter.php
```diff
@@ -264,6 +264,14 @@ public function onFilterUpdate()
...
+                break;
```
bennothommo: @robinbonnes A `break` directive is not needed if there's a `return` above it in the same scope.
```suggestion
```

bennothommo (dismissed):

modules/backend/widgets/Filter.php
LukeTowers: @bennothommo personally I don't mind the oneliner array in some cases, especially when there is only one element. Thoughts?

bennothommo: @LukeTowers It's subjective, but I justify it in a couple of ways:

- Although it's easier to tell, at a glance, the above as being an array with the `=>` symbol - it's harder when you have two parameters instead, ie. `return ['.control-filter', $this->render()];`
- We already require arrays to split over multiple lines when there's more than one item, so this is just consistent with that.
- It's similar formatting to method signatures that span longer lines, ie:

```
public function someMethod(
    $someReallyLongAttribute = 'A really long default parameter'
) {
```

LukeTowers: @bennothommo counterpoint to your points :)
> Although it's easier to tell, at a glance, the above as being an array with the => symbol - it's harder when you have two parameters instead, ie. return ['.control-filter', $this->render()];

If it's two elements, it should be two lines, no matter how long it is.

> We already require arrays to split over multiple lines when there's more than one item, so this is just consistent with that.

Not quite, since this is technically one item, with a key provided.

> It's similar formatting to method signatures that span longer lines

I sincerely hope that example does not exist in October anywhere; I do not ever want to see that added in the future. If something is that long of a method parameter / default value, you're doing something wrong with naming / defaults assignment.

bennothommo: @LukeTowers they're all fair points (as for point 3, I haven't seen any yet, but rest assured I will deal with that if I ever saw it too).

I'll probably still do it the way I've suggested, because I do find it spaces out the definitions a bit more clearer, but I won't ping anyone if they have implemented it as a one-liner.

LukeTowers:
@robinbonnes please also post a screenshot of what this looks like and add a PR to the [test plugin](https://github.com/octoberrain/test-plugin)

bennothommo:
@robinbonnes Are you still interested in finishing off this PR?

msimkunas:
I'm not sure if this PR is still being worked on, but here goes...

The `Backend\Widgets\Filter` class extends the `WidgetBase` class which uses a trait that provides the `resetSession` method which is supposed to reset all session data related to the current widget. Perhaps we can use this method instead of manually setting the scope values back to their defaults?

robinbonnes:
> @robinbonnes please also post a screenshot of what this looks like and add a PR to the [test plugin](https://github.com/octoberrain/test-plugin)

image

Although the PR needs some adjustments... Can we reopen this?

msimkunas:
In my testing, [this](https://github.com/octobercms/october/blob/2917482b35414bc8fa424022b91abce6fee855c9/modules/backend/widgets/Filter.php#L273) line only updates the scope container. The content of the backend list, however, remains the same.

Removing this line inverts the issue: the list contents are refreshed but the scope container no longer updates.

msimkunas:
In order to circumvent the above issue, we should not abruptly return the partial update before the `filter.update` event has had its chance to run at the end of the `onFilterUpdate` method. 

The following works (and is a bit cleaner):

```
$updateScopePartial = false;

switch($scope->type) {
    ...
    case 'clear':
        foreach ($this->getScopes() as $scope) {
            $this->setScopeValue($scope, null);
        }

        $updateScopePartial = true;
        break;
}

$params = func_get_args();

$result = $this->fireEvent('filter.update', [$params]);

if($updateScopePartial) {
    $this->prepareVars();
    $result[] = ['.control-filter' => $this->makePartial('filter_scopes')]; 
}

if ($result && is_array($result)) {
    return call_user_func_array('array_merge', $result);
}
```

The `$updateScopePartial` controls whether the scope container should be updated since we don't want to trigger partial updates on every scope request. Also, only the inner partial of the scope container is rerendered.

@robinbonnes Currently I don't have time to make a PR for this. Would you update your PR with these changes?

msimkunas:
I have also noticed that there is a bug regarding scopes with dropdown options. This PR with the changes I've suggested above clears the actual filters in the session, however, the previously selected dropdown items are still selected in the backend UI. This selection is only applied when filtering the list again though.

msimkunas:
This PR would also benefit from an optional `resetDefault: true` flag which, if true, would reset the scope values back to their **defaults** or clear them otherwise.

If you don't have time to continue working on this PR @robinbonnes, I could give it a go soon so do let me know.

msimkunas:
One more thing. The `checkbox` and `switch` scopes must be handled differently, e.g.:

```
foreach($this->getScopes() as $scope) {
    $value = null;

    switch($scope->type) {
        case 'checkbox':
            $value = false;
            break;
        case 'switch':
            $value = '0';
            break;
    }

    $this->setScopeValue($scope, $value);
}
```

Using `null` as a value for these scopes does not properly reset them - those scopes are reset to their default values instead which may or may not be the desired behavior. This can be controlled with a flag (see my comment above).

robinbonnes:
@msimkunas Thanks for your suggestions, really appreciate it! I've added it to the PR.

Some questions:
- Why add `$this->prepareVars();`?
- Can't a switch/checkbox be default: `true`?

Ofcourse you are free to work on this PR.

modules/backend/widgets/Filter.php
```diff
@@ -264,6 +265,24 @@ public function onFilterUpdate()
                 $value = post('options.value.' . $scope->scopeName) ?: null;
                 $this->setScopeValue($scope, $value);
                 break;
+
+            case 'clear':
+                foreach ($this->getScopes() as $scope) {
```
LukeTowers: I think this variable name would have to be different because otherwise it would be overriding the earlier $scope variable wouldn't it?

msimkunas:
@robinbonnes The `prepareVars` method defines the scopes which are used inside the partial.

As for the default switch/checkbox value – I believe that the user setting should be respected here and the default (if provided) should be retrieved from the scope configuration.

msimkunas:
Also bear in mind that this PR needs to reset the filter values on the client side in order to be fully functional. This requires one to somehow reinitialize the JS scope widget and IIRC the existing JS widget requires significant changes for this to be possible (i.e. it needs to unbind event handlers and properly dispose of itself upon destruction).

Another approach would be to somehow detect partial updates from within the widget and clear the scopes accordingly.

LukeTowers:
@robinbonnes 

> The Backend\Widgets\Filter class extends the WidgetBase class which uses a trait that provides the resetSession method which is supposed to reset all session data related to the current widget. Perhaps we can use this method instead of manually setting the scope values back to their defaults?

What was the problem with that approach?

msimkunas:
@LukeTowers The problem, I think, is that it's only one part of the solution. See [my comment](https://github.com/octobercms/october/pull/4744#issuecomment-761726267) for more details.

robinbonnes:
Doesn't the `resetSession` function also reset the list settings etc.? Maybe for now setting the filters back to their defaults and refreshing the filter partial is enough to merge this? Then in the future someone can make the JS adjustements and maybe modify the `resetSession` function?

daftspunk:
Thanks for this @robinbonnes and your insights @msimkunas 

This has been merged for the next release in v2.2

PR #4748, opened 2019-11-07
Adds the ability to override the Lists prepareQuery in list export

fansaien:
The controller can create or modify the QueryBuilder.

We already can load other columns.yaml to add new columns in the `makeLists()` method.  We also need to change the default Lists widget QueryBuilder to export the records.

### DIFF
--- a/modules/backend/behaviors/ImportExportController.php
+++ b/modules/backend/behaviors/ImportExportController.php
@@ -640,7 +640,8 @@ class ImportExportController extends ControllerBehavior
             ? 'getColumnValueRaw'
             : 'getColumnValue';
 
-        $query = $widget->prepareQuery();
+        $query = $this->controller->methodExists('prepareQuery') ? $this->controller->prepareQuery()
+            : $widget->prepareQuery();
         $results = $query->get();
 
         if ($event = $widget->fireSystemEvent('backend.list.extendRecords', [&$results])) {


LukeTowers:
@fansaien you already have access to the Lists widget & therefore its query builder through the controller. The `$widget` instance is pulled directly from `$controller->makeLists()` which initializes the widget through the regular ListController behavior methods meaning that both `listExtendQueryBefore()` and `listExtendQuery()` apply to the query in this instance as well. 

Please provide more information on why you think this is a necessary change and what exactly you are trying to achieve that can't be achieved with the existing code.

fansaien:
@LukeTowers Yes, I can create the widget from `makeLists()`. But I do not want to reuse the `prepareQuery()` in the Lists widget. 

For example, the Bookstore has many books and each book has many authors.
In the books list page, I would like to add the "Export authors" button on each row. 
In this case, the widget's model should be Author, not Book. (This can be set on the makeLists())
Furthermore, I also want to override the prepareQuery() in the controller, because this function focuses on the Books List not the Staff list.

LukeTowers:
Screenshots please of what you're trying to do along with a sample of what you get as output right now vs a sample of what output you're trying to get.

fansaien:
@LukeTowers 
Please review this sample:

staff

- This event has two shifts, and it lists all the shifts with the Relation Controller.
- The "Test Shift One" has two shift slots, and the "Test Shift Two" has three shift slots.
- Each shift slot has been assigned only one staff.

I added the "export staff" button on each shift. This button only exports the staff info, not for shifts or shift slots.

So, I do not want to run the `widget->prepareQuery()` method in the ImportExportController.php, and try to find a way to override this method.

LukeTowers:
@fansaien what you should do then is [create a separate list definition](https://octobercms.com/docs/backend/lists#multiple-list-definitions) and [tell the export behavior to use that definition](https://octobercms.com/docs/backend/import-export#list-behavior-integration). Then you can (in your controller's `listExtendQuery($query, $definition)` method) use the provided definition to decide what modifications you wish to make to the query.

Example:
```
export:
    useList:
        definition: staff_on_shift
```

```php
public function listExtendQuery($query, $definition)
{
     if ($definition ===  'staff_on_shift') {
        // Do custom query extension stuff here
    }
}
```

fansaien:
@LukeTowers, thank you for your suggestions. 

I can not just do that because the widget's model is not a Staff model and it should be the Shift model.
I can override the `makeLists()` to write the right model, but I would like to create my QueryBuilder. Sometimes, it may join many other tables and not based on the current list, and I do not need to check the filters and searching words.

LukeTowers:
@fansaien I think you're missing what I'm saying. You can create a separate list definition that is exactly the same as the original (so it would be referencing a Shift model & not a Staff model) and just change its name so you can distinguish between it when modifying the query. You can also return an entirely different query object from the listExtendQuery method, you don't have to just modify the one provided to you. As such, this change is not necessary (and as a side note is too vague to be accepted in the first place, prepareQuery as a method on the controller does not provide enough information on what is happening, although the main reason for denying this change is that it's unnecessary since you can achieve what you're trying to do with the existing methods available to you).

fansaien:
@LukeTowers 
I understand the `listExtendQuery()` function is working.
But I mean I do not want to run the whole `prepareQuery()` function. It is almost 230 lines and tries to check lots of things which unnecessary.

LukeTowers:
@fansaien the amount of lines that a given method contains has nothing to do with anything. Just to prove my point, if you remove all comments and empty lines from that method you end up with less than 100. Attempting to justify something based on the number of lines of code has no bearing on reality. The actual work being done by that method does not include anything that would be a hinderance on performance whatsoever.

There are actually a number of things done in that method that are essential for all of October's features to work correctly, and as I said before if you truly don't care at all about the query object generated by that method then all you have to do is return your own query object from the listExtendQuery() method and the widget will use that query object instead.

PR #4750, opened 2019-11-08
Dynamic TagList Options

DanHarrin:
This PR implements support for a `get*FieldName*Options()` method that returns dynamic options for the TagList widget, much like it does on the dropdown form field (see [the documentation](https://octobercms.com/docs/backend/forms#field-dropdown)).

For example, this method returns dynamic options for the `tags` TagList widget:
```
public function getTagsOptions($value, $formData)
{
    return ['one', 'two', 'three', 'four', 'five'];
}
```

### DIFF
--- a/modules/backend/formwidgets/TagList.php
+++ b/modules/backend/formwidgets/TagList.php
@@ -1,7 +1,6 @@
 <?php namespace Backend\FormWidgets;
 
 use Backend\Classes\FormWidgetBase;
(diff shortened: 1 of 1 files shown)


bennothommo:
@DanHarrin was there any particular reason for closing this PR?

DanHarrin:
@bennothommo I have no idea why but I must have misread the documentation, it seems like this has already been implemented?

PR #4751, opened 2019-11-08
Avoid array_intersect error if argv is not defined

ChVuagniaux:
I rencontered this issue because, I'm working on a way to run [Laravel HTTP tests](https://laravel.com/docs/6.x/http-tests) on OctoberCMS and for Backend work properly, I have to rerun few october Service provider.

I know my usage is not commun (and not very recommended), but this fix seems legit for avoid fatal PHP error.

Anyway thanks to have take few minutes to read me ;)

### DIFF
--- a/modules/system/ServiceProvider.php
+++ b/modules/system/ServiceProvider.php
@@ -153,7 +153,7 @@ class ServiceProvider extends ModuleServiceProvider
         /*
          * CLI
          */
-        if (App::runningInConsole() && count(array_intersect($commands, Request::server('argv'))) > 0) {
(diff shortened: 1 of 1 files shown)


LukeTowers:
LGTM! Let me know how the HTTP tests stuff works out for you, sounds pretty interesting.

ChVuagniaux:
@LukeTowers Thanks for merging. 

We have made a plugin for be able to run Dusk, HTTP & Unit tests on OC as we do on a Laravel App.
https://github.com/inetis-ch/oc-testing-plugin

PR #4753, opened 2019-11-08
Removed double checking if file is protected in FileUpload widget

tomaszstrojny:
Don't know why there are this many commits (those were merged in ~March), but as it's one file change I guess it's all right.

The change fixes two things:

1. Getting path from the file model using `$file->getPath()` and `$file->getThumb(...)` already checks if the file is protected. It's been done in `System/Models/File` so it doesn't have to be checked again here,
1. `FilesController` was hardcoded here. Without the fix, we couldn't change the default file model for protected files (and that's how I found the bug here).

Regards,

Tomasz Strojny


LukeTowers:
@tomaszstrojny looks good, catches up the FileUpload widget with the changes made here: https://github.com/octobercms/october/commit/4f7c5cc1e753817beda9b9c914d7d8b28b00dd2c.

Could you remove the `FilesController` import at the top since it's no longer used?

tomaszstrojny:
Sure, totally forgot about it. Done.

PR #4754, opened 2019-11-09
Linkable form tabs

Rike-cz:
This PR enables linking form tabs via URL fragments. When URL fragment has `#tab-name`, tab with `data-tab-url="#tab-name"` is shown on page load. URL fragment is changed by clicking on tab nav. I am using part of solution used in backend/system/mailtemplates (data-tab-url and history) here.

This feature has to be switch on in form.yaml setting:
```
tabs:
    linkable: true
```


modules/backend/widgets/form/partials/_form_tabs.htm
```diff
@@ -12,10 +12,10 @@
     }
 ?>
 <div class="<?= $navCss ?>">
-    <ul class="nav nav-tabs">
+    <ul class="nav nav-tabs" <?= $tabs->linkable ? 'data-linkable' : '' ?>>
         <?php $index = 0; foreach ($tabs as $name => $fields): ?>
             <li class="<?= $index++ == 0 ? 'active' : '' ?>">
-                <a href="#<?= $type . 'tab-' . $index ?>">
+                <a href="#<?= $type . 'tab-' . $index ?>" <?= $tabs->linkable ? 'data-tab-url="#' . str_slug(trans($name)) . '"' : '' ?>>
```
LukeTowers: @Rike-cz I think it might actually be a better idea to not translate the `$name` before slugging it because that would create different deep links for users with different backend languages.

Rike-cz: ok, but something like #company_plugin_lang_tabs_tab1 doesn't look good. So do you prefer something like icons have? https://octobercms.com/docs/backend/forms#form-tab-options

LukeTowers: @Rike-cz I don't think it really matters what it looks like, so long as it works. I'd prefer not to add a separate config section for it because then you'd have to add each linkable tab to the config in order for that tab's link to work. A lot of modern browsers completely hide that information from the end users in the first place.

Rike-cz: ok, last attempt - remove lang prefix at least: `'data-tab-url="#' . preg_replace('#::lang\.(.+)$#', '\\1', $name) . '"'` :-)

modules/system/assets/ui/js/tab.js
```diff
@@ -84,6 +84,10 @@
         })
 
         this.updateClasses()
+
+        if (this.$tabsContainer.is('[data-linkable]') && location.hash) {
```
w20k: @Rike-cz would be better to switch those, micro speed optimization: `if (location.hash && this.$tabsContainer.is('[data-linkable]') ) { `

Wouldn't it be better to use has `hasClass` instead of `is`? just to check if  `tabsContainer` has`data-linkable` class?

LukeTowers: @w20k `data-linkable` is an attribute, not a class.

Rike-cz: OK, so I will swap conditions only, no problem. But bellow I see "This branch has conflicts that must be resolved". Whats wrong there?

seanthepottingshed: Oooo - this excites me!

LukeTowers: Fixed it for you @Rike-cz

w20k:
LGTM, needs testing ;)

bennothommo:
Tested it and it works well. It only covers one set of tabs (ie. you can make both the primary and secondary linkable, but it will only cover the last clicked tab) but that's fine.

LukeTowers:
Great, glad to hear it! @bennothommo @w20k which of you two want to sponsor the merge?

LukeTowers:
Hmm, wait a sec. Why is there an href for the tab and a separate data-tab-url?

Rike-cz:
@LukeTowers Because I wanted something readable previously and url is automatically generated only. I can change it.

modules/backend/widgets/form/partials/_form_tabs.htm
```diff
@@ -12,13 +12,14 @@
...
 ?>
 <div class="<?= $navCss ?>">
-    <ul class="nav nav-tabs">
+    <ul class="nav nav-tabs"  <?= $tabs->linkable ? 'data-linkable' : '' ?>>
         <?php $index = 0; foreach ($tabs as $name => $fields):
             $lazy = in_array($name, $tabs->lazy);
         ?>
             <li class="<?= ($index++ === 0) ? 'active' : '' ?> <?= $lazy ? 'tab-lazy' : '' ?>">
                 <a
                     href="#<?= $type . 'tab-' . $index ?>"
```
LukeTowers: How about just use the href and change it to `#<= $type . 'tab-' . str_slug($name); ?>`

Rike-cz: Does not it break something else? My solution is based on existing backend solution (`/backend/system/mailtemplates`) - why is `data-tab-url` used there?

LukeTowers: @Rike-cz I don't know. @daftspunk what is `data-tab-url` used for in `/backend/system/mailtemplates`?

Rike-cz: @daftspunk ...and why was `data-tab-url` used instead of `href`? Maybe there is reason which we do not see.

mrgswift: I second what @LukeTowers  said, using at least 'tab-' as a prefix to data-tab-url name.  This way if some one wants to use an anchor hashtag with the same name as a tab, it doesn't conflict.

mrgswift: @Rike-cz @LukeTowers 

Searching the entire codebase of OctoberCMS, the only reference to the data attribute tabUrl is in tabs.js.  All it appears to do is add the tab url hashtag to the address bar using history.replaceState.  

https://github.com/octobercms/october/blob/a68f3abf1c53cc702b6e748d927d95d25c59b31a/modules/system/assets/ui/js/tab.js#L67-L69

I just tested using href instead of data-tab-url and it works with no problems.  Anchor code used:
```
<a
<?= $tabs->linkable ?  'href="#'. $type . 'tab-' . str_slug($name) . '"' : 'href="#'. $type . 'tab-' . $index . '"' ?>
```

In my opinion, there is no need to preserve the tab url in the address bar when a tab is clicked.  Whether or not it should be preserved when navigated to directly is up for debate.

Rike-cz: Directly navigating is important for me. First situation: I have many tabs and on some of the tab is button to another part of backend. There shoud be something ended with "Save and return back" and better is returning to previous tab instead of first one. Second situation: Customer couldn't find something in backend. I can send him link to specific tab directly.

mrgswift: @Rike-cz Totally agree.  Direct linking to tabs is important.  I was more commenting on whether the hashtag remains once the tab becomes active.  After thinking about it more from what you said, I think the hashtag should be preserved even after the tab becomes active.    I created a pull request a week ago without realizing this one existed to have linkable tabs, but it used history to remove the hashtag after the tab is selected.  Your PR is a lot simpler, as mine was completely in JS code adding a few more prototypes to tabs.js.  I definitely prefer your PR to mine.

Rike-cz:
I have changed `data-tab-url` to common `href` but implicit `data-tab-url` behavior used in some parts of backend stayed untouched.
Please make new review and tests.

LukeTowers:
@bennothommo what do you think about making this behaviour enabled by default / non-configurable? Can you think of any reasons against that?

bennothommo:
@LukeTowers I definitely think it should be configurable, as some people may have designed their forms in a way that the first tab is the most important, and may be concerned if people can skip steps, or inadvertently go to other tabs if, say, someone less knowledgeable bookmarks a backend form with a tab hash in place.

As for making it enabled by default, I don't see any disadvantages to that.

Rike-cz:
So there has to be something like _non-linkable_ instead of _linkable_. What do you prefer?

LukeTowers:
@Rike-cz keeping it `linkable` is fine, just change the default to `true` so that people have to set `linkable: false` to disable it.

LukeTowers:
@w20k can you test this out?

w20k:
@LukeTowers will add it to my long-long  Todo list :) Switching jobs does impact on the free-time, will try to make a test/smoke run on this week with few prior tasks!

PR #4757, opened 2019-11-10
Add mail port settings hint

ayumi-cloud:
Relates to this github issue: https://github.com/octobercms/october/issues/4755

### Before

image

### After

image


LukeTowers:
We'll go with @bennothommo's suggestion instead, of auto switching the port when the different protocol is detected.

PR #4758, opened 2019-11-10
Update .htaccess standard file

ayumi-cloud:
This pull request adds three things:

1. Disable PHP in Storage Folder (added security protection).

2. Whitelist the `favicon.ico` file in root.

3. Cleanup the standard routes, remove the `index.php` and redirect to `/ ` the issue here was October was creating two exact same web pages which caused a duplicate content issue with search engines, now we get the following:

```html
https://www.example.com/index.php redirects to https://www.example.com/
```

instead of having two exact same web pages:

```html
https://www.example.com/index.php
https://www.example.com/
```

Ok I found a coding error in October stopping the **mod-rewrite** functions from working!

I have removed part 3 from the `.HTaccess` and corrected the `server.php` file.

Ready for testing now!

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -60,7 +60,6 @@
     ## Standard routes, remove index.php and redirect to /
     ##
     RewriteCond %{REQUEST_FILENAME} !-f
-    RewriteCond %{THE_REQUEST} ^.*/index\.php
-    RewriteRule ^index.php(.*)$ /$1 [R=301,L]
+    RewriteRule ^ index.php [L]
 
 </IfModule>
--- a/.htaccess
+++ b/.htaccess
@@ -34,6 +34,11 @@
     ## Disable PHP in storage folder
     ##
     RewriteRule ^storage/.*\.(?:php[1-6]?)$ - [NC,F]
+    
+    ##
+    ## Whitelist the favicon.ico file in root
+    ##
+    RewriteCond %{REQUEST_FILENAME} !/favicon\.ico
 
     ##
     ## White listed folders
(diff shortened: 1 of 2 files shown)


ayumi-cloud:
=== don't merge please - still testing out the redirect ===

ayumi-cloud:
Ok I found a coding error in October stopping the mod-rewrite functions from working!

I have removed part 3 from the `.HTaccess` and corrected the `server.php` file.

ayumi-cloud:
My last commit I put the wrong order, it should be like this:

```
RewriteCond
RewriteCond
RewriteCond
..
RewriteRule (at the end)
```

I moved the `favicon.ico` whitelist to the top as it used a `RewriteCond` rule. Have tested it and passes.

.htaccess
```diff
@@ -29,6 +29,16 @@
     RewriteRule ^storage/framework/.* index.php [L,NC]
     RewriteRule ^storage/temp/protected/.* index.php [L,NC]
     RewriteRule ^storage/app/uploads/protected/.* index.php [L,NC]
+    
+    ##
+    ## Disable PHP in storage folder
+    ##
+    RewriteRule ^storage/.*\.(?:php[1-6]?)$ - [NC,F]
```
LukeTowers: This shouldn't be needed since all PHP files should be blocked from running except for index.php, and those files should not be allowed to be uploaded in the first place.

```
 ##
    ## Block all PHP files, except index
    ##
    RewriteCond %{REQUEST_FILENAME} -f
    RewriteCond %{REQUEST_FILENAME} \.php$
    RewriteRule !^index.php index.php [L,NC]
```

ayumi-cloud: Done

server.php
```diff
@@ -8,11 +9,14 @@
...
-// application without having installed a "real" web server software here.
-if ($uri !== '/' and file_exists(__DIR__.'/'.$uri))
-{
+
+/**
+ * This file allows us to emulate Apache's "mod_rewrite" functionality from the
+ * built-in PHP web server. This provides a convenient way to test a Laravel
+ * application without having installed a "real" web server software here.
+ */
+if ($uri !== '/' && file_exists(__DIR__.'/'.$uri)) {
```
bennothommo: @ayumi-cloud Would you mind adding some spacing around the concatenation parts?
```suggestion
if ($uri !== '/' && file_exists(__DIR__ . '/' . $uri)) {
```

Issue #4752, opened 2019-11-08
Browser update for cookies starting in 2020 (notes for admins and everyone)


ayumi-cloud:
I write this post to help the admins and everyone out.

I will update it when I get new infomation from different browsers, from time to time.

(Maybe the admins want to delete this post and move it to the correct location - I leave it up to them to decide).

### New changes

You should set the october_session cookie to the following:

- SameSite - Lax or Strict
- Secure - true
- HttpOnly - true

If you are using October to create an API or embed etc. which will be used on a third party website you should set the following:

- SameSite - None
- Secure - true
- HttpOnly - true

### Google Chrome

Deadline for stable version is 4th feb 2020 (Chrome V80). Chrome dev and Chrome Canary are already running this update, for more information see here: https://www.chromium.org/updates/same-site

To test your website you can use the following things:

![image](https://user-images.githubusercontent.com/57409060/68503755-87b09e80-025b-11ea-929c-c72adf7db996.png)

Above is what experiments I have turned on.

```
chrome://flags/#same-site-by-default-cookies
chrome://flags/#enable-removing-all-third-party-cookies
chrome://flags/#cookies-without-same-site-must-be-secure
```

### Microsoft Edge

Microsoft are going ahead and updating their browser and this update will take place when Edge V80 comes out. Please note, Edge versions come out one version before Chrome versions so this update should go live before feb 2020.

### Firefox

Today I have seen in Firefox Nightly, firefox's setup here are some screenshots of what I'm seeing:

![image](https://user-images.githubusercontent.com/57409060/68503930-e8d87200-025b-11ea-89a7-f1eaa746f45a.png)

and

![image](https://user-images.githubusercontent.com/57409060/68503967-f68df780-025b-11ea-8b9b-0d0e6c6e69d4.png)

Note: Firefox is doing the same as Chrome and Microsoft and setting Lax as the default and Strict is a higher secure version! I have been running SameSite Strict for 1-2 years now with October websites and had no problems. I recommend setting `Strict` and testing to see if you have any problems and then moving down to `Lax`.

### Safari

Safari team have said they are going to update their cookies in their browser as well. I can't give an exact date, but expect them to follow quickly.

### PHP

[...]

PR #4762, opened 2019-11-12

ayumi-cloud:
Discussion found here: https://github.com/octobercms/october/issues/4752

Basically done the following:

1. Removed the `SameSite = null` as default. Under new rules this **will cause an issue** (see point 2).

2. `SameSite = blank`. Under new rules will become `SameSite = None`.

3. Cookies are made up of two main groups `First-party` and `Third-party`. 

3.a). `First-party` cookies are set to strict or lax. Strict has an extra layer of security and I have been running October websites between 1-2 years will zero problems with Strict. (Our company has over 300 October websites, so we have tested it in a big'ish trail for over a year).

For new websites the default will be set to strict.

3.b). `Third-party` cookies are set to `SameSite None` and `Secure true`. 

### Testing guide

First-Party uses flag: `chrome://flags/#same-site-by-default-cookies`

Third-Party uses flag: `chrome://flags/#cookies-without-same-site-must-be-secure`

### Thoughts

Maybe need to update October's min php version to 7.3 or above with regards to SameSite support.

### DIFF
--- a/config/session.php
+++ b/config/session.php
@@ -163,10 +163,15 @@ return [
     | even if the user follows a link to another website. Lax cookies are
     | only sent with a top-level get request.
     |
-    | Supported: "lax", "strict"
+    | First-party cookies that match the domain, be set to 'Strict' or 'Lax'.
+    |
+    | Third-party cookies from other domains, be set to 'None' and need
+    | the Secure set to true.
+    |
+    | Supported: "lax", "strict" and "none"
     |
     */
 
-    'same_site' => null,
+    'same_site' => 'Strict',
 
 ];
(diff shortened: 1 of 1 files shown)


LukeTowers:
@ayumi-cloud can you resubmit this?

ayumi-cloud:
Note: Issue with Windows 10 and SameSite Google Chrome V80, users need to install KB4534273

See here for link: https://support.microsoft.com/en-gb/help/4534273/windows-10-update-kb4534273

ayumi-cloud:
@LukeTowers @daftspunk 

Done, see here: https://github.com/octobercms/october/pull/4881

Please remove this pr from the github projects and re-link the new pr.

Thanks.

PR #4764, opened 2019-11-12
Initial WIP on implementing dependsOn support for filter scopes.

LukeTowers:
Still need to resolve an issue where if the slave filter has values set when the master filter updates, thus triggering a change of the available options to the slave, the original values are still set on the slave but not actually visible in the popup as options because they're no longer valid options. To fix this we'll need the ability to get the browser to refresh the slave filter's selected values (count icon basically since it already forces the options popup to refresh) when its masters update; while at the same rechecking the slave's scope values set on the server to ensure that they're all valid and there aren't values left over from the previous request that are no longer valid but are still being applied to the query.

### DIFF
--- a/modules/backend/classes/FilterScope.php
+++ b/modules/backend/classes/FilterScope.php
@@ -52,6 +52,11 @@ class FilterScope
     public $options;
 
     /**
+     * @var array Other scope names this scope depends on, when the other scopes are modified, this scope will update.
+     */
+    public $dependsOn;
+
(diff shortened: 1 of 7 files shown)


LukeTowers:
@LarBearrr @bennothommo @daftspunk @huynhnamegal this should be ready to test now

LukeTowers:
Added docs to https://github.com/octobercms/docs/pull/419

PR #4766, opened 2019-11-14
Turn on redirects with HTTPS and also suggest to use HSTS with HTTPS

ayumi-cloud:
This pr does two things:

1. Instead of blocking HTTP, the website should 301 redirect a user or bot from HTTP to HTTPS.

2. When you use HTTPS you should be using HSTS at the same time, otherwise you allow hackers to bypass HTTPS.

### What is HSTS?

HTTP Strict Transport Security (HSTS) is a web server directive that informs user agents and web browsers how to handle its connection through a response header sent at the very beginning and back to the browser.

This sets the Strict-Transport-Security policy field parameter. It forces those connections over HTTPS encryption, disregarding any script's call to load any resource in that domain over HTTP. HSTS is but one arrow in a bundled sheaf of security settings for your web server or your web hosting service.

### Why Should Your Company Implement HSTS?

You never close your physical store or home without locking the doors, right? You may even have metal detectors at the door to control shrinkage. Data can be just as valuable as physical items in your shop or house, so it’s just as important to keep them locked up and secure. Padlocking your website is sometimes not enough as people will still find a way to reach your website over http://. HSTS forces browsers and app connections to use HTTPS if that is available. Even if someone just types in the www or http://.

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -15,8 +15,15 @@
     ##
     ## Uncomment following lines to force HTTPS.
     ##
-    # RewriteCond %{HTTPS} off
-    # RewriteRule (.*) https://%{SERVER_NAME}/$1 [R,L]
+    # RewriteCond %{HTTPS} !=on
+    # RewriteRule ^ https://%{HTTP_HOST}%{REQUEST_URI} [R=301,L]
+
+    ##
+    ## Uncomment following lines, turn on HSTS when using HTTPS.
+    ##
+    # <IfModule mod_headers.c>
+    #     # Header always set Strict-Transport-Security "max-age=31536000; includeSubDomains; preload" env=HTTPS
+    # </IfModule>
 
     ##
     ## Black listed folders
--- a/.htaccess
+++ b/.htaccess
@@ -11,19 +11,6 @@
     ## if you have installed to a subdirectory, enter the name here also.
     ##
     # RewriteBase /
-    
-    ##
-    ## Uncomment following lines to force HTTPS.
-    ##
-    # RewriteCond %{HTTPS} !=on
-    # RewriteRule ^ https://%{HTTP_HOST}%{REQUEST_URI} [R=301,L]
-
-    ##
-    ## Uncomment following lines, turn on HSTS when using HTTPS.
-    ##
-    # <IfModule mod_headers.c>
-    #     Header always set Strict-Transport-Security "max-age=31536000; includeSubDomains; preload" env=HTTPS
-    # </IfModule>
 
     ##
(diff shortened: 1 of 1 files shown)


bennothommo:
@ayumi-cloud Number 1 is already covered by some plugins (such as https://octobercms.com/plugin/popcornphp-redirecttohttps or my own https://octobercms.com/plugin/bennothommo-urlnormaliser).

Number 2 should not be enforced by October itself, as there's the potential for people to get locked out of their own website if, say, their SSL certificate expires or the browsers decide to deplatform an SSL provider. Instead, it should be applied at the server configuration level.

ayumi-cloud:
@bennothommo There's many things wrong with your comment - no offence!

https://octobercms.com/plugin/popcornphp-redirecttohttps plugin I take it you haven't used as people are reporting errors with it, see here: https://github.com/VoroninWD/oc-redirect-to-https-plugin/issues/7

Your number 2 point, you are mixing up Search Results HTTP Public Key Pinning (HPKP) with HSTS. When a cert expires, a user can't use a HTTP connection until the HSTS date has expired. It doesn't stop you from changing to another cert. So you're completely wrong there.

If you are not happy with the following code, then let's completely remove the HTTPS from the .HTaccess, rather that than having an incorrect code there.

36864:
I think the thought process is that if someone using October knows how to go into the .htaccess and uncomment a few lines to enable HTTPS, they are probably also capable of adding new lines to enable HSTS if they so desire.

Plugins to enable this functionality exist for users who or don't want to modify server files, and prefer to use the web interface for everything.

HSTS is a much less common requirement for websites, and the specific configuration may vary for different websites. For example, your implementation was including subdomains. Some people route specific subdomains in their local network for development and may not want to enable https for those. (dev.example.com points to the dev server, staging.example.com points to the staging server, and www.example.com is the only publicly-accessible domain that uses HTTPS and HSTS)

LukeTowers:
@ayumi-cloud I agree with @36864. If we were to provide a default way of enabling HSTS in the .htaccess file by uncommenting some lines there would have to be a BIG WARNING surrounding it because HSTS if implemented incorrectly without fully considering the implications can cause some serious problems. Apart from that, this seems more like an issue for the documentation than one for the default .htaccess file.

ayumi-cloud:
@LukeTowers @bennothommo no problem, thanks for taking the time to look at this pr. 👍

LukeTowers:
@ayumi-cloud no problem, thanks for taking the time to submit improvement ideas to October!

PR #4769, opened 2019-11-17
safemode fix

mjauvin:
- modify current `safemode` behavior by changing the `codeeditor` field to `read-only` if `safemode` is enabled
- add isSafeMode() method to `CmsCompoundObject` class
- add dismissable info callout when accessing code editor field


mjauvin:
I think it makes sense to still allow new template pages to be created with PHP code, but if you think this is wrong, please let's discuss this.

LukeTowers:
@mjauvin safe mode should apply to everything, regardless of whether or not the file already exists. Also note that having a static method on CmsCompoundObject doesn't make sense.

mjauvin:
Why doesn't the static method make sense?

mjauvin:
how do you like the idea of using the backend.form.extendFieldsBefore event?

bennothommo:
@mjauvin Should we just hide the code editor completely if safe mode is on, as opposed to making it read only? Thinking on it from the mindset of a client - I would think something is wrong with the system if I can see this "Code" tab with a code editor, but I can't seem to be able to edit it.

mjauvin:
That would be an option, but I think it's still interesting to see the field's content. Unless there's a way to show a kind of tool tip when hovering mouse cursor on a readonly field?

LukeTowers:
@mjauvin it doesn't make sense because CmsCompoundObject is a class to inherit from, it's not a helper class.

mjauvin:
Ok, I tried to make the field "hidden" instead of readOnly and the problem is that if a change is done to the markup while in safemode, the code field will get lost after the save.

mjauvin:
Look with a dismissable callout, I think this is good.
Screenshot from 2019-11-18 09-44-56

modules/backend/formwidgets/CodeEditor.php
```diff
@@ -167,6 +167,7 @@ public function prepareVars()
         $this->vars['highlightActiveLine'] = $this->highlightActiveLine;
         $this->vars['useSoftTabs'] = $this->useSoftTabs;
         $this->vars['showGutter'] = $this->showGutter;
+        $this->vars['safeMode'] = $this->parentForm->model->isSafeMode();
```
LukeTowers: This can't exist here either because the code editor can be used on other models than CMS templates. You could probably implement this as a `hint` property (so `hint: cms::lang.cms_object.safe_mode_enabled`) on the codeeditor formwidget instead though and I'd be fine with that.

modules/cms/controllers/Index.php
```diff
@@ -59,6 +60,18 @@ public function __construct()
     {
         parent::__construct();
 
+        Event::listen('backend.form.extendFieldsBefore', function ($widget) {
```
LukeTowers: No need to add the event listener here, the form controller behavior provides a `formExtendFieldsBefore($widget)` method on the controller that you can override instead for the same end result.

modules/backend/formwidgets/CodeEditor.php
mjauvin: done

modules/cms/helpers/Cms.php
```diff
@@ -35,4 +36,14 @@ public function url($path = null)
 
         return Url::to($path);
     }
+
+    public static function isSafeMode()
```
LukeTowers: Should be safeModeEnabled()

modules/backend/formwidgets/CodeEditor.php
```diff
@@ -102,6 +102,11 @@ class CodeEditor extends FormWidgetBase
      */
     public $showPrintMargin = false;
 
+    /**
+     * @var string Hint to show above the code editor
+     */
+    public $codeHint = '';
```
LukeTowers: This should either be `hint` that takes a string to render inside of the partial container or probably better you should just render a separate hint field above the code field in the same tab.

modules/cms/helpers/Cms.php
mjauvin: ok

modules/backend/formwidgets/CodeEditor.php
mjauvin: That's pretty much what I'm doing, just passing the rendered content from the Index controller.

How do you see this exactly?

LukeTowers: The problem is that you're passing rendered content, that's not done anywhere else in October AFAIK. I'm thinking the cleanest approach is probably to add the hint field as a separate hint field to the same tab that the codeeditor field is in so that it's right next to the codeeditor field but is separate so it requires no additional changes to the API.

mjauvin: Ohhh, I see what you mean now. Ok, thanks.

modules/cms/controllers/Index.php
```diff
@@ -59,6 +61,31 @@ public function __construct()
...
+                $codeField = $widget->secondaryTabs['fields']['code'];
+                unset($widget->secondaryTabs['fields']['code']);
+
+                // add hint field first
+                $widget->secondaryTabs['fields']['hint'] = [
+                    'tab' => 'cms::lang.editor.code',
+                    'type' => 'hint',
+                    'path' => '~/modules/cms/controllers/index/_hint_field.htm',
+                    'cssClass' => 'p-b-0',
+                ];
```
LukeTowers: @mjauvin the other thing that you could do is include this field in the default form, but mark it as hidden: true and then just unhide it under the proper conditions.

mjauvin: I thought about it but then I'd have to include it in 4 seperate forms... which do you prefer?

LukeTowers: I'm fine with it being in the separate fields.yaml

mjauvin:
Screenshot from 2019-11-19 10-52-50

That's what it looks like now

modules/cms/classes/layout/fields.yaml
```diff
@@ -32,6 +32,13 @@ secondaryTabs:
             type: codeeditor
             language: twig
 
+        notice:
+            tab: cms::lang.editor.code
+            type: partial
+            hidden: true
+            path: ~/modules/cms/controllers/index/_notice_field.htm
```
LukeTowers: You should be able to change the field name to safemode_notice and rename the partial `_safemode_notice.htm` and then remove the path property from all these configs. What is the p-b-0 class for?

mjauvin:
the `p-b-0` is a class to remove bottom padding, it's ugly without this

mjauvin:
Here's what it looks like without it:

Screenshot from 2019-11-19 12-28-59

LukeTowers:
Perfect, thanks!

mjauvin:
would it make sense to set the cssClass from the Event instead so it's easier to change later?

LukeTowers:
@mjauvin no, I don't think so. I can't think of a reason why it would need to be changed.

mjauvin:
@LukeTowers I tested this PR using OctoDock (amazing tool @petehalverson) and this works as expected.

Debug mode is "false" by default on the launched container, and I got the warning an read-only mode for the CMS templates.

Changed to Debug = true and warning disappeared, was able to edit the CMS templates again.

mjauvin:
Previous tests were done with enableSafeMode = null
I also tested forcing enableSafeMode to false/true and behavior is as expected.

mjauvin:
If anyone is willing to test this, the container can be found here:

https://a41ee025fa67.octodock.com/backend

user/pass = admin/admin

modules/cms/controllers/index/_safemode_notice.htm
```diff
@@ -0,0 +1,6 @@
+<div class="callout fade in callout-info no-subheader">
```
LukeTowers:
```suggestion
<div class="callout callout-warning no-subheader">
```

modules/cms/controllers/index/_safemode_notice.htm
```diff
@@ -0,0 +1,6 @@
...
+    <div class="header">
```
LukeTowers:
```suggestion
    <div class="header" style="border-radius: 0">
```

modules/cms/controllers/index/_safemode_notice.htm
```diff
@@ -0,0 +1,6 @@
...
+        <i class="icon-info"></i>
```
LukeTowers:
```suggestion
        <i class="icon-warning"></i>
```

modules/cms/controllers/index/_safemode_notice.htm
```diff
@@ -0,0 +1,6 @@
...
+        <h3><?= e(trans('cms::lang.cms_object.safe_mode_enabled')) ?></h3>
```
LukeTowers: Can you change the trans key to say "Safe mode is currently enabled. Editing the PHP code of CMS templates is disabled."

mjauvin: Who's going to handle all the translations for the other languages?

LukeTowers: Whoever wants to / notices that it changes. It's pretty much the same string, just more descriptive. The way October handles translations in the core is entirely community driven, the only language we actively keep up to date (besides moving keys around) is English

mjauvin:
Crap, didn't notice they added a "add to batch" for the suggested changes... next time.

bennothommo:
Looks great @mjauvin. Because it's a new piece of functionality, I think we'll have to hold off merging it until 461 has begun, as we're trying to stabilise 460 at the moment.

mjauvin:
Can I remove the OctoDock instance I share above for testing this?
Are there more testing needed?

LukeTowers:
@mjauvin no more testing required, should be good to remove it

LukeTowers:
FYI: Browsers will automatically always convert newlines into `\r\n` when POSTing data, so this can cause issues with templates that use unix newlines in their code section because they will always trigger the isDirty check: https://github.com/ajaxorg/ace/issues/1515#issuecomment-20971401

PR #4770, opened 2019-11-17
1. Fix paths to plugins and themes for update manager

LeMaX10:
Fix paths to plugins and templates for update manager.

All parts of the system work with settings, but the manager always looks at the root directories.

### DIFF
--- a/modules/system/classes/UpdateManager.php
+++ b/modules/system/classes/UpdateManager.php
@@ -582,7 +582,7 @@ class UpdateManager
         $fileCode = $name . $hash;
         $filePath = $this->getFilePath($fileCode);
 
-        if (!Zip::extract($filePath, $this->baseDirectory . '/plugins/')) {
+        if (!Zip::extract($filePath, App::pluginsPath())) {
             throw new ApplicationException(Lang::get('system::lang.zip.extract_failed', ['file' => $filePath]));
         }
 
@@ -624,7 +624,7 @@ class UpdateManager
         $fileCode = $name . $hash;
         $filePath = $this->getFilePath($fileCode);
 
-        if (!Zip::extract($filePath, $this->baseDirectory . '/themes/')) {
+        if (!Zip::extract($filePath, App::themesPath())) {
             throw new ApplicationException(Lang::get('system::lang.zip.extract_failed', ['file' => $filePath]));
         }
 
--- a/modules/system/classes/UpdateManager.php
+++ b/modules/system/classes/UpdateManager.php
@@ -582,7 +582,7 @@ class UpdateManager
         $fileCode = $name . $hash;
         $filePath = $this->getFilePath($fileCode);
 
-        if (!Zip::extract($filePath, App::pluginsPath())) {
+        if (!Zip::extract($filePath, plugins_path())) {
             throw new ApplicationException(Lang::get('system::lang.zip.extract_failed', ['file' => $filePath]));
         }
 
--- a/modules/system/classes/UpdateManager.php
+++ b/modules/system/classes/UpdateManager.php
@@ -624,7 +624,7 @@ class UpdateManager
         $fileCode = $name . $hash;
         $filePath = $this->getFilePath($fileCode);
 
-        if (!Zip::extract($filePath, App::themesPath())) {
+        if (!Zip::extract($filePath, themes_path())) {
             throw new ApplicationException(Lang::get('system::lang.zip.extract_failed', ['file' => $filePath]));
         }
 


LukeTowers:
@LeMaX10 how are you changing these values and what are you using instead?

modules/system/classes/UpdateManager.php
```diff
@@ -582,7 +582,7 @@ public function extractPlugin($name, $hash)
         $fileCode = $name . $hash;
         $filePath = $this->getFilePath($fileCode);
 
-        if (!Zip::extract($filePath, $this->baseDirectory . '/plugins/')) {
+        if (!Zip::extract($filePath, App::pluginsPath())) {
```
LukeTowers:
```suggestion
        if (!Zip::extract($filePath, plugins_path())) {
```

modules/system/classes/UpdateManager.php
```diff
@@ -624,7 +624,7 @@ public function extractTheme($name, $hash)
         $fileCode = $name . $hash;
         $filePath = $this->getFilePath($fileCode);
 
-        if (!Zip::extract($filePath, $this->baseDirectory . '/themes/')) {
+        if (!Zip::extract($filePath, App::themesPath())) {
```
LukeTowers:
```suggestion
        if (!Zip::extract($filePath, themes_path())) {
```

modules/system/classes/UpdateManager.php
LeMaX10: I am a little not biased towards helpers, given that they can be removed in some of the releases. Using the method of the Application object is due to the presence and use of this method in the core as a whole. I could be wrong. Based on your offer, I see no reason to refuse.

LeMaX10:
> @LeMaX10 how are you changing these values and what are you using instead?

In one of my projects, I had to remove plug-in and template directories from my home directory to comply with internal security rules, giving me the ability to manage this in the settings. However, I must keep all the standard features of the system, including the ability to update and install extensions from the market.

As a result of the implementation of this feature, I came across a problem solved in this PR.

LukeTowers:
@LeMaX10 could you explain that more? Have you removed the directories from October entirely or are you just hosting october in your user account's home directory but then storing the themes / plugins in an external directory? That would cause issues when attempting to load any assets from the themes / plugins unless you've also setup your server to forward those paths to the real paths.

Does such a setup even work? I'm really curious to know more details about your specific setup.

mjauvin:
@LukeTowers but this fix allows using the same paths defined in config/cms.php for pluginsPath / themesPath, isn't that what we want?

LukeTowers:
@mjauvin the reason I'm asking these questions is because themesPath & pluginsPath in config/cms.php isn't used for the local directory path, it's used in generating URLs. The actual configs are themesPathLocal & pluginsPathLocal, and they're undocumented for a reason. I want to learn more about a use case that actually needs to change those paths.

mjauvin:
Oh, I see the difference now. Right.

LeMaX10:
> @LeMaX10 could you explain that more? Have you removed the directories from October entirely or are you just hosting october in your user account's home directory but then storing the themes / plugins in an external directory? That would cause issues when attempting to load any assets from the themes / plugins unless you've also setup your server to forward those paths to the real paths.
> 
> Does such a setup even work? I'm really curious to know more details about your specific setup.

Good day!

Yes, I deleted the directories from the root directory of October, thereby transferring them to separate protected directories. I also managed to implement the storage of plugins and themes on external servers.

The catalogs of plugins and templates were moved from the root directory to a directory that is not accessible to the user of the site, thereby plugins and templates are isolated from the physical core of the system by the set scope. This allows you to very effectively cement the system to protect it by eliminating the possibility of interference from outside, through a low-quality contractor code.

To maintain themesPath & pluginsPath settings from config / cms.php, the nginx configuration is used to return only static files. Overriding paths is the responsibility of a simple module registered in loadModules.

In principle, these edits are not critical, given that all parts of the system work with plugin_path () including PluginManager, however, UpdateManager for some reason uses only the root directory for its work.


I sketched an approximate scheme as it is now implemented.
schemes

LukeTowers:
@LeMaX10 So are you using the pluginsPathLocal and themesPathLocal config items to do that then?

LeMaX10:
> @LeMaX10 So are you using the pluginsPathLocal and themesPathLocal config items to do that then?

Yes, I use for this purpose.

modules/system/classes/UpdateManager.php
LukeTowers: I suggest using the helpers because they are used elsewhere in the core while App::pluginsPath() isn't used elsewhere anywhere. It is extremely unlikely that October will ever remove helpers (in fact we're re-adding Laravel's helpers that they removed in our upgrade to L6) since we're big on maintaining compatibility.

LeMaX10: > I suggest using the helpers because they are used elsewhere in the core while App::pluginsPath() isn't used elsewhere anywhere. It is extremely unlikely that October will ever remove helpers (in fact we're re-adding Laravel's helpers that they removed in our upgrade to L6) since we're big on maintaining compatibility.

Yes OK! I do not see any contradictions with this option.

daftspunk:
Thanks @LeMaX10. This looks fine to refactor this in but will need to bump the milestone to a less risky build--461 has to be stable

FlusherDock1:
Does whitelisting `favicon.ico` is necessary? I mean in modern web if you need favicon, you just can specify it in the head of your layout.

I don't want whitelist non existing `favicon.ico` on my websites, since there are plenty of viruses that can be hidden inside `.ico` format.

LukeTowers:
Again will have to bump this, 466 has to be stable since 467 will be the Laravel 6 upgrade.

LukeTowers:
Unbumping because I have time to test this

PR #4773, opened 2019-11-18
Czech translation in /modules/system/lang improved

maraerben:
I've just added some Czech translations required in backend list filters. No serious changes were made.

### DIFF
--- a/modules/system/lang/cs/client.php
+++ b/modules/system/lang/cs/client.php
@@ -33,7 +33,6 @@ return [
         'fullscreen' => 'Celá obrazovka',
         'preview' => 'Náhled',
     ],
-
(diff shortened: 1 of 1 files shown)


modules/system/lang/cs/client.php
```diff
@@ -65,13 +63,24 @@
         'group' => [
             'all' => 'Vše'
         ],
+        'scopes' => [
+            'apply_button_text' => 'Filtrovat',
+            'clear_button_text'  => 'Zrušit'
```
LukeTowers: Please ensure that the last item in an array has a comma too. This isn't enforced anywhere right now but it will trigger the checks to run again and I prefer it to be that format anyways.

PR #4777, opened 2019-11-19
do not use jquery-sortable on mobile, it breaks the touch events

### LINKED ISSUE #3755: List of checkboxes do not work on the touch screen
##### Expected behavior
It should work

##### Actual behavior
It does not work. Pressing does not give any result

##### Reproduce steps
Go to any list and open its settings. Try turning off or on any checkboxes.
screenshot_2018-09-05-01-04-15-205_com android chrome

It also does not work on the computer in the view mode through the device
image

##### October build
442

To solve this problem - just delete the list from the file: /modules/backend/widgets/lists/partials/_setup_form.htm

image

And everything will work. But can there be another reason?

mjauvin:
fixes #3755

### DIFF
--- a/modules/backend/assets/js/october.simplelist.js
+++ b/modules/backend/assets/js/october.simplelist.js
@@ -15,10 +15,11 @@
     var SimpleList = function (element, options) {
 
         var $el = this.$el = $(element)
+        var is_mobile = /Mobi/.test(navigator.userAgent);
 
         this.options = options || {}
 
-        if ($el.hasClass('is-sortable')) {
+        if ($el.hasClass('is-sortable') && !is_mobile) {
 
             /*
              * Make each list inside sortable
--- a/modules/backend/assets/js/october.simplelist.js
+++ b/modules/backend/assets/js/october.simplelist.js
@@ -15,11 +15,11 @@
     var SimpleList = function (element, options) {
 
         var $el = this.$el = $(element)
-        var is_mobile = /Mobi/.test(navigator.userAgent);
+        var isMobile = $('html').hasClass('mobile');
 
         this.options = options || {}
 
-        if ($el.hasClass('is-sortable') && !is_mobile) {
+        if ($el.hasClass('is-sortable') && !isMobile) {
 
             /*
              * Make each list inside sortable
--- a/modules/backend/assets/js/october.simplelist.js
+++ b/modules/backend/assets/js/october.simplelist.js
@@ -15,11 +15,10 @@
     var SimpleList = function (element, options) {
 
         var $el = this.$el = $(element)
-        var isMobile = $('html').hasClass('mobile');
 
         this.options = options || {}
 
-        if ($el.hasClass('is-sortable') && !isMobile) {
+        if ($el.hasClass('is-sortable') && !$('html').hasClass('mobile')) {
 
             /*
              * Make each list inside sortable
--- a/modules/backend/assets/js/october-min.js
+++ b/modules/backend/assets/js/october-min.js
@@ -1084,8 +1084,8 @@ $(document).ready(function(){if(Modernizr.touchevents||(typeof(localStorage)!=='
 $(window).trigger('resize')}
 else if(localStorage.ocSidePanelFixed==1){$(document.body).removeClass('side-panel-not-fixed')
 $(window).trigger('resize')}}})}(window.jQuery);+function($){"use strict";var SimpleList=function(element,options){var $el=this.$el=$(element)
-var isMobile=$('html').hasClass('mobile');this.options=options||{}
-if($el.hasClass('is-sortable')&&!isMobile){var sortableOptions={distance:10}
+this.options=options||{}
+if($el.hasClass('is-sortable')&&!$('html').hasClass('mobile')){var sortableOptions={distance:10}
 if(this.options.sortableHandle)
 sortableOptions[handle]=this.options.sortableHandle
 $el.find('> ul, > ol').sortable(sortableOptions)}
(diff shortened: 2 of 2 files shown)


modules/backend/assets/js/october.simplelist.js
```diff
@@ -15,10 +15,11 @@
     var SimpleList = function (element, options) {
 
         var $el = this.$el = $(element)
+        var is_mobile = /Mobi/.test(navigator.userAgent);
```
LukeTowers: Note that the HTML element in the backend is populated with a number of "feature flag classes", so you could just detect if that element has a .mobile class on it, although I'm uncertain what is being tested to trigger that

mjauvin:
Thanks @LukeTowers , that works very well.

Tested on the following mobile browser on Android:

- Chrome
- Firefox
- Edge
- Dolphin
- Opera

LukeTowers:
@mjauvin After looking into this further I see why the sortable plugin is used in this case. If we just disable it we're breaking the ability for those columns to be re-ordered period. While this does restore the ability to toggle columns on and off, it's not really an acceptable problem because it leaves the functionality that is supposed to exist here broken on mobile. Note that this is also used for the import / export behavior and a few other places. 

I think we might just have to look at either fixing the jquery plugin's support for touch events or switching to the other library.

mjauvin:
I agree, but short term,  toggling check-boxes is really basic behavior and this is currently broken.

At least this restore the toggling behavior in the mean time.

LukeTowers:
@mjauvin this is only toggling checkboxes on the list setup column isn't it? That's more of an advanced feature that's less used.

mjauvin:
It is. I do use this a lot.

bennothommo:
@LukeTowers I might be wrong, but I don't think Sortable works on mobile anyway. There's no way to determine whether a user is sorting or scrolling with  touch gestures.

LukeTowers:
@bennothommo are you able to test the import / export behavior on mobile? That uses this jQuery plugin but it doesn't use checkboxes, and the sorting is pretty important to it

bennothommo:
@LukeTowers Tested it on the Blog plugin on a client's site, and yeah, doesn't work at all on mobile. The checkboxes also fail to toggle, as it does with the original issue and the List Setup modal.

LukeTowers:
@bennothommo why would checkboxes on the blog plugin be sortable?

bennothommo:
@LukeTowers I mean the Import / Export functionality of the Blog plugin. :)

LukeTowers:
Ah, gotcha @bennothommo. What I was talking about was the custom control that didn't use checkboxes but did use this sortable plugin.

bennothommo:
@LukeTowers Is there a core page I can view that?

LukeTowers:
@bennothommo The Import columns list isn't using checkboxes but is using sortable.

LukeTowers:
FYI, the sortable behavior even for the checkboxes in the list setup works fine on mobile (iOS Firefox), it's just that you can't toggle a checkbox on mobile.

bennothommo:
Hmmm, interesting. I'm testing Chrome on Android.

mjauvin:
@LukeTowers It's definitely not working on Android, have tested with Chrome, Firefox and Edge

LukeTowers:
@mjauvin @bennothommo are you testing whether you can drag items around or just if you can click the checkboxes? Clicking the checkboxes is broken on iOS but you can still drag things around.

mjauvin:
Dragging around. Do you have a trick or do you just drag? Do you hold a bit before it can be dragged? I tried everything I could think of.

bennothommo:
Dragging, as well :)

LukeTowers:
Nope, no special tricks, just dragging them around. Are you able to look deeper into what is causing the breakdown?

mjauvin:
It's a known issue with jquery-sortable:

https://github.com/johnny/jquery-sortable/issues/209

w20k:
Yep, dragging doesn't work :) And checkboxes are - now)
Maybe adding `http://touchpunch.furf.com`, would solve checkbox issue?

mjauvin:
I tried but it didn't work. You wanna try?

LukeTowers:
FYI not going to accept this in its current state since it would break dragging import / export columns on iOS. We should really just fix the root cause, i.e. replace the no longer function jquery plugin.

daftspunk:
Can you check that this commit doesn't fix the issue? 18398a8318ac6e03af9ae28c33a6b5996d363f83

The checkboxes seem to work on iOS but there is some areas that appear to have conflicting events. For example, if you double-tap you can get the checkbox selected.

mjauvin:
Ok, this definitely does not help things on Andoid, any browser... further more, some other recent commit prevents horizontal scrolling of table header to show hidden content to the right which still worked fine recently on android mobile browsers.

I'll try to find the exact commit that broke this.

mjauvin:
Ok, looks like some JS caching was involved in previous horizontal scrolling no working....

So everything is back to normal, but the modernizr commit does not change anything on Android.

iOS people might want to test more with this?

mjauvin:
We could look at replacing the jquery-sortable with this one which seems to work very well on Android (can ios users test this with the examples on their site) :

https://shopify.github.io/draggable/

mjauvin:
This one is even simpler, pure js:

https://github.com/SortableJS/Sortable/blob/master/README.md

mjauvin:
This one even has a jquery binding:

 https://sortablejs.github.io/Sortable/

jquery binding:

https://github.com/SortableJS/jquery-sortablejs

mjauvin:
Ok, so I used SortableJS to replace our current sortable and it works on all mobile browsers, although the checkbox toggling is for only working in Firefox mobile for now... still, big progress here.

LukeTowers:
Sortable JS looks pretty powerful, and it seems like it could be used to solve some other issues I've had clients complain about with regards to reordering menu items and pages in RainLab.Pages

mjauvin:
obsoleted by @daftspunk's commit 566f138

PR #4780, opened 2019-11-21
Update TagList.php

huynhnamegal:
Add a space after each item.
Old: item1,item2
New: item1, item2

### DIFF
--- a/modules/backend/formwidgets/TagList.php
+++ b/modules/backend/formwidgets/TagList.php
@@ -207,8 +207,10 @@ class TagList extends FormWidgetBase
     protected function getSeparatorCharacter()
     {
         switch (strtolower($this->separator)) {
-            case 'comma':
+            case 'spacedcomma':
                 return ', ';
+            case 'comma':
+                return ',';
             case 'space':
                 return ' ';
         }
--- a/modules/backend/formwidgets/TagList.php
+++ b/modules/backend/formwidgets/TagList.php
@@ -208,7 +208,7 @@ class TagList extends FormWidgetBase
     {
         switch (strtolower($this->separator)) {
             case 'comma':
-                return ',';
+                return ', ';
             case 'space':
                 return ' ';
         }


LukeTowers:
Before and after screenshot please

LukeTowers:

modules/backend/formwidgets/TagList.php
```diff
@@ -208,7 +208,7 @@ protected function getSeparatorCharacter()
     {
         switch (strtolower($this->separator)) {
             case 'comma':
-                return ',';
+                return ', ';
```
bennothommo: @huynhnamegal @LukeTowers If this is to be accepted, I think it should be a different option as opposed to changing the format of `comma` separated tags, so it doesn't break any current implementations.
```suggestion
                return ',';
            case 'spacedComma':
                return ', ';
```

huynhnamegal:
@LukeTowers  Sorry about this mistake. I will pay attention to the branch before creating PRs.

modules/backend/formwidgets/TagList.php
huynhnamegal: That's great! In my country, it's look great if we have a space after comma.

LukeTowers: @bennothommo haven't looked into this too deeply, how would it break existing setups?

LukeTowers:
@huynhnamegal can you post a before / after gif / screenshot?

modules/backend/formwidgets/TagList.php
bennothommo: @LukeTowers It probably won't - I'm just guessing from the intention of having a separate result for comma or space. It may be that some people use the value of the TagList for comma-separated values.

huynhnamegal:
@LukeTowers 
Before adding space in code
image
After adding space in code
image

I have used taglist which saved array, and used loop in order to show like that before.

DanHarrin:
This is a worthy option, I'm currently using a select statement in my list configurations when displaying a taglist -
```
select: 'REPLACE(field, ",", ", ")'
```

bennothommo:
@DanHarrin I agree - worthy option, but I reckon it should be an *additional* option, not replacing one that's already out there.

DanHarrin:
Yes, I agree @bennothommo

modules/backend/formwidgets/TagList.php
LukeTowers: @huynhnamegal please implement the above suggestion and add a PR to the docs as well.

daftspunk:
This modifies how the raw data is stored and we want to constrain this as much as possible. More diversity in the data storage format creates more technical debt

I think here you want simply want to use a twig filter to manipulate the data for display

First [split the data in to an array](https://twig.symfony.com/doc/2.x/filters/split.html):
```
{% set foo = "one,two,three"|split(',') %}
```

Then [reassemble to the desired display format](https://twig.symfony.com/doc/3.x/filters/join.html):

```
{{ [1, 2, 3]|join(', ', ' and ') }}
{# outputs 1, 2 and 3 #}
```

Hope this helps!

LukeTowers:
I agree with @daftspunk's point, willing to reopen if a logical argument against it is presented though

DanHarrin:
I also concur with @daftspunk, and, as I mentioned above, you are able to achieve the same effect (to that with Twig) for lists, using a select statement like so:
```
select: 'REPLACE(field, ",", ", ")'
```

PR #4789, opened 2019-11-26
Improve number range filter

### LINKED ISSUE #3982: Number range filter doesn't work if either 'min' or 'max' is set to 0
https://github.com/octobercms/october/blob/826e72d3a969c2307e83e59c28056b26e488723c/modules/backend/widgets/Filter.php#L764

bennothommo:
Fixes #3982.

- Allow minimum or maximum to be unspecified, meaning you want everything up to maximum, or everything above minimum.
- Allow for zero values to work as either minimum or maximum values.
- Tweak display of infinite values when refreshing filter.


bennothommo:
To test:

- Using the Test plugin, go to **Users** page.
- Change the values of the new "ID Range" number filter at the top of the page.

You should be able to specify a minimum value only, a maximum value only, or both.

modules/backend/widgets/Filter.php
```diff
@@ -145,18 +145,19 @@ public function renderScopeElement($scope)
                 break;
 
             case 'numberrange':
-                if ($scope->value && is_array($scope->value) && count($scope->value) === 2 &&
-                    $scope->value[0] &&
-                    $scope->value[1]
+                if (
+                    $scope->value
+                    && (is_array($scope->value) && count($scope->value) === 2)
```
w20k: Do you really need brackets here @bennothommo or it's just for readability purpose?

bennothommo: @w20k Just for readability. When it's all on one line without brackets, it's quite hard to discern what is happening :)

w20k:
LGTM - will do the testing later on ;) Thanks  @bennothommo!

LukeTowers:
@bennothommo how does this treat the user entering the literal value `0`? That should be passed on and not treated as infinite

bennothommo:
@LukeTowers `0` is an acceptable value and is treated as is. The absence of a value entirely is treated as infinite (effectively `null`)

PR #4803, opened 2019-12-04
register cms twig environment

mjauvin:
One can now use `App::get('twig.environment')` to get CMS twig environment.

### DIFF
--- a/modules/cms/classes/Controller.php
+++ b/modules/cms/classes/Controller.php
@@ -621,6 +621,13 @@ class Controller
         if ($isDebugMode) {
             $this->twig->addExtension(new DebugExtension($this));
         }
+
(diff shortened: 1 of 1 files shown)


mjauvin:
`Twig::parse()` can now render CMS Twig Extensions as well as system ones.

LukeTowers:
@mjauvin see #3575, that's the one that should be polished up and tested. Note that we don't want the CMS Twig environment to pollute the global one, I believe that causes issue with the Mail system's Twig environment, especially since it also has "partial" tags.

mjauvin:
I find #3575 too complex and I don't really see the point. My simple fix is enough. I agree we should use `cms.twig.environment`, though.

Issue #4799, opened 2019-12-03
Browser client side validation not running on backend forms


multiwebinc:
If I have the following yaml for a plugin's registered settings:

```
tabs:
    fields:
        your_age:
            label: Your Age
            type: number
            step: 1  # defaults to 'any'
            min: 1   # defaults to not present
            max: 100 # defaults to not present
```

This does not perform any validation; either client-side (using [form data validation](https://developer.mozilla.org/en-US/docs/Learn/HTML/Forms/Form_validation)) or server-side. In the example above, I can submit any number (e.g. -1000, 0, 1000) and it will save it to the database. I also tried sending a string in the HTTP request, but this gets cast to `0`.

bennothommo:
@multiwebinc I believe the only validation rule implemented for Settings is "required" at this stage. Feel free to submit a PR that allows the insertion of validation rules in the `modules/system/behaviors/SettingsModel.php` file.

daftspunk:
If I understand correctly, you should be using a dedicated Model for these settings. Here are some examples of using validation with the settings model:

- System
https://github.com/octobercms/october/blob/91d7b980a9abd299f0300b878b43410934a696b4/modules/system/models/MailSetting.php#L41-L47

- GA plugin
https://github.com/rainlab/googleanalytics-plugin/blob/master/models/Settings.php

Hope this helps!

multiwebinc:
@daftspunk You're absolutely right. I somehow missed the section on https://octobercms.com/docs/backend/forms#validate-form-fields.

However what about client-side validation? This doesn't appear to be happening for me. Is this as designed?

LukeTowers:
@multiwebinc I think something's missing that's supposed to trigger the browser's built in client side validation. I believe form save buttons were recently changed from type="submit" to type="button" to prevent accidental form submissions when hitting enter inside of fields, so you can start looking there.

daftspunk:
Yes, client-side validation is somewhat redundant because the server is the only point that matters (clients can be tricked to bypass validation)

Instead, we emulate client-side validation with custom responses coming from the server. The UX is mostly the same and is intentionally preferred over "rolling validation" where the input validates as you enter data, which is seen as micromanaging the user

multiwebinc:
@daftspunk I get your point, which is why I was wondering if this was intentional behavior or not.

Anyway see the draft PR I submitted since it's pretty trivial to implement in case you decide to.

LukeTowers:
[...]

PR #4804, opened 2019-12-04
Validate backend forms client-side

multiwebinc:
Fixes https://github.com/octobercms/october/issues/4799


modules/system/assets/js/framework.js
```diff
@@ -30,12 +30,23 @@ if (window.jQuery.request !== undefined) {
...
+         * Prepare the options
          */
         var $form = options.form ? $(options.form) : $el.closest('form'),
             $triggerEl = !!$form.length ? $form : $el,
             context = { handler: handler, options: options }
 
+        /*
+         * Validate the form client-side
+         */
+        if (!$form[0].checkValidity()) {
```
LukeTowers:
```suggestion
        if ($form[0] && !$form[0].checkValidity()) {
```
`$form` isn't always defined

modules/system/assets/js/framework.js
```diff
@@ -30,12 +30,23 @@ if (window.jQuery.request !== undefined) {
...
+         * Prepare the options
          */
         var $form = options.form ? $(options.form) : $el.closest('form'),
             $triggerEl = !!$form.length ? $form : $el,
             context = { handler: handler, options: options }
 
+        /*
+         * Validate the form client-side
+         */
+        if ($form[0] && !$form[0].checkValidity()) {
```
w20k: It's even better to check form before making an assumption it's an array ;) 
Like so: `if ($form && $form[0] && !$form[0].checkValidity()) {`

multiwebinc: Good catch.

LukeTowers:
```suggestion
        if ($form && $form[0] && !$form[0].checkValidity()) {
```

LukeTowers:
@w20k or @bennothommo can you test this please so we can merge it?

w20k:
LGTM, will test @LukeTowers !

modules/system/assets/js/framework.js
```diff
@@ -30,12 +30,23 @@ if (window.jQuery.request !== undefined) {
...
          */
         var $form = options.form ? $(options.form) : $el.closest('form'),
             $triggerEl = !!$form.length ? $form : $el,
             context = { handler: handler, options: options }
 
+        /*
+         * Validate the form client-side
+         */
+        if ($form && $form[0] && !$form[0].checkValidity()) {
+            $form[0].reportValidity();
```
LukeTowers: @multiwebinc I really want to merge this but IE doesn't support reportValidity(), is it possible for this case to be handled smoother? I mostly don't want a console error message to appear in a browser that doesn't support that method.

multiwebinc: What would be your preferred method for handling legacy browsers? Just skip the check and send the form like in the current behavior, or implement some sort of custom legacy validation function like they did [here](https://www.sitepoint.com/html5-forms-javascript-constraint-validation-api/), and if so, how would you want the errors to be displayed to the user (e.g. beside the input, or in an alert)? Would the errors need to be translatable as well?

LukeTowers: @multiwebinc if the polyfill for supporting legacy browsers is simple enough then perhaps we could go that route, but if not then I'd rather just skip the check on the legacy browsers since that's the current behavior and this is more of a progressive enhancement.

multiwebinc: A polyfill for supporting legacy browsers is not trivial, so I just added a check if the browser supports validation, and if not it behaves as before. Please test because I don't have access to any legacy browsers.

multiwebinc: Also, I wasn't sure exactly where the function should be placed, so I put it with a few other ones.

LukeTowers: @w20k are you able to test the polyfill? I don't think it's going to help in this case because IE11 partially supports this API AFAIK, it just doesn't support reportValidity(). Also it doesn't need to be a separate function

multiwebinc: > it doesn't need to be a separate function

True. I'm just used to breaking long lines up, but I guess that doesn't make much sense this case.

> I don't think it's going to help in this case because IE11 partially supports this API AFAIK

Then we could do `typeof document.createElement( 'input' ).checkValidity == 'function'` [[ref](https://stackoverflow.com/a/8550926)].

multiwebinc:
I'm wondering if it's better to do what I've done here with the "required" attribute in each of the partials, or else just make `$field->getAttributes()` output it instead.

modules/system/assets/js/framework.js
LukeTowers: @multiwebinc it needs to check the `reportValidity()` method, see https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/reportValidity vs https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement/checkValidity. IE 10 + supports `checkValidity()` but doesn't support `reportValidity()`

LukeTowers:
@multiwebinc do it in `getAttributes()` instead

mjauvin:
Something like this:
```
    public function getAttributes($position = 'field', $htmlBuild = true)
    {   
        $result = array_get($this->attributes, $position, []);
        $result = $this->filterAttributes($result, $position);

        if ($position === 'field' && $this->required && !isset($result['required'])) {
            $result[] = 'required => false';
        }   

        return $htmlBuild ? Html::attributes($result) : $result;
    }   
```

LukeTowers:
Just note that developers should have the ability to override the default required attribute if desired.

multiwebinc:
@mjauvin Thanks for the code snippet. I changed it up a bit to account for "required" being added to the "attributes" option.

@LukeTowers Developers can still control whether the field is required or not. There's nothing hard-coded.

modules/system/assets/js/framework.js
multiwebinc: @LukeTowers Thanks for the info. Take another look.

LukeTowers:
Absolutely stellar work @multiwebinc! Thanks very much for pushing this through! And big thanks to @petehalverson as always for the fantastic Octodock to test this out rapidly! ❤️

multiwebinc:
@LukeTowers Thanks dude! Glad I could help.

mjauvin:
This works very well @multiwebinc !

One thing I noted while testing, if the Model uses the Validation trait and a field is added to the $rules with "required" as in:
```
$rule = [
  'myField' => 'required'
];
```
The "required" parameter will also be added to the html form input, as if the field config's "required" key had been set... Is that intentional behavior? Personally I think it's fine, not sure if that was intended though... @LukeTowers

multiwebinc:
@mjauvin I'm sure that's by design. If the field is required in the model, it would make sense that the form to add/update that model would also make that field required to the user. I see putting it in the yaml more as overriding the model's default behavior. The precedence in generating the backend form is the yaml `attributes` : `required` > yaml `required` > model rules.

LukeTowers:
@mjauvin yup, that's intended behaviour. It's how the red dot gets added to the label automatically based on the model validation rules.

PR #4807, opened 2019-12-06
Fixed DataTable dropdowns in Modals

tobias-kuendig:
This is a minor bug I encountered when using the DataTable widget in a modal.

The dropdown inside the datatable uses some strange `:focus` selectors to pass along the selected dropdown value.

When the DataTable is inside the modal, it seems like the modal itself is stealing the focus. I figured this out by logging the focused element to the console, immediately after the `.focus()` call in the `onItemClick` handler:

```js
    DropdownProcessor.prototype.onItemClick = function(ev) {
        var target = this.tableObj.getEventTarget(ev)

        if (target.tagName == 'LI') {
            target.focus();
            console.log(document.querySelectorAll(':focus'))
            this.updateCellFromFocusedItem()
            this.hideDropdown()
        }
    }
```

As you can see in this GIF, the `li` is not focused correctly. It works as long as the DataTable is not inside the modal.

popup

The easiest workaround for this problem is to pass along the target element explicitly since it is already known in this case. I'm aware that this is not solving the actual problem but it seems to be a reasonble fix that prevents bigger changes to this component.

You can reproduce this bug using my fork of the test plugin (Playground -> People -> Show DataTable):

https://github.com/OFFLINE-GmbH/test-plugin/tree/database-dropdown-bug

https://github.com/OFFLINE-GmbH/test-plugin/commit/3faa07513c3e2c8cb063f76c841ecd0181f8f2bc

### DIFF
--- a/modules/backend/widgets/table/assets/js/build-min.js
+++ b/modules/backend/widgets/table/assets/js/build-min.js
@@ -871,7 +871,8 @@ return cachingKey}
 DropdownProcessor.prototype.getAbsolutePosition=function(element){var top=document.body.scrollTop,left=0
 do{top+=element.offsetTop||0;top-=element.scrollTop||0;left+=element.offsetLeft||0;element=element.offsetParent;}while(element)
 return{top:top,left:left}}
-DropdownProcessor.prototype.updateCellFromFocusedItem=function(){var focusedItem=this.findFocusedItem();this.setSelectedItem(focusedItem);}
+DropdownProcessor.prototype.updateCellFromFocusedItem=function(focusedItem){if(!focusedItem){focusedItem=this.findFocusedItem();}
+this.setSelectedItem(focusedItem);}
 DropdownProcessor.prototype.findSelectedItem=function(){if(this.itemListElement)
 return this.itemListElement.querySelector('ul li.selected')
 return null}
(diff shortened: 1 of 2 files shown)


bennothommo:
@tobias-kuending Thanks for the report. Note however that people can also use the keyboard to navigate the dropdown items as well, either by using the down and up arrow keys, or by typing in words to act like an invisible query. Do these also exhibit the same focus issue with the popup as well?

Have you been able to determine why the popup is taking away the focus during your tests?

daftspunk:
The bootstrap modal implementation enforces the user focus at all times

https://github.com/octobercms/october/blob/30f7c037182efbb21d4668f9f6d0468f438dff40/modules/system/assets/ui/vendor/bootstrap/js/modal.js#L115-L131

The functionality is designed to prevent you from focusing on the content behind the modal/popup when its open. So any content that exists outside the popup contents (dropdowns, color pickers, etc.) is never allowed to have focus

We overcome it with a `popup-allow-focus` class within the popup contents

daftspunk:
Confirmed this fix works. I have immortalized this test in the [test plugin](https://github.com/octoberrain/test-plugin/commit/6ba8bd28938ba212bc45a57f99e2b0d6849cc4ee), its good example of manual form and popup construction, thanks @tobias-kuendig !

@bennothommo unfortunately the text searching feature appears to be broken inside modals, grab the latest test plugin and navigate to People > **Show DataTable**

I think we can merge this and address the searching problem in another PR

tobias-kuendig:
FYI, I tried editing the DataTable data in the modal using only the keyboard and it seems like the focus stealing interferes with it a lot.

In this example, I tab through the fields. First, ony the checkboxes get focused every second "tab". When I explicitly set focus using the mouse, I am able to tab through all fields. Dropdown values can be selected using up and down arrows, but once you hit space to open the menu, you can no longer select any of the values.

This all works if the DataTable is not mounted inside the modal.

focus

PR #4811, opened 2019-12-09
Remove JS layout calculations of the backend navbar by using flexbox

tobias-kuendig:
By using flexbox for the backend navbar, we can prevent the usage of the `data-calculate-width` helper which is run on every page load and visibly fixes the navbar:

#### Before

before

#### After

after

* :heavy_check_mark: Drag scroll works 
* :heavy_check_mark: If the items overflow, the active item is scrolled into view
* :heavy_check_mark: Works in IE11

Let me know if I forgot a feature that needs testing!

As a side note: What is the current IE version October targets? There is [some `for .. of` code](https://github.com/octobercms/october/blob/882ca6019c9c1b124e6e3f052b6aec02b6fa826b/modules/backend/layouts/_head.htm#L63) in the backend that won't work in IE11:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of#Browser_compatibility

### DIFF
--- a/modules/backend/assets/css/october.css
+++ b/modules/backend/assets/css/october.css
@@ -660,7 +660,7 @@ ul.mainmenu-nav li {}
 ul.mainmenu-nav li .svg-icon {-webkit-backface-visibility:hidden;backface-visibility:hidden}
 ul.mainmenu-nav li span.counter {display:block;position:absolute;top:.143em;right:0;padding:.143em .429em .214em .286em;background-color:#d9350f;color:#fff;font-size:.786em;line-height:100%;-webkit-border-radius:3px;-moz-border-radius:3px;border-radius:3px;opacity:1;filter:alpha(opacity=100);-webkit-transform:scale(1,);-ms-transform:scale(1,);transform:scale(1,);-webkit-transition:all 0.3s;transition:all 0.3s}
 ul.mainmenu-nav li span.counter.empty {opacity:0;filter:alpha(opacity=0);-webkit-transform:scale(0,);-ms-transform:scale(0,);transform:scale(0,)}
-nav#layout-mainmenu {background-color:#000;padding:0 0 0 20px;line-height:0;white-space:nowrap;vertical-align:top}
+nav#layout-mainmenu {background-color:#000;padding:0 0 0 20px;line-height:0;white-space:nowrap;display:flex}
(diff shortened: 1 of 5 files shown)


bennothommo:
@tobias-kuendig I believe we are still supporting IE11 in the Backend, so please feel free to convert those `for..of` loops when spotted.

ayumi-cloud:
I really like this pr idea, I hope you dont mind if I add some comment suggestions.

1. Your flexbox won't work correctly in IE and other browsers as you are not using correct prefixes, for example:

```css
display: flex;
```

Needs to be prefixed to this:

```css
display:-webkit-box;
display:-ms-flexbox;
display:flex;
```

Likewise with the rest of the flexbox code.

2. There are so many Safari and IE bugs with `flex-basis` I strongly suggest you remove all that code and switch to using `width` and/or `min-width` instead. There are around a dozen open issues with these two browsers on `flex-basis`.

3. There is also a zero bug in IE with flexbox, instead of writing `0` you need to write `0px`, for example:

```css
flex: 0 0 auto;
```

would have to be written as:

```css
flex: 0px 0px auto;
```

...stupid bug, but that's the official solution.

4. I only see LESS code and no complied CSS yet.

Hope that helps!

p.s. thanks for fixing this issue, it's been annoying us for a while.

tobias-kuendig:
Thank you for this feedback. I did test the code in IE11 and everything works fine without prefixes. As IE11 is the lowest targeted version I think we don't need to add vendor prefixes.

I did not test the PR in Safari as I currently do not have access to a Mac. Maybe someone can provide feedback for Safari here.

@LukeTowers should I compile the CSS for this PR? IIRC the compilation is done by the core maintainers to prevent merge conflicts.

bennothommo:
@tobias-kuendig Feel free to compile the CSS - if it conflicts with another PR, we'll get them to merge the latest changes from `develop` and re-compile.

bennothommo:
@tobias-kuendig Just a ping on this to compile the asssets - we'll test and then merge once done.

tobias-kuendig:
Sorry, didn't see your comment back in december! I have added the compiled backend JS and CSS files.

bennothommo:
@tobias-kuendig Thanks heaps!

PR #4816, opened 2019-12-10
Add option to set custom content type for cms page

Samuell1:
This adds option to set custom Content-Type header per CMS page. It allows to simplify job of creating dynamic robots.txt file (https://octobertricks.com/tricks/dynamic-robotstxt) or any other files that needs to be other content type then text/html.

obrázok

### DIFF
--- a/modules/cms/classes/Controller.php
+++ b/modules/cms/classes/Controller.php
@@ -252,6 +252,13 @@ class Controller
         }
 
         /*
+         * Apply custom content type header
+         */
+        if ($page->contentType) {
+            $this->setResponseHeader('Content-Type', $page->contentType);
+        }
+
+        /*
          * Prepare and return response
          * @see \System\Traits\ResponseMaker
          */
--- a/modules/cms/classes/page/fields.yaml
+++ b/modules/cms/classes/page/fields.yaml
@@ -43,6 +43,12 @@ tabs:
             type: dropdown
             options: getLayoutOptions
 
+        settings[contentType]:
+            tab: cms::lang.editor.settings
+            span: left
+            label: cms::lang.editor.contenttype
+            comment: cms::lang.editor.contenttype_comment
+
         settings[description]:
             tab: cms::lang.editor.settings
             label: cms::lang.editor.description
--- a/modules/cms/lang/en/lang.php
+++ b/modules/cms/lang/en/lang.php
@@ -194,6 +194,8 @@ return [
         'resetting' => 'Resetting',
         'commit_success' => 'The :type has been committed to the filesystem',
         'reset_success' => 'The :type has been reset to the filesystem version',
+        'contenttype' => 'Content Type',
+        'contenttype_comment' => 'If not set defaults to "text/html".'
     ],
     'asset' => [
         'menu_label' => 'Assets',
--- a/themes/demo/pages/test.htm
+++ b/themes/demo/pages/test.htm
@@ -1,0 +1,6 @@
+title = "test"
+url = "/test"
+contentType = "text/plain"
+is_hidden = 0
+==
+test
--- a/themes/demo/pages/test.htm
+++ b/themes/demo/pages/test.htm
@@ -1,6 +1,0 @@
-title = "test"
-url = "/test"
-contentType = "text/plain"
-is_hidden = 0
-==
-test


bennothommo:
@Samuell1 I'm not sure it is wise to give people that much control over the definition of a file. It also doesn't account for variable pages, which may require different media types. It's probably a good candidate for a plugin though.

Samuell1:
@bennothommo What is variable pages? It should be backwards compatible with everything.

AniketMagadum:
@Samuell1 but will it support syntax highlighting as files are saved with .htm extension

LukeTowers:
@daftspunk thoughts?

Samuell1:
@AniketMagadum It should? Not sure what you mean now.

bennothommo:
@Samuell1 Sorry only just saw your response - when I mean variable pages, I mean pages that have variables in the URL (ie. /test/:id). Extreme edge case - I know, but you could theoretically have different media types, if say, you had an API response being generated on a page that can be delivered as JSON or as XML.

Samuell1:
@bennothommo Yes, it should work with variable pages, most likely this will be used to return pages like robots.txt, ads.txt, sitemap.xml or any files that doesnt need to be changed. Or maybe you can use it to return json api for testing etc.

Target is for devs to prepare for example robots.txt file already in theme.

ayumi-cloud:
@Samuell1 see here: https://octobercms.com/plugin/xeor-contenttype

Samuell1:
@ayumi-cloud yes i know i have plugin too but this feature should be in core.

ayumi-cloud:
@Samuell1 

Yah I agree it should be in the core.

Plugin works fine in build 460, I see no problems, could you share me a link of the Response changes please. I want to have a look at the code change.

p.s. Aren't you going to keep the same design pattern?

image

I like having to select a list for quick access and if need be add a custom one (saves time).

mjauvin:
Personally, I prefer the way @Samuell1 did it. It's cleaner and does not add an extra tab.

Samuell1:
@ayumi-cloud This was changed it calls now trait that handles response with headers https://github.com/octobercms/october/blob/193dac80bdee7e521d35ae1637272eadff9bbb49/modules/cms/classes/Controller.php#L265

I am not sure what saves time when you set content type once per page. But maybe if there is option to add select with custom typed text then yes, i dont like to have 2 inputs for one.

ayumi-cloud:
@Samuell1 thanks for the link reference.

ayumi-cloud:
p.s. Is there anyway to code it like this:

image

Where we can add a set of pre-defined mime-types and add custom ones to a list. This way it reduces the chances of human error having to write the mime-type for every single web page template we need to change. In my case I would need to change over 500 templates for a single website that has over 10 million web pages.

Samuell1:
@ayumi-cloud taglist supports something like that, but for multiple maybe if there is way to limit it for one then it will be fine.

ayumi-cloud:
@Samuell1 

Here's an expanded version of my comment:

### Have a single dropdown field - span-full

```html
<select>
  <optgroup label="Custom Mime-Types">
      <option value="manifest">Manifest</option>
  </optgroup>
  <optgroup label="Common Mine-Types">
      <option value="css">Css</option>
      <option value="html">Html</option>
      <option value="js">Javascript</option>
      <option value="mjs">Javascript Modules</option>
  </optgroup>
  <optgroup label="Edit">
      <option value="edit">Edit</option>
  </optgroup>
</select>
```

### When user selects `edit`

When a user selects `edit` a modal window opens with two fields:

```yaml
type: text
span: left
label: 'Label'

type: text
span: right
label: 'Mime-Type'
```

Then user clicks save and it's add to the `custom mime-type` list.

bennothommo:
@Samuell1 Following on from our discussion last night, and for anyone else interested, you can change the content type via the "Code" section of a CMS page or layout:
```php
<?php
function onStart()
{
    $this->setResponseHeader('Content-Type', 'text/plain');
}
```

LukeTowers:
@bennothommo could you add the available ResponseMaker traits methods to the section of the docs for the Page / Layout "code" section?

LukeTowers:
@Samuell1 as far as this goes for inclusion into the core, I'm kind of on the fence about it. It's something that can (and has been) done in a plugin, and in the code section now, so I'm not sure we should be including it as a field by default. October's creed is that everything should be as simple as possible, no simpler; so we want to be very careful when considering adding new fields to the default forms.

@bennothommo @w20k @daftspunk thoughts?

bennothommo:
@LukeTowers will do.

Also, I'm still not convinced it needs to be included. It has the potential to create issues with page display and security. The sort of people who need control over the content type are the sort of people who can work out a way around this limitation :)

ayumi-cloud:
https://bitbucket.org/sozonovalexey/oc-contenttype-plugin/issues/4/update-plugin-to-work-with-octobers-new

Simple fix to this issue.

PR #4820, opened 2019-12-11
1. Register helpers in plugins

LeMaX10:
Registration of helpers from plugins, so that plugins can implement their own helpers.

Refactoring in the string methods plugin manager. Since the system should have mbstring available by default, it is recommended to use the string methods namely mb_ *.

### DIFF
--- a/modules/system/classes/PluginManager.php
+++ b/modules/system/classes/PluginManager.php
@@ -203,7 +203,7 @@ class PluginManager
         }
 
         $pluginPath = $this->getPluginPath($plugin);
-        $pluginNamespace = strtolower($pluginId);
+        $pluginNamespace = mb_strtolower($pluginId);
 
         /*
          * Register language namespaces
@@ -251,6 +251,14 @@ class PluginManager
         $initFile = $pluginPath . '/init.php';
         if (!self::$noInit && File::exists($initFile)) {
             require $initFile;
+        }
+
+        /*
+         * Add helpers
+         */
+        $helpersFile = $pluginPath . '/helpers.php';
+        if (File::exists($helpersFile)) {
+            require $helpersFile;
         }
 
         /*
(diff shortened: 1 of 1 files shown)


modules/system/classes/PluginManager.php
```diff
@@ -253,6 +253,14 @@ public function registerPlugin($plugin, $pluginId = null)
             require $initFile;
         }
 
+        /*
+         * Add helpers
+         */
+        $helpersFile = $pluginPath . '/helpers.php';
+        if (File::exists($helpersFile)) {
+            require $helpersFile;
```
LukeTowers: Have you tested this? I wouldn't think that it would include the helpers.php file globally since this is happening within the scope of the registerPlugin() method.

LeMaX10: Yes, I tested this before posting. In this case, the logic works by analogy with connecting routes, the file is require into the system during the registration of the plug-in if it exists.

LukeTowers: routes.php and init.php are different in that they are including code that should be executed immediately, not defining global functions. I will have to test this myself before I can merge.

LukeTowers:
@LeMaX10 I don't think we want to support multi-btye strings in the Plugin ID.

modules/system/classes/PluginManager.php
LeMaX10: During testing, I used a tinker, and also implemented a dependent plugin.

> routes.php and init.php are different in that they are including code that should be executed immediately, not defining global functions. I will have to test this myself before I can merge.

Oh sure. I will be glad to your check

LeMaX10:
> @LeMaX10 I don't think we want to support multi-btye strings in the Plugin ID.

I think these edits can be rolled back.

LukeTowers:
@LeMaX10 go ahead and revert them please

LeMaX10:
> @LeMaX10 go ahead and revert them please

Rollback mb_strtolower to strtolower

bennothommo:
@LeMaX10 Thanks for your contribution, however, I feel this change should not be included as it will blindly require any `helpers.php` file that is contained in a plugin - this might cause issues with plugins that already contain one of these files and are including them in a different way.

Note that the better way to do this would be to do something like this in your `Plugin.php` file:

```php
public function boot()
{
    require_once __DIR__ . '/helpers.php';
    
    // Any other boot functionality below...
}
```

This way, the `helpers.php` script is covered by our exception catcher that will disable plugins if they are unable to boot, allowing sites to continue to work even with a faulty plugin.

PR #4822, opened 2019-12-12
Fix saving of translations in relation controller by re-using the form widget's model

tobias-kuendig:
This PR is an addition to https://github.com/octobercms/october/pull/4193

Re-using the pivotWidget's and manageWidget's models for saving form data fixes https://github.com/rainlab/translate-plugin/issues/209, where translations are not saved for the model when editing it in a relation widget.

I did some manual testing with this change and everything was working as expected: Translations are now properly saved when creating or editing related models with pivot data. Creating related models in `single` mode now works as well.

Credits to @alxy for his previous work on this issue!

### DIFF
--- a/modules/backend/behaviors/RelationController.php
+++ b/modules/backend/behaviors/RelationController.php
@@ -1106,7 +1106,7 @@ class RelationController extends ControllerBehavior
             $this->relationObject->add($newModel, $sessionKey);
         }
         elseif ($this->viewMode == 'single') {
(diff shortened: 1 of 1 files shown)


LukeTowers:
@mjauvin could you test this out?

mjauvin:
> @mjauvin could you test this out?

@LukeTowers I confirm this solves the issue!

mjauvin:
This can be merged as far as I'm concerned.

PR #4824, opened 2019-12-13
Feature/relation controller check contains

SebastiaanKloos:
Added a check to prevent the relation controller from adding a new link between master and slave when link already made by using afterSave model event

### DIFF
--- a/modules/backend/behaviors/RelationController.php
+++ b/modules/backend/behaviors/RelationController.php
@@ -1099,7 +1099,9 @@ class RelationController extends ControllerBehavior
                 $modelToSave->save(null, $this->manageWidget->getSessionKey());
             }
 
-            $this->relationObject->add($newModel, $sessionKey);
+            if (!$this->model->{$this->relationName}->contains($newModel)) {
+                $this->relationObject->add($newModel, $sessionKey);
+            }
         }
         elseif ($this->viewMode == 'single') {
             $newModel = $this->viewModel;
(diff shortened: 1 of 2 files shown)


modules/backend/behaviors/RelationController.php
```diff
@@ -1099,7 +1099,9 @@ public function onRelationManageCreate()
                 $modelToSave->save(null, $this->manageWidget->getSessionKey());
             }
 
-            $this->relationObject->add($newModel, $sessionKey);
+            if (!$this->model->{$this->relationName}->contains($newModel)) {
```
36864: This will load all related models from the database to check for duplicates.

If you have a large dataset, the process is likely to run out of memory.

SebastiaanKloos: Allright. Do you have a suggestion for a fix? 😃

36864: From your issue description, you appear to be trying to solve a problem in your code at the framework level. Why are you saving the relationship in the afterSave event?

SebastiaanKloos:
I think I found a workaround for this issue

modules/backend/behaviors/RelationController.php
SebastiaanKloos: Because the repeater should tell the parent where to link to

CleanShot 2019-12-13 at 15 30 57@2x

LukeTowers:
@SebastiaanKloos could you post it here for future explorers?

SebastiaanKloos:
I think you are able to exclude the activity_id from the parent model in the afterSave event. Simply exclude the id from the array used by the sync method. I will create the fix next week and post an example then

Issue #4828, opened 2019-12-14
Expand the character limit of the error message shown in the list view of system errors


ayumi-cloud:
Currently event logs look like this:

![image](https://user-images.githubusercontent.com/57409060/70855512-cde4c780-1ec3-11ea-883d-1f07a4dbe412.png)

The problem I find is that, I can't read the whole message and have to open the error log, just to be able to read the message everytime!

So is the same annoying thing Google Search does! They cut the search result and stop you from hovering a mouse to read the full title or description.

This bug is creating an extra click for users which is not needed.

So I click to read this:

![image](https://user-images.githubusercontent.com/57409060/70855539-52374a80-1ec4-11ea-9277-934bf25ba633.png)

When really all I need is the cms to do this:

![image](https://user-images.githubusercontent.com/57409060/70855559-a6422f00-1ec4-11ea-936b-f4681f3a4da7.png)

I just want to see the first error line in full.

I take it this is the code line that's annoying me:

https://github.com/octobercms/october/blob/56eab502606c2c1bb2e64f08e27db1467d211b4f/modules/system/models/EventLog.php#L82-L89

Please, please, please can we remove this character limit.

LukeTowers:
@ayumi-cloud try looking at the error log page on a smaller screen, say 1200px wide and you'll see why it's required.

ayumi-cloud:
@LukeTowers I don't get what you mean?

Here's a width 1200px:

![image](https://user-images.githubusercontent.com/57409060/70870165-d1418700-1f87-11ea-9df8-7da774047b73.png)

- Seems fine to me.

iPad example:

![image](https://user-images.githubusercontent.com/57409060/70870199-1e255d80-1f88-11ea-9e4d-ebc50f976865.png)

- Seems fine to me.

Smart phone example:

![image](https://user-images.githubusercontent.com/57409060/70870212-3eedb300-1f88-11ea-9c3e-91f973282b1a.png)

- I only issue an issue with the list not being responsive, but this issue would be the same if we cut the character limit to 100 anyway.

Is there any other issue and reason?

I would like to give full error summary and screen sizes under 756px make the list widget responsive and allow user to scroll left and right.

LukeTowers:
Fair enough @ayumi-cloud, it was probably just an arbitrary number chosen to prevent the whole thing from being displayed which would definitely cause problems. Your regex change seems fine to me, just needs to get tested.

bennothommo:
Fixed by #4830.

PR #4830, opened 2019-12-16
Expand the character limit of the error message shown in the list view of system errors

summercms:
Github issue: https://github.com/octobercms/october/issues/4828

My regex:

image

and

image

The regex matches all the error before it finds the exact match word of `Stack trace` and then ends the regex after the exact match word of `Stack trace` on the `:`

Giving the following result:

image

### DIFF
--- a/modules/system/models/EventLog.php
+++ b/modules/system/models/EventLog.php
@@ -84,8 +84,8 @@ class EventLog extends Model
         if (preg_match("/with message '(.+)' in/", $this->message, $match)) {
             return $match[1];
         }
-        elseif (preg_match("/^((.|\n)*)(\bStack trace\b)(\:)$/im", trim($this->message), $match)) {
-            return $match[1];
+        elseif (preg_match("/^(.*|\n)/im", trim($this->message), $match)) {
+            return str_replace('<br>', '', $match[1]);
         }
 
         return Str::limit($this->message, 100);
--- a/modules/system/models/EventLog.php
+++ b/modules/system/models/EventLog.php
@@ -84,10 +84,9 @@ class EventLog extends Model
         if (preg_match("/with message '(.+)' in/", $this->message, $match)) {
             return $match[1];
         }
-        elseif (preg_match("/^(.*|\n)/im", trim($this->message), $match)) {
+        elseif (preg_match("/^(.*|\r\n|\r|\n)/im", trim($this->message), $match)) {
             return str_replace('<br>', '', $match[1]);
         }
-
         return Str::limit($this->message, 100);
     }
 }
--- a/modules/system/models/EventLog.php
+++ b/modules/system/models/EventLog.php
@@ -88,6 +88,6 @@ class EventLog extends Model
         // Get first line of message
         preg_match('/^([^\n\r]+)/m', $this->message, $matches);
 
-        return Str::limit($matches[1], 500);
+        return Str::limit($matches[1] ?? '', 500);
     }
 }
--- a/modules/system/models/EventLog.php
+++ b/modules/system/models/EventLog.php
@@ -81,12 +81,15 @@ class EventLog extends Model
      */
     public function getSummaryAttribute()
     {
-        if (preg_match("/with message '(.+)' in/", $this->message, $match)) {
-            return $match[1];
+        if (Config::get('develop.limitEventLogSummary', true)) {
+            if (preg_match("/with message '(.+)' in/", $this->message, $match)) {
+                return $match[1];
+            }
+            return Str::limit($this->message, 100);
+        } else {
+            if (preg_match("/^(.*|\r\n|\r|\n)/im", trim($this->message), $match)) {
+                return str_replace('<br>', '', $match[1]);
+            }
         }
-        elseif (preg_match("/^(.*|\r\n|\r|\n)/im", trim($this->message), $match)) {
-            return str_replace('<br>', '', $match[1]);
-        }
-        return Str::limit($this->message, 100);
     }
 }
--- a/modules/system/models/EventLog.php
+++ b/modules/system/models/EventLog.php
@@ -84,6 +84,9 @@ class EventLog extends Model
         if (preg_match("/with message '(.+)' in/", $this->message, $match)) {
             return $match[1];
         }
+        elseif (preg_match("/^((.|\n)*)(\bStack trace\b)(\:)$/im", trim($this->message), $match)) {
+            return $match[1];
+        }
 
         return Str::limit($this->message, 100);
     }
--- a/modules/system/models/EventLog.php
+++ b/modules/system/models/EventLog.php
@@ -84,11 +84,14 @@ class EventLog extends Model
         if (Config::get('develop.limitEventLogSummary', true)) {
             if (preg_match("/with message '(.+)' in/", $this->message, $match)) {
                 return $match[1];
+            } else {
+                return Str::limit($this->message, 100);
             }
-            return Str::limit($this->message, 100);
         } else {
             if (preg_match("/^(.*|\r\n|\r|\n)/im", trim($this->message), $match)) {
                 return str_replace('<br>', '', $match[1]);
+            } else {
+                return Str::limit($this->message, 100);
             }
         }
     }
--- a/modules/system/models/EventLog.php
+++ b/modules/system/models/EventLog.php
@@ -84,9 +84,10 @@ class EventLog extends Model
         if (preg_match("/with message '(.+)' in/", $this->message, $match)) {
             return $match[1];
         }
-        elseif (preg_match("/^(.*|\r\n|\r|\n)/m", trim($this->message), $match)) {
-            return Str::limit(str_replace('<br>', '', $match[1]), 500);
-        }
-        return Str::limit($this->message, 100);
+
+        // Get first line of message
+        preg_match('/^([^\n\r]+)/m', $this->message, $matches);
+
+        return Str::limit($matches[1], 500);
     }
 }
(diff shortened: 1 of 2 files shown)


ayumi-cloud:
Reason for commit: https://github.com/octobercms/october/pull/4830/commits/6ec9d838500e059ea07140688a23e12e61bb986d

We have been running this pr at work for several days, we found some issues, so we updated the pr to fix them. The new regex finds all possible error messages correctly and there is a hidden `<br>` at the end of the first sentence in the dom code, so we have removed that to give an improved displayed output.

ayumi-cloud:
Reason for commit: https://github.com/octobercms/october/pull/4830/commits/6ec9d838500e059ea07140688a23e12e61bb986d

We have been running this pr in the office at our company for several days, we found some issues, so we updated the pr to fix them. The new regex finds all possible error messages correctly! There is a hidden `<br>` at the end of the first sentence in the dom code, so we have removed that to give an improved displayed output, for example:

### old

```html
<td class="list-cell-index-3 list-cell-name-message list-cell-type-partial   column-break-word">
ErrorException: Undefined property: Acme\Example\Something::$0 in /home/example/public_html/plugins/acme/something.php:116<br></td>
```

Note: The `<br></td>` at the end.

### new

```html
<td class="list-cell-index-3 list-cell-name-message list-cell-type-partial   column-break-word">
ErrorException: Undefined property: Acme\Example\Something::$0 in /home/example/public_html/plugins/acme/something.php:116</td>
```

Note: The `</td>` at the end.

modules/system/models/EventLog.php
```diff
@@ -84,6 +84,9 @@ public function getSummaryAttribute()
         if (preg_match("/with message '(.+)' in/", $this->message, $match)) {
             return $match[1];
         }
+        elseif (preg_match("/^(.*|\n)/im", trim($this->message), $match)) {
```
w20k: @ayumi-cloud it's a really bad idea to have `.*` regexp in here. And your newline part - won't work. Mainly you'd have a carriage return at the beginning. Long story short, is a better regexp for a newline character (sort of): `(\r\n|\r|\n)`

And if you'd like to have `<br>` after a newline character better to find those and insert before.

Details:
- in Unix and all Unix-like systems, \n is the code for end-of-line, \r means nothing special
- as a consequence, in C and most languages that somehow copy it (even remotely), \n is the standard escape sequence for end of line (translated to/from OS-specific sequences as needed)
- in old Mac systems (pre-OS X), \r was the code for end-of-line instead
-  in Windows (and many old OSs), the code for end of line is 2 characters, \r\n, in this order
- as a (surprising;-) consequence (harking back to OSs much older than Windows), \r\n is the standard line-termination for text formats on the Internet
- for electromechanical teletype-like "terminals", \r commands the carriage to go back leftwards until it hits the leftmost stop (a slow operation), \n commands the roller to roll up one line (a much faster operation) -- that's the reason you always have \r before \n, so that the roller can move while the carriage is still going leftwards!-) Wikipedia has a more detailed explanation.
- for character-mode terminals (typically emulating even-older printing ones as above), in raw mode, \r and \n act similarly (except both in terms of the cursor, as there is no carriage or roller;-)

ayumi-cloud: @w20k Thanks, I totally forgot about adding support for Mac's. Going to save this comment in my notes, thanks for going the extra mile with all the details, very useful. I remember those days using `\r`

I'm going to add your regex in a test website and run it for a little while to test it and then I will update the pr.

Thanks for your review 👍

w20k:
@ayumi-cloud could I ask you to describe your issue in detail? Because from your PR point of view, the limit part (last part after if statement) won't' be executed ever! :)
There might be a better option then just chewing everything from the input string with your regexp: `preg_match("/^(.*|\r\n|\r|\n)/im",`

ayumi-cloud:
@w20k @LukeTowers 

1st happy new year ...

2nd sorry you lost me !

This is the result from the pr:

image

The code is running using that if statement.

To me the regex is: match everything up to the first line break.I choose `.*` instead of `.+` because it can display nothing.

ayumi-cloud:
> But the easiest solution would be to hover over the message and show a full message in a popup (of some sort), maybe?

Not really feeling that idea, my goal for this pr was to display the first line of code which tells users the summary of the error, if they want more info they can then click on the link. One reason I hate the idea of a pop-up is adding animation (which October is bad with Jank at the present) and how are you going to display this pop-up with a user hovering their finger over the list item on a tablet device? Won't work nicely on touch screens.

w20k:
@ayumi-cloud just an idea, nothing else. Could you ping @LukeTowers one more time about the `limit`, maybe it'd be wise to add it as an configurable option? Like: `on/off the limit`?

ayumi-cloud:
@w20k @LukeTowers I'm up for letting users be able to choose as they like, so how about I create an option in one of the config files and wrap that code to turn it on and off.

I leave Luke to decide the name and location if you both are happy with the idea and then let me know and I will update the pr.

ayumi-cloud:
@w20k please review my last commit when you free and let me know your thoughts, thanks.

ayumi-cloud:
@w20k please review my last commit when you free and let me know your thoughts, thanks.

LukeTowers:
@w20k what're your concerns with making the entire first line be displayed instead of the first 100 characters?

w20k:
@LukeTowers my only concern is that the regexp will work every time (`"/^(.*|\r\n|\r|\n)/im"`). Every new doesn't matter how long it would be and limit part won't even be called.

ayumi-cloud:
@w20k I see your point, but surely there must be some limiting going on by the Laravel error handling vendor files that deal with that, to stop a single line going on forever.

I wait for what you guys decide and then I will update the pr. I'll probably revert the last two commits.

mjauvin:
I would prefer the option's value to be true by default

mjauvin:
Also, why is there a "i" in the regex options? It's not needed and will take more CPU to run

mjauvin:
But the best way to handle this, in my opinion, would be to fire an event to allow a plugin to modify that string.

mjauvin:
I suggest something like this:
```
    public function getSummaryAttribute()
    {   
        if (preg_match("/with message '(.+)' in/", $this->message, $match)) {
            return $match[1];
        }

        /**
         * @event system.eventlog.overrideSummaryValue
         * Provides an opportunity to modify the EventLog message summary value
         *
         * Example usage:
         *
         *     Event::listen('system.eventlog.overrideSummaryValue', function ((string) $message) {
         *         if (preg_match("/^(.*|\r\n|\r|\n)/m", trim($message), $match)) {
         *    ￼       return str_replace('<br>', '', $match[1]);
         *         }
         *     });
         */
        if ($message = Event::fire('system.eventlog.overrideSummaryValue', [$this->message])) {
            return $message;
        }   

        return Str::limit($this->message, 100);
    }
```

ayumi-cloud:
I was thinking of just doing this:

```php
    public function getSummaryAttribute()
    {
        if (preg_match("/with message '(.+)' in/", $this->message, $match)) {
            return $match[1];
        }
        elseif (preg_match("/^(.*|\r\n|\r|\n)/im", trim($this->message), $match)) {
            return Str::limit(str_replace('<br>', '', $match[1]), 500);
        }
        return Str::limit($this->message, 100);
    }
```

- Creates a max limit of 500 characters, which would be more than enough and then remove the config file.

LukeTowers:
I don't really see a benefit to a separate config option for this either, we should just have sane / useful defaults.

ayumi-cloud:
When a decision is made with this pls tag me .... for now everyone at our company is hard coding this into their october cms while we develop plugins and themes etc. we all need this feature!

LukeTowers:
@bennothommo can you take a look at this when you have time?

bennothommo:
@LukeTowers will do.

bennothommo:
@ayumi-cloud I would suggest using this for the `getSummaryAttribute()`:
```php
    public function getSummaryAttribute()
    {
        if (preg_match("/with message '(.+)' in/", $this->message, $match)) {
            return $match[1];
        }

        // Get first line of message
        preg_match('/^([^\n\r]+)/m', $this->message, $matches);

        return Str::limit($matches[1], 500);
    }
}
```
It simplifies the regex right down to basically "get everything up until the first line break", and doesn't capture the first line break it hits, so you don't need to the `str_replace` for the `<br>` tag.

It also drops the original final line of the method, as with your regex in place, there's no way that line would be hit.

bennothommo:
Otherwise, it's all tested and confirmed working.

ayumi-cloud:
@bennothommo thanks, looks good to me, have uploaded your commit. 👍

bennothommo:
@ayumi-cloud Just made a quick tweak to echo an empty string if the regex doesn't make any matches (however extremely unlikely that would be).

ayumi-cloud:
@w20k @LukeTowers @bennothommo 

Sorry to ask this! When this pr has been merged. We were working on line breaks with regex today and the following combinations we ended up using:

```
(\\r|\\n|%0A|%0D|%0a|%0d)
```

I was just wondering if it was a good idea to add the hexidecimal line breaks as well e.g. `%0A` and `%0D` to cover all possible outcomes or this would not be needed - thought I'd ask just to be on the safe side.

p.s. Thanks bennothommo  for the tweak, I keep missing adding `?? ''` to lots of things, but always highly needed in many cases! I really love that php 7 feature.

mjauvin:
no, hexadecimal values are not needed, this is covered by the escape sequence (\r\n)

ayumi-cloud:
@mjauvin perfect, thanks 👍

PR #4834, opened 2019-12-19
Minor fixes for getting a error-free codebase

mdougllas:
The Todo app on the CMS demonstration was throwing an exception if you tried to insert more than 3 tasks instead of showing the message about limit for tasks. Additionally, there was an error on the namespace for the plugin test that was fixed.

The benefit for the end users is a Todo app example without throwing exceptions.

All tests were ran after the fixing and they just work fine.

### DIFF
--- a/plugins/october/demo/components/Todo.php
+++ b/plugins/october/demo/components/Todo.php
@@ -1,7 +1,7 @@
 <?php namespace October\Demo\Components;
 
 use Cms\Classes\ComponentBase;
-use ApplicationException;
+use October\Rain\Exception\ApplicationException;
 
 class Todo extends ComponentBase
 {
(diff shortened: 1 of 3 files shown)


bennothommo:
@mdougllas Thanks for your efforts, but unfortunately we cannot accept this.

I have just tested the Todo pluign on a fresh install and I saw the limit error just fine.

Demo Test

`use ApplicationException` can be called like that as `ApplicationException` is loaded as a global alias, so it does not need to be fully qualified in its namespace.

Regarding the "Goto" namespace, that is intentional - it is to test that plugins are gracefully disabled if they contain parse errors.

mdougllas:
Yes, my bad. I’m sorry for the mistake. Will learn the CMS and the codebase better. Just found it and was too exited. Thank you for the great work!

PR #4835, opened 2019-12-19
Allow users to skip seeding when running migrations for modules.

### LINKED ISSUE #2360: Seed admin user with random password on initial system seed
Hi,

Just wondering if there is a way to not seed the database on initial `php artisan october:up`?

I have found the below in `/modules/system/classes/UpdateManager.php` line 150:

```
public function update()
{

...

if ($firstUp) {
    $modules = Config::get('cms.loadModules', []);
    foreach ($modules as $module) {
        $this->seedModule($module);
    }
}

...
```

I would like this not to run. I can only think to create my own `october:up` command, and my own `UpdateManager` class, extend the original and only modify the `update()` method. Seems like there might be a better way? If not will do that for now.

Thanks,
Mat

bennothommo:
Adds the "--no-seed" option for the "october:up" and "october:install" Artisan commands.

Fixes #2360.

### DIFF
--- a/modules/system/classes/UpdateManager.php
+++ b/modules/system/classes/UpdateManager.php
@@ -128,9 +128,10 @@ class UpdateManager
 
     /**
      * Creates the migration table and updates
+     * @param bool $seed If true, the tables will also be seeded.
      * @return self
      */
-    public function update()
+    public function update(bool $seed = true)
     {
         $firstUp = !Schema::hasTable($this->getMigrationTableName());
         if ($firstUp) {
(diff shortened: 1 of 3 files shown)


LukeTowers:
@bennothommo if we're doing this can we also have a command to add an admin account?

bennothommo:
@LukeTowers I had a look at what actually is seeded, and literally the only two things seeded are the admin account, and the mail layouts.

I'm half-tempted to change the scope of this to be a flag to disable the creation of an admin account (ie. `--no-admin`) when doing the installation or running migrations.

What do you reckon?

LukeTowers:
@bennothommo after talking it over with @daftspunk some more, I think the solution to the original issue is to add a command to change admin credentials from the command line (`october:passwd  $username $newPass`)

Additionally we should also change the seeder to generate a random password (30 characters) and output it to the console to solve the issue of insecure credentials by default.

bennothommo:
@LukeTowers Sounds good, however, I might tackle that as a separate PR then. This one could probably be closed, unless there's a compelling reason to not create mail layouts as part of the initial seed.

Issue #3521, opened 2018-04-20
Make a CLI command available to set, change and delete backend users


CptMeatball:
##### Expected behavior
Add, Change and/or Delete backend users from the command line. In this way you can create users without needing to login to the backend (or if you've locked yourself out of the website for example). 

This could be very useful during development or other scenarios. 

##### Actual behavior
There is no way to create backend users at this moment.

##### October build
All.

##### Proposed Ways
- Using tinker , but not sure that's a great solution as passing commands to tinker might be a problem
- Setting user and password in the `.env` file that is picked up by the inital `october:up`.
- Creating CLI commands like: `october:admin create` `change` or `delete`

bennothommo:
I feel this is a bit of security risk in having the command available. Someone who has CLI access could effectively add themselves with full privileges to any October site accessible through CLI, even if they per se don't have access to the database.

SebastiaanKloos:
You can already using Tinker

36864:
@SebastiaanKloos Tinker is an optional package. 
That said, if a user can run an october command (php artisan october:etc), they can also run arbitrary php code (php -s), or run php in interactive mode (php -a).
I'm having trouble coming up with a server configuration that would allow a user to run october commands but not open an interactive php prompt, load the config files to get database credentials, and then do whatever they want to do.

bennothommo:
The ability to change the password of a Backend user from the CLI has been added: https://github.com/octobercms/october/pull/4836

PR #4836, opened 2019-12-20
Allow changing of Backend user passwords through CLI.

bennothommo:
Adds an Artisan command "october:passwd" to change the password of a Backend User through CLI. This command may only be run through CLI - it will not work if called through a web handler.

Refs: #3521, #4835

### DIFF
--- a/modules/system/console/OctoberPasswd.php
+++ b/modules/system/console/OctoberPasswd.php
@@ -17,23 +17,17 @@ use Symfony\Component\Console\Question\Question;
 class OctoberPasswd extends Command
 {
     /**
-     * The console command name.
-     *
-     * @var string
+     * @var string The console command name.
      */
     protected $name = 'october:passwd';
 
     /**
-     * The console command description.
-     *
-     * @var string
+     * @var string The console command description.
      */
     protected $description = 'Change the password of a Backend user.';
 
     /**
-     * Was the password automatically generated?
-     *
-     * @var bool
+     * @var bool Was the password automatically generated?
      */
     protected $generatedPassword = false;
 
@@ -43,14 +37,14 @@ class OctoberPasswd extends Command
     public function handle()
     {
         // Must be in CLI environment
-        if (php_sapi_name() !== 'cli') {
+        if (!App::runningInConsole()) {
             $this->error('This command may only be run through the command-line.');
             exit(1);
         }
 
         $username = $this->argument('username')
             ?? $this->ask(
-                'Which user would you like to change the password for'
+                'Which user would you like to change the password for?'
             );
 
         // Check that the user exists
@@ -76,13 +70,10 @@ class OctoberPasswd extends Command
         $user->password = $password;
         $user->forceSave();
 
-        if (!$this->generatedPassword) {
-            $this->info('Password successfully changed.');
-            exit(0);
-        }
-
         $this->info('Password successfully changed.');
-        $this->output->writeLn('The new password is <info>' . $password . '</info>.');
+        if (!$this->generatedPassword) {
+            $this->output->writeLn('The new password is <info>' . $password . '</info>.');
+        }
         exit(0);
     }
 
--- a/modules/system/console/OctoberPasswd.php
+++ b/modules/system/console/OctoberPasswd.php
@@ -107,6 +107,6 @@ class OctoberPasswd extends Command
     {
         $this->generatedPassword = true;
 
-        return Str::random(14);
+        return Str::random(22);
     }
 }
--- a/modules/system/console/OctoberPasswd.php
+++ b/modules/system/console/OctoberPasswd.php
@@ -1,0 +1,129 @@
+<?php namespace System\Console;
+
+use Str;
+use Backend\Models\User;
+use Illuminate\Console\Command;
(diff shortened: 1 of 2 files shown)


modules/system/console/OctoberPasswd.php
```diff
@@ -0,0 +1,129 @@
...
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Change the password of a Backend user.';
+
+    /**
+     * Was the password automatically generated?
+     *
+     * @var bool
```
LukeTowers:
```suggestion
     * @var bool Was the password automatically generated
```

For class properties I prefer the shorter docblock format, same goes for the above properties as well

modules/system/console/OctoberPasswd.php
```diff
@@ -0,0 +1,129 @@
...
+    /**
+     * Execute the console command.
+     */
+    public function handle()
+    {
+        // Must be in CLI environment
+        if (php_sapi_name() !== 'cli') {
+            $this->error('This command may only be run through the command-line.');
+            exit(1);
+        }
```
LukeTowers: Why is this command restricted to the CLI? Additionally, I would prefer the use of `App::runningInConsole()` if we do continue to restrict it in this way

modules/system/console/OctoberPasswd.php
```diff
@@ -0,0 +1,129 @@
...
+
+        $username = $this->argument('username')
+            ?? $this->ask(
+                'Which user would you like to change the password for'
```
LukeTowers:
```suggestion
                'Which user would you like to change the password for?'
```

modules/system/console/OctoberPasswd.php
```diff
@@ -0,0 +1,129 @@
...
+        $user->forceSave();
+
+        if (!$this->generatedPassword) {
+            $this->info('Password successfully changed.');
+            exit(0);
+        }
+
+        $this->info('Password successfully changed.');
+        $this->output->writeLn('The new password is <info>' . $password . '</info>.');
+        exit(0);
```
LukeTowers: Revise above to 
```php
$this->info('Password successfully changed');
if ($this->generatedPassword) {
    $this->output->writeLn('The new password is <info>' . $password . '</info>.');
}
exit(0);
```

modules/system/console/OctoberPasswd.php
```diff
@@ -0,0 +1,129 @@
...
+    /**
+     * Generate a password and flag it as an automatically-generated password.
+     *
+     * @return string
+     */
+    protected function generatePassword()
+    {
+        $this->generatedPassword = true;
+
+        return Str::random(14);
```
LukeTowers:
```suggestion
        return Str::random(30);
```

modules/system/console/OctoberPasswd.php
daftspunk: When I saw this I assumed it to be an extra security precaution to prevent the `Artisan::call()` from somehow programatically calling this and resetting an account. You never know what people will do and this is a "better safe, than sorry" type check

bennothommo: @daftspunk is correct - I didn't want to allow people to use `Artisan::call()` and sneak it a plugin or something. People who have access to the CLI are very safe to assume that they are trusted, so I felt it was best to restrict it to there.

modules/system/console/OctoberPasswd.php
bennothommo: Sounds good

modules/system/console/OctoberPasswd.php
bennothommo: @LukeTowers is that a bit overkill? You can try some of those randomly-generated 14 character passwords on here and it shows them as secure: https://howsecureismypassword.net/ :)

modules/system/console/OctoberPasswd.php
LukeTowers: While this is nice in theory there is actually zero security benefit in practice to doing this since the there's absolutely nothing stopping a plugin from doing the exact same thing that this command does internally. Hence I don't think there's any reason to include that check. The only reason I can think is if you wanted to protect against a theoretical plugin that allowed any backend users to call artisan commandos through the backend, but such a plugin has a whole host of other reasons why it shouldn't exist.

modules/system/console/OctoberPasswd.php
LukeTowers: Is there any reason why it shouldn't be that long? 😜

modules/system/console/OctoberPasswd.php
bennothommo: From the other point of view, I don't think there's any legitimate reason to make it available outside of CLI. But I'll remove it shortly.

modules/system/console/OctoberPasswd.php
bennothommo: I suppose it just looks comical haha. Meet you in the middle at 22? :P

LukeTowers: Sure @bennothommo 😂

LukeTowers:
LGTM!

PR #4837, opened 2019-12-21
Normalize PluginManager  storage/access

mjauvin:
makes accessing plugins with PluginManager easier (solves case-sensitivity related errors)

e.g. 
```
$manager = PluginManager::instance();

$manager->hasPlugin('Rainlab.Blog');
$manager->exists('RainLab.Blog');
$manager->isDisabled('rainlab.blog');

All access the same plugin without confusion.
```


mjauvin:
replaced by #4838

PR #4838, opened 2019-12-21
PluginManager normalizations

mjauvin:
replaces #4837 

fix isDisabled() to be case insensitive
fix findByIdentifier() to be case insensitive


mjauvin:
This is in line with earlier improvements made in #4337

mjauvin:
bump!

modules/system/classes/PluginManager.php
```diff
@@ -348,7 +348,8 @@ public function findByNamespace($namespace)
     public function findByIdentifier($identifier)
     {
         if (!isset($this->plugins[$identifier])) {
-            $identifier = $this->normalizeIdentifier($identifier);
+            $code = $this->getIdentifier($identifier);
+            $identifier = $this->normalizeIdentifier($code);
```
LukeTowers, line 352: Why is the getIdentifier() call required here?

mjauvin: @LukeTowers I understand the reason you ask the question as it doesn't seem to make sense considering the parameter names... we might want to rewrite part of this class' methods at some point, but I was just trying to emulate what the isDisabled() method does with its identifier.

Am I doing it wrong here, please feel free to let me know.

Issue #4845, opened 2019-12-23
Add an unique identifier to the filter container popup box


ayumi-cloud:
At the moment the filter container popup box doesn't have an unique identifier.

Actually, the `id` tag it currently uses is not unique and should actually be placed in the `class` tag instead!

The id tag looks like this:

```html
id="controlFilterPopover"
```

Every single filter pop-up container box has that id-tag. It's just that people open the filter container one at a time that they don't notice the possible error.

Our company wants to request adding an unique identifier to the filter container pop-up box.

Currently the javascript code line:

https://github.com/octobercms/october/blob/a66310bd5682d08582c1d72ac429e6f5329bbefe/modules/system/assets/ui/js/filter.js#L50

Creates an unique identifier value attribute that matches the unique filter name from the YAML file.

```js
value="{{ scopeName }}"
```

We suggest also adding this `{{ scopeName }}` to the id-tag to create unique identifier.

Currently October hard-codes the css styling for the filter pop-up box giving a terrible result, see here the code:

```css
.control-filter-popover.control-filter-box-popover {
    min-width: 190px;
}
```

We want to be able to adjust each of the filter pop-up container boxes to specific widths, in order to do this we need that container to carry an unique identifier tag.

Therefore we would like to see the following:

Code line: https://github.com/octobercms/october/blob/a66310bd5682d08582c1d72ac429e6f5329bbefe/modules/system/assets/ui/js/filter.js#L49

Old:

```html
<form>
```

New:

```html
<form id="controlFilterPopover-{{ scopeName }}">
```

Then we can add custom styles to each different filter pop-up container box, for example:

```css
form#controlFilterPopover-example {
    width: 300px;
}
```

PR #4846, opened 2019-12-23
Add an unique identifier to all the filter container popup boxes

ayumi-cloud:
See github issue: https://github.com/octobercms/october/issues/4845

Best to show screenshots, so you can understand.

### Before

image

Fixed width no way to adjust each different filter pop-up container box to different widths. Because October's ID tag is the same for every filter box!

### After

image

Now can adjust each filter pop-up container boxes to different widths. Because now we got a true unique identifier on every filter box!

### DIFF
--- a/modules/system/assets/ui/js/filter.dates.js
+++ b/modules/system/assets/ui/js/filter.dates.js
@@ -104,7 +104,7 @@
      */
     FilterWidget.prototype.getPopoverDateTemplate = function () {
         return '                                                                                                        \
-                <form>                                                                                                  \
+                <form id="controlFilterPopoverDate-{{ scopeName }}">                                                    \
                     <input type="hidden" name="scopeName" value="{{ scopeName }}" />                                    \
                     <div id="controlFilterPopoverDate" class="control-filter-popover control-filter-box-popover">       \
                         <div class="filter-search loading-indicator-container size-input-text">                         \
@@ -136,7 +136,7 @@
      */
     FilterWidget.prototype.getPopoverRangeTemplate = function () {
         return '                                                                                                          \
-                <form>                                                                                                    \
+                <form id="controlFilterPopoverRange-{{ scopeName }}">                                                     \
                     <input type="hidden" name="scopeName" value="{{ scopeName }}" />                                      \
                     <div id="controlFilterPopoverDate" class="control-filter-popover control-filter-box-popover --range"> \
                         <div class="filter-search loading-indicator-container size-input-text">                           \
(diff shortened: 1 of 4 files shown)


w20k:
Simple - looks good to me.

bennothommo:
@ayumi-cloud I reckon it might be better seeing if the ID of the filter `<div>` could be made unique as opposed to adding an ID to the encompassing `<form>` tag.

ayumi-cloud:
@bennothommo 1st happy new year 👍 

Actually I did think of that at the time, the reason I choose the `<form>` tag and not the `<div>` tag was because there is already an id tag there, which is linked to the javascript code. Instead of changing a big amount of javascript, I was lazy and added the id tag to the parent container element.

If it's a major issue, I guess I will have to re-write the javascript code and all the files that use that id tag.

bennothommo:
@ayumi-cloud Happy new year to you too.

I personally think it's the better fix - even if it means more work - as duplicate IDs in general are invalid HTML.

ayumi-cloud:
@bennothommo I don't think it would be a duplicate id as they are on different code lines, the only issue I've ever seen is adding an id-tag to a HTML5 element, some old browsers don't recognise it, I think IE11 maybe in that list?

Guess have to do the longer route 🥇

LukeTowers:
@ayumi-cloud personally I don't mind the current approach, what're your concerns @bennothommo?

bennothommo:
@LukeTowers I'm certain that duplicate IDs are invalid HTML, even with HTML5. While it's not an issue when people use different types of filters for a single List widget, it presents when someone uses 2 of the *same* type of filter. The browsers work around it just fine, but it still is "cleaner" to have them use unique identifiers.

I personally don't think it's too big a change to make the IDs unique and change the selectors used in the JS files to point to the new IDs (or change them over to using the class `control-filter-popover`.

ayumi-cloud:
@LukeTowers @bennothommo 

Before I continue with this pr, there is no duplicate IDs going on here! The javascript is loading the HTML into the dom not the source code! Also the javascript html code snippet is being loaded only once per filter. You can't open two filter boxes at the same time as the first one auto closes. **There is no duplicate ID issue here.** I can prove it with some screenshots if needed.

See here as proof:

image

LukeTowers:
@bennothommo @ayumi-cloud is correct, there is no issue of duplicate IDs. The popup window itself is only present in the DOM when visible, and the div has `controlFilterPopoverDate` whereas the form would have `controlFilterPopoverDate-{{ scopeName }}`

LukeTowers:
My main argument against changing it to just the form is all of the JS changes required to make that happen

ayumi-cloud:
@LukeTowers @bennothommo going to revert my commit and not change the javascript code. I worry there is a tiny chance it may cause a breaking to some users. I rather keep it in the `<form>` tag.

bennothommo:
@LukeTowers @ayumi-cloud Alright then - proceed as is.

ayumi-cloud:
I think this pr is ready for testing now, if anyone wants to add a testing label to it.

PR #4851, opened 2019-12-27
Yaml standards

sspooky13:
Hello,
I would like to offer you library which helps you observe some standards. As first, I understand I dropped 7.0 support what is not desirable from your side but you can accept this when you will drop 7.0 support.
As you can see I added 4 commits:

1. just drop PHP 7.0 support
2. just added library to project
3. added indent check: this will help you observe y(a)ml file pretty
4. added empty line at the end check: this will help you observe only one empty line at the end...with empty line show you better conflicts

[Library](https://github.com/sspooky13/yaml-standards#implemented-checkers) of course has more standards but some don't have fixer or it's not relevant for you, so I didn't add it.

Do you have any questions? Something doesn't like you? Feel free to ask 😊

Thank you in advance 🤝


LukeTowers:
@sspooky13 I'm not convinced of the benefits of adding this to our project and it seems slightly suspicious to be perfectly honest that you're trying to add your own relatively unknown & unused dependency into a project used by tens of thousands of sites around the world.

sspooky13:
I created library to help programmers observe some yaml standards so I try to offer it.
I understand reason this library has not to add benefits for your repository but attitude it's "relatively unknown and unused" seems for me weird.
But idea is still same: I only offer you my library, if you don't want it, then ok.

LukeTowers:
@sspooky13 I appreciate the thought, but at this point in time we don't have a need for it.

PR #4854, opened 2019-12-28
Add event to override URL for a record in a List widget

### LINKED ISSUE #4784: Editing soft deleted child from relation manager
- OctoberCMS Build: 458
- PHP Version: 7.2
- Database Engine: MySQL

### Description:

Clicking on a soft deleted entry in a relation manager list gives an error...

image

### Steps To Reproduce:
- Create a plugin
- Create 2 models for this plugin, Booking and Attendee
- Add the soft delete trait to theme
- Add database tables (with deleted_at columns
- Define columns.yaml and fields.yaml etc... for these models
- Set up relation *Booking belongsToMany Attendees* with scope `withTrashed()`
- Implement relation behavior tand define $relationConfig property on Bookings controller 
- Setup bookings config_relation.yaml as follows
```
attendees:
  label: Attendees
  view:
    toolbarButtons: create|add|remove
    list: $/author/plugin/models/attendee/columns.yaml
  manage:
    form: $/author/plugin/models/attendee/fields.yaml
```
- Create an attendee
- Create a booking
- Go to attendees list 
- Delete an attendee (Do not remove it from the booking)
- Go back to booking
- Click on the now soft deleted attendee
- You should now see the error `Call to member function getId() on null` 

### Expected Behavior
I'd expect one of three things to happen.
1. Don't allow the soft deleted entries in the relation manager list to be clickable.
2. If the soft deleted item is clicked, show a popup that it can't be edited because it's been deleted / or can't be found
3. Load the soft deleted item into the popup form / Act as though it not been deleted

bennothommo:
Provides an opportunity to define the URL for a particular record in a list widget.

You may either override the $url parameter with a string to define a custom URL for a record, or you may override it with `false` to specify the record to have no URL and thus become unclickable.

~~Resolves #4784.~~

### DIFF
--- a/modules/backend/widgets/Lists.php
+++ b/modules/backend/widgets/Lists.php
@@ -671,11 +671,48 @@ class Lists extends WidgetBase
             return 'javascript:;';
         }
 
-        if (!isset($this->recordUrl)) {
+        $recordUrl = $this->recordUrl;
+
+        /**
+         * @event backend.list.recordUrl
+         * Provides an opportunity to define the URL for a particular record in a list.
+         *
+         * You may either return a string to define a custom URL for a record, or you may define `false` to force the
+         * record to have no URL, and thus become unclickable.
+         *
+         * Example usage:
+         *
+         *     Event::listen('backend.list.recordUrl', function ($listWidget, $record, string &$url) {
+         *         // Only for the User controller
+         *         if (!$listWidget->getController() instanceof \Backend\Controllers\Users) {
+         *             return;
+         *         }
+         *
+         *         // Only for the User model
+         *         if (!$listWidget->model instanceof \Backend\Models\User) {
+         *             return;
+         *         }
+         *
+         *         // If record is soft-deleted, make it unclickable.
+         *         if ($record->trashed()) {
+         *             $url = false;
+         *             // Return false to halt the event and prevent other plugins from modifying the URL.
+         *             return false;
+         *         }
+         *
+         *         // Define custom URL for banned records - this URL could potentially be overriden by other plugins.
+         *         if ($record->is_banned) {
+         *             $url = '/users/banned';
+         *         }
+         *     });
+         */
+        $this->fireSystemEvent('backend.list.recordUrl', [$record, &$recordUrl]);
+
+        if (!is_string($recordUrl) || empty($recordUrl)) {
             return null;
         }
 
-        $url = RouterHelper::replaceParameters($record, $this->recordUrl);
+        $url = RouterHelper::replaceParameters($record, $recordUrl);
         return Backend::url($url);
     }
 
--- a/modules/backend/widgets/Lists.php
+++ b/modules/backend/widgets/Lists.php
@@ -677,8 +677,8 @@ class Lists extends WidgetBase
          * @event backend.list.recordUrl
          * Provides an opportunity to define the URL for a particular record in a list.
          *
-         * You may either return a string to define a custom URL for a record, or you may define `false` to force the
-         * record to have no URL, and thus become unclickable.
+         * You may either override the $url parameter with a string to define a custom URL for a record, or you may
+         * override it with `false` to specify the record to have no URL and thus become unclickable.
          *
          * Example usage:
          *
--- a/tests/unit/backend/widgets/ListsTest.php
+++ b/tests/unit/backend/widgets/ListsTest.php
@@ -1,5 +1,6 @@
 <?php
 
+use Backend;
 use Backend\Models\User;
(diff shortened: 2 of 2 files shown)


LukeTowers:
Hmmm...

I'm not sure this is the best way of achieving the desired end result. Typically I would recommend that if you need custom URL's on a per record basis you should probably do that handling as redirects from the actual URL that each record has based on the record provided in the URL. That way it solves two problems at once, getting the initial click to the right place and getting a direct navigation to the correct place as well.

Additionally, fireSystemEvent fires a global and a local event, so inline docs for the local event would be required as well.

@daftspunk thoughts?

daftspunk:
> Typically I would recommend that if you need custom URL's on a per record basis you should probably do that handling as redirects from the actual URL that each record has based on the record provided in the URL. 

This is exactly how I handle it in the past; provide a link to a single redirector URL that handles this business logic and forwards the request on. This solution is preferred over the current proposal

bennothommo:
@LukeTowers @daftspunk A redirector does not cover these specific scenarios that this addition intends to solve:

- Making a particular row in a List widget unclickable *without* having to have some form of accessor in the model or manipulating the model externally.
- Allowing an external plugin to redirect certain records of a List widget

daftspunk:
Ok, so long as we make the API nice to use by providing an override interface in ListController behavior

```
    /**
     * Replace a records URL with something else, return null to disable the record URL
     * @param  Model $record The populated model used for the row
     * @param  string $recordUrl The record url to override
     * @param  string $definition List definition (optional)
     * @return string|null
     */
    public function listOverrideRecordUrl($record, $recordUrl, $definition = null)
    {
    }
```


And update the event name to `list.overrideRecordUrl` to bring it [in line with the others](https://github.com/octobercms/october/blob/9fa7cbf70e8f60ea170bd45dac9585fd0a7d1ee5/modules/backend/behaviors/ListController.php#L190-L196)

daftspunk:
Also, how does this resolve #4784?

bennothommo:
@daftspunk it resolves [this comment](https://github.com/octobercms/october/issues/4784#issuecomment-557810660) and the one below it. Essentially, the reporter was showing soft deleted records in a List widget inside a Relation manager, and it was failing when clicked through. I figured that given soft deleted records should not be manipulated, the error itself didn't need resolving at this stage, but that some mechanism to allow records to be unclickable would suffice.

bennothommo:
We could perhaps do a check for soft deleted records in `getRecordUrl()` and handle them that way, but I thought that might be too specific a fix.

daftspunk:
> Set up relation Booking belongsToMany Attendees with scope withTrashed()

It would appear they also need to extend the relation manage query to include `withTrashed` to make those records editable. Displaying those records with nowhere to go might be confusing to the end-user. The OP seems to be making a report about a dirty error

Turns out the original problem is due to a workflow issue, see 321f7eaa1f1a49cae80fda152d111f329f25dbcc for details on what was happening

This proposal would not get us closer to a fix since RelationController targets the `recordOnClick` property, not the `recordUrl` property. I'm not sure I can see a use case for it anymore :(  Putting aside the UX concern of making some rows clickable, and other rows unclickable, could be perceived as an antipattern by the end-user

bennothommo:
@daftspunk thanks for the further info. You're right then - it won't properly cover the original issue.

I do have specific use cases that unclickable rows could be used for - one example is informational rows. The other is for trashed rows like the original issue. While it might be antipattern if the rows looks the same as clickable rows, I feel it would be fine in combination with UI hints that the row is not clickable such as greyed text or using warning icons.

LukeTowers:
Note that there already is a mechanism in place for assigning specific classes / styling to specific rows, so I could be on board with that use case as long as the API is simple and makes sense

daftspunk:
Makes sense. The only issue is that a row is considered clickable if it specifies `recordUrl` _or_ `recordOnClick`. This means if `recordOnClick` is defined, the solution doesn't cover the use case

I wonder if a `disabled` attribute on the list row would be better...

bennothommo:
@daftspunk Would it be necessary for this change to handle `recordOnClick`, given that options effectively makes the JavaScript responsible for handing clicks?

bennothommo:
After discussions with the team, this is better achieved by injecting the `nolink` class into row via the `backend.list.injectRowClass` event. The class is configured to force rows to not be clickable.

```php
$listWidget->bindEvent('list.injectRowClass', function ($record, $value) {
    return 'nolink';
});
```

If you need custom routing depending on the record, this can be achieved by sending all requests to an intermediate controller action, which can then forward the request on to the proper page.

PR #4861, opened 2020-01-03
Add Paragraph Formats to Editor Settings

pboi20:
Hi everyone!

Please consider this PR more like a request for comments, as it's my first attempt to contribute to October CMS.

This adds a new field in the Editor Settings to customize the options available in the Paragraph Formats dropdown of the Rich Editor:

**Editor Settings**

paragraph-format-settings

**Paragraph Formats dropdown**

paragraph-format-dropdown

The main use case I have in mind is to prevent the use of specific Headings (mostly h1, maybe h2), in situations where those are handled by the underlying page layout.

I did my best to follow what was already there in the html_style_* group of settings, as this is fairly similar.

The default values were pulled from the Froala documentation:
https://www.froala.com/wysiwyg-editor/docs/options#paragraphFormat

Is this of any interest for the October CMS core? Would it be better to rework this into an external plugin?

Many thanks and Happy New Year!

P.


w20k:
I think you'd also need to have a plugin for that: 

Plugin: | paragraph_format.min.js
https://www.froala.com/wysiwyg-editor/docs/options#paragraphFormat

pboi20:
That's right. I believe it's included in the current richeditor build: 
https://github.com/octobercms/october/blob/develop/modules/backend/formwidgets/richeditor/assets/js/build.js#L17

bennothommo:
@pboi20 Would you be able to implement the "Paragraph Formats" table somewhere in the "Markup Tags" tab? It seems a bit excessive having another tab just for this table, and its purpose is to provide configuration for available tags so it fits that tab.

You may need to adjust the other fields already in that tab in order to find a suitable place for it.

pboi20:
@bennothommo Makes sense, thanks for the feedback!

Moving the "Paragraph Formats" table to the bottom of the "Markup Tags" tab seems like an acceptable choice to my eye:

bottom

Other ideas I find visually interesting:

right

top-right

Any thoughts?

mjauvin:
I prefer #1 personally

pboi20:
Moved the "Paragraph Formats" table to the bottom of the "Markup Tags" tab (Screenshot 1)

bennothommo:
TBH, I think #3 looks the cleanest.

w20k:
Same here, #3 looks better!

mjauvin:
In a way, yes, but the rows don't align and that's ugly... 😁

pboi20:
@mjauvin Agreed, about the unaligned rows.

Another suggestion:

04 top-left

Side note: I stuck to basic `span: auto|left|right` for the above suggestions. Would [`span: storm`](https://octobercms.com/docs/backend/forms#form-field-options) be an interesting option to preserve field alignment?

mjauvin:
Now, THAT is nice! :smile:

LukeTowers:
@bennothommo @mjauvin @pboi20 I think that this new field would make more sense in Toolbar buttons since it affects the styles that popup in the toolbar button for selecting paragraph formats.

pboi20:
@LukeTowers Good point. Would it be interesting to rename the tab to "Toolbar Settings" (or similar), which would contain the "Toolbar Buttons" and "Paragraph Formats" fields?

LukeTowers:
@pboi20 "Toolbar Options" works for me

pboi20:
Moved the "Paragraph Formats" table to the "Toolbar Options" tab:

05 toolbar_options

bennothommo:
@pboi20 Looks good. Just needs to be tested and then it's good to merge.

w20k:
Hmmm...looks good, but doesn't seem to change `value` from settings for me. Only take into account - default value. Will do more test.

pboi20:
@w20k Thanks for taking some time to test! I just did a quick one on my end, using the Static Pages rich editor :

    1) Drop all tables and run `php artisan october:up`

    2) in Settings > Editor Settings > Toolbar Options

       - Confirm the default paragraph formats:
       - N, H1, H2, H3, H4, PRE

    3) in Pages > My new page

       - Confirm the default items in the Paragraph Format dropdown: 
         Normal, Heading 1, Heading 2, Heading 3, Heading 4, Code

    4) in Editor Settings > Toolbar Options

       - Remove the following formats: H1, H2
       - Save your changes

    5) in Pages > My Page

       - Confirm the modified items in the Paragraph Format dropdown: 
         Normal, Heading 3, Heading 4, Code


Let me know if there's anything else I can help testing.

pboi20:
I did some more testing and... I think I was able to replicate the bug you described 😅️

In my previous message, I described a test starting from a fresh DB, with an empty `system_settings` table. The bug seems to happen when introducing the new editor setting on top of a populated `system_settings` table, one that has a saved `backend_editor_settings` value.

Steps to reproduce:

    1) Drop all tables and run `php artisan october:up`

    2) Run `git checkout cd4f56c6`

        - This is the state of `origin/develop`, before this PR.

    3) in Settings > Editor Settings

        - Click the 'Save' button
        - This adds the `backend_editor_settings` item to the `system_settings` table

    4) Run `git checkout richeditor-paragraph-format`

        - This branch and this PR

    5) in Settings > Editor Settings > Toolbar Options

        - Notice that the Paragraph Formats table is empty

    6) in Pages > My Page

        - The default items appear the Paragraph Format dropdown:
          Normal, Heading 1, Heading 2, Heading 3, Heading 4, Code
        - So far, so good
    
    7) in Settings > Editor Settings

        - Click the 'Save' button again
        - This updates `backend_editor_settings` in the DB with an empty paragraph format selection
        ** This is the bug **

    8) in Pages > My Page

       - The Paragraph Format dropdown is now empty

    9) in Settings > Editor Settings

        - Click the 'Reset to default' button
        - This resets the `backend_editor_settings` in the DB with the proper default paragraph formats

Apologies for my tedious "grocery list" bug report.

So, it seems to me that a data migration could solve this problem. It could load the saved `backend_editor_settings` state, add in the default paragraph formats and save it back to the DB... But, I don't know. 

Would it be preferable to simply add and extra client-side check, to prevent feeding the Paragraph Formats dropdown with an empty array?

modules/backend/models/EditorSetting.php
```diff
@@ -111,6 +111,13 @@ public function initSettingsData()
         $this->html_paragraph_formats = $this->makeFormatsForTable($this->defaultHtmlParagraphFormats);
     }
 
+    public function afterFetch()
+    {
+        if (!isset($this->value['html_paragraph_formats'])) {
+            $this->html_paragraph_formats = $this->makeFormatsForTable($this->defaultHtmlParagraphFormats);
+        }
+    }
+
```
pboi20: This is a quick fix to initialize the Paragraph Formats table with the default values, if the field doesn't exist in the current editor settings state.

Does it make sense? Am I missing another obvious way of handling default values?

Eoler:
It would be nice to have option to remove H1,H2 from paragraph formats.

modules/backend/models/EditorSetting.php
```diff
@@ -99,6 +108,14 @@ public function initSettingsData()
         $this->html_style_paragraph = $this->makeStylesForTable($this->defaultHtmlStyleParagraph);
         $this->html_style_table = $this->makeStylesForTable($this->defaultHtmlStyleTable);
         $this->html_style_table_cell = $this->makeStylesForTable($this->defaultHtmlStyleTableCell);
+        $this->html_paragraph_formats = $this->makeFormatsForTable($this->defaultHtmlParagraphFormats);
+    }
+
+    public function afterFetch()
+    {
+        if (!isset($this->value['html_paragraph_formats'])) {
+            $this->html_paragraph_formats = $this->makeFormatsForTable($this->defaultHtmlParagraphFormats);
```
LukeTowers:
```suggestion
            $this->html_paragraph_formats = $this->makeFormatsForTable($this->defaultHtmlParagraphFormats);
            $this->save();
```
Make sure the defaults are saved in the settings so this doesn't continue to get run on every request.

modules/backend/models/EditorSetting.php
```diff
@@ -115,27 +132,57 @@ protected function makeStylesForTable($arr)
...
+            if (array_has($value, ['format_tag', 'format_label'])) {
+                return [
+                    array_get($value, 'format_tag'),
+                    array_get($value, 'format_label')
+                ];
+            }
+        });
+    }
+
+    private static function getConfiguredArray($key, $default = null, $callback = null)
```
LukeTowers:
```suggestion
    protected static function getConfiguredArray($key, $default = null, $callback = null)
```
October pretty much never uses the private scope since everything should be extensible.

pboi20:
@LukeTowers Thanks for the feedback!

LukeTowers:
@pboi20 no problem. Ping me again in a couple weeks if I haven't gotten to this by then, need to focus on pushing out Build 466 right now.

pboi20:
@LukeTowers Synced this PR with the develop branch. Fixed the field positioning to account for the new toolbar presets :

toolbar-options

bennothommo:
@pboi20 Thanks for your work on this. We're going to push this out to build 467 as build 466 is pretty much completed and we don't want to delay it any further. We'll definitely include it with 467 if it's all working good.

damsfx:
Hi all,  
I think it might also be interesting also to have the possibility to set markup classes for lists (UL/OL) .   
Same things that is actualy present for paragraphes, tables, images ...

bennothommo:
@damsfx Would you be willing to provide a separate PR that adds in this functionality? Note that we'll have to delay this to build 468.

modules/backend/formwidgets/richeditor/assets/js/richeditor.js
```diff
@@ -115,6 +116,17 @@
               'oc-text-uppercase': 'Uppercase'
             }
 
+        froalaOptions.paragraphFormat = this.options.paragraphFormat
+            ? this.options.paragraphFormat
```
bennothommo, line 120: Fix misspelt setting name.

```suggestion
        froalaOptions.paragraphFormat = this.options.paragraphFormats
            ? this.options.paragraphFormats
```

modules/backend/formwidgets/richeditor/assets/js/richeditor.js
```diff
@@ -115,6 +116,17 @@
               'oc-text-uppercase': 'Uppercase'
             }
 
+        froalaOptions.paragraphFormat = this.options.paragraphFormats
+            ? this.options.paragraphFormats
```
bennothommo, line 120: Sorry, it was correct.
```suggestion
        froalaOptions.paragraphFormat = this.options.paragraphFormat
            ? this.options.paragraphFormat
```

bennothommo:
Tested, and confirmed working.

PR #4863, opened 2020-01-03
Apply custom secondary color to Media

gergo85:
Before:
before

After:
after


daftspunk:
LGTM! Thanks @gergo85

PR #4866, opened 2020-01-06
Generate random password for seeded admin account through october:up

### LINKED ISSUE #2360: Seed admin user with random password on initial system seed

bennothommo:
Follows on from work done for https://github.com/octobercms/october/pull/4836.

When `october:up` runs for the first time on a fresh install, it currently seeds an admin account with the credentials `admin / admin` for username and password.

This PR randomises the password assigned for the seeded admin account, to provide extra security for fresh installs.

The PR also includes the beginnings of a mechanism to allow seeders (and potentially migrations) to provide deferred messages which are reported at the *end* of an update process.

Fixes https://github.com/octobercms/october/issues/2360


LukeTowers:
@bennothommo could you post a screenshot of it?

bennothommo:
@LukeTowers sure thing:

image

modules/system/classes/UpdateManager.php
```diff
@@ -429,7 +437,11 @@ public function seedModule($module)
         }
 
         $seeder = App::make($className);
-        $seeder->run();
+        $return = $seeder->run();
```
LukeTowers, line 451: Does this collect only seeder messages or all plugin migration file results of run() too?

bennothommo: @LukeTowers just the seeders at the moment.

LukeTowers: Might as well make it for all of them

bennothommo: @LukeTowers I've taken a look at what would be involved with allowing the migrations to defer messages too. The plugin migrations can be handled easily, but the module migrations are currently using `Illuminate\Database\Migrations\Migrator` to handle migrations. If we want the same functionality for them, we'll have to extend and augment that class. Is that something you would like done as well?

LukeTowers: We're fine with not having it available on module migrations

LukeTowers: Still want plugin migrations handled though

LukeTowers:
@bennothommo what's left on this?

bennothommo:
@LukeTowers still need to allow messages from the plugin migrations.

LukeTowers:
@bennothommo is that something you need to do or do you need my help for it? Do we need it for this PR to work well or can we create a separate issue for it and deal with it later, after 466 and the L6 build?

bennothommo:
@LukeTowers it's definitely not a blocker for 466/L6 - we have `october:passwd` if people wish to change their passwords automatically. I can finish it off after the dust settles with them.

LukeTowers:
@bennothommo sounds good, let's just implement this in a much smaller build after the L6 upgrade.

LukeTowers:
@bennothommo could you resolve the merge conflict on this and then let me know if it's good to go? Would like to get it into 1.1.1 if possible.

bennothommo:
@LukeTowers sure thing!

Issue #3811, opened 2018-09-21
taglist odd behavior on remove first value


mberizzo:
##### Expected behavior
Add and remove values without problems

##### Actual behavior
You can add coma separeted values without problems but if you want to delete the first value, you can't because when input field lose focus, the first value reappear. Only happend with first value and when you are updating

##### Reproduce steps

1. In your fields.yaml create a field with type=taglist
2. Add diferents values. ie: val1, val2, val3, val4
3. Click Save button
4. Click to edit
5. Delete val1 (click "x")
6. Click "Save" (now the val1 reappear) 

##### October build
434

LukeTowers:
Can you attach a GIF of this happening along with the list of fields used to recreate it?

mberizzo:
![bug_taglist](https://user-images.githubusercontent.com/2679457/45917870-7983a100-be54-11e8-99c5-8815190fc6aa.gif)

fields.yaml
```
tabs:
    fields:
        name:
            label: 'mberizzo.beers::lang.field.name'
            span: auto
            type: text
            tab: 'mberizzo.beers::lang.tab.details'
        category:
            label: 'mberizzo.beers::lang.field.category'
            nameFrom: name
            descriptionFrom: description
            span: auto
            type: relation
            tab: 'mberizzo.beers::lang.tab.details'
        description:
            label: 'mberizzo.beers::lang.field.description'
            size: small
            span: full
            type: textarea
            tab: 'mberizzo.beers::lang.tab.details'
        properties:
            label: 'mberizzo.beers::lang.field.properties'
            span: full
            type: taglist
            tab: 'mberizzo.beers::lang.tab.features'
        specs:
            label: 'mberizzo.beers::lang.field.specs'
            prompt: 'backend::lang.form.add'
            span: full
            containerAttributes: {  }
            type: repeater
            maxItems: 4
            tab: 'mberizzo.beers::lang.tab.features'
            form:
                fields:
                    name:
                        label: 'mberizzo.beers::lang.field.name'
                        span: left
                        type: text
                        comment: 'mberizzo.beers::lang.misc.spects.name_comment'
                    value:
                        label: 'mberizzo.beers::lang.field.value'
                        span: right
                        type: text
                        comment: 'mberizzo.beers::lang.misc.spects.hops_comment'
        launch_date:
            label: 'mberizzo.beers::lang.field.launch_date'
            mode: date
            span: auto
            type: datepicker
            tab: 'mberizzo.beers::lang.tab.config'
[...]

PR #4867, opened 2020-01-06
Remove first item in taglist

jacobdekeizer:
Fix issue #3811 
First item in tag list can now be removed.


daftspunk:
@jacobdekeizer thanks for your contribution!

This fix seemed to work better: ec0918e28458afce85905d89163ca7bf4b79044f 

Pls check that it works for you

jacobdekeizer:
@daftspunk That doesn't work for me

daftspunk:
To confirm when you navigate to Playground > Pages > Complex Page using the [Test plugin](https://github.com/octoberrain/test-plugin) and click the SEO tab and try to remove `OctoberCMS`, it doesn't work?

jacobdekeizer:
@daftspunk Just tested it with the test plugin and your fix, but it doesn't work :(
The js is in the source code so it isn't a caching issue.
I remove the first item in the taglist on the update screen. When I click save the first item will come back.

daftspunk:
@jacobdekeizer Try this one, seems to work better 8ce9525b44885a8d6b937420c9e3845d29e1687d

jacobdekeizer:
@daftspunk That works :)

jacobdekeizer:
@daftspunk I do get this javscript error when removing an taglist item:
`Uncaught TypeError: Cannot read property 'element' of undefined at DecoratedClass.SelectOnClose._handleSelectOnClose (storm-min.js?v=462:1509)`

daftspunk:
@jacobdekeizer Third time's a charm d46accfc4f309f5f071cee41ac5bf58092559392

This should prevent the exception. Thanks!

daftspunk:
Proposed fix sets `selectOnClose` to always false (feature only ever used by taglist) and disables key functionality for custom tags. An alternative fix is in place that appears to be working

PR #4871, opened 2020-01-09
Fix registerSchedule typehint in PluginBase

jacobdekeizer:
Change type `string` to type `Schedule` in registerSchedule docblock so we have proper type hinting for this method.


bennothommo:
Nice one @jacobdekeizer.

Issue #4801, opened 2019-12-04
Repeater: Collapse all by default and accordion style


robinbonnes:
# Feature request

My client requested two features for the repeater formwidget and I would like to know if there is a bigger demand for this.

1. Collapse all repeater items by default. This is needed for the pages plugin, to give more overview to a static page with a lot of sections (repeater items) active. Example:

![image](https://user-images.githubusercontent.com/5117166/70124554-5918be80-1675-11ea-9f04-efa9794180a2.png)

2.  Collapse other repeater items when opening one of them. Just like in Bootstrap accordions (https://getbootstrap.com/docs/4.1/components/collapse/#accordion-example), when you open a repeater item it should collapse any other open repeater item.

# Proposal

Add two options to the repeater formwidget:

- `collapsedByDefault`: Collapse all repeater items in the same formwidget by default. Type: boolean, default: false.
- `accordionStyle`: Collapse other repeater items in the same formwidget when opening an item. Type: boolean, default: false.

# References

- https://github.com/rainlab/builder-plugin/issues/165
- https://github.com/octobercms/october/issues/2631

LukeTowers:
That sounds like a good idea to me @robinbonnes. Not sure about the config names but that's a minor detail that could be improved on later.

robinbonnes:
For now I can live with this quick-and-dirty fix to collapse repeaters by default on static pages:

```
$(document).ready(function() {
    $('form[data-object-type="page"] [data-control="fieldrepeater"]').data('oc.repeater').collapseAll();
});
```

SebastiaanKloos:
@robinbonnes Sounds like a good improvement. Do you need any help for creating a PR?

roojai:
> $(document).render(function() {
>     $('form[data-object-type="page"] [data-control="fieldrepeater"]').each(function() { 
>         $(this).fieldRepeater().data('oc.repeater').collapseAll(); 
>     });
> });

Hi there, was wondering where you put this code. Thanks

bennothommo:
Added in https://github.com/octobercms/october/pull/4877.

daslicht:
So this is added inteh current version ? How to use it please?

mightyhaggis:
@daslicht Somehow I missed this update! And a great one it is.
Looking through the Issues and having just tested to find it works  :)

`{repeater name="" tab="" prompt="" style="accordion"} ... {/repeater}`
Options for style: https://github.com/octobercms/october/pull/4877#issue-362492258

[...]

PR #4877, opened 2020-01-14
Add "styles" for repeater widgets

bennothommo:
Implements  #4801. Refs: https://github.com/rainlab/builder-plugin/issues/165, #2631

This adds a new configuration option for Repeater widgets, `style`, which defines the style of repeater on load. This configuration option accepts one of the following values:
- `default`: Shows all the repeater items as expanded on load. This is the default current behavior, and is used if `style` is not defined.
- `collapsed`: Shows all the repeater items as collapsed on load. The user can collapse or expand items as they wish.
- `accordion`: Shows only the first repeater item as expanded on load, all others are collapsed. When another item is exanded, any other expanded item is collapsed, effectively making it so that only one item is expanded at a time.

An additional tweak has been made to the "collapse all" and "expand all" functionality as well. These functions now only apply the collapse and expand on items in the same level as the one which the function was triggered - it no longer changes the state of any child or parent items. This matches most nested item interactions (such as those in Windows/OS X and file explorers like in VSCode or PHPStorm).

### DIFF
--- a/modules/backend/formwidgets/Repeater.php
+++ b/modules/backend/formwidgets/Repeater.php
@@ -43,6 +43,15 @@ class Repeater extends FormWidgetBase
      */
     public $maxItems;
 
+    /**
+     * @var string The style of the repeater. Can be one of three values:
+     *  - "default": Shows all repeater items expanded on load.
+     *  - "collapsed": Shows all repeater items collapsed on load.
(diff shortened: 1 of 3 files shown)


robinbonnes:
@bennothommo I really appreciate that you work on this! At the moment I don't have time to fix those things myself. I will donate asap to keep you guys going!

LukeTowers:
Love it @bennothommo! ❤️

PR #4881, opened 2020-01-15

summercms:
Discussion found here: https://github.com/octobercms/october/issues/4752

Basically done the following:

1. Removed the `SameSite = null` as default. Under new rules this **will cause an issue** (see point 2).

2. `SameSite = blank`. Under new rules will become `SameSite = None`.

3. Cookies are made up of two main groups `First-party` and `Third-party`. 

3.a). `First-party` cookies are set to strict or lax. Strict has an extra layer of security and I have been running October websites between 1-2 years will zero problems with Strict. (Our company has over 300 October websites, so we have tested it in a big'ish trail for over a year).

For new websites the default will be set to strict.

3.b). `Third-party` cookies are set to `SameSite None` and `Secure true`. 

### Testing guide

First-Party uses flag: `chrome://flags/#same-site-by-default-cookies`

Third-Party uses flag: `chrome://flags/#cookies-without-same-site-must-be-secure`

### Thoughts

Maybe need to update October's min php version to 7.3 or above with regards to SameSite support.

---

Note: **_Issue with Windows 10 and SameSite Google Chrome V80_**, users need to install KB4534273

See here for link: https://support.microsoft.com/en-gb/help/4534273/windows-10-update-kb4534273

### DIFF
--- a/config/session.php
+++ b/config/session.php
@@ -163,10 +163,14 @@ return [
     | even if the user follows a link to another website. Lax cookies are
     | only sent with a top-level get request.
     |
-    | Supported: "lax", "strict"
+    | First-party cookies that match the domain, be set to 'Strict' or 'Lax'.
+    |
+    | Third-party cookies from other domains, be set to 'None' and need
+    | the Secure set to true.
+    |
+    | Supported: "Lax", "Strict" and "None"
     |
     */
-
-    'same_site' => null,
+    'same_site' => 'Strict',
 
 ];


Eoler:
This one is getting back into action on July 14: https://blog.chromium.org/2020/05/resuming-samesite-cookie-changes-in-july.html

ayumi-cloud:
Quick update: The browsers have been rolling this out for a few weeks now, see more info here: https://www.chromium.org/updates/same-site

summercms:
Quick update: The browsers have been rolling this out for a few weeks now, see more info here: https://www.chromium.org/updates/same-site

> Aug 11, 2020
> The target rollout population has been increased to 100% of users on Chrome Stable versions 80 and above, and the actual proportion of users with the new behavior enabled is now ramping up to 100% gradually. Users will receive the new behavior when they restart Chrome.

@LukeTowers @daftspunk 

Request to add this pr as the rollout is now confirmed and in full swing.

config/session.php
```diff
@@ -163,10 +163,14 @@
     | even if the user follows a link to another website. Lax cookies are
     | only sent with a top-level get request.
     |
-    | Supported: "lax", "strict"
+    | First-party cookies that match the domain, be set to 'Strict' or 'Lax'.
+    |
+    | Third-party cookies from other domains, be set to 'None' and need
+    | the Secure set to true.
+    |
+    | Supported: "Lax", "Strict" and "None"
```
LukeTowers, line 171: Can you explain what each mode does?

RodrigoGalter: @LukeTowers see this article of [MDN](https://developer.mozilla.org/pt-BR/docs/Web/HTTP/Headers/Set-Cookie/SameSite)

LukeTowers: Can you explain it to me in the context of an admin trying to figure out what value they need to set for their specific site?

summercms: @LukeTowers I see what you are saying (let me try and write a clearer description aimed at a clueless person looking at it for the first time). Please review this and I will try and create a commit for an updated description:

> This option determines how your cookies behave when cross-site requests take place and can be used to mitigate CSRF attacks.
>
> Cookies that match the domain of the current site, i.e. what's displayed in the browser's address bar, are referred to as first-party cookies. Similarly, cookies from domains other than the current site are referred to as third-party cookies.
>
> Cookies without a `SameSite` attribute will be treated as `SameSite=Lax`, meaning the default behaviour will be to restrict cookies to first party contexts only.
>
> Cookies for cross-site usage must specify `SameSite=None` and  `Secure=true` to enable inclusion in third party context.
>
> `Lax` - Cookies are allowed to be sent with top-level navigations and will be sent along with GET request initiated by third party website. This is the default value in modern browsers.
>
> `Strict` - Cookies will only be sent in a first-party context and not be sent along with requests initiated by third party websites.
>
> Supported: "Lax", "Strict" and "None"

I think this is a bit clearer and more custom to October CMS config settings.

summercms: The branch name is not there? So, if you need the description changing I would have to create a fresh new pr.

LukeTowers: Okay, so if I understand it correctly the following is what happens?

If you are on example.com and you make a request to example.com, both Strict & Lax cookies belonging to example.com will be attached to that request
If you are on evil.net and you make a request to example.com then only Lax cookies belonging to example.com will be sent? Strict cookies made in this case will not be sent?

Also, how strict are the browsers when looking at things like subdomains when factoring in if a strict cookie will be removed from the request or not?

summercms: > If you are on example.com and you make a request to example.com, both Strict & Lax cookies belonging to example.com will be attached to that request
> If you are on evil.net and you make a request to example.com then only Lax cookies belonging to example.com will be sent? Strict cookies made in this case will not be sent?

Yes correct.

> Also, how strict are the browsers when looking at things like subdomains when factoring in if a strict cookie will be removed from the request or not?

Browsers will treat the main domain and subdomains exactly the same, **they use the same setup as the cross-origin api rules**. The main domain and the subdomain (on the same website) will be under first-party and not third-party cookies.

For cross-origin api rules its best you read about COOP and COEP (the rules work for subdomains and other websites), see this short video: https://www.youtube.com/watch?v=XLNJYhjA-0c

p.s. my above description I removed the part:

> By default, we do not enable this as other CSRF protection services are in place.

Because browsers have said using the SameSite you do not need to use CSRF anymore (but it can't hurt to still use both to have backwards compatibility).

Let me know if I need to re-submit a new pr or not?

ayumi-cloud: To fully answer your question, its best you read about COOP and COEP, see this short video: https://www.youtube.com/watch?v=XLNJYhjA-0c

LukeTowers: @ayumi-cloud yes, please resubmit a new PR.

summercms:
Please see new pr here: https://github.com/octobercms/october/pull/5293

PR #4883, opened 2020-01-15
Add depedency injection to components, controllers and behaviours

jacobdekeizer:
# Problem
In the current components, controllers and behaviours we don't have dependency injection. This forces us to use facades or resolve the dependencies manually with the `app` or `resolve` helper methods.
# Solution
Resolve the dependencies like laravel does with their controller actions for:
- Ajax handlers in components, behaviours and backend controllers
- Actions in backend controllers
- Methods defined with `addDynamicMethod` for components and backend controllers

This allows us to have an ajax handler like this:
```php
public function onTest(Illuminate\Http\Request $request){
     dd($request->all());
}
```
# Related PR
For this functionality to work I made this PR octobercms/library#453

# Discussion
I would also like to see dependency injection for the `init` method for a component, but this requires us to remove the init method from the ComponentBase class.

### DIFF
--- a/modules/backend/classes/Controller.php
+++ b/modules/backend/classes/Controller.php
@@ -22,6 +22,7 @@ use October\Rain\Exception\ApplicationException;
 use Illuminate\Database\Eloquent\MassAssignmentException;
 use Illuminate\Http\RedirectResponse;
 use Illuminate\Routing\Controller as ControllerBase;
+use System\Classes\DependenciesResolver;
 
 /**
  * The Backend base controller class, used by Backend controllers.
@@ -116,6 +117,11 @@ class Controller extends ControllerBase
     protected $guarded = [];
 
     /**
+     * @var DependenciesResolver Used to resolve dependencies for ajaxHandlers and controller actions
+     */
+    private $dependenciesResolver;
+
+    /**
      * Constructor.
      */
     public function __construct()
@@ -157,6 +163,8 @@ class Controller extends ControllerBase
             $manager = new MediaManager($this, 'ocmediamanager');
             $manager->bindToController();
         }
+
+        $this->dependenciesResolver = resolve(DependenciesResolver::class);
 
         $this->extendableConstruct();
     }
@@ -402,7 +410,10 @@ class Controller extends ControllerBase
         }
 
         // Execute the action
-        $result = call_user_func_array([$this, $actionName], $parameters);
+        $result = call_user_func_array(
+            [$this, $actionName],
+            $this->dependenciesResolver->resolveForExtendableObject($this, $parameters, $actionName)
+        );
 
         // Expecting \Response and \RedirectResponse
         if ($result instanceof \Symfony\Component\HttpFoundation\Response) {
@@ -617,7 +628,10 @@ class Controller extends ControllerBase
             $pageHandler = $this->action . '_' . $handler;
 
             if ($this->methodExists($pageHandler)) {
-                $result = call_user_func_array([$this, $pageHandler], $this->params);
+                $result = call_user_func_array(
+                    [$this, $pageHandler],
+                    $this->dependenciesResolver->resolveForExtendableObject($this, $this->params, $pageHandler)
+                );
                 return $result ?: true;
             }
 
@@ -625,7 +639,10 @@ class Controller extends ControllerBase
              * Process page global handler (onSomething)
              */
             if ($this->methodExists($handler)) {
-                $result = call_user_func_array([$this, $handler], $this->params);
+                $result = call_user_func_array(
+                    [$this, $handler],
+                    $this->dependenciesResolver->resolveForExtendableObject($this, $this->params, $handler)
+                );
                 return $result ?: true;
             }
 
@@ -662,7 +679,10 @@ class Controller extends ControllerBase
     {
         $this->addViewPath($widget->getViewPaths());
 
-        $result = call_user_func_array([$widget, $handler], $this->params);
+        $result = call_user_func_array(
+            [$widget, $handler],
+            $this->dependenciesResolver->resolveForExtendableObject($widget, $this->params, $handler)
+        );
 
         $this->vars = $widget->vars + $this->vars;
 
--- a/modules/cms/classes/ComponentBase.php
+++ b/modules/cms/classes/ComponentBase.php
@@ -5,6 +5,7 @@ use Lang;
 use Config;
 use October\Rain\Extension\Extendable;
 use BadMethodCallException;
+use System\Classes\DependenciesResolver;
 
 /**
  * Component base class
@@ -77,6 +78,11 @@ abstract class ComponentBase extends Extendable
     protected $externalPropertyNames = [];
 
     /**
+     * @var DependenciesResolver Used to resolve dependencies for ajaxHandlers
+     */
+    private $dependenciesResolver;
+
+    /**
      * Component constructor. Takes in the page or layout code section object
      * and properties set by the page or layout.
      * @param null|CodeBase $cmsObject
@@ -94,6 +100,8 @@ abstract class ComponentBase extends Extendable
         $className = Str::normalizeClassName(get_called_class());
         $this->dirName = strtolower(str_replace('\\', '/', $className));
         $this->assetPath = Config::get('cms.pluginsPath', '/plugins').dirname(dirname($this->dirName));
+
+        $this->dependenciesResolver = resolve(DependenciesResolver::class);
 
         parent::__construct();
     }
@@ -184,7 +192,10 @@ abstract class ComponentBase extends Extendable
             return $event;
         }
 
-        $result = $this->$handler();
+        $result = call_user_func_array(
+            [$this, $handler],
+            $this->dependenciesResolver->resolveForExtendableObject($this, [], $handler)
+        );
 
         /**
          * @event cms.component.runAjaxHandler
--- a/modules/system/classes/DependenciesResolver.php
+++ b/modules/system/classes/DependenciesResolver.php
@@ -30,7 +30,7 @@ class DependenciesResolver
      * @param string $method
      * @return array
      */
-    public function resolveForExtendableObject(object $object, array $parameters, string $method): array
+    public function resolveForExtendableObject($object, array $parameters, string $method): array
     {
         if (method_exists($object, $method)) {
             try {
--- a/modules/system/classes/DependenciesResolver.php
+++ b/modules/system/classes/DependenciesResolver.php
@@ -1,0 +1,51 @@
+<?php
+
+namespace System\Classes;
+
+use Illuminate\Contracts\Container\Container;
+use Illuminate\Routing\RouteDependencyResolverTrait;
+
+class DependenciesResolver
+{
+    use RouteDependencyResolverTrait;
+
+    /**
+     * @var Container
+     */
+    private Container $container;
+
+    /**
+     * @param Container $container
+     */
+    public function __construct(Container $container)
+    {
+        $this->container = $container;
+    }
+
+    /**
+     * Resolve the dependencies for an object which uses the extendable trait
+     *
+     * @param object $object
+     * @param array $parameters
+     * @param string $method
+     * @return array
+     */
+    public function resolveForExtendableObject(object $object, array $parameters, string $method): array
+    {
+        if (method_exists($object, $method)) {
+            try {
+                return $this->resolveMethodDependencies($parameters, new \ReflectionMethod($object, $method));
+            } catch (\ReflectionException $exception) {
+                // do nothing
+            }
+        }
+
+        if (method_exists($object, 'getExtendableCallReflection')) {
+            $reflectionFunction = $object->getExtendableCallReflection($method);
+
+            return $this->resolveMethodDependencies($parameters, $reflectionFunction);
+        }
+
+        return $parameters;
+    }
+}
--- a/modules/system/classes/DependenciesResolver.php
+++ b/modules/system/classes/DependenciesResolver.php
@@ -12,7 +12,7 @@ class DependenciesResolver
     /**
      * @var Container
      */
-    private Container $container;
+    private $container;
 
     /**
      * @param Container $container
--- a/tests/fixtures/plugins/october/tester/components/Archive.php
+++ b/tests/fixtures/plugins/october/tester/components/Archive.php
@@ -1,6 +1,7 @@
 <?php namespace October\Tester\Components;
 
 use Cms\Classes\ComponentBase;
+use Illuminate\Http\Request;
 
 class Archive extends ComponentBase
 {
@@ -37,5 +38,10 @@ class Archive extends ComponentBase
     public function onTestAjax()
     {
         $this->page['var'] = 'page';
+    }
+
+    public function onTestDependencyInjection(Request $request)
+    {
+        return $request->method();
     }
 }
--- a/tests/unit/cms/classes/ControllerTest.php
+++ b/tests/unit/cms/classes/ControllerTest.php
@@ -370,6 +370,20 @@ ESC;
         $this->assertEquals('page', $content['ajax-result']);
     }
 
+    public function testComponentAjaxDependencyInjection()
+    {
+        Request::swap($this->configAjaxRequestMock('testArchive::onTestDependencyInjection'));
+
+        $theme = Theme::load('test');
+        $controller = new Controller($theme);
+        $response = $controller->run('/with-component');
+        $content = $response->getOriginalContent();
+
+        $this->assertInstanceOf('Symfony\Component\HttpFoundation\Response', $response);
+        $this->assertArrayHasKey('result', $content);
+        $this->assertEquals('POST', $content['result']);
+    }
+
     /**
      * @expectedException        October\Rain\Exception\SystemException
      * @expectedExceptionMessage is not registered for the component
--- a/tests/unit/system/classes/DependencyResolverTest.php
+++ b/tests/unit/system/classes/DependencyResolverTest.php
@@ -49,7 +49,7 @@ class DependencyResolverTest extends TestCase
         );
 
         $this->assertInstanceOf(Request::class, $result[0]);
-        $this->assertInstanceOf('test', $result[1]);
+        $this->assertEquals('test', $result[1]);
     }
 
     ///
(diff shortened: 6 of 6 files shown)


LukeTowers:
Personally I've never been a massive fan of dependency injection as it pertains to type hinting variables you want in method signatures, it's always been a bit too far on the "magic" side of things for me to care to use. However I would like to get @daftspunk's and @bennothommo's thoughts on this.

jacobdekeizer:
@LukeTowers Dependency injection does not have a lot of magic at all. I recently watched [this video](https://www.youtube.com/watch?v=y7EbrV4ChJs) which explains it very well and shows how "simple" a container really is.

nathan-van-der-werf:
Want to leave this here: https://symfony.com/doc/current/service_container/calls.html

Also facades is way more magic than dependency injection...

LukeTowers:
Thanks for the links, I'll take a further look into it. As a side note, we already pass arguments to AJAX handlers in some cases (in the backend all controller parameters are passed as arguments to the handler called)

jacobdekeizer:
@LukeTowers That functionality still works, this feature is backwards compatible.

bennothommo:
TBH, the DI through parameters is something that I missed a whole lot less than I thought I would when moving between Laravel and October.

I'm not opposed to it being brought in, especially to assist people from Laravel coming over to October, but I would like to see some unit tests for it, and it has to be thoroughly tested across the Backend and CMS sections.

jacobdekeizer:
@LukeTowers @bennothommo I added tests for dependency injection. Two tests in the DependencyResolverTest are commented out, because they only work when octobercms/library#453 is merged.

martendej:
@bennothommo @LukeTowers What is the current status of this PR?

LukeTowers:
@martendej waiting for @daftspunk's input.

LukeTowers:
@jacobdekeizer @nathan-van-der-werf I spoke to @daftspunk and we're still leaning on side of not really liking DI from a personal preference side of things. It would help if you could expand on why you would like to see support for DI added to the core with as much detail as possible so that we can be better informed as we consider the proposed changes.

nathan-van-der-werf:
> @jacobdekeizer @nathan-van-der-werf I spoke to @daftspunk and we're still leaning on side of not really liking DI from a personal preference side of things. It would help if you could expand on why you would like to see support for DI added to the core with as much detail as possible so that we can be better informed as we consider the proposed changes.

https://en.wikipedia.org/wiki/SOLID
https://en.wikipedia.org/wiki/Dependency_inversion_principle
https://en.wikipedia.org/wiki/Inversion_of_control

https://phptherightway.com/#dependency_injection

jacobdekeizer:
@LukeTowers We don't use facades at our company. We create classes which are bind to interfaces. To use these interfaces we need to inject them in to the component. The only way to do this is in an octobercms component is to use the ugly (personal preference) `resolve` helper method.
For example if we want to fire an event we resolve the `Illuminate\Contracts\Events\Dispatcher` interface. If we want to overwrite the event dispatcher for whatever reason, we don't have to change it everywhere in our components, but just override the implementation in the IOC container.

Octobercms is based on the laravel framework, but programming principles from laravel don't work in octobercms.

LukeTowers:
@nathan-van-der-werf 

> Good coding principles vs. personal preference 👀

October practices those good coding principles, DI via type hinting method parameters is merely a form of syntactic sugar, not required to adhere to those principles.

@jacobdekeizer can you show me an example of what your code looks like using DI via type hinting in method parameters vs the resolve method?

jacobdekeizer:
@LukeTowers 

With dependency injection:
```php
public function onTest(Flashbag $flash, Request $request, MyAwesomeService $service): void {
    $success = $service->doSomething($request->get('something'));

    $flash->success($success ? 'yes' : 'no');
}
```

How we have to do it currently in october:
```php
public function onTest(): void {
    /** @var FlashBag $flash This is needed for typehinting */
    $flash = resolve(FlashBag::class);

    /** @var Request $request */
    $request = resolve(Request::class);

    /** @var MyAwesomeService $service */
    $service = resolve(MyAwesomeService::class);

    $success = $service->doSomething($request->get('something'));

    $flash->success($success ? 'yes' : 'no');
}
```

As you can see my function with DI is **3 times smaller, easier to read and understand.** 


Ps: flashbag can only be resolved when this is added to the container. I will make a PR for this in the future.
```php
$this->app->alias('flash', FlashBag::class);
```

LukeTowers:
Thanks for the info @jacobdekeizer. 

Side note, I don't think we're interested in integrating Laravel's Form Requests into October, we already have a number of ways of validating forms that are a lot nicer in my opinion.

martendej:
Obviously October CMS has its advantages (or we wouldn't be using it) but imho Form Requests (with the use of DI) are a neat way of mapping your requests onto objects with working type-hinting, which I think is fundamental to understanding the flow of code. Form Requests can also hold more specific code or mutators relevant to a specific request. Moreover, validation is now done at Model level, whereas a Model should be agnostic on how a request is formed because that is simply not a concern of a Model. Anyway, those are my two cents on your comment.

LukeTowers:
@martendej I'd disagree with validation being done anywhere but the lowest level (i.e. the model). If you don't have validation on the model itself then there is always the possibility for invalid data to make it in through some other means of directly interacting with the model.

Could you provide an example of a Form Request doing something that October can't do / is clunky to do in October?

jacobdekeizer:
@LukeTowers Not every request is directly saved in to a model.
So sometimes you need to validate the current request. In octobercms we have to use the validator facade or the validation factory in our component like this:

```php
public function onSubmit(): void {
    /** @var Illuminate\Contracts\Validation\Factory $validationFactory*/
    $validationFactory= resolve(Illuminate\Contracts\Validation\Factory::class);

    /** @var Request $request */
    $request = resolve (Request::class);

    $validator= $validationFactory->make($request->all(), [
         'order_id' => 'required|integer',
    ], [], [
         'order_id' => 'Order id',
    ]);

   throw new ValidationException($validator);

   $orderId = $request->get('orderId');
}
```

While a FormRequest with DI can look like this
```php
class OrderRequest extends FormRequest
{
    public function authorize(): bool
    {
        return true
    }

    public function rules(): array
    {
        return [
            'order_id' => 'required|integer',
        ];
    }

    public function messages(): array
    {
        return [
            'order_id.required' => 'Order id'
        ];
    }

    public function getOrderId(): int
    {
        return (int) $this->request->get('order_id');
    }
}
```
Now our component can look like this:
```php
public function onSubmit(OrderRequest $request): void {
    $orderId = $request->getOrderId(); // strict typing
}
```

LukeTowers:
@jacobdekeizer all that extra stuff you use though is your choice to use. It can also just be as simple as:
```php
public function onSubmit()
{ 
    $data = post();
    $validator = \Validator::make($data, ['order_id' => 'required|integer']);
    if ($validator->fails()) {
        throw new \ValidationException($validator);
    }

    $orderId = $data['order_id'];
}
```

jacobdekeizer:
@LukeTowers I still find my FormRequest solution much cleaner and better, look at [this](https://medium.com/@kamerk22/the-smart-way-to-handle-request-validation-in-laravel-5e8886279271) medium article about form requests.

A quote from the article:
> The controller should do only one thing handle request from the route and return an appropriate response. Writing validation logic in the controller will break The Single Responsibility Principle.

nathan-van-der-werf:
> @jacobdekeizer all that extra stuff you use though is your choice to use. It can also just be as simple as:
> 
> ```
> public function onSubmit()
> { 
>     $data = post();
>     $validator = \Validator::make($data, ['order_id' => 'required|integer']);
>     if ($validator->fails()) {
>         throw new \ValidationException($validator);
>     }
> 
>     $orderId = $data['order_id'];
> }
> ```

Did you read the contents of the links I sent? Separation of concerns? Also this ain't reusable, whereas the FormRequest is reusable.

LukeTowers:
@nathan-van-der-werf a difference of opinions and approaches. My suggestion optimizes for readability at the expense of some reusability, yours optimizes for reusability at the expense of some readability. Neither is more correct than the other, it just comes down to personal preferences.

At any rate, thanks for all the information, I'll get @daftspunk to review this when he's got some time and he can make the final decision.

Klaasie:
The good vs bad code discussion aside. I don't believe OctoberCMS as a framework should enforce one approach over the other. This PR is backwards compatible meaning people who like facades can keep on using them, whereas people preferring DI can now use that approach.

bennothommo:
@nathan-van-der-werf 

> Facades are an anti-pattern, it breaks all SOLID principles.

And yet they are very intrinsic to the Laravel architecture, so for the sake of your (and @jacobdekeizer's, as I presume you are colleagues) arguments, you cannot decry them whilst simultaneously saying:

> Octobercms is based on the laravel framework, but programming principles from laravel don't work in octobercms.

As I stated before, I personally don't mind this being implemented - I know for a fact that colleagues of mine who came to October from a Laravel background were thrown off by this not being implemented - but it's gotta be carefully considered with established October - not Laravel - paradigms.

daftspunk:
I hate to be the bearer of bad news here (cue the confused 😕 and thumbs down 👎  emojis...)

This is not the first time [a contentious issue like this](https://octobercms.com/forum/post/why-twig) has been raised and we are fully aware of the proposed principles, so it is not a matter of education (please move past this). It is also not a matter of personal preference, but a matter of our own principles and methodology; ideas that this platform was originally founded upon. October works very hard to avoid too much magic, spaghetti code and black boxes, even if it means the developer writing more verbose code. This design decision should be evident by the fact that no DI code exists in the October code base, this is not an accident and not due to a lack of education.

> https://phptherightway.com/#dependency_injection
> 
> Good coding principles vs. personal preference 👀

This is just one set of coding principles, with an assertion that they are good ("PHP the right way"). We assert that these principles are not so good and we actively seek to avoid them. While we respect these alternative principles, there is no rule that says we must follow them.

> Did you read the contents of the links I sent? Separation of concerns? 

Nobody is saying you cannot separate concerns and has nothing to do with dependency injection at the method level.

> The controller should do only one thing handle request from the route and return an appropriate response. Writing validation logic in the controller will break The Single Responsibility Principle.

We agree with this position and has nothing to do with dependency injection at the method level.

> In the current components, controllers and behaviours we don't have dependency injection. This forces us to use facades or resolve the dependencies manually with the app or resolve helper methods.

It is worthwhile noting that the component and controller constructors already have support for dependency injection. This is where we draw the line for support as it is not a zero-sum game at this point.

```
// Component constructor
public function __construct(Flashbag $flash, Request $request, MyAwesomeService $service, CodeBase $cmsObject = null, $properties = [])
{
    $this->flash = $flash;
    $this->request = $request;
    $this->service = $service;

    parent::__construct($cmsObject, $properties);
}
```

Allowing DI in methods brings a different set of problems...

### This proposal breaks POD principles

October follows the principles akin to [package oriented design](https://github.com/ardanlabs/gotraining/tree/master/topics/go/design/packaging), with an emphasis on extensibility. The programming language Go is based on these principles, likewise, October is based on these same principles.

> **You can think of [plugins] as applying the idea of microservices on a source tree.**

This means when our plugin defines its public interface (API) in PHP code, it does so by mimicking a HTTP endpoint along with an RPC-style definition. Take the following `onPerformAction` handler that passes three arguments:

```
function onPerformAction($arg1, $arg2, $arg3)
```

This loosely translates to a HTTP endpoint of the following:

```
/onPerformAction/arg1/arg2/arg3
```

Now consider what would happen if the method signature included injected objects:

```
function onPerformAction(Flashbag $flash, Request $request, MyAwesomeService $service, $arg1, $arg2, $arg3)
```

Expressed as an endpoint

```
/onPerformAction/[Object]/[Object]/[Object]/arg1/arg2/arg3
```

We see that the principle is functionally broken. October's API design (method signature) is critical to its success as not only an enjoyable tool but an accessible one. This is where developer performance matters most, not by counting lines of code, but by the time it takes for a developer to comprehend new code.

### This proposal breaks extensibility

This change breaks extensibility and October's API in general. Take the following method found in a component called `myComponent`

```
public function onRegister()
{
    // Register the user
}
```

If we wish to extend this code to include a registration email, we could extend it as follows

```
public function onRegisterWithEmail()
{
    // Call the components logic
    $this->myComponent->onRegister();

    // Send the registration email
}
```

Now consider what would happen if we changed the method signature to this

```
public function onRegister(Flashbag $flash, Request $request, MyAwesomeService $service)
{
    // Register the user
}
```

It is no longer easy to extend or override this action call. The external interface has been polluted by a local resolver definition. The procedure is no longer fully encapsulated, since it relies on an external resolver as a dependency (dependency injector dependency).

### Conclusion

This is where October steers away from OOP and leans more to the realm of functional programming. A critical distinction in the architecture of October that must be understood to see the impact of this change.

We want to acknowledge the effort and time that has gone into this proposed solution. However, it is not a question of "can we", but more a question of "should we". Instead, we should find a solution that does not hurt the integrity of the design pattern.

Given these design constraints, we should steer the conversation towards finding a different solution where everyone can be happy. We welcome any attempt that still provides a nice interface without departing from the very thing that makes October great.

### Ideas

For commonly used objects, it is worthwhile registering a facade for these instances, as we do in the core files, which allows for IoC. For less common, local objects, an instance resolver can be used instead.

```
class OrderRequest extends FormRequest
{
    public static function instance()
    {
        return App::make(static::class);
    }
}
```

This has a similar effect to a facade, although, a constructor method is needed since we aren't able to proxy through a shell class

```
public function onSubmit(): void {
    $orderId = OrderRequest::instance()->getOrderId();
}
```

You could even write a simple helper function returning a tuple/array

```
public function onTest(): void {
    [$flash, $request, $service] = $this->resolveDeps();
}

protected function resolveDeps() {
    /** @var FlashBag $flash This is needed for typehinting */
    $flash = resolve(FlashBag::class);

    /** @var Request $request */
    $request = resolve(Request::class);

    /** @var MyAwesomeService $service */
    $service = resolve(MyAwesomeService::class);

    return [$flash, $request, $service];
}
```

Or even

```
// Need some way to resolve these
protected function resolveDeps(Flashbag $flash, Request $request, MyAwesomeService $service) {
    return [$flash, $request, $service];
}
```

These are by no means perfect solutions but ones we should be discuss

martendej:
```
// Component constructor
public function __construct(Flashbag $flash, Request $request, MyAwesomeService $service, CodeBase $cmsObject = null, $properties = [])
{
    $this->flash = $flash;
    $this->request = $request;
    $this->service = $service;

    parent::__construct($cmsObject, $properties);
}
```

Just tested this code example, and it does NOT work because the cmsObject and properties arguments are passed first (I tried to 'inject' the Request instance).

`Type error: Argument 1 passed to Component::__construct() must be an instance of Illuminate\Http\Request, instance of Cms5e183b317da04534288690_6bbbf6a2a81b5207edd483756ef98dceClass given`

It does work when putting the Request argument in third position, but that would produce bad code because the first two arguments are optional.

daftspunk:
@martendej Apologies, I have this around the wrong way. I can confirm the functionality was introduced in 2016 when this was last discussed: 

https://github.com/octobercms/october/commit/c94706ed057504cb6d3fa42b2bcb6c4249f33284

I have edited my original description

jacobdekeizer:
@daftspunk I disagree

This code below doesn't work. DI in components does only work if you have the DI parameters as last, but then you get a `Optional parameter is provided before required` warning.

> // Component constructor
> public function __construct(Flashbag $flash, Request $request, MyAwesomeService $service, CodeBase $cmsObject = null, $properties = [])
> {
>     $this->flash = $flash;
>     $this->request = $request;
>     $this->service = $service;
> 
>     parent::__construct($cmsObject, $properties);
> }

October doesn't use this route behaviour:
> Expressed as an endpoint
> 
> ```
> /onPerformAction/[Object]/[Object]/[Object]/arg1/arg2/arg3
> ```


> This proposal breaks extensibility

You can still extend the extend a component like this:

```php
public function onRegisterWithEmail(Flashbag $flash, Request $request, MyAwesomeService $service)
{
    $this->myComponent->onRegister($flash, $request, $service);
}
```

daftspunk:
@jacobdekeizer Apologies, the code provided was incorrect. See my previous message and [this commit from when the functionality was introduced](https://github.com/octobercms/october/commit/c94706ed057504cb6d3fa42b2bcb6c4249f33284).

> October doesn't use this route behaviour:

This is a design principle, not a routing behaviour.

> You can still extend the extend a component like this:

I'm sure you can, but this misses the point. It breaks the aforementioned design principle of not depending on complex objects. For reasons of portability, simplicity and extensibility.

jacobdekeizer:
@daftspunk I know, but then you get a `Optional parameter is provided before required warning.`.
If I use a facade or resolve my dependency is still depend on the so called "complex object", but then I hide it in my code.

daftspunk:
> Optional parameter is provided before required warning

```
public function __construct(CodeBase $cmsObject = null, $properties = [], Flashbag $flash = null, Request $request = null, MyAwesomeService $service = null)
```

Note the dependencies default to `null` to overcome this.

jacobdekeizer:
@daftspunk Then I give the impression that I don't need that class. Why use workarounds when it is wrong in the first place?

daftspunk:
We could adjust the code to this if it makes you happier?

```
public function __construct(CodeBase $cmsObject, $properties, Flashbag $flash, Request $request, MyAwesomeService $service)
```

I'm not sure that these first two args even use their defaults

jacobdekeizer:
@daftspunk It's an improvement, but I can't say I'm happy.

daftspunk:
It appears we are at a stalemate with this. October doesn't use DI natively (although it supports it where possible), this was a decision made from the start and it is unlikely to change. It is currently impossible to support DI at the method level, where routes are directly wired in via page actions or ajax handlers. As mentioned, we would be happy to discuss solutions that don't alter this pattern

daftspunk:
> The good vs bad code discussion aside. I don't believe OctoberCMS as a framework should enforce one approach over the other. This PR is backwards compatible meaning people who like facades can keep on using them, whereas people preferring DI can now use that approach.

Normally I would agree with this, except what we are discussing here is the promotion of complex dependency mapping between plugins, which breaks extensibility potential, where themes and plugins become fragile and can break easily due to modifications in services and interfaces of their extended parents. October is an inherently stable platform and we work hard to ensure it remains as stable as possible. These are tough decisions to make, but we must protect the values behind October and that is primarily geared toward business-oriented, stable code, that avoids breaking changes. Laravel has a different very different outlook to this, so not all of their ideas are appropriate to transfer across

martendej:
`public function __construct(CodeBase $cmsObject, $properties, Flashbag $flash, Request $request, MyAwesomeService $service)`

@daftspunk If the defaults of these arguments are indeed never used, then I'd like for this code change to happen. Having DI parameters after parameters with default values would mean the first arguments would not be optional after all when constructing the class in a regular way (without DI, that is). Maybe we can all still get something out of this pull request after all? :)

PR #4885, opened 2020-01-16
Protect server logs files when server errors are created

ayumi-cloud:
This is part of the work we are doing with our October CMS Version II Proposal - Security Module, found here: https://github.com/ayumi-cloud/oc-security-module

We feel this update should be added now to October CMS version 1.

Basically when a php error is created the server will create an `errors.log` file at the root of the server. This is kind of bad as the file contains your server full url past the `public` folder.

To protect this important file, we decided to block all access to all `.log` files.

So if the file does get created, then October protects the webmaster/developer anyway.

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -21,6 +21,7 @@
     ##
     ## Black listed folders
     ##
+    RewriteRule .log$ index.php [L,NC]
     RewriteRule ^bootstrap/.* index.php [L,NC]
     RewriteRule ^config/.* index.php [L,NC]
     RewriteRule ^vendor/.* index.php [L,NC]
@@ -55,12 +56,5 @@
     ##
     RewriteCond %{REQUEST_FILENAME} !-f
     RewriteRule ^ index.php [L]
-    
-    ##
-    ## Block browser access to log files
-    ##
-    <Files ~ "\.log$">
-        Require all denied
-    </Files>
 
 </IfModule>
--- a/.htaccess
+++ b/.htaccess
@@ -55,5 +55,12 @@
     ##
     RewriteCond %{REQUEST_FILENAME} !-f
     RewriteRule ^ index.php [L]
+    
+    ##
+    ## Block browser access to log files
+    ##
+	<Files  ~ "\.log$">
+  		Require all denied
+	</Files>
 
 </IfModule>
--- a/.htaccess
+++ b/.htaccess
@@ -59,8 +59,8 @@
     ##
     ## Block browser access to log files
     ##
-	<Files  ~ "\.log$">
-  		Require all denied
-	</Files>
+    <Files ~ "\.log$">
+        Require all denied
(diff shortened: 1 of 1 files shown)


LukeTowers:
@ayumi-cloud this should probably go under the black listed section and redirect to index.php like the other black listed locations

Samuell1:
Hey, is this caused by apache or its same for nginx?

LukeTowers:
@Samuell1 the location and name of error.log is typically the result of a cPanel instance, really this is just protecting people from crappy hosts with crappy defaults.

ayumi-cloud:
Sorry this statement is bugging me:

> really this is just protecting people from crappy hosts with crappy defaults.

That's totally not true!

Here's an example, I turn on php 7.4 with October CMS and get a php conflict error, php will generate a log file with the error! This has nothing to do with using a `crappy sever`.

~ Sorry I just want to make sure people understand this and understand this file can be generate to everyone no matter what server they use.

LukeTowers:
@ayumi-cloud You can configure where the PHP error log is located. Crappy servers have that location set to the web root, which is a terrible idea. Better servers have it located outside of the web root for the reason you're wanting to add this to the default .htaccess. 

Either way we're probably going to add this, but my point still stands 😉

ayumi-cloud:
@LukeTowers I see what you mean now, yah I agree with you.

bennothommo:
@ayumi-cloud Are you sure that the log files are web-accessible if they are in the root folder of the project?

These lines in `.htaccess` are supposed to prevent that:

https://github.com/octobercms/october/blob/ec75929678b127d5ddd272ea7138678fe1e89480/.htaccess#L37-L45

Effectively, the rules say that if the requested URL is a legitimate file, and it does not exist in those listed folders, and it is not an `index.php` file, then it will redirect to `index.php`. I don't see how a properly configured Apache server would display a log file from root in that scenario.

ayumi-cloud:
@bennothommo 

To be honest, the security module in our October II proposal we have actually changed most of what's going on with that. The simple reason is that an attacker can inject a open redirect attack and the attack will get stuck in the cache because October is redirecting and not blocking the attack. When you redirect the attack it gets stuck into the cache. This is what happened to Slack this month and they had to patch their systems to stop automatic account stealing.

Secondary the well-known folder is allowing everything with a star which is very dangerous and we changed that with some proper regex lines.

Third the log files don't just need to be in the root, what's to stop an attacker injecting them into a hidden folder etc.

These three things we addressed a few weeks ago in our proposal.

Not sure how to proceed with this pr as we've pretty much changed everything now anyway.

bennothommo:
@ayumi-cloud That's conflating the issue though - if someone is able to inject files in your install, you have bigger problems than the log files being web-accessible.

In normal operation of October, there should be no log files in those directories specified.

PR #4889, opened 2020-01-16
Add all the themes and language packages for ace editor v1.2.6

ayumi-cloud:
Ok, let me explain my thinking!

Last year, @w20k and myself tried to update Ace Editor from v1.2.6 to the latest version and we had a problem trying to update the vendor package. The themes didn't display correctly and produced a white background.

To get around this and give users the full features, I have decided to create a pr that uploads all the themes and language files for the current version Ace Editor v1.2.6

Then you can use more `mode` files.

You can see the current list here: https://github.com/octobercms/october/tree/develop/modules/backend/formwidgets/codeeditor/assets/vendor/ace

For a full list of language files to use in Ace Editor you can check out their doc's found here: https://cloud9-sdk.readme.io/docs/language-mode

p.s. Our company needs these files as we are using the code editor for different coding languages that aren't in the current list.


mjauvin:
Why would we add all this to the base installation if a plugin can do this?

ayumi-cloud:
@mjauvin because the codeeditor widget is in the core? 

So instead of needing to write a plugin, we can just write a single line of code in the yaml file, e.g.

```yaml
language: apache_conf
```

Link: https://octobercms.com/docs/backend/forms#widget-codeeditor

ayumi-cloud:
Maybe the admins can have a discussion about expanding the list of languages.

A perfect example is with Apache, just look at how many plugins in the marketplace use that code with the codeeditor and end up just using a `text` markup because there isn't enough `language` and `theme` support currently.

Yes, we going to build a different code editor we already talked to daftspunk 1-2 months ago about that. It's just October continues to be annoying with development. Yes we can waste time building a plugin, when in a perfect world October just works out of the box. The quick option is to add more `language` and `theme` support to ace editor.

p.s. we are coding a Monaco Editor widget into the core with our october v2 proposal, we just wanted to have something to bridge the gap in the mean time and not have to spend time fixing too many things in v1.

LukeTowers:
@ayumi-cloud you're using the CMS section to edit code files that are not covered by the default installation right now?

ayumi-cloud:
@LukeTowers Hopefully I understand what you mean, not the `CMS Section`. Let me draw you a picture to show you better.

image

In the above picture, were we thinking about allowing users to edit .htaccess code in the code editor which when saving will link up the various code editor files to create the full .htaccess file.

Basically we wanted to add the code theme styling to the code editor widget in the above case that would be `Apache_conf`

We also want to save the file to the location `/views/something.htm` instead of into the mysql database for example.

This is what we are thinking.

Currently October supports these at the moment:

image

We were thinking of just adding more themes, there are a few other coding languages we want to integrate into October such as WebAssembly (we're currently working on) - as an example.

ayumi-cloud:
Same thing like Wordpress Wordfence does I think? 

~We not use Wordpress since 2015 (October all the way now!)

But I believe this is what they do with their software, to allow users to edit and update their Apache files.

mjauvin:
If a plugin you write needs more themes for the Ace editor, why not add this theme in the plugin itself instead of increasing the core October code base?

ayumi-cloud:
@mjauvin because we wanted to ask the admins to do a quick re-check on everything and see if we could update the number of `mode` files from a dozen to two dozen. Just add some commonly used ones such as `.xml` and `.json` etc.

We're also thinking that we might have to create a code editor plugin in the long run - due to the codeeditor only saving to the database at the moment anyway.

p.s. I'm trying to split this issue into two things.

1. We will create a new code editor plugin

2. Double the number of `mode` theme files in the core.

ayumi-cloud:
Full list is here to re-review:

```
ABAP
ABC
ActionScript
ADA
Apache_Conf
AsciiDoc
Assembly_x86
AutoHotKey
BatchFile
C9Search
C_Cpp
Cirru
Clojure
Cobol
coffee
ColdFusion
CSharp
CSS
Curly
D
Dart
Diff
Dockerfile
Dot
Dummy
DummySyntax
Eiffel
EJS
Elixir
Elm
Erlang
Forth
FTL
Gcode
Gherkin
Gitignore
Glsl
golang
Groovy
HAML
Handlebars
Haskell
haXe
HTML
HTML_Ruby
INI
Io
Jack
Jade
Java
JavaScript
JSON
JSONiq
JSP
JSX
Julia
LaTeX
LESS
Liquid
Lisp
LiveScript
LogiQL
LSL
Lua
LuaPage
Lucene
Makefile
Markdown
Mask
MATLAB
MEL
MUSHCode
MySQL
Nix
ObjectiveC
OCaml
Pascal
Perl
pgSQL
PHP
Powershell
Praat
Prolog
Properties
Protobuf
Python
R
RDoc
RHTML
Ruby
Rust
SASS
SCAD
Scala
Scheme
SCSS
SH
SJS
Smarty
snippets
Soy_Template
Space
SQL
Stylus
SVG
Tcl
Tex
Text
Textile
Toml
Twig
Typescript
Vala
VBScript
Velocity
Verilog
VHDL
XML
XQuery
YAML
```

ayumi-cloud:
These are the ones we would use:

```
Apache_Conf
CSS
HTML
Javascript
JSON
LESS
Markdown
MySQL
PHP
Python
SASS
SCSS
Test
Twig
Typescript
XML
YAML
```

LukeTowers:
@ayumi-cloud I would say that @mjauvin is correct, it would be better if your plugin specifically added the mode files that it wanted to use. Additionally, you don't have to save anything in the DB if you don't want to, just utilize the beforeSave() and get*Attribute() methods on the model to handle writing and reading values to/from files instead of the DB respectively.

ayumi-cloud:
@LukeTowers @mjauvin 

No problem, going to close this pr. 

Our company needs to have a private discussion about creating this custom code editor plugin. To cover everyone's required features and needs in our office. I'll probably message Luke in private later when we decided on somethings.

PR #4893, opened 2020-01-20
Rebase October CMS on Laravel 6

### LINKED ISSUE #4381: Move to Laravel 6
https://www.bountysource.com/issues/75392602-move-to-laravel-6

Laravel 6 has been released and is a LTS release (https://laravel.com/docs/6.x/releases#laravel-6). As such, October should look at moving from 5.5 to 6. This issue will contain discussions relevant to that change and related issues.

## Upgrade Guides:
- [x] 5.5 -> 5.6: https://laravel.com/docs/5.6/upgrade
- [x] 5.6 -> 5.7: https://laravel.com/docs/5.7/upgrade
- [x] 5.7 -> 5.8: https://laravel.com/docs/5.8/upgrade
- [x] 5.8 -> 6.x: https://laravel.com/docs/6.x/upgrade

## Related Issues: (so far)
- [x] https://github.com/octobercms/october/issues/4182 **([No longer necessary](https://github.com/octobercms/october/issues/4182#issuecomment-577502907))**
- [x] https://github.com/owncloud/core/issues/13052 & https://www.happyassassin.net/2014/12/29/adventures-in-php-web-asset-minimization/ (Change from JSMin to JSqueeze) - **(implemented in https://github.com/octobercms/october/pull/4893/commits/627c0aa6431ad67c6575851b85630856eb5448d8)**
- [x] #4797 will be moving to a fork of Assetic maintained by the core October CMS team, also will need to ensure that we use wikimedia's fork of LessPHP which will require PHP 7.2.9 **(Fixed by https://github.com/octobercms/library/pull/443)**
- [x] https://github.com/rainlab/builder-plugin/issues/212#issuecomment-489566704 (Upgrade doctrine/dbal) **(Fixed by https://github.com/octobercms/library/pull/443)**
- [x] https://github.com/octobercms/october/pull/4280 (Update autoloading) **(Implemented in https://github.com/octobercms/october/commit/e099b7bdeea9f9619196cb288da9603aae016ea7)**
- [x] https://github.com/octobercms/october/pull/3051 (Laravel Dusk replacing Selenium) **(Fixed by https://github.com/octobercms/october/pull/4919)**
- [x] https://github.com/octobercms/october/issues/3532 (Update PHPUnit) **(Fixed by https://github.com/octobercms/library/pull/443)**
- [x] https://github.com/octobercms/october/issues/4649 (HasOneThrough relationship) **(Implemented in https://github.com/octobercms/library/commit/90718d8c6def5c86fd9479e031f38fb8c5534d29, unit tests https://github.com/octobercms/october/commit/26944a5f68ae34723d9ecfdeac8be4bd799950d4)**
- [x] https://github.com/octobercms/october/issues/3893 (SameSite cookie not working) **(Fixed by build 462)**
- [x] https://github.com/octobercms/october/issues/3227 (Bring October's Auth system more in line with Laravel's) **(This will be investigated for inclusion at a later stage - at the moment, our authentication process is largely incompatible with Laravel's Auth system)**
- [x] https://github.com/octobercms/october/issues/4184 (Update Carbon to Carbon2)
- [x] https://github.com/octobercms/library/commit/c8e2740d974a4409f9990d64f19ef493acf7c0e6 (Revisit cookie serialization and encryption serialization in general)
- [x] https://github.com/octobercms/library/pull/443 (Unit testing upgrading) **(Merged)**
- [x] L6 also includes a switch to the Ignition error page handler (https://github.com/facade/ignition), we should look into integrating that into October **(We will not be including this in October at this stage.)**
- [ ] Support added back for serializable closures, also check if route caching works with closures (https://github.com/octobercms/docs/issues/425)
- [x] Update marketplace's list of pre-installed composer dependencies: https://github.com/octobercms/october/issues/2831
- [x] October's Translation service doesn't match Laravel's https://github.com/octobercms/october/issues/5092, fixed by https://github.com/octobercms/library/commit/b8e400b6a3a076a9a70dba42631bfe13bff30cd3

## Breaking changes: (so far)
- [x] storage/framework/cache/data directory needs to be added and `!data/` added to the .gitignore file under `storage/framework/cache` **(Fixed by https://github.com/octobercms/october/pull/4893/commits/25511bfd34e338bd95f2b4d1721d2ca50d8a551d)**
- Cms\Classes\CmsObjectCollection->where() method signature changed
- [x] Require PHP 7.2 **(Fixed by https://github.com/octobercms/library/pull/443)**
- [x] Require PHP Unit 8.0 (unit tests will need to have return type hints added) - also remove https://github.com/octobercms/october/commit/72f85ebc558a8e6fe514f6496f5393e535bac41b when updating **(Fixed by https://github.com/octobercms/library/pull/443)**
- [x] Cache TTL has been changed from minutes to seconds when using integers, recommend using datetime objects instead
- [x] ~~May need to swap https://github.com/meyfa/phpunit-assert-gd for https://github.com/lupka/phpunit-compare-images as assertGD doesn't support PHPUnit 7.0~~ (No longer required, `phpunit-assert-gd` [supports PHPUnit 8 now](https://github.com/meyfa/phpunit-assert-gd/pull/2)) **(Fixed by https://github.com/meyfa/phpunit-assert-gd/pull/2)**
- [x] `setUp()` and `tearDown()` methods in unit tests require adding ` : void` as a return type hint to their definitions. **(Fixed by https://github.com/octobercms/library/pull/443)**
- [x] If you're utilizing the core unit tests you will need to grab the latest copies of the base unit tests from octobercms/october on GitHub and drop them into your project
- [x] Carbon will be updated from v1 to v2, see https://carbon.nesbot.com/docs/#api-carbon-2 (Still need to go through and upgrade in core)
- [x] Sparkpost & Mandrill mail drivers removed from core Laravel, we should probably add them back to October (i.e. revert https://github.com/laravel/framework/commit/e2461f19c3b951f7cbb4b424601cad1f6553774a#diff-c1d5bf1c77457840314f00f3837a2b2a in octobercms/library) **(Implemented in https://github.com/octobercms/library/commit/c8e34ea17579b5c8a7a4f9d9414d713425e6f30c)**
- [x] The Input facade has been removed as it was mostly a mirror of the Request facade, we need to add it back in (https://laravel.com/docs/6.x/upgrade#the-input-facade) **(Fixed by https://github.com/octobercms/library/pull/443)**
- [x] The default redis client has changed from `predis` to `phpredis`. Configuration change may be required by users wishing to retain predis, or potentially we need to dynamically set the configuration for people (by adding a client config value defaulting to phpredis and doing Config::set() if the value isn't detected). See more here: https://laravel.com/docs/6.x/upgrade#redis-default-client **(Fixed by https://github.com/octobercms/library/commit/4aeaf54d7e3d12a916a3e096cdbf730fe2aee992, https://github.com/octobercms/october/compare/0394494fbe43...a2966da395bb)**
- [x] Rackspace storage support was removed. Potentially needs to be re-added? **(Re-added in https://github.com/octobercms/library/commit/3ee82fd1af518fac829d780fcedf2cbf723efff0)**
- [x] `email` validation rule changed internal behaviour to use a broken version of RFC validation instead of PHP's built in validation, need to change it back. See https://github.com/octobercms/october/issues/5070
- [x] `trans()` and `transChoice()` methods have been removed from the Translation service, fixed by https://github.com/octobercms/library/commit/a364d6455698b4786efc6870cf097a8677b2dccc

## Other tasks:
- [x] Integrate old helpers (https://github.com/laravel/helpers/blob/master/src/helpers.php) into Rain library. **(Implemented in https://github.com/octobercms/library/pull/456/commits/af29bbfc32230140d4deac1a76b5b81d38f401f1)**
- [x] Update the Unit Testing documentation in the docs repo and https://github.com/octobercms/october/tree/wip/laravel-6/tests#plugin-testing
- [x] Confirm Selenium tests in `tests/functional` directory still work, update README if necessary. **(Replaced with Dusk testing: https://github.com/octobercms/october/pull/4919)**
- [x] [Low priority] Add Selenium tests to GitHub Actions workflow.
- [ ] Go through all of the documentation under octobercms/docs that originally mirrored the Laravel docs and update it to reflect the latest changes to the Laravel docs.
- [x] [Update Validator to include latest rules.](https://github.com/octobercms/october/pull/4893#issuecomment-591152927) **(fixed by https://github.com/octobercms/october/commit/9911c6b5817088a99ac65d9f2ac76346402e1d47)**
- [x] Update Laravel Tinker to 2.0 to improve compatibility of dependencies [https://github.com/octobercms/library/commit/6fae9cd2baed001f9aa69415f2f130fc104eb573](https://github.com/octobercms/library/commit/6fae9cd2baed001f9aa69415f2f130fc104eb573)

## Known issues:
- [x] [Signed URLs do not seem to work.](https://github.com/octobercms/october/pull/4893#issuecomment-576755152) **(fixed by https://github.com/octobercms/library/pull/456/commits/c258e666b9c66174e773a57317764209665f2703)**
- [x] `php artisan cache:clear` removes the required `storage/framework/cache/data` folder.

bennothommo:
Fixes #4381. Related Rain library PR: https://github.com/octobercms/library/pull/456. Related docs PR: https://github.com/octobercms/docs/pull/446

This PR represents the current progress for rebasing October CMS on the Laravel 6 framework. This PR can be considered an "early beta" release. It is not recommended for use on production installations, however, we would encourage users to test this out on development copies in preparation for the upgrade.

**[Release notes and required changes can be found here.](https://github.com/octoberrain/meta/blob/master/l6-upgrade-notes.md)**

## New minimum requirements
- PHP 7.2 or above.

## Testing
We would appreciate any assistance with testing from the community. If you are interested in testing this out, please follow the below steps to quickly get a fresh instance of the CMS running.

### Testing on existing projects

Follow the Upgrade Notes here: https://github.com/octoberrain/meta/blob/master/l6-upgrade-notes.md

### Testing on fresh instance

1. Download and unzip the Laravel 6 branch into a directory, or clone the repository and checkout the **wip/laravel-6** branch.
    - Download: https://github.com/octobercms/october/archive/wip/laravel-6.zip
    - Clone: `git clone git@github.com:octobercms/october.git`
2. Install the Composer dependencies within the newly created folder.
    - Run `composer install`
3. Generate a **.env** file.
    - Run `php artisan october:env`
4. Change the following details in your **.env** file.
    - **APP_URL**: `http://127.0.0.1:8000`
    - **APP_KEY**: Run `php artisan key:generate --show` and copy and paste the value generated here.
    - **DB_CONNECTION**: `sqlite`
    - **DB_DATABASE**: `storage/database.sqlite`
5. Create a new SQLite file in **storage/database.sqlite**.
    - Run `touch storage/database.sqlite`
6. Run all the migrations to set up the database.
    - Run `php artisan october:up`
7. Finally, serve the website.
    - Run `php artisan serve`

This will allow you to review the CMS in the following contexts:
- **Frontend:** http://127.0.0.1/
- **Backend:** http://127.0.0.1/backend (log in using "admin" as username, "admin1234" as password)

## Reporting bugs and issues
At this stage, do not create a new issue on GitHub if you spot any issues. Please report the issue as a comment to https://github.com/octobercms/october/issues/4381, which we are using to track progress on the task.


composer.json
```diff
@@ -31,21 +31,19 @@
...
-        "php": ">=7.0.8",
+        "php": "^7.2",
         "ext-mbstring": "*",
         "ext-openssl": "*",
-        "october/rain": "~1.0",
-        "october/system": "~1.0",
-        "october/backend": "~1.0",
-        "october/cms": "~1.0",
-        "laravel/framework": "^5.5.40",
+        "october/rain": "dev-wip/laravel-6",
```
LukeTowers:
```suggestion
        "october/rain": "dev-wip/laravel-6 as 1.0",
```
The backend modules require 1.0 so I think this is required @bennothommo (correct me if I'm wrong)

bennothommo: @LukeTowers you're correct - however, I removed the modules from composer.json temporarily so I didn't have to deal with sub-splitting the branch. My plan was to re-add it back before merging it all in.

LukeTowers: Ah, I see, makes sense

modules/cms/classes/CmsObjectCollection.php
```diff
@@ -35,14 +36,38 @@ public function withComponent($components, $callback = null)
...
+     * Returns objects whose properties match the supplied value.
+     *
+     * This is a wrapper for our custom `applyWhereFilter` method, as the signature of this method changed in Laravel 6.
+     *
+     * @param  string  $key
+     * @param  mixed  $operator
+     * @param  mixed  $value
+     * @return static
+     */
+    public function where($key, $operator = null, $value = null)
```
LukeTowers: @bennothommo is this a breaking change technically? How does this affect the Halycon datasources?

modules/backend/models/User.php
```diff
@@ -24,10 +24,10 @@ class User extends UserBase
...
      */
     public $rules = [
-        'email' => 'required|between:6,255|email|unique:backend_users',
-        'login' => 'required|between:2,255|unique:backend_users',
-        'password' => 'required:create|between:4,255|confirmed',
-        'password_confirmation' => 'required_with:password|between:4,255'
+        'email' => 'required|between:6,191|email|unique:backend_users',
+        'login' => 'required|between:2,191|unique:backend_users',
+        'password' => 'required:create|between:8,191|confirmed',
+        'password_confirmation' => 'required_with:password|between:8,191'
```
LukeTowers: @daftspunk @bennothommo I wonder if we need to go as deep as detecting what the default varchar length is with a special character to be parsed by the validation trait since we've introduced the config for it.

modules/cms/routes.php
```diff
@@ -22,7 +22,7 @@
      * The CMS module intercepts all URLs that were not
      * handled by the back-end modules.
      */
-    Route::any('{slug}', 'Cms\Classes\CmsController@run')->where('slug', '(.*)?')->middleware('web');
+    Route::any('{slug?}', 'Cms\Classes\CmsController@run')->where('slug', '(.*)?')->middleware('web');
```
LukeTowers, line 25: Why was this required?

tests/unit/system/classes/AutoDatasourceTest.php
```diff
@@ -7,7 +7,7 @@
 
 class CmsThemeTemplateFixture extends Model
 {
-    protected $fillable = ['*'];
+    protected $guarded = [];
```
LukeTowers, line 10: What was the reason for this change?

LukeTowers (dismissed):

modules/backend/models/User.php
daftspunk: Bit of a complex idea, 191 is a reasonable length. Note password does not need this limitation because the end result is hashed

modules/cms/classes/CmsObjectCollection.php
bennothommo: @LukeTowers I believe it's not a breaking change - the new method signature is technically less restrictive than the old signature. I've made that particular method a wrapper which replicates the restrictions of the old method and forwards the call on to a new method which is essentially our old method. `CmsObjectCollection` doesn't appear to be extended by another class, so I believe it should be fine.

modules/cms/routes.php
bennothommo: The home page wouldn't load without it, as the home page has no slug. :)

modules/backend/models/User.php
LukeTowers: Fair enough. @daftspunk what's the length of the hashed end result? Should we be concerned about 191 being too short? @bennothommo you can change this to be a min instead of a range.

modules/cms/routes.php
LukeTowers: But home pages work fine right now, so what changed?

tests/unit/system/classes/AutoDatasourceTest.php
bennothommo: `protected $fillable = ['*']` is actually incorrect. What happened was in our unit tests, when the seeders run on set up, they set all Eloquent models to be unguarded, so this was never picked up. However, Laravel must've improved their guarding, because the unit test that tests fillable attributes was failing. I've set the guarding for that fixture to use the correct method for making a single model unguarded.

modules/backend/models/User.php
daftspunk: Password hash is 60 characters

modules/cms/classes/CmsObjectCollection.php
LukeTowers: @bennothommo what used to call this method though? Whatever called this method would still be sending the old signature. Additionally, it seems like this is misleading, since we haven't actually changed the behavior of the internal filter to take the operator into account.

tests/unit/system/classes/AutoDatasourceTest.php
LukeTowers: Perfect, thanks for the explanation!

modules/cms/routes.php
bennothommo: This is what happens without the change:
image
I can only surmise that Laravel has fixed something with regards to the difference between required URL parameters and optional ones.

modules/cms/classes/CmsObjectCollection.php
bennothommo: @LukeTowers I think it's mainly called when using a query against Halcyon objects. The above change will work with the old signature. It would only fail if someone wants to use the *new* signature for Halcyon objects (ie. adding an operator to `where` queries) - as far as I can tell, that's not happening as it's never been supported previously.

I felt that since Halcyon objects are something completely unique to October, it would be safe to keep the behaviour as is.

LukeTowers: @bennothommo do we have to change the parameter names in order to pass the matching signature check? If not I'd prefer to keep the old names even if it means we have to change their defaults. 

Either way there should be a comment about how it doesn't actually match the new laravel signature

bennothommo: @LukeTowers it appears we can keep the property names, but they have to remain with `null` as default in order to match the signature from Laravel. I've changed the method around to suit, and added a comment with the details of the change.

modules/backend/models/User.php
bennothommo: @LukeTowers I've changed the rules to use `min` instead. Should we do the same here: 
https://github.com/octobercms/library/blob/wip/laravel-6/src/Auth/Models/User.php#L27-L29
They are some pretty conservative rules.

composer.json
```diff
@@ -28,24 +28,23 @@
...
         "wikimedia/composer-merge-plugin": "dev-master"
     },
     "require-dev": {
-        "fzaninotto/faker": "~1.7",
-        "phpunit/phpunit": "~6.5",
-        "phpunit/phpunit-selenium": "~1.2",
-        "meyfa/phpunit-assert-gd": "1.1.0",
+        "fzaninotto/faker": "^1.9",
+        "phpunit/phpunit": "^8.0|^9.0",
+        "phpunit/phpunit-selenium": "dev-master",
```
LukeTowers: Is this temporary?

modules/cms/routes.php
LukeTowers: Hmm, odd. Oh well, should be fine.

composer.json
bennothommo: @LukeTowers I believe so - they seem to be on the cusp of releasing v8.0 which supports PHPUnit 8 and 9. TBH though, I'm not sure if our Selenium tests still work.

LukeTowers: Perhaps you can add the selenium tests back into the unit testing automation?

bennothommo: @LukeTowers I can give it a try - I recall these sort of tests requiring us to fire up a headless version of Chrome to run the tests on, which was painful to install. Were these tests ever part of the automated testing?

LukeTowers: Not sure, but if we can make it work with github actions that would be pretty sweet. I know there's a couple of plugins that make integrating with Selenium / Dusk in October a bit better, perhaps we should just try and get the current tests working standalone (and publish instructions for doing so) and then later on we can revisit automating them.

bennothommo: No worries, I'll add it as a task to the list. I'll leave this thread open as well to remind me.

Teranode:
Getting an error when using Signed urls:
[log.txt](https://github.com/octobercms/october/files/4091806/log.txt)

Following: https://laravel.com/docs/6.x/urls#signed-urls

I've tried using:
```
Route::get('episode/download/{id}', ['as' => 'download', 'middleware' => 'signed', 'Batotwo\Show\Http\Controllers\Shows\ShowController@downloadFile']);
```
Using `->name('download')` will generate an error so i have to use `'as' => 'download'` ive also tried middleware `->middleware('Spatie\UrlSigner\Laravel\Middleware\ValidateSignature')` and `'middleware' => 'Spatie\UrlSigner\Laravel\Middleware\ValidateSignature'`

Also used 'Illuminate\Routing\Middleware\ValidateSignature'

but still ended up with that error.

LukeTowers:
@Teranode huh, didn't know that feature existed! Pretty nifty. @bennothommo once we've got it properly supported in October do you think we should add it to our docs?

bennothommo:
@LukeTowers sounds good to me. I'll add it to the list.

modules/backend/models/User.php
LukeTowers: @bennothommo hmm, I would say so yes. That might break some uses but nothing should really use the base model unmodified anyways

Teranode:
Artisan commands giving off errors:
`php artisan cache:clear` doesn't work
`php artisan cache:forget` does work

but gives off that error. On further inspecting you're suppose to specify a `key` or a `store`, when using it. When I don't it defaults to removing all keys and all stores and then giving that error.

rajakhoury:
@Teranode both commands are working fine on my ends, been using this this with the file and redis stores. Isn't this related to folder permissions?

bennothommo:
@Teranode I would check the permissions of your `storage` folder to determine if that is causing `php artisan cache:clear` to fail. `cache:forget` is for removing a single item from the cache, so a key and a store is meant to be defined for that command.

I've just tested both on my own local copy and not had an issue.

LukeTowers:
To anyone interested, we've made a lot of progress and the L6 upgrade is now ready for active testing. Please review https://github.com/octoberrain/meta/blob/master/l6-upgrade-notes.md for information on how you can test the changes in your own projects and any changes that may be required in your code bases.

Teranode:
@LukeTowers 
Upon changing the composer package via:
```
"october/rain": "dev-wip/laravel-6 as 1.0",
"october/system": "dev-wip/laravel-6",
"october/backend": "dev-wip/laravel-6",
"october/cms": "dev-wip/laravel-6",
```

I get this:


It happens for every backend url except for the ones appended with: `system/settings/update`

bennothommo:
@Teranode I've only used and tested the file cache at the moment.

modules/system/ServiceProvider.php
```diff
@@ -94,6 +94,7 @@ public function boot()
             }
         }
 
+        Paginator::useBootstrapThree();
         Paginator::defaultSimpleView('system::pagination.simple-default');
```
FlusherDock1, line 98: Can we finally convert html markup of pagination from Bootstrap 3 to 4 ?

LukeTowers: @FlusherDock1 you can feel free to override this in your own plugins but changing this away from Bootstrap 3 will break existing sites using the paginate()->render().

FlusherDock1: @LukeTowers I respect that, but using this kind of logic we can't update everything because there will be always one website that can be broken because of the new changes. 

P.S. Bootstrap 4 was released 6 years ago, I don't think someone really using Bootstrap 3 codebase right now.

LukeTowers: @FlusherDock1 there's no-one stopping you from using whatever markup you want in your own projects. I can tell you right now that there are active sites with millions of views annually that utilize the default Bootstrap 3 markup. If you want to use Bootstrap 4, then just do the following in your own plugins:

```php
Paginator::defaultView('pagination::bootstrap-4');
Paginator::defaultSimpleView('pagination::simple-bootstrap-4');
```

bennothommo: @FlusherDock1 The `useBootstrapThree` call is mainly for the Backend, which uses the Storm UI and thus uses Bootstrap 3 classes. Changing this would break the pagination on the backend. As @LukeTowers has indicated, you are more than free to use Bootstrap 4 for your own plugins.

FlusherDock1: >If you want to use Bootstrap 4, then just do the following in your own plugins:
Paginator::defaultView('pagination::bootstrap-4');
Paginator::defaultSimpleView('pagination::simple-bootstrap-4');

Thanks, i didn't know about that.

>The useBootstrapThree call is mainly for the Backend, which uses the Storm UI and thus uses Bootstrap 3 classes. Changing this would break the pagination on the backend. As @LukeTowers has indicated, you are more than free to use Bootstrap 4 for your own plugins.

About backend bootstrap 3 I know, I just suggested changes only for frontend. 

NVM, i think this question is closed due to storm ui limitations

.github/workflows/tests.yml
```diff
@@ -83,6 +83,12 @@ jobs:
           path: ${{ steps.composer-cache.outputs.dir }}
           key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.json') }}
           restore-keys: ${{ runner.os }}-composer-
+      - name: Install Dusk plugin
```
LukeTowers: Why don't we pull this in via composer?

bennothommo: @LukeTowers we definitely can once it's been added to Packagist. I just wanted to get it working first.

martendej:
> Artisan commands giving off errors:
> `php artisan cache:clear` doesn't work
> `php artisan cache:forget` does work
> 
> but gives off that error. On further inspecting you're suppose to specify a `key` or a `store`, when using it. When I don't it defaults to removing all keys and all stores and then giving that error.
> 

I have the same issue, I have traced it down to the Illuminate\Cache\Console\ClearCommand expecting a boolean return value on the flush() method of the cache driver, which the October\Rain\Halcyon\MemoryRepository (default?) does not return.

bennothommo:
@martendej Thanks for the info. I will investigate it further.

multiwebinc:
https://github.com/octobercms/october/blob/wip/laravel-6/modules/system/lang/en/validation.php needs to be updated with the changes made in https://github.com/laravel/laravel/blob/master/resources/lang/en/validation.php. There are several entries missing (e.g. `gt`, `gte`, `lt`, `lte`).

bennothommo:
Thanks for the report @multiwebinc.

bennothommo:
@Teranode signed URLs should now work as of https://github.com/octobercms/library/pull/456/commits/c258e666b9c66174e773a57317764209665f2703.

bennothommo:
@multiwebinc Language files for validation have been updated: https://github.com/octobercms/october/commit/9911c6b5817088a99ac65d9f2ac76346402e1d47

multiwebinc:
Excellent work, @bennothommo

Teranode:
Almost forgot, Laravel Horizon will not function if the code below is not present in the app.php file.

```
    /*
    |--------------------------------------------------------------------------
    | Application Environment
    |--------------------------------------------------------------------------
    |
    | This value determines the "environment" your application is currently
    | running in. This may determine how you prefer to configure various
    | services your application utilizes. Set this in your ".env" file.
    |
    */
    'env' => env('APP_ENV'),
```

config/app.php
```diff
@@ -129,6 +129,26 @@
...
+    | Load automatically discovered packages
+    |--------------------------------------------------------------------------
+    |
+    | By default, October CMS disables the loading of discovered packages
+    | through Laravel's package discovery service, in order to allow packages
+    | used by plugins to be disabled if the plugin itself is disabled.
+    |
+    | Set this to `true` to enable automatic loading of these packages. This
+    | will result in packages being loaded, even if the plugin using them is
+    | disabled.
```
LukeTowers:
```suggestion
    | disabled. This is NOT RECOMMENDED.
```

modules/backend/routes.php
```diff
@@ -25,7 +25,7 @@
             'middleware' => ['web'],
             'prefix' => Config::get('cms.backendUri', 'backend')
         ], function () {
-            Route::any('{slug}', 'Backend\Classes\BackendController@run')->where('slug', '(.*)?');
+            Route::any('{slug?}', 'Backend\Classes\BackendController@run')->where('slug', '(.*)?');
```
LukeTowers, line 28: Is this a change that needs to be documented? It affected RainLab.Translate so the possibility exists that other plugins could be affected.

modules/cms/twig/DebugExtension.php
```diff
@@ -8,7 +8,7 @@
 use Cms\Classes\ComponentBase;
 use Illuminate\Pagination\Paginator;
 use Illuminate\Support\Collection;
-use Illuminate\Support\Debug\HtmlDumper;
```
LukeTowers, line 11: Does this require a class alias setup in the library?

bennothommo: @LukeTowers good idea - https://github.com/octobercms/october/commit/0dee0e54b100a22cdd36647161e951af04effcce

modules/backend/routes.php
bennothommo: @LukeTowers done - https://github.com/octoberrain/meta/commit/81507a443a9027fcb9a03bb82b2b2b3f7c1652fb.

Issue #4892, opened 2020-01-18
Chrome 81 has some updated styling features


ayumi-cloud:
Just to let the admins know and do some A-B testing with October CMS.

Google Chrome 81 have updated many styling features, a few months ago they did a full video on the features, I think this was the video I watched: https://www.youtube.com/watch?v=-oyeaIirVC0

Basically, in October CMS you get a very heavy black line around things when you select objects now.

It's actually a little bit annoying!

See example:

![image](https://user-images.githubusercontent.com/57409060/72667282-066f4800-3a12-11ea-917f-b7861ead09aa.png)

Also another thing is the checkboxes are different in the new chrome but that's actually a nice thing and looks much better in github.

I thought I'd let the admins know and they can decide if they want it to look like this or fix some different cross-browser styling.

Firefox looks like this in Nightly:

![image](https://user-images.githubusercontent.com/57409060/72667302-56e6a580-3a12-11ea-87e0-df3f62ec77c0.png)

As per github issue: https://github.com/octobercms/october/issues/4819

Thought it best to tell you (before people start using Chrome 81 stable).

ayumi-cloud:
p.s. Not going to create a pr until the admins give me feedback. Doing styling pr's is hard because one person says one thing and another person says another thing. So best to wait for a decision to be made first and then I be happy to code the pr.

Note: Google Chrome for many years added a soft blue line around things! Now it's a hard black line (which kind of looks ugly) I would guess it's to make developers a bit more pro-active in addressing accessibility coding issues.

ayumi-cloud:
One idea could be to create a new config file called: `accessibility` and have various options users can select, for example one user may want to turn this feature off, another user may want to style it blue etc. Then the config file could connect to a `dynamic` css file to style the backend, instead of the current setup which is hard-coded.

Samuell1:
@ayumi-cloud I think best will be to remove this it looks ugly and if somebody wants it he is free to overwrite css by his custom backend styles.

ayumi-cloud:
@Samuell1 yah I agree with you. Will wait for confirmation from the admins.

LukeTowers:
Personally I don't mind the dotted line, it does make the default styling more accessible. I'll leave this one up to @daftspunk though.

bennothommo:
My two cents: Dotted-line looks fine to me in the nav, the big black line looks gross 🤢. Whoever thought that would improve accessibility should be [insert awful event here].

ayumi-cloud:
[...]

PR #4895, opened 2020-01-21
Fix and standardized cross-browser focus ring issues

ayumi-cloud:
This fixes the github issue: https://github.com/octobercms/october/issues/4892

Google Chrome 1 - 80 used a light blue focus ring and Chrome 81 and above uses a dark black line. I have a theory they changed this for people with poor eye sight?

### Chrome 81 (before)

image

### Chrome 81 (after)

image

p.s. In firefox it will remove the dotted line and put the same blue line as seen in the after picture in chrome into firefrox.

### DIFF
--- a/modules/system/assets/ui/less/site.normalize.less
+++ b/modules/system/assets/ui/less/site.normalize.less
@@ -1,3 +1,13 @@
+/*! Make focus ring standard on every browser */
+
+// See github issue https://github.com/octobercms/october/issues/4892
+
+&:focus {
+    outline: none;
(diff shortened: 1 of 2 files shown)


Samuell1:
@ayumi-cloud This was nice change but now issue is focus is now visible even with mouse clicking maybe there should be way to enable focus only when mouse isnt used? Something like this https://github.com/ten1seven/what-input#basic-styling

LukeTowers:
@Samuell1 we're not adding any more JS dependencies for this

Samuell1:
@LukeTowers it can be done like this: https://stackoverflow.com/a/51093815

PR #4904, opened 2020-01-22
Load mail templates from view files if not found in the database

fansaien:
The new layout should be saved in the system_mail_layouts table.

The  `MailLayout::createLayouts();` only be called in the MailTemplate Controller. So, the new mail layout can be added to the database only after opened the Mail templates page in the settings.

The current code will auto run `MailLayout::createLayouts();` when the layout code not found in the database.

### DIFF
--- a/modules/system/models/MailLayout.php
+++ b/modules/system/models/MailLayout.php
@@ -68,13 +68,18 @@ class MailLayout extends Model
 
     public static function getIdFromCode($code)
     {
-        $layoutId = array_get(self::listCodes(), $code);
-        if ($layoutId === null) {
-            self::createLayouts();
-            self::$codeCache = null;
-            $layoutId = array_get(self::listCodes(), $code);
+        return array_get(self::listCodes(), $code);
+    }
+
+    public static function findOrMakeLayout($code)
+    {
+        $layout = self::whereCode($code)->first();
+        if (!$layout && View::exists($code)) {
+            $layout = new self;
+            $layout->code = $code;
+            $layout->fillFromView($code);
         }
-        return $layoutId;
+        return $layout;
     }
 
     /**
--- a/modules/system/models/MailLayout.php
+++ b/modules/system/models/MailLayout.php
@@ -74,11 +74,13 @@ class MailLayout extends Model
     public static function findOrMakeLayout($code)
     {
         $layout = self::whereCode($code)->first();
+
         if (!$layout && View::exists($code)) {
             $layout = new self;
             $layout->code = $code;
             $layout->fillFromView($code);
         }
+
         return $layout;
     }
 
--- a/modules/system/models/MailLayout.php
+++ b/modules/system/models/MailLayout.php
@@ -68,13 +68,13 @@ class MailLayout extends Model
 
     public static function getIdFromCode($code)
     {
-        $styleId = array_get(self::listCodes(), $code);
-        if ($styleId === null) {
+        $layoutId = array_get(self::listCodes(), $code);
+        if ($layoutId === null) {
             self::createLayouts();
             self::$codeCache = null;
-            $styleId = array_get(self::listCodes(), $code);
+            $layoutId = array_get(self::listCodes(), $code);
         }
-        return $styleId;
+        return $layoutId;
     }
 
     /**
--- a/modules/system/models/MailLayout.php
+++ b/modules/system/models/MailLayout.php
@@ -68,7 +68,13 @@ class MailLayout extends Model
 
     public static function getIdFromCode($code)
     {
-        return array_get(self::listCodes(), $code);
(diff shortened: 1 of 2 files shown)


LukeTowers:
Please provide more information on why this is required.

fansaien:
The developer can add a new mail layout in plugin.php and set the layout in the Mail Views: https://octobercms.com/docs/services/mail

Then the MailTemplate.php try to get the layoutId by the layoutCode
```php
$this->layout_id = MailLayout::getIdFromCode($layoutCode);
```

However, the layout_id will be null unless access the /system/mailtemplates page first.
Because the MailTemplates Controller will sync all the layout into the database in the inex() method.

So, I did this change to make sure the layout file be saved to the table auto and do not need to open the template page again.

LukeTowers:
@fansaien thanks for the explanation! Have you tested this locally? There's an in-request memory cache of DB queries so I suspect that the second ::lists() would return the same result as the first.

fansaien:
@LukeTowers 
Hi, yes, I tested in my local env.
There are two new Query function in the `Illuminate\Database\Eloquent\Model`
```php
// No Cache
public function newQuery()

// This funcation may enable the in memory cache
public function newModelQuery() 
```
The `lists()` will call newQuery() in this way
```php
public function __call($method, $parameters)
{
    if (in_array($method, ['increment', 'decrement'])) {
        return $this->$method(...$parameters);
    }

    return $this->newQuery()->$method(...$parameters);
}
```

So, there is no in-memory cache for the lists() function

LukeTowers:
Perfect, thanks for checking that! @bennothommo I'm not sure I'm a huge fan of the rescanning happening when attempting to find a layout and failing, but tbh I can't really think of a better place to do that. Could you take a look at this PR and let me know what you think?

bennothommo:
@LukeTowers hmmm, I would've thought the database would only be storing templates and layouts that have *changed*, otherwise it should use the default template and layout provided by the plugin?

LukeTowers:
@bennothommo that's a good point. 

@fansaien why do you need a layout's ID instead of just the code? What are you using that value for?

fansaien:
I add my layout like this

* in plugin.php

```php
public function registerMailLayouts()
{
    return [
        'marketing'    => 'acme.blog::layouts.marketing'
    ];
}
```

* set the layout id in the mail.htm

```htm
subject = "Sample Email"
layout = "marketing"
==
xxxx
```

When I send the email, I found it try to get the layout id in `MailTemplate.php`

```php
protected function fillFromSections($sections)
{
    $this->content_html = $sections['html'];
    $this->content_text = $sections['text'];
    $this->subject = array_get($sections, 'settings.subject', 'No subject');

    $layoutCode = array_get($sections, 'settings.layout', 'default');
    $this->layout_id = MailLayout::getIdFromCode($layoutCode);
}
 ```
 The layout_id will be null because getIdFromCode method only read the layout id from database.

So, I modified the `getIdFromCode` method to load the custom layout.

**Comment by LukeTowers** [comment]
The problem then is that the layout is not being loaded properly from code. Layouts, just like views, should not have to exist in the database in order to be used in code. Are you able to look into a fix for that?

**Comment by fansaien** [comment]
If the system_mail_layouts table only designed for the system and default layout, I agree with you.

I submit the new codes and follow the MailTemplate::findOrMakeTemplate method

File: modules/system/models/MailLayout.php [hunk 1]
```diff
@@ -68,13 +68,18 @@ public static function listCodes()
...
+    {
+        $layout = self::whereCode($code)->first();
+        if (!$layout && View::exists($code)) {
+            $layout = new self;
+            $layout->code = $code;
+            $layout->fillFromView($code);
         }
-        return $layoutId;
+        return $layout;
     }
```
LukeTowers: This is a great solution, although a couple minor changes required. First, we'll have to retain the `getIdFromCode()` method as it's public and could be relied on by current plugins, so just add the `findOrMakeLayout()` as an additional method. 

Secondly, could you add an extra line of whitespace before and after the `if (!$layout && View::exists($code))` block to match the `findOrMakeTemplate()` method on the MailTemplate class?

Thanks for your work on this!

fansaien: I added the space lines. 
But I can not change the method getIdFromCode to protected because it is used in `MailTemplate::syncAll()` method

LukeTowers: That's fine, I wanted it to stay public

LukeTowers:
LGTM! Just awaiting testing of the original bug and confirmation this resolves the issue. @w20k if you have time could you take a look?

w20k:
@LukeTowers, when `labs` get back to business, will do the full `todo list` testing :)
From first glance - LGTM!

LukeTowers:
@w20k any updates on this?

PR #4908, opened 2020-01-23
Updated getParameter method

martendej:
Fixed typehinting for getParameter method in Router class (this method returns a string or null, not an array) and simplified its body.

### DIFF
--- a/modules/cms/classes/Router.php
+++ b/modules/cms/classes/Router.php
@@ -300,15 +300,13 @@ class Router
 
     /**
      * Returns a routing parameter.
-     * @return array
+     * @param  string $name
+     * @param  string|null $default
+     * @return string|null
      */
-    public function getParameter($name, $default = null)
+    public function getParameter(string $name, string $default = null)
     {
-        if (isset($this->parameters[$name]) && ($this->parameters[$name] === '0' || !empty($this->parameters[$name]))) {
-            return $this->parameters[$name];
-        }
-
-        return $default;
+        return $this->parameters[$name] ?? $default;
     }
 
     /**
--- a/modules/cms/classes/Router.php
+++ b/modules/cms/classes/Router.php
@@ -306,7 +306,8 @@ class Router
      */
     public function getParameter(string $name, string $default = null)
     {
-        return $this->parameters[$name] ?? $default;
+        $value = $this->parameters[$name] ?? '';
+        return $value !== '' ? $value : $default;
     }
 
     /**


martendej:
With this update empty strings would now be returned instead of the 'default' parameter value however. Are empty route parameter values possible within the current setup?

LukeTowers:
I know there's something with the blog plugin where `default` will be added to one of the URL parameters as a fallback, not sure if this breaks that or not.

w20k (approved):
LGTM, but needs testing.

modules/cms/classes/Router.php
```diff
@@ -300,15 +300,13 @@ public function getUrl()
...
      */
-    public function getParameter($name, $default = null)
+    public function getParameter(string $name, string $default = null)
     {
-        if (isset($this->parameters[$name]) && ($this->parameters[$name] === '0' || !empty($this->parameters[$name]))) {
-            return $this->parameters[$name];
-        }
-
-        return $default;
+        return $this->parameters[$name] ?? $default;
```
bennothommo: @martendej I think it is possible for parameters to be provided as an empty string depending on routing rules in place, eg. optional parameters - previously, this would use the default value. This change will prevent that from happening. Would you be able to revert it to using the condition block that was there before? The typehint changes above are fine to leave in.

bennothommo (changes requested):

modules/cms/classes/Router.php
martendej: I changed the code to account for this situation, can you check if this is ok with you?

bennothommo (approved):

modules/cms/classes/Router.php
bennothommo: Thanks @martendej.

LukeTowers:
LGTM. Has anyone tested this yet on Octodock?

bennothommo:
Thanks @martendej. Tested and works fine.

marcbrouwer:
I'm having issue with this change in v1.1.1. If an optional parameter is not set it always results into 0 (false). In the case of an optional parameter `:param?`, which is not set, the results are always 0/false for the following cases:

- $this->param('param')
- $this->param('param', 'Some default value')

Is anyone else having this issue?

mjauvin:
@marcbrouwertue what does the page URL setting looks like? Did you define a default value for this URL's param?

marcbrouwer:
@mjauvin the url is set to `/:param?`, nothing else, so no default value

mjauvin:
right, if no default value is set, it will get assigned false which means it's not an empty string which means the default value passed to getParameter() will not be used.

Why not set a default value in the url parameter?

LukeTowers:
@marcbrouwertue what was your code expecting previously?

marcbrouwer:
@mjauvin @LukeTowers I'm updating from 1.0.496 iirc. Here not setting a default value results into null, wheras after the change it results into false. The code below simulates the behaviour:

```
<?php

// simulating call to getParameter(string $name, string $default = null)
$parameters = [ 'param' => false ]; // simulating $this->parameters
$name = 'param';
$default = null;

// code before change
if (isset($parameters[$name]) && ($parameters[$name] === '0' || !empty($parameters[$name]))) {
	var_dump($parameters[$name]);
}

var_dump($default);
```

The above results into null and the below results into false

```
<?php

// simulating call to getParameter(string $name, string $default = null)
$parameters = [ 'param' => false ]; // simulating $this->parameters
$name = 'param';
$default = null;

// code after change
$value = $parameters[$name] ?? '';
var_dump($value !== '' ? $value : $default);
```

The code after the change also will never return the default value as the parameter is set to false by default

mjauvin:
@marcbrouwertue you're right, this created a behavior change, this probably should not have been merged as is since it's basically a breaking change.

LukeTowers:
@marcbrouwertue please submit a separate issue with details so we can track it separately

PR #4909, opened 2020-01-23
Add info about adding custom rules to .htaccess

ayumi-cloud:
As per conversion with @LukeTowers 

It's important to place custom rules in `.htaccess` **before** the last code line and not **after** as they will not work. Below is proof of concept:

### After example

image

Above: Not working returns a `404` error code.

### Before example

image

Above: Working returns a `403` error code.

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -52,6 +52,7 @@
 
     ##
     ## Standard routes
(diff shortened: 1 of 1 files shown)


LukeTowers:
@ayumi-cloud please resubmit to the develop branch instead.

PR #4912, opened 2020-01-27
Add missing backend translations for SK language

Samuell1:

### DIFF
--- a/modules/backend/lang/sk/lang.php
+++ b/modules/backend/lang/sk/lang.php
@@ -2,7 +2,8 @@
 
 return [
     'auth' => [
-        'title' => 'Administrácia'
+        'title' => 'Administrácia',
+        'invalid_login' => 'Údaje, ktoré ste zadali sú nesprávne. Prosím skontrolujte údaje a skúste to znova.'
     ],
     'field' => [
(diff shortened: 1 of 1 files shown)


modules/backend/lang/sk/lang.php
```diff
@@ -15,6 +16,11 @@
     ],
     'page' => [
         'untitled' => 'Bez názvu',
+        '404' => [
+            'label'     => 'Stránka nenájdená',
+            'help'      => "Hľadali sme a hľadali, ale požadovanú adresu URL jednoducho nebolo možné nájsť. Možno ste hľadali niečo iné?",
+            'back_link' => 'Späť na predchodzú stránku',
```
bennothommo: @Samuell1 Would you be able to remove the extra spaces before the arrow indicators? The rest of the document seems to be using the `space arrow space` format.

Samuell1: I can but this was copied from en lang.php file

LukeTowers: I'm fine with the aligned => format, eventually I'd like it everywhere in the lang files

Samuell1: Anything needed? @LukeTowers

PR #4913, opened 2020-01-28
Add variable helpers for ComponentBase

Samuell1:
Using this helpers for a long time now and i was thinking maybe it will be good to make it as PR for core. What you think?

###  Usage of methods:

* `setVar` - if we want to set component variable and page variable
* `setPageVar` - or if we want to set only page variable

```php
class ProductsList extends ComponentBase
{
    public $myVariable;

    public function init()
    {
        $this->setVar('myVariable', 'value');
        $this->setPageVar('myPageVariable', 'value2');
    }
}
```

### DIFF
--- a/modules/cms/classes/ComponentBase.php
+++ b/modules/cms/classes/ComponentBase.php
@@ -264,6 +264,28 @@ abstract class ComponentBase extends Extendable
     }
 
     /**
+     * Sets an variable for component and page
+     * @param string $name Variable name
+     * @param mixed $value Value of variable
+     * @return mixed
+     */
+    protected function setVar($name, $value)
+    {
+        return $this->$name = $this->page[$name] = $value;
+    }
+
+    /**
+     * Sets an variable for page
+     * @param string $name Variable name
+     * @param mixed $value Value of variable
+     * @return mixed
+     */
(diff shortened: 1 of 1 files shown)


bennothommo:
@Samuell1 I'm not sure I can see the applicability of this, sorry. The `$this->page` separation is deliberate - it's so that the page variable scope is not polluted by component variables. The `setPageVar` method seems redundant when you can just use `$this->page['variable'] = 'value';`.

Samuell1:
@bennothommo Mostly is to make code look at lot better and i forgot to say this allows autocomplete. But still this is just idea that can maybe help new people and simplify a lot of things.

jacobdekeizer:
@Samuell1 You can write `{{ __SELF__.myVariable }}` to access a public variable in your plugin component partial.

LukeTowers:
@Samuell1 sorry, going to have to agree with @bennothommo on this one. I don't really see the value of those as helper functions. Feel free to release a trait that does that though that other people can utilize if they so desire!

Samuell1:
Its here if anybody wants: https://octobertricks.com/tricks/component-helper-methods-to-set-variables

PR #4917, opened 2020-01-28
Allow middle mouse click to open list row in new window

Samuell1:
This allows to use middle click to open list row in new window. 

If anybody knows how to make it to stop focusing new window, like middlemouse natively does on links. Ideas will be appreciated.

### DIFF
--- a/modules/system/assets/ui/js/list.rowlink.js
+++ b/modules/system/assets/ui/js/list.rowlink.js
@@ -62,7 +62,7 @@
                 }
             }
 
-            $(this).not('.' + options.excludeClass).find('td').not('.' + options.excludeClass).click( function (e) {
+            $(this).not('.' + options.excludeClass).find('td').not('.' + options.excludeClass).click(function (e) {
                 handleClick(e)
             }).mousedown(function (e) {
                 if (e.which == 2) {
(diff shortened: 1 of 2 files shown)


LukeTowers:
@w20k any ideas on how to prevent the focusing on the new window?

w20k:
@LukeTowers should be something similar. One of the options definitely should work:
```
window.open(href).focus();
//Or
var win = window.open(href);
win.blur();
win.opener.focus();
window.focus();
```

bennothommo:
@w20k @Samuell1 I did a quick bit of research into this - apparently it's frowned upon by browsers because it's a behaviour that is used to sneak in unwanted ads, and other nasty stuff such as bitcoin miners.

If it's something that is really required, we'd have to look into replacing the click functionality of the Lists widget with some form of hyperlinking (if the target is a URL, not JavaScript).

Samuell1:
I think for now this is fine. But if there is way to use anchor links on table columns then it will be cool to have it.

modules/system/assets/ui/js/list.rowlink.js
```diff
@@ -62,8 +62,12 @@
                 }
             }
 
-            $(this).not('.' + options.excludeClass).find('td').not('.' + options.excludeClass).click(function(e) {
+            $(this).not('.' + options.excludeClass).find('td').not('.' + options.excludeClass).click( function (e) {
```
LukeTowers:
```suggestion
            $(this).not('.' + options.excludeClass).find('td').not('.' + options.excludeClass).click(function (e) {
```

PR #4918, opened 2020-01-29
Fix pagination reset if search input is filled

### LINKED ISSUE #4914: List pagination is reseted to first page if search is filled on refresh
- OctoberCMS Build: 458
- PHP Version: 7.3
- Database Engine: MySQL

### Description:
List pagination is set to first page if search is already filled and page is refreshed. Looks like its because this line https://github.com/octobercms/october/blob/develop/modules/backend/widgets/Lists.php#L308
Shouldnt this reset pagination only if you search for first time?

### Steps To Reproduce:
Just fill something in search then move to second page and then refresh.

Samuell1:
Fixes #4914

### DIFF
--- a/modules/backend/widgets/Lists.php
+++ b/modules/backend/widgets/Lists.php
@@ -640,7 +640,7 @@ class Lists extends WidgetBase
     protected function getCurrentPageNumber($query)
     {
         $currentPageNumber = $this->currentPageNumber;
-        if (empty($this->currentPageNumber)) {
+        if (empty($currentPageNumber)) {
             $currentPageNumber = $this->getSession('lastVisitedPage');
         }
 
--- a/modules/backend/widgets/Lists.php
+++ b/modules/backend/widgets/Lists.php
@@ -1403,25 +1403,19 @@ class Lists extends WidgetBase
     //
 
     /**
-     * Applies a search term to the list results, same as setSearchTerm method does
-     * but it will reset pagination to page 1 after new submited search term.
-     * @param string $term
-     */
-    public function submitSearchTerm($term)
-    {
-        $this->setSearchTerm($term);
-        $this->currentPageNumber = 1;
-    }
-
-    /**
      * Applies a search term to the list results, searching will disable tree
      * view if a value is supplied.
      * @param string $term
+     * @param boolean $resetPagination
      */
-    public function setSearchTerm($term)
+    public function setSearchTerm($term, $resetPagination = false)
     {
         if (!empty($term)) {
             $this->showTree = false;
+        }
+
+        if ($resetPagination) {
+            $this->currentPageNumber = 1;
         }
 
         $this->searchTerm = $term;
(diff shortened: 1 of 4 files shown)


modules/backend/widgets/Lists.php
```diff
@@ -1415,6 +1415,7 @@ public function setSearchTerm($term)
             $this->showTree = false;
         }
 
+        $this->currentPageNumber = 1;
```
bennothommo: @Samuell1 does setting this to page 1 here also change `lastVisitedPage` in the session to page 1 as well? If it doesn't, it'll need to, otherwise a refresh would take them to a potentially different page.

Samuell1: Found out that pagination needs to be totally reworked somehow because this PR still doesnt fix that issue. I think you are true we need to save lastVisitedPage on every call where pagination chages because now its only saved when prepareVars method is called.

Samuell1: @bennothommo Found a fix! PR is now updated.

modules/backend/widgets/Lists.php
```diff
@@ -640,9 +640,7 @@ protected function getRecords()
     protected function getCurrentPageNumber($query)
     {
         $currentPageNumber = $this->currentPageNumber;
-
-        if (!$currentPageNumber && empty($this->searchTerm)) {
-            // Restore the last visited page from the session if available.
+        if (empty($this->currentPageNumber)) {
```
LukeTowers:
```suggestion
        if (empty($currentPageNumber)) {
```

modules/backend/widgets/Lists.php
```diff
@@ -1404,6 +1402,17 @@ public function addFilter(callable $filter)
     // Searching
     //
 
+    /**
+     * Applies a search term to the list results, same as setSearchTerm method does
+     * but it will reset pagination to page 1 after new submited search term.
+     * @param string $term
+     */
+    public function submitSearchTerm($term)
```
LukeTowers: I would rather that `setSearchTerm($term)` becomes `setSearchTerm($term, $resetPagination = false)` instead of adding an entirely new method that does basically the same thing.

Samuell1: Yep, thats better idea!

modules/backend/widgets/lists/partials/_list_pagination_simple.htm
```diff
@@ -29,7 +29,8 @@
         <select
             disabled
             name="page"
-            class="form-control input-sm custom-select select-no-search">
+            class="form-control input-sm custom-select select-no-search"
+            autocomplete="off">
```
LukeTowers: Why does this need to be added?

Samuell1: Its to fix long standing issue when browser caches input values after you refresh page.

Samuell1: Same issue is with checkboxes in list but thats for separate PR

PR #4919, opened 2020-01-29
[Laravel 6] Add support for Laravel Dusk tests

bennothommo:
Forms part of work done in https://github.com/octobercms/october/pull/4893. Replaces https://github.com/octobercms/october/pull/3051.

This implements support for [Laravel Dusk](https://laravel.com/docs/6.x/dusk), a more friendly way for doing comprehensive browser-based testing.

This PR significantly refactors the testing base to allow these tests to be included for both core, as well as a part of plugins.


tests/Browser/Pages/Backend/Dashboard.php
```diff
@@ -0,0 +1,44 @@
...
+
+    /**
+     * Get the global element shortcuts for the site.
+     *
+     * @return array
+     */
+    public function elements()
+    {
+        return [
+        ];
```
LukeTowers: This is formatted weird

tests/Browser/Pages/Page.php
```diff
@@ -0,0 +1,17 @@
...
+{
+    /**
+     * Get the global element shortcuts for the site.
+     *
+     * @return array
+     */
+    public static function siteElements()
+    {
+        return [
+        ];
```
LukeTowers: Formatted weird

tests/Browser/console/.gitignore
```diff
@@ -0,0 +1,2 @@
+*
```
LukeTowers: What's this folder used for?

tests/BrowserTestCase.php
```diff
@@ -0,0 +1,121 @@
...
+    // OctoberCMS Helpers
+    //
+
+    protected function signInToBackend()
+    {
+        $this->browse(function ($browser) {
+            $browser->visit('backend')
+                ->assertTitle('Administration Area')
+                ->type("login", 'admin')
+                ->type("password", 'admin')
```
LukeTowers: Does this take into account the new password generation scheme? Wouldn't the setup portion need to include a call to `october:passwd admin admin1234`?

tests/Browser/console/.gitignore
bennothommo: @LukeTowers Dusk will print out error logs in that folder when running Dusk tests. It's similar to the screenshots subfolder in the same folder.

tests/BrowserTestCase.php
bennothommo: @LukeTowers I'm probably going to be removing this method. Dusk supposedly allows you to pre-auth a user when running Dusk tests, so the login helper is redundant.

tests/Browser/Pages/Page.php
bennothommo: Penalized for Laravel's mistakes ;P

tests/BrowserTestCase.php
bennothommo: I've converted this to a macro which can be used with the Browser helper in these tests. It will allow environment variables to override the user login information, so that should cover your case.

LukeTowers:
@bennothommo 
> If you are manually registering Dusk's service provider, you should never register it in your production environment, as doing so could lead to arbitrary users being able to authenticate with your application.

How can we protect users from this?

bennothommo:
@LukeTowers There's a couple of things in play already to prevent this from happening:

- We're using automatic registration of the service provider, via using `autoload-dev` in `composer.json`.
- The tests run in a different environment to the live site by replacing `.env` with `.env.dusk` on the fly, as well as the configuration in `config/dusk/`. The database runs in SQLite for the tests by default, and requires manual intervention in order to use any other database.

It could be possible that with that message, they're referring to the [Dusk routes](https://github.com/laravel/dusk/blob/5.0/src/DuskServiceProvider.php#L16-L32) they provide in their service provider, which can basically let you log in as any user. They don't actually work with October, so we could potentially extend the Service Provider and just remove those routes.

LukeTowers:
@bennothommo I would feel most comfortable if we removed those routes entirely.

bennothommo:
@LukeTowers instead of extending the service provider, I've just disabled auto-discovery for Laravel Dusk, and am using the `system` module's service provider to load the console commands as necessary. Those routes should no longer be loaded.

bennothommo:
This is ready to review and merge into the main L6 branch. I'll add the tests/README.md changes to the October docs once reviewed.

LukeTowers:
LGTM!

LukeTowers:
@BennoThommo actually have we dealt with re: unit tests related to https://github.com/rainlab/translate-plugin/pull/509 & https://github.com/rainlab/translate-plugin/issues/546
`DatabaseServiceProvider` does the following when registering:
```php
Model::clearBootedModels();
Model::clearExtendedClasses();
Model::flushDuplicateCache();
```
Should we also add `Model::flushEventListeners();` (used in https://github.com/octobercms/october/blob/master/tests/unit/cms/classes/CmsCompoundObjectTest.php for Halcyon, but also applicable to Eloquent) and then also implement a `Model::flushDuplicateCache()` in Halcyon models and add those four lines to `HalcyonServiceProvider`?

PR #4920, opened 2020-01-29
Allow DI in console command constructor

martendej:
When registering a console command in your Plugin.php by using the registerConsoleCommand method it is currently not allowed to have any parameters the constructor of your command. This pull request addresses that issue by simply binding the class name in the container instead of an object that is constructed without parameters.

### DIFF
--- a/modules/system/classes/PluginBase.php
+++ b/modules/system/classes/PluginBase.php
@@ -262,11 +262,7 @@ class PluginBase extends ServiceProviderBase
     public function registerConsoleCommand($key, $class)
     {
         $key = 'command.'.$key;
-
-        $this->app->singleton($key, function ($app) use ($class) {
-            return new $class;
-        });
-
+        $this->app->singleton($key, $class);
         $this->commands($key);
     }
 
--- a/modules/system/classes/PluginBase.php
+++ b/modules/system/classes/PluginBase.php
@@ -256,13 +256,13 @@ class PluginBase extends ServiceProviderBase
      * Registers a new console (artisan) command
      *
      * @param string $key The command name
-     * @param string $class The command class
+     * @param string|\Closure $command The command class or closure
      * @return void
      */
-    public function registerConsoleCommand($key, $class)
+    public function registerConsoleCommand($key, $command)
     {
         $key = 'command.'.$key;
-        $this->app->singleton($key, $class);
+        $this->app->singleton($key, $command);
         $this->commands($key);
     }
 


bennothommo:
@martendej Is this going to be backwards compatible with commands already implemented in plugins out there?

You'll also need to adjust the docblock above to indicate that `$class` would now accept a closure or a string.

martendej:
@bennothommo Yes, this should be backwards compatible. I have updated the method docblock.

modules/system/classes/PluginBase.php
```diff
@@ -256,17 +256,13 @@ public function registerMailPartials()
...
-    public function registerConsoleCommand($key, $class)
+    public function registerConsoleCommand($key, $command)
     {
         $key = 'command.'.$key;
-
-        $this->app->singleton($key, function ($app) use ($class) {
-            return new $class;
-        });
-
+        $this->app->singleton($key, $command);
```
bennothommo: @martendej Would you mind changing this to the following, it will be more backwards-compatible:
```suggestion
        if (is_string($command)) {
            $closure = function ($app) use ($command) {
                return new $command;
            };
        } else {
            $closure = $command;
        }

        $this->app->singleton($key, $closure);
```

bennothommo (changes requested):

modules/system/classes/PluginBase.php
bennothommo: @martendej Ah never mind, I see where this all goes. Please disregard my requested change.

bennothommo (approved):

martendej:
@bennothommo @LukeTowers Can anyone of you test this? Furthermore, what is the reason that a key has to be supplied when registering a console command and why are they bound as singletons?

bennothommo:
@martendej I've tested it a little previously, but we're currently focusing the efforts on getting the Laravel 6 upgrade out finally.

The key is used because I believe the plugin architecture doesn't use the command definition in the Command class itself, and instead relies on this key to define the name of the command. It's been kept this way to maintain backwards compatibility.

As for each command being a singleton, is there any particular scenario where it shouldn't be?

martendej:
Well, correct me if im wrong, but wouldn't it be so that if you fire a command in your terminal, and within that command you would fire another command of the same type that would run immediately, the state of those commands would be shared amongst eachother? Albeit very theoretical, I don't think that is wishful behaviour, especially not if you are storing values in properties?

bennothommo:
@martendej I'd need to see a scenario like that to fully grasp it, but I would think that if state needs to be separate between subsequent runs of the same command, you'd use something else to track the state.

A command is just an endpoint for the CLI, and should have no state on its own.

LukeTowers:
@martendej what's the status on this? Are you still wanting it merged?

daftspunk:
Apologies for the time it took to review this. LGTM and thanks for using October CMS 🙏

PR #4921, opened 2020-01-29
Micro performance improvement of Lists widget

Rike-cz:
This issue was discussed on Slack with @bennothommo and @mjauvin.

If I have `showSetup == false` in **Lists widget**, it's pointless to make SQL questions for getting user preferences, because user cannot change them. In this case default setting from YAML is used.

### DIFF
--- a/modules/backend/widgets/Lists.php
+++ b/modules/backend/widgets/Lists.php
@@ -210,7 +210,9 @@ class Lists extends WidgetBase
         /*
          * Configure the list widget
          */
-        $this->recordsPerPage = $this->getUserPreference('per_page', $this->recordsPerPage);
+        if ($this->showSetup) {
+            $this->recordsPerPage = $this->getUserPreference('per_page', $this->recordsPerPage);
+        }
 
         if ($this->showPagination == 'auto') {
             $this->showPagination = $this->recordsPerPage && $this->recordsPerPage > 0;
@@ -728,7 +730,7 @@ class Lists extends WidgetBase
         /*
          * Supplied column list
          */
-        if ($this->columnOverride === null) {
+        if ($this->showSetup && $this->columnOverride === null) {
             $this->columnOverride = $this->getUserPreference('visible', null);
         }
 


mjauvin:
Looks good to me.

LukeTowers:
Hmm, on further thought this actually places a limitation on the configurability of the lists. There are other ways for user preferences to be set; specifically any plugin could choose to set any preference under any condition. If we make this change this would prevent other sources from changing the user preferences for lists and having an effect.

Thoughts @bennothommo @daftspunk?

mjauvin:
@LukeTowers good point. Out of curiosity, how would a plugin change those user preferences ?

LukeTowers:
@mjauvin by interacting with the UserPreference model directly

bennothommo:
@LukeTowers Are there any examples of plugins that do that?

I feel that given that UserPreferences are scoped and to have a plugin manipulating the preferences externally seems a bit against the point of that function, as well as the idea that the primary way of manipulating Lists is through events (which wouldn't be affected by this change), this might be an acceptable break.

LukeTowers:
@bennothommo Not as far as I'm aware right now. So long as there are events that fire that enable the same configuration flexibility I'm fine with this change then.

mjauvin:
@bennothommo how would you change this using an event?

mjauvin:
Ok, this simple example works:
```
        Event::listen('backend.list.extendQuery', function ($listWidget, $query) {
            if ($listWidget->showPagination == true) {
                $listWidget->recordsPerPage = 5;
            }
        });
```

bennothommo:
@Rike-cz @mjauvin @LukeTowers I have tested this - it works, but it's literally saving one query which, in my tests, took 210 microseconds to complete. So, it's a pretty negligible difference.

With that being said, one thing to consider with this change is that user preferences are no longer applied if `showSetup` is false. This might actually make more sense to have it this way - it means that should someone have preferences on a list widget that at some point in the future has the setup modal disabled, they are not stuck with preferences that they can no longer change or reset to default.

On this basis - I'd be happy to merge this, unless anyone has any objections?

LukeTowers:
@bennothommo nope, LGTM

Issue #4924, opened 2020-02-02
Safe mode enabled is stuck on


ayumi-cloud:
October v464

Browser: Latest Firefox.

See screenshot:

![image](https://user-images.githubusercontent.com/57409060/73613403-3cedbb00-45ed-11ea-827b-8e099ed80e0c.png)

Above: config settings is turned on yet cms is saying it's on and blocking total access to the php section!

Also the message is not the greatest, could it not mention to edit the `config file: cms/enableSafeMode`

ayumi-cloud:
Closed issue forgot I had 0600 file permission turned on with config files.

PR #4926, opened 2020-02-02
Improve error message for safe mode enabled

ayumi-cloud:
As per my comment here: https://github.com/octobercms/october/issues/4924

> Also the message is not the greatest, could it not mention to edit the config file: cms/enableSafeMode

Picture a newbie user using October CMS for the first time, my error message does the following:

1. Tell them the exact issue.
2. Tell them the exact solution.

### DIFF
--- a/modules/cms/lang/en/lang.php
+++ b/modules/cms/lang/en/lang.php
@@ -11,7 +11,7 @@ return [
         'error_deleting' => "Error deleting the template file ':name'. Please check write permissions.",
         'delete_success' => 'Templates deleted: :count.',
         'file_name_required' => 'The File Name field is required.',
-        'safe_mode_enabled' => 'Safe mode is currently enabled (editing the PHP section is disabled) to turn this off, edit the config file: \'cms/enableSafeMode\'.'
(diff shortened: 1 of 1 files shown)


modules/cms/lang/en/lang.php
```diff
@@ -11,7 +11,7 @@
         'error_deleting' => "Error deleting the template file ':name'. Please check write permissions.",
         'delete_success' => 'Templates deleted: :count.',
         'file_name_required' => 'The File Name field is required.',
-        'safe_mode_enabled' => 'Safe mode is currently enabled. Editing the PHP code of CMS templates is disabled.'
+        'safe_mode_enabled' => 'Safe mode is currently enabled (editing the PHP section is disabled) to turn this off, edit the config file: \'cms/enableSafeMode\'.'
```
LukeTowers:
```suggestion
        'safe_mode_enabled' => 'Safe mode is currently enabled. Editing the PHP code of CMS templates is disabled. To disable safe mode, set the `cms.enableSafeMode` configuration value to `false`.
```

PR #4929, opened 2020-02-05
Modify menu structure to objects

Klaasie:
### Introduction
This PR simplifies some parts of the navigation manager items management.

### Issue
Recently I've ran into the issue where I wanted to modify the icon of an existing menu item. This was more complicated that I had expected and ended up copying the current definition of the menu item, removing it and then adding it again with my own icon.

### Changes
By creating 2 classes, `MainMenuItem` and `SideMenuItem`, that will be instantiated while registering the menu items, it  will be much clearer as to what objects we are dealing with and what information they hold (opposed to casting an array to object).

I've modified the `NavigationManager` to use the 2 new classes. All properties of the new classes have been made public to ensure backwards compatibility. The `NavigationManager` now has a new method `getMainMenuItem` to retrieve a `MainMenuItem` based on owner and code. This will simplify retrieving the proper main menu item to modify.

In addition to those changes I've cleaned up the `NavigationManager` class a bit. Adding missing DocBlocks and typehinting wherever possible.

### Example
So with these changed I could now do the following to remove the default svg of the dashboard menu item and have it fall back to the font-icon one.
```
\Event::listen('backend.menu.extendItems', static function (NavigationManager $navigationManager) {
    $item = $navigationManager->getMainMenuItem('october.backend', 'dashboard');
    $item->setIconSvg(null);
});
```

### DIFF
--- a/modules/backend/classes/MainMenuItem.php
+++ b/modules/backend/classes/MainMenuItem.php
@@ -1,0 +1,293 @@
+<?php namespace Backend\Classes;
+
+use October\Rain\Exception\SystemException;
+
+/**
+ * Class MainMenuItem
+ *
+ * @package Backend\Classes
+ */
+class MainMenuItem
+{
+    /**
+     * @var string
+     */
+    public $code;
+
+    /**
+     * @var string
+     */
+    public $owner;
+
+    /**
+     * @var string
+     */
+    public $label;
+
+    /**
+     * @var null|string
+     */
+    public $icon;
+
+    /**
+     * @var null|string
+     */
+    public $iconSvg;
+
+    /**
+     * @var mixed
+     */
+    public $counter;
+
+    /**
+     * @var null|string
+     */
+    public $counterLabel;
+
+    /**
+     * @var string
+     */
+    public $url;
+
+    /**
+     * @var array
+     */
+    public $permissions = [];
+
+    /**
+     * @var int
+     */
+    public $order = 500;
+
+    /**
+     * @var SideMenuItem[]
+     */
+    public $sideMenu = [];
+
+    /**
+     * @return string
+     */
+    public function getCode(): string
+    {
+        return $this->code;
+    }
+
+    /**
+     * @param string $code
+     */
+    public function setCode(string $code)
+    {
+        $this->code = $code;
+    }
+
+    /**
+     * @return string
+     */
+    public function getOwner(): string
+    {
+        return $this->owner;
+    }
+
+    /**
+     * @param string $owner
+     */
+    public function setOwner(string $owner)
+    {
+        $this->owner = $owner;
+    }
+
+    /**
+     * @return string
+     */
+    public function getLabel(): string
+    {
+        return $this->label;
+    }
+
+    /**
+     * @param string $label
+     */
+    public function setLabel(string $label)
+    {
+        $this->label = $label;
+    }
+
+    /**
+     * @return string|null
+     */
+    public function getIcon()
+    {
+        return $this->icon;
+    }
+
+    /**
+     * @param string|null $icon
+     */
+    public function setIcon($icon)
+    {
+        $this->icon = $icon;
+    }
+
+    /**
+     * @return string|null
+     */
+    public function getIconSvg()
+    {
+        return $this->iconSvg;
+    }
+
+    /**
+     * @param string|null $iconSvg
+     */
+    public function setIconSvg($iconSvg)
+    {
+        $this->iconSvg = $iconSvg;
+    }
+
+    /**
+     * @return mixed
+     */
+    public function getCounter()
+    {
+        return $this->counter;
+    }
+
+    /**
+     * @param mixed $counter
+     */
+    public function setCounter($counter)
+    {
+        $this->counter = $counter;
+    }
+
+    /**
+     * @return string|null
+     */
+    public function getCounterLabel()
+    {
+        return $this->counterLabel;
+    }
+
+    /**
+     * @param string|null $counterLabel
+     */
+    public function setCounterLabel($counterLabel)
+    {
+        $this->counterLabel = $counterLabel;
+    }
+
+    /**
+     * @return string
+     */
+    public function getUrl(): string
+    {
+        return $this->url;
+    }
+
+    /**
+     * @param string $url
+     */
+    public function setUrl(string $url)
+    {
+        $this->url = $url;
+    }
+
+    /**
+     * @return array
+     */
+    public function getPermissions(): array
+    {
+        return $this->permissions;
+    }
+
+    /**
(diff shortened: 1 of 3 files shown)


LukeTowers:
@Klaasie none of the getters / setters that just interface directly 1:1 with the public properties need to exist. Only keep the ones that "massage" the data a bit first for convenience.

modules/backend/classes/NavigationManager.php
```diff
@@ -16,15 +18,15 @@
  */
 class NavigationManager
 {
-    use \October\Rain\Support\Traits\Singleton;
```
LukeTowers: It's October's style to use the full namespace for model traits instead of importing it at the top, so please revert this.

modules/backend/classes/NavigationManager.php
```diff
@@ -1,7 +1,9 @@
 <?php namespace Backend\Classes;
 
+use Backend\Models\User;
```
LukeTowers: Why is this being included?

modules/backend/classes/NavigationManager.php
Klaasie: I've reverted this in a new commit, guess this is something my IDE does by itself.

modules/backend/classes/NavigationManager.php
Klaasie: I've changed the documentation for the method `filterItemPermissions` to properly reference the User class. Then my IDE imported it. I've removed the import and modified the documentation to reference the full namespace.

Klaasie:
@LukeTowers I've removed the getters and setters and their implementation in the `NavigationManager` per your suggestion.

modules/backend/classes/MainMenuItem.php
```diff
@@ -0,0 +1,125 @@
...
+        unset($this->sideMenu[$code]);
+    }
+
+    /**
+     * @param array $data
+     * @return static
+     */
+    public static function createFromArray(array $data): self
+    {
+        $instance = new self();
```
LukeTowers:
```suggestion
        $instance = new static();
```

modules/backend/classes/NavigationManager.php
```diff
@@ -24,7 +24,7 @@ class NavigationManager
     protected $callbacks = [];
 
     /**
-     * @var array List of registered items.
+     * @var MainMenuItem[] List of registered items.
```
LukeTowers: I've never seen this syntax before, it's still just an array, the difference is that it's an array of MainMenuItems instead of an array of arrays.

modules/backend/classes/SideMenuItem.php
```diff
@@ -0,0 +1,117 @@
...
+        unset($this->permissions[$permission]);
+    }
+
+    /**
+     * @param array $data
+     * @return static
+     */
+    public static function createFromArray(array $data): self
+    {
+        $instance = new self();
```
LukeTowers:
```suggestion
        $instance = new static();
```

modules/backend/classes/NavigationManager.php
Klaasie: This syntax is a way to let your editor know which type to expect from an array. It will allow autocompletion and functions as "jump to definition". In this particular case I think it's useful in future debugging or code modifications.
See https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc.md#arrays for more information.

modules/backend/classes/MainMenuItem.php
```diff
@@ -0,0 +1,125 @@
...
+    public function removeSideMenuItem(string $code)
+    {
+        unset($this->sideMenu[$code]);
+    }
+
+    /**
+     * @param array $data
+     * @return static
+     */
+    public static function createFromArray(array $data): self
```
bennothommo: @Klaasie While I think `self` is technically valid here as a return type as this returns an instance of the same class, the correct return type is specifically `static`, however, that won't be coming until PHP 8. It's best to remove this for now. (The docblock is fine to keep though).

Refs:
- https://stackoverflow.com/questions/39068983/php-7-interfaces-return-type-hinting-and-self
- https://wiki.php.net/rfc/static_return_type
```suggestion
    public static function createFromArray(array $data)
```

modules/backend/classes/NavigationManager.php
```diff
@@ -118,7 +119,7 @@ protected function loadItems()
         /*
          * Sort menu items
          */
-        uasort($this->items, function ($a, $b) {
+        uasort($this->items, static function ($a, $b) {
```
bennothommo: @Klaasie is there a particular reason to change this to a static function?

Klaasie: @bennothommo The function in this case doesn't require anything from the object scope (`$this`). So declaring this as a static method is a small optimization. (Actually, it's considered a micro optimization)

bennothommo: Thanks for the info - I did not know that, but it makes sense.

daftspunk:
Awesome work all round. Thanks!

modules/backend/classes/MainMenuItem.php
```diff
@@ -0,0 +1,125 @@
...
+        $instance->owner = $data['owner'];
+        $instance->label = $data['label'];
+        $instance->url = $data['url'];
+        $instance->icon = $data['icon'] ?? null;
+        $instance->iconSvg = $data['iconSvg'] ?? null;
+        $instance->counter = $data['counter'] ?? null;
+        $instance->counterLabel = $data['counterLabel'] ?? null;
+        $instance->permissions = $data['permissions'] ?? $instance->permissions;
+        $instance->order = $data['order'] ?? $instance->order;
+        return $instance;
```
LukeTowers, line 123: @bennothommo here we'd just need to merge $data with $defaultConfig = ['permissions' => [], 'order' => 500] and then loop over the array and set the properties based on the provided keys.

LukeTowers: Then either remove those explicitly defined properties or add a one liner inline comment explaining them for the API docs. Otherwise they serve no purpose.

modules/backend/classes/NavigationManager.php
```diff
@@ -256,20 +258,39 @@ public function addMainMenuItem($owner, $code, array $definition)
...
             'code'  => $code,
             'owner' => $owner
-        ]));
+        ]);
+
+        $this->items[$itemKey] = MainMenuItem::createFromArray($item);
+
+        if (array_key_exists('sideMenu', $item)) {
+            $this->addSideMenuItems($owner, $code, $item['sideMenu']);
+        }
```
LukeTowers, line 270: Should probably  also move this into the MainMenuItem class @bennothommo

PR #4933, opened 2020-02-09
implement image/file attachMany field types to theme data

mjauvin:
If a field type is `image-multi` or `file-multi` in the theme form fields, it will be added to $attachMany[] relation instead of $attachOne[]

### DIFF
--- a/modules/cms/models/ThemeData.php
+++ b/modules/cms/models/ThemeData.php
@@ -46,6 +46,7 @@ class ThemeData extends Model
      * @var array Relations
      */
     public $attachOne = [];
+    public $attachMany = [];
 
     /**
      * @var ThemeData Cached array of objects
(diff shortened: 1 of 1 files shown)


modules/cms/models/ThemeData.php
```diff
@@ -125,7 +126,11 @@ public function afterFetch()
                 $this->jsonable[] = $id;
             }
             elseif ($field['type'] === 'fileupload') {
-                $this->attachOne[$id] = File::class;
+                if (str_contains($field['mode'], '-multi')) {
```
bennothommo: @mjauvin Might be better to make this more explicit:
```suggestion
                if (in_array($field['mode'], ['file-multi', 'image-multi'])) {
```

LukeTowers:
@mjauvin I'm not sure I'm a fan of this, `image` and `file` are the only supported documented modes available for the fileupload formwidget, I'm not really a fan of adding special `image-multi` & `file-multi` modes that are only able to be used by theme data fields. Isn't this something that a plugin could handle?

mjauvin:
@LukeTowers you're right, the "-multi" suffix is something added in the FileUpload formwidget's `getDisplayMode()` method.

I agree a plugin could fix this, but if you're creating a theme that needs an $attachMany form field, should you really force a plugin to be installed just for this?

mjauvin:
I guess that's what the "require:" field would be used for in theme.yaml... doh! :)

PR #4935, opened 2020-02-11
Add ability to force a specific datasource for AutoDatasource operations

### LINKED ISSUE #4887: Theme:sync --target=database not working
- OctoberCMS Build: 458
- PHP Version: 7.2
- Database Engine: MySQL 5.7.26
- Plugins Installed: 
Indikator.Backend	1.6.11
RainLab.Pages	1.3.2
RainLab.Builder	1.0.26

### Description:
I save the templates in database (databaseTemplates is true in config). This works fine, but when I changed some layout files on filesystem and sync this files to database, this doesn't work as expected.

### Steps To Reproduce:
 For example I use the command:
`artisan theme:sync --target=database --force --paths=layouts/about.htm`

But the changes are not in database, although the message shows `The theme mytheme has been successfully synced from the filesystem to the database.`

I have debug the process and I think the source is not the correct one. When I change some content in database and look at the variable `$content` in [modules/cms/classes/AutoDatasource.php#L156](https://github.com/octobercms/october/blob/882ca6019c9c1b124e6e3f052b6aec02b6fa826b/modules/cms/classes/AutoDatasource.php#L156) , I see the content from database and not from filesystem.

What's going wrong?

bennothommo:
Fixes #4887.

When retrieiving templates, AutoDatasource will use the first datasource it finds with the template 
available. This is fine most of the time, but when running "theme:sync", we need an ability to force the template to be loaded from a specific source, regardless of priority or active datasource.

In the case of #4887, the reporter was trying to synchronise from the filesystem to the database. The filesystem is lower in priority than the database, so when it was loading the Halcyon model for the layout, AutoDatasource was returning the database version as it was the higher priority - even though the `activeDatasourceKey` had been set to `filesystem`. This meant that their changes from the filesystem were not being sync'd to the database.

This PR adds a mechanism to force a specific datasource to be used in a closure, so that any operation run in it is sure to be using the selected datasource.

### DIFF
--- a/modules/cms/classes/AutoDatasource.php
+++ b/modules/cms/classes/AutoDatasource.php
@@ -2,6 +2,7 @@
 
 use Cache;
 use Exception;
+use October\Rain\Exception\ApplicationException;
 use October\Rain\Halcyon\Model;
 use October\Rain\Halcyon\Processors\Processor;
 use October\Rain\Halcyon\Datasource\Datasource;
@@ -35,6 +36,11 @@ class AutoDatasource extends Datasource implements DatasourceInterface
      * @var string The key for the datasource to perform CRUD operations on
      */
     public $activeDatasourceKey = '';
+
+    /**
+     * @var string Force operations on a selected datasource
+     */
+    protected $forcedDatasourceKey = null;
 
     /**
      * Create a new datasource instance.
@@ -192,6 +198,10 @@ class AutoDatasource extends Datasource implements DatasourceInterface
      */
     protected function getDatasourceForPath(string $path)
     {
+        if (isset($this->forcedDatasourceKey)) {
+            return $this->datasources[$this->forcedDatasourceKey];
+        }
+
         // Default to the last datasource provided
         $datasourceIndex = count($this->datasources) - 1;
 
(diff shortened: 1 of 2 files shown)


LukeTowers:
@bennothommo The point of the activeDatasource property is to achieve exactly this, in my mind we should just fix it so it works correctly rather than implement a whole other method for specifying the active datasource. Thoughts?

bennothommo:
@LukeTowers I did think that, but `activeDatasourceKey` seems to only be used for the write and delete operations, the `select` and `selectOne` methods do not use the `activeDatasourceKey` property in any way. They simply traverse the datasources in order of priority and find the first one that can provide the requested template.

There's no way to say "I want to select something from this source and *only* this source" through AutoDatasource currently, and given that `activeDatasourceKey` is [always set](https://github.com/octobercms/october/blob/9c2662f9d6a3d457c24a32a7d29eec312a12e6ee/modules/cms/classes/AutoDatasource.php#L55) on construction, we can't reliably use it to determine whether we want to look at one source or at all sources when running a select. That's essentially why I opted for a separate flag.

LukeTowers:
Closing in favour of #5017, will re-open if that one doesn't pan out.

PR #4940, opened 2020-02-12
Added ability to set active backend tab using a URL hashtag

mrgswift:
**New Feature/Enhancement:** Added ability to use two new URL hashtags to set a tab active during the on render event.  When either of these hashtags are used, they will be automatically removed from the URL as soon as the tab is made active. 

This is useful when you want to perform a manual refresh in a controller but want a specific tab to be active after the refresh.  

**The hashtags are as follows:**

- **#primarytab-[index]** - Add this hashtag to the end of the backend URL to set active tab by numeric index. For example, using the example URL `/backend/acme/example/tabexample/update/1#primarytab-2`, sets the tab with index equal to 2 active by default when loading the tabexample controller in update context. 

- **#tab-[title]** -  Add this hashtag to the end of the backend URL to set active tab by a tab's title.  For example, a tab with the title 'Information' can be set active by default (see example URL below). 
  Example URL: `/backend/acme/example/tabexample/update/1#tab-information` 

  All tab titles must be converted to lowercase when using hashtags and spaces should be replaced by underscores "_" (without quotes).  

  Example URL containing the tab title with a space 'More info':  `/backend/acme/example/tabexample/update/1#tab-more_info`

Only lowercase letters can be used when using either hashtag.  This update does not effect other URL hashtags that do not match the above hashtags.  Hashtags not following the correct convention are ignored, but further script execution by other scripts is not prevented.  Additionally, using these hashtags only effects the parent form widget's tabs.

### DIFF
--- a/modules/system/assets/ui/docs/tab.md
+++ b/modules/system/assets/ui/docs/tab.md
@@ -7,6 +7,7 @@ This plugin is a wrapper for the Twitter Bootstrap Tab component. It provides th
 - Removing tabs with the Close icon, or with triggering an event from inside a tab pane or tab. The removing can be canceled if the confirm.oc.tab event handler returns false.
 - Scrolling tabs if they do not fit the screen
 - Collapsible tabs
+- Backend active tab on render using a URL hashtag (#tab-[title] OR #primarytab-[index])
 
 ### Supported CSS modifiers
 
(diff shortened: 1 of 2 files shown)


LukeTowers:
@mrgswift this seems pretty similar to https://github.com/octobercms/october/pull/4754, how is this different?

mrgswift:
@LukeTowers Apologies, Didn't even see that PR.  That PR looks a lot more elegant and useful than mine.

PR #4954, opened 2020-02-22
Add formGetRedirectUrl method. Fixes #4946

### LINKED ISSUE #4946: Allow custom FormController redirects
There is currently no easy way to grab the saved model and redirect to a URL based on the data inside of it when saving a FormController form.

### Use Case

I hit the 'Save & Preview' button on a blog post create page in admin. It should save the blog post, then redirect the browser to $model->permalink - a dynamically generated URL based on the slug assigned during the save process. 

### Proposal

Rename `FormController.makeRedirect` to `FormController.formMakeRedirect` and change all calls to `$this->controller->formMakeRedirect`. This change would be backwards compatible and allow overriding of the redirect behavior.

Alternatively, we could just add a POST variable to define the redirect URL override. Unfortunately 'redirect' is already being used as a simple boolean so we'd need to come up with another name that fits.

If you're happy with this proposal I can whip up a PR.

### In Depth Issue

When FormController's `create_onSave()` method is called, the model is saved and then 3 calls are made:
```php
$this->controller->formAfterSave($model);
$this->controller->formAfterCreate($model);
if ($redirect = $this->makeRedirect('create', $model)) {
    return $redirect;
}
```
* The first two lines contain my hydrated model but I can't return a Redirect::to($model->permalink) from them because it won't have any affect.
* Inside `makeRedirect` method there is no interaction with the controller and no POST value I can send to set the redirect URL.
* I can make the redirect happen from formAfterSave like so:
```php
public function formAfterSave($model)
{
    if (post('preview')) {
        // Grab the current config
        $config = $this->asExtension('FormController')->getConfig();
        // Overwrite the redirect value with our new model's permalink
        $config->{$this->formGetContext()}['redirect'] =
            $model->permalink.'?preview=true';
        // Save the new config
        $this
            ->asExtension('FormController')
            ->setConfig($config);
    }
}
```
but obviously this isn't a great solution. It's messy and not very intuitive.

Flynsarmy:
Implementation of what was discussed in #4946.

Adds a `formGetRedirectUrl` method to `FormController` overrideable through a controller method of the same name. This method returns a URL to redirect to upon save. 

This PR is backwards compatible.

### DIFF
--- a/modules/backend/behaviors/FormController.php
+++ b/modules/backend/behaviors/FormController.php
@@ -441,19 +441,6 @@ class FormController extends ControllerBehavior
     }
 
     /**
-     * Returns a URL based on supplied context. Relative URLs are treated as
-     * backend URLs.
-     *
-     * @param string $context Redirect context, eg: create, update, delete
-     * @param Model $model The active model.
-     * @return string
-     */
-    public function formGetRedirectUrl($context = null, $model = null)
-    {
-        return '';
-    }
-
-    /**
      * Internal method used to prepare the form model object.
      *
      * @return October\Rain\Database\Model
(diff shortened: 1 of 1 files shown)


Flynsarmy:
Documentation PR here https://github.com/octobercms/docs/pull/432

LukeTowers:
@Flynsarmy the idea was to rename `getRedirectUrl` as `formGetRedirectUrl` and

mjauvin:
@LukeTowers don't text and drive! 🤣

Flynsarmy:
I'm not sure I understand the problem @LukeTowers

LukeTowers:
I was wanting the existing `getRedirectUrl()` method to be renamed to `formGetRedirectUrl()`: https://github.com/octobercms/october/blob/d46accfc4f309f5f071cee41ac5bf58092559392/modules/backend/behaviors/FormController.php#L474 & https://github.com/octobercms/october/blob/d46accfc4f309f5f071cee41ac5bf58092559392/modules/backend/behaviors/FormController.php#L499-L516. Also would need to be made public (thus overridable)

PR #4960, opened 2020-02-29
Backend theme localization

### LINKED ISSUE #4308: Make theme translations available under the themes.themecode::lang namespace in the backend translation system
October CMS doesn't support the themes internationalization at the moment. I mean that you can't add more than one (default) language to the theme settings screens. It's an unsolvable issue for the development of the themes for the international market.

We at LOVATA just started to develop themes for our Shopaholic. And we plan to add a huge amount of themes to the marketplace in the near future. And we want to make themes using more easy for the non-English speaking users. Not only we faced this problem, please, look at this [tweet](https://twitter.com/MateThemes/status/1069862295373799424).

GinoPane:
Fixes https://github.com/octobercms/october/issues/4308

I utilized the existing logic for plugin backend translations with no changes. Therefore language files must be located in a `lang` directory inside the theme directory. As for me, it is a good point because there are no surprises  for developers.

Translations are populated only for active theme when the theme options are opened. 

Here's the example.

Language file in the theme directory:
image

Theme settings:

```
tabs:
  fields:
    website_name:
      tab: Info
      label: theme.ginopane-mdblog::lang.options.website_name.label
      type: text
      default: Gino Pane - MDBlog
      translatable: true
```

The result:
image

Works nicely.

If everything is fine I'll make a PR for docs.

### DIFF
--- a/modules/cms/controllers/ThemeOptions.php
+++ b/modules/cms/controllers/ThemeOptions.php
@@ -1,8 +1,11 @@
 <?php namespace Cms\Controllers;
 
+use File;
+use Lang;
 use Backend;
 use BackendMenu;
 use ApplicationException;
+use Cms\Classes\Theme;
 use Cms\Models\ThemeData;
 use Cms\Classes\Theme as CmsTheme;
 use System\Classes\SettingsManager;
@@ -43,6 +46,7 @@ class ThemeOptions extends Controller
         parent::__construct();
 
         $this->pageTitle = 'cms::lang.theme.settings_menu';
+        $this->registerLocalization($this->getDirName());
 
         BackendMenu::setContext('October.System', 'system', 'settings');
         SettingsManager::setContext('October.Cms', 'theme');
@@ -103,8 +107,12 @@ class ThemeOptions extends Controller
 
     /**
      * Default to the active theme if user doesn't have access to manage all themes
+     *
+     * @param string $dirName
+     *
+     * @return string
      */
-    protected function getDirName($dirName = null)
+    protected function getDirName(string $dirName = null): string
     {
         /*
          * Only the active theme can be managed without this permission
@@ -142,5 +150,23 @@ class ThemeOptions extends Controller
         }
 
         return $theme;
+    }
+
+    /**
+     * Register language namespace if available for the theme
+     *
+     * @param string $dirName
+     *
+     * @throws ApplicationException
+     */
+    protected function registerLocalization(string $dirName)
+    {
+        /** @var Theme $theme */
+        $theme = $this->findThemeObject($dirName);
+
+        $langPath = $theme->getPath() . '/lang';
+        if (File::isDirectory($langPath)) {
+            Lang::addNamespace("theme.$dirName", $langPath);
+        }
     }
 }
--- a/modules/cms/controllers/ThemeOptions.php
+++ b/modules/cms/controllers/ThemeOptions.php
@@ -109,7 +109,6 @@ class ThemeOptions extends Controller
      * Default to the active theme if user doesn't have access to manage all themes
      *
      * @param string $dirName
-     *
      * @return string
      */
     protected function getDirName(string $dirName = null): string
--- a/modules/cms/controllers/ThemeOptions.php
+++ b/modules/cms/controllers/ThemeOptions.php
@@ -159,7 +159,6 @@ class ThemeOptions extends Controller
      */
     protected function registerLocalization(string $dirName)
     {
-        /** @var Theme $theme */
         $theme = $this->findThemeObject($dirName);
 
         $langPath = $theme->getPath() . '/lang';
--- a/modules/cms/controllers/ThemeOptions.php
+++ b/modules/cms/controllers/ThemeOptions.php
@@ -155,7 +155,7 @@ class ThemeOptions extends Controller
      * Register language namespace if available for the theme
      *
      * @param string $dirName
-     * @throws ApplicationException
+     * @throws ApplicationException if the theme object cannot be loaded
      */
     protected function registerLocalization(string $dirName)
     {
--- a/modules/cms/controllers/ThemeOptions.php
+++ b/modules/cms/controllers/ThemeOptions.php
@@ -163,7 +163,7 @@ class ThemeOptions extends Controller
 
         $langPath = $theme->getPath() . '/lang';
         if (File::isDirectory($langPath)) {
-            Lang::addNamespace("theme.$dirName", $langPath);
+            Lang::addNamespace("theme.{$theme->getId()}", $langPath);
         }
     }
 }
--- a/modules/cms/controllers/ThemeOptions.php
+++ b/modules/cms/controllers/ThemeOptions.php
@@ -155,7 +155,6 @@ class ThemeOptions extends Controller
      * Register language namespace if available for the theme
      *
      * @param string $dirName
-     *
      * @throws ApplicationException
      */
     protected function registerLocalization(string $dirName)
--- a/modules/cms/controllers/ThemeOptions.php
+++ b/modules/cms/controllers/ThemeOptions.php
@@ -111,7 +111,7 @@ class ThemeOptions extends Controller
      * @param string $dirName
      * @return string
      */
-    protected function getDirName(string $dirName = null): string
+    protected function getDirName(string $dirName = null)
     {
         /*
          * Only the active theme can be managed without this permission
--- a/modules/cms/controllers/ThemeOptions.php
+++ b/modules/cms/controllers/ThemeOptions.php
@@ -46,7 +46,6 @@ class ThemeOptions extends Controller
         parent::__construct();
 
         $this->pageTitle = 'cms::lang.theme.settings_menu';
-        $this->registerLocalization($this->getDirName());
 
         BackendMenu::setContext('October.System', 'system', 'settings');
         SettingsManager::setContext('October.Cms', 'theme');
@@ -149,21 +148,5 @@ class ThemeOptions extends Controller
         }
 
         return $theme;
-    }
-
-    /**
-     * Register language namespace if available for the theme
-     *
-     * @param string $dirName
-     * @throws ApplicationException if the theme object cannot be loaded
-     */
-    protected function registerLocalization(string $dirName)
-    {
-        $theme = $this->findThemeObject($dirName);
-
-        $langPath = $theme->getPath() . '/lang';
-        if (File::isDirectory($langPath)) {
-            Lang::addNamespace("theme.{$theme->getId()}", $langPath);
-        }
     }
 }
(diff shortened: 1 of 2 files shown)


LukeTowers:
This seems fine to me. @GinoPane can you submit a PR to the docs?

modules/cms/controllers/ThemeOptions.php
```diff
@@ -103,8 +107,12 @@ public function formExtendFieldsBefore($form)
 
     /**
      * Default to the active theme if user doesn't have access to manage all themes
+     *
+     * @param string $dirName
+     *
+     * @return string
```
LukeTowers:
```suggestion
     *
     * @param string $dirName
     * @return string
```

modules/cms/controllers/ThemeOptions.php
```diff
@@ -103,8 +107,12 @@ public function formExtendFieldsBefore($form)
...
      */
-    protected function getDirName($dirName = null)
+    protected function getDirName(string $dirName = null): string
```
LukeTowers:
```suggestion
    protected function getDirName(string $dirName = null)
```
We don't really use return type hints a lot in October core.

modules/cms/controllers/ThemeOptions.php
```diff
@@ -143,4 +151,22 @@ protected function findThemeObject($name = null)
 
         return $theme;
     }
+
+    /**
+     * Register language namespace if available for the theme
+     *
+     * @param string $dirName
+     *
+     * @throws ApplicationException
```
LukeTowers:
```suggestion
     * @param string $dirName
     * @throws ApplicationException
```
Also when would an application exception be thrown?

modules/cms/controllers/ThemeOptions.php
```diff
@@ -143,4 +151,22 @@ protected function findThemeObject($name = null)
...
+     */
+    protected function registerLocalization(string $dirName)
+    {
+        /** @var Theme $theme */
```
LukeTowers:
```suggestion
```
This doesn't add anything useful.

modules/cms/controllers/ThemeOptions.php
GinoPane: You'd better start using them 😄

modules/cms/controllers/ThemeOptions.php
GinoPane: It is being thrown in `findThemeObject` method. And as long as it is not caught here exception is bubbling and this method automatically starts throwing the same.

modules/cms/controllers/ThemeOptions.php
GinoPane: It does for IDE (autocompletion will be enabled) and for devs who won't need to guess which object is there.

GinoPane:
Though I don't really agree with them I accepted the suggestions @LukeTowers

lautsevich:
@GinoPane, thx! As for me, it looks great, when we're talking about DX.

mjauvin:
LGTM!

modules/cms/controllers/ThemeOptions.php
```diff
@@ -143,4 +150,20 @@ protected function findThemeObject($name = null)
 
         return $theme;
     }
+
+    /**
+     * Register language namespace if available for the theme
+     *
+     * @param string $dirName
+     * @throws ApplicationException
```
bennothommo: @GinoPane As with https://github.com/octobercms/october/pull/4972, if we are going to start adding **@throws** declarations in the docs, we need to specify some context as to *why* they are thrown. Would you mind adding in an example of why `ApplicationException` would be thrown for this method?

GinoPane: @bennothommo do you mean adding it into a docblock here? Basically the exception is inherited from findThemeObject method where it is thrown when the theme name is not passed via POST for any reason or when the theme cannot be loaded.

GinoPane: Like 
```
@throws ApplicationException if the theme object cannot be loaded
```

bennothommo: @GinoPane sorry for the delay in responding - yes, please add the example in this docblock.

GinoPane: @bennothommo done!

LukeTowers:
Looking for your input on this @daftspunk

mjauvin:
@LukeTowers do you think that could introduce bugs/backward compatibility issues that we don't see? This looks fairly clean to me.

LukeTowers:
@mjauvin just want @daftspunk's stamp of approval on the proposed API.

LukeTowers:
@GinoPane can you make the theme namespace be theme.themeCode instead of theme.themeDir? Minor difference, but pulling it from theme code (defined in theme.yaml) will make it so that the translations will continue to work even if the user installs the theme under a different directory.

GinoPane:
@LukeTowers I'm happy to change it, but this directory-based solution seems to be a standard. The `findThemeObject` method expects the directory. Or did I use it in a wrong way?

mjauvin:
@LukeTowers ThemeOptions' getDirName() calls Theme's getActiveThemeCode() method so I think this *is* actually using the theme code. Not sure why that method is called getDirName though... unless theme code is inextricably linked to the theme directory?

GinoPane:
Ah, I understood what was required. 

This line was updated in the recent commit:

image

mjauvin:
that should be reopened. What was blocking this?

bennothommo:
@mjauvin Only thing that's really blocking this is that @LukeTowers wanted to get @daftspunk's approval for the change.

modules/cms/controllers/ThemeOptions.php
```diff
@@ -43,6 +46,7 @@ public function __construct()
         parent::__construct();
 
         $this->pageTitle = 'cms::lang.theme.settings_menu';
+        $this->registerLocalization($this->getDirName());
```
LukeTowers: Having the registration of theme language files here means that these translations would only be available specifically on this controller for specifically the currently active theme. @bennothommo what do you think about making them more widely available in the backend in general? Is there any benefit to that?

I would probably say that they definitely shouldn't be made available in the frontend given that might encourage developers to use this as a feature to localize themes rather than support with RainLab.Translate, which would be a worse UX overall for the end users; however I'm not 100% sure if locking them just to this particular context is the right option or not. 

@mjauvin @bennothommo any thoughts?

mjauvin: I see what you mean @LukeTowers but I personally think this is perfect as is since the intended use is to allow theme options to be localizable WITHOUT requiring an extra plugin to do this (self contained theme).

What other controller do you have in mind that could use this?

bennothommo: @LukeTowers I think the only other area where there would be any need for localisation from the a theme would be the Syntax Parser fields which get rendered into a form in the Static Pages plugin - not sure if that's already being localised in some way already?

Otherwise, the Theme Options controller would be the only place where localisation would be needed.

I agree with you that the chief method of translation on the theme content itself should be done through the Translate plugin.

LukeTowers: That's a good point re: the syntax parser fields. @GinoPane can you take a stab at that?

GinoPane: So what did you guys decide? I'm not sure I know anything about syntax parser fields. Is it another controller?

LukeTowers: @GinoPane it's in the RainLab.Pages plugin, but what should work the best is to just do a simple App::isRunningInBackend() around the language registration code for all present themes in the CMS ServiceProvider.

GinoPane: Hm, I didn't plan to change any of the plugins. This is purely for vanilla system backend.
Do you suggest to surround this line with the `isRunningInBackend` condition?

image

LukeTowers: Yes, but also move it to the CMS ServiceProvider rather than the theme options controller

wtraub:
Unsubscribe

Luke Towers <notifications@github.com> schrieb am Do., 5. Nov. 2020, 16:31:

modules/cms/controllers/ThemeOptions.php
GinoPane: Ok, moved it

modules/cms/controllers/ThemeOptions.php
```diff
@@ -1,8 +1,11 @@
 <?php namespace Cms\Controllers;
 
+use File;
+use Lang;
 use Backend;
 use BackendMenu;
 use ApplicationException;
+use Cms\Classes\Theme;
```
LukeTowers: These extra import statements shouldn't be required anymore

GinoPane: Oops, indeed

LukeTowers:
My only other concern with this is that now there would be two ways to register translations for a theme, frontend translations would be registered in theme.yaml (https://github.com/rainlab/translate-plugin#message-translation) and backend translations would be registered through a lang folder. Any thoughts on this as a potential annoyance / gotcha for theme developers? @mariavilaro @bennothommo @mjauvin?

mjauvin:
@LukeTowers `theme.yaml` is introduced by RainLab.Translate... this PR is for the base Laravel localization strings.

LukeTowers:
@mjauvin I know that, but a developer that's just trying to figure out how to offer their theme in multiple languages having two completely different ways of doing so doesn't really seem to stick with October's ethos of "things should be as simple as possible, no simpler"

mjauvin:
@LukeTowers isn't it the same with plugins having their backend translation through the lang folder and their front-end (e.g. components) having them through RainLab.Translate's Message Translation ?

I'm not sure I see the difference here...

LukeTowers:
Just playing devil's advocate here @mjauvin 😄 I personally don't have a problem with the two separate locations, but I'm also extremely familiar with how October works so I'm just trying to look at it from the perspective of someone that's new to October trying to figure out how to get things done. I've reviewed the docs PR and made some changes though so now it seems fine to me.

mjauvin:
@LukeTowers @GinoPane 
Shouldn't the lang file path preix be `themes.my-theme::` instead of `theme.my-theme::` to more closely match the themes path ?

LukeTowers:
Good catch @mjauvin I thought it already was `themes`. Submit a PR to the docs and here and I'll merge it

lautsevich:
Good job! Thx 2 all!

bennothommo:
@GinoPane @mjauvin @LukeTowers I've started having intermittent exceptions on login with this PR in place. I think that having this registration occur during the `register` process for the CMS service provider is resulting in it sometimes being run before the DB connection is established.

I've made a PR to move this to the `boot` process instead here - https://github.com/octobercms/october/pull/5349

Would you mind giving it a test and letting me know if the localization is still working?

PR #4966, opened 2020-03-03
Add option to ignore timestamps for date range filters

LarryBarker:
Date columns and form fields have the option to ignore timezones, which displays the true value from the database in the column or form field. This PR adds the option for date and daterange filters to support the same feature.

The current behavior of daterange filters can be misleading, especially if the user has a backend timezone set in their preferences. For example, setting the timezone to America/Chicago can cause the date range filter to return values outside of the selected date range. 

image
image

Inspecting the scope values shows different dates than what are selected: 
image

We can confirm these values are being passed from the client side by inspecting the network request:
image

I have confirmed that this change produces the desired results by testing with the Test plugin:
image
image
image
image

### DIFF
--- a/modules/backend/widgets/filter/partials/_scope_daterange.htm
+++ b/modules/backend/widgets/filter/partials/_scope_daterange.htm
@@ -10,7 +10,7 @@
         'firstDay' => $scope->firstDay,
         'yearRange' => $scope->yearRange,
     ])) ?>"
-    <?php if ($scope->ignoreTimezone): ?>data-ignore-timezone<?php endif ?>
+    <?php $scope->ignoreTimezone ? 'data-ignore-timezone' : ''; ?>
 >
     <span class="filter-label"><?= e(trans($scope->label)) ?>:</span>
     <span class="filter-setting"><?= isset($afterStr) && isset($beforeStr) ? ($afterStr . ' → ' . $beforeStr) : e(trans('backend::lang.filter.date_all')) ?></span>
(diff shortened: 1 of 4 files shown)


modules/backend/widgets/filter/partials/_scope_daterange.htm
```diff
@@ -10,7 +10,8 @@
         'firstDay' => $scope->firstDay,
         'yearRange' => $scope->yearRange,
     ]))
-    ?>">
+    ?>"
+    <?php if ($scope->ignoreTimezone): ?>data-ignore-timezone<?php endif ?>>
```
LukeTowers:
```suggestion
    <?php if ($scope->ignoreTimezone): ?>data-ignore-timezone<?php endif ?>
>
```

LukeTowers:
LGTM, I'll wait for @daftspunk to review it though

LarryBarker:
@LukeTowers help 🤓

modules/backend/widgets/filter/partials/_scope_daterange.htm
```diff
@@ -9,8 +9,9 @@
         'maxDate' => $scope->maxDate,
         'firstDay' => $scope->firstDay,
         'yearRange' => $scope->yearRange,
-    ]))
-    ?>">
+    ])) ?>"
+    <?php if ($scope->ignoreTimezone): ?>data-ignore-timezone<?php endif ?>
```
LukeTowers, line 13:
```suggestion
    <?php $scope->ignoreTimezone ? 'data-ignore-timezone' : ''; ?>
```

PR #4970, opened 2020-03-05
Allows model scope with join in relation form widget

RomainMazB:
With this lines inversion, it allows the programmers to use model scope with join statement.

More info in [this forum post](https://octobercms.com/forum/post/displays-a-related-model-with-scope-which-uses-join-doesnt-work-is-that-an-october-issue)

### DIFF
--- a/modules/backend/formwidgets/Relation.php
+++ b/modules/backend/formwidgets/Relation.php
@@ -127,13 +127,13 @@ class Relation extends FormWidgetBase
                 $query->where($relationModel->getKeyName(), '<>', $model->getKey());
             }
 
-            if ($scopeMethod = $this->scope) {
-                $query->$scopeMethod($model);
-            }
-
             // Even though "no constraints" is applied, belongsToMany constrains the query
             // by joining its pivot table. Remove all joins from the query.
             $query->getQuery()->getQuery()->joins = [];
+
+            if ($scopeMethod = $this->scope) {
+                $query->$scopeMethod($model);
+            }
 
             // Determine if the model uses a tree trait
             $treeTraits = ['October\Rain\Database\Traits\NestedTree', 'October\Rain\Database\Traits\SimpleTree'];


bennothommo:

bennothommo:
@RomainMazB I am fairly certain the joins are removed for a reason, so I'm not sure we can accept this fix.

RomainMazB:
@bennothommo Sorry for the branch.

The changes I made as you can see doesn't do anything but allowing the joins in the scope. The joins are still removed everywhere else, for sure it's here for a reason but removing the joins in scopes is kind of a non sense.

How to achieve what I'm trying to without join?

For those who didn't read the forum thread and in an easier version to understand:
I'm trying to scope an Article models query to remove all the Article which has no tag linked to it (in a Many To Many relationship, so I'm forced to use a join on pivot table).

bennothommo:
@RomainMazB given that the relation widget and controller tightly control the handling of related data, I presume that the joins are removed to prevent conflicts or other connected relation data from affecting the functionality of the widget. As I said, if there's a line there specifically removing joins *after* allowing a developer to modify the query, then it's a clearly an intentional step.

RomainMazB:
@bennothommo I understand your judgement, but since it creates a limitation, we should create a workaround for these edge case to allow the developer to create a scope with a join.

For now I can't see any proper way to fix my problem without this modification: the data I need is in the pivot table, so I need a join query in my scope.

An other way (but not a good one imo) would be to create an accessor in the model which need the scoped Articles to return a filtered list of Articles just to fill a select field instead of a relation one. It would work but we lose all the benefits of relationwidget, such as the  and we add a "useless" query:
```
    public function getArticlesWithTagOptions() {
        return Article::join('articles_tags', 'articles.id', '=', 'articles_tags.article_id')->get();
    }
```

mjauvin:
@RomainMazB You would need to submit a PR that allow extending the base functionality without introducing a breaking change.

Maybe a new field option for the Relation formwidget? (e.g. "forcedScope:" which would be called AFTER the joins are removed but regular "scope:" would still be called BEFORE the joins are removed?)

RomainMazB:
@mjauvin It's a solution but I don't like the "force" way since it's seems like a hacky way to achieve that and create a new behavior which "complicates" the widget by adding a new parameters just to escape an apparently wanted security.

And creating a "force" way would have the same results as the actual pull request: it allows the user to exec scope with join. So all the scope could use the "forceScope" to ensure it pass the security and the actual scope parameters would become useless, no? So why this PR would be refused and the other one merged?

If all the OctoberCMS maintainers are ok with that solution, I'll work on it.

I'm convinced since I faced it that allows join in scope is useful in many cases and need to be implemented in OctoberCMS.

mjauvin:
@RomainMazB Maybe a cleaner way would be to add an event that would allow the query to be modified before it gets executed?

mjauvin:
@RomainMazB an alternative for you would be to create a new formwidget based on the Relation formwidget and use the plugin's `registerFormWidgets()` method to allow you to use this "special" widget in your project(s) ?

See: https://octobercms.com/docs/backend/widgets#form-widget-registration

RomainMazB:
@mjauvin yeah it's a bit cleaner but it still complicates the widget and create a "backdoor" for a wanted security.

IMO the real question is "in which cases do we need to secure the query by deleting all the join statement and is it the only way to securize that cases?" For now I don't see any case where it's needed but I'm not the best guy to answer that question.

Maybe @luketowers or @daftspunk can?

I'll work on it as soon as I have a guideline for a PR which could be merged which satisfies anybody.

RomainMazB:
For those who will land here with the same problem in the future, in the case where the PR is refused, here is the trick with a custom FormWidget:

`plugins/romainmazb/realestate/formwidgets/RelationScopeJoin.php`
```
<?php

namespace RomainMazB\RealEstate\FormWidgets;

use Backend\FormWidgets\Relation;
use October\Rain\Database\Relations\Relation as RelationBase;

class RelationJoinScope extends Relation
{
    // Render the form widget with the initial relation partial
    public function render()
    {
        $this->prepareVars();
        return $this->makePartial('~/modules/backend/formwidgets/relation/partials/_relation.htm');
    }
    
    // The exact same function from the parent Relation widget with the lines inversion
    protected function makeRenderFormField()
    {
        return $this->renderFormField = RelationBase::noConstraints(function () {

            $field = clone $this->formField;
            $relationObject = $this->getRelationObject();
            $query = $relationObject->newQuery();

            list($model, $attribute) = $this->resolveModelAttribute($this->valueFrom);
            $relationType = $model->getRelationType($attribute);
            $relationModel = $model->makeRelation($attribute);

            if (in_array($relationType, ['belongsToMany', 'morphToMany', 'morphedByMany', 'hasMany'])) {
                $field->type = 'checkboxlist';
            }
            elseif (in_array($relationType, ['belongsTo', 'hasOne'])) {
                $field->type = 'dropdown';
            }

            // Order query by the configured option.
            if ($this->order) {
                // Using "raw" to allow authors to use a string to define the order clause.
                $query->orderByRaw($this->order);
            }

            // It is safe to assume that if the model and related model are of
            // the exact same class, then it cannot be related to itself
            if ($model->exists && (get_class($model) == get_class($relationModel))) {
                $query->where($relationModel->getKeyName(), '<>', $model->getKey());
            }

            // Even though "no constraints" is applied, belongsToMany constrains the query
            // by joining its pivot table. Remove all joins from the query.
            $query->getQuery()->getQuery()->joins = [];

            if ($scopeMethod = $this->scope) {
                $query->$scopeMethod($model);
            }

            // Determine if the model uses a tree trait
            $treeTraits = ['October\Rain\Database\Traits\NestedTree', 'October\Rain\Database\Traits\SimpleTree'];
            $usesTree = count(array_intersect($treeTraits, class_uses($relationModel))) > 0;

            // The "sqlSelect" config takes precedence over "nameFrom".
            // A virtual column called "selection" will contain the result.
            // Tree models must select all columns to return parent columns, etc.
            if ($this->sqlSelect) {
                $nameFrom = 'selection';
                $selectColumn = $usesTree ? '*' : $relationModel->getKeyName();
                $result = $query->select($selectColumn, Db::raw($this->sqlSelect . ' AS ' . $nameFrom));
            }
            else {
                $nameFrom = $this->nameFrom;
                $result = $query->getQuery()->get();
            }

            // Some simpler relations can specify a custom local or foreign "other" key,
            // which can be detected and implemented here automagically.
            $primaryKeyName = in_array($relationType, ['hasMany', 'belongsTo', 'hasOne'])
                ? $relationObject->getOtherKey()
                : $relationModel->getKeyName();

            $field->options = $usesTree
                ? $result->listsNested($nameFrom, $primaryKeyName)
                : $result->lists($nameFrom, $primaryKeyName);

            return $field;
        });
    }
}
```
`plugins/romainmazb/realestate/Plugin.php`:
```
<?php namespace RomainMazB\RealEstate;

use System\Classes\PluginBase;

{
    public function registerFormWidgets()
    {
        return [
            'RomainMazB\RealEstate\FormWidgets\RelationJoinScope' => 'relation-join-scope'
        ];
    }
}
```
`fields.yaml`
```
        compatible_house_models:
            label: 'Modèles constructible'
            span: full
            nameFrom: name
            type: relation-join-scope
            scope: withDescription
            cssClass: 'inline-options'
            tab: Compléments
```

Hope it helps.

LukeTowers:
@RomainMazB @bennothommo As always, the solution to issues like this is to look through the git blame:

The join clearing code was added 6 years ago in https://github.com/octobercms/october/commit/5f12d9fc9b78023c2877f7dc31d8735c280c0544 and does exactly what the comment says, bypasses a constraint that the BelongsToMany relationship adds that breaks the widget.

The model scope extension ability was added 3 years ago in https://github.com/octobercms/october/commit/8ee7d3f5ae0e56b1e332ad18a8e6162365e4e4d7 and appears to be located where it is purely as a random decision by the original author.

As long as someone tests this with a BelongsToMany relationship to verify everything is still working fine this change looks acceptable to me.

RomainMazB:
@LukeTowers The fact is I tested it in my plugin, with or without the join clearing: the formwidget works perfectly. Can't figure why this was added. I'm not saying it's useless, it's better with simply due to performance optimization: the pivot table data are useless in this widget, and the query is faster without a join statement.

Here are 3 screenshots with three different situations:
 - 1./ Initial state
 - 2./ With the line inversion, only the scope join statement is performed
 - 3./ With the join clearing line commented, both scope and belongsToMany statement are performed
Initial state
With lines inversion
With lines commented

You can see the queries and their performance in the debug bar, and here is my relation and scope:

```PHP
public $belongsToMany = [
    'descriptions' => [
        Description::class,
        'table' => 'romainmazb_realestate_descriptions_house_models'
    ]
];

public function scopeWithDescription($query)
{
    return $query->join('romainmazb_realestate_descriptions_house_models', 'romainmazb_realestate_house_models.id', '=', 'romainmazb_realestate_descriptions_house_models.house_model_id');
}
```

The lines inversion doesn't screw anything: no performance issue, no widget issue.

I also tested with a left join even if it's useless, the results doesn't create any issue, it's just what the developer doesn't want to accomplish since no scope is applied, all the rows are returned.
Capture d’écran de 2020-03-07 14-20-42

PR #4972, opened 2020-03-06
Fix docblocks and imports in FormController behaviour.

jacobdekeizer:
Add correct parameter, return type and exception docblocks for FormController behaviour.
Import whole namespace for facade so we can follow the code in the editor.

### DIFF
--- a/modules/backend/behaviors/FormController.php
+++ b/modules/backend/behaviors/FormController.php
@@ -1,16 +1,16 @@
 <?php namespace Backend\Behaviors;
 
+use Db;
+use Str;
+use Lang;
+use Flash;
+use Event;
+use Redirect;
+use Backend;
 use Backend\Classes\ControllerBehavior;
-use Backend\Facades\Backend;
-use Exception;
-use Illuminate\Support\Facades\DB;
-use Illuminate\Support\Facades\Event;
-use Illuminate\Support\Facades\Lang;
-use Illuminate\Support\Facades\Redirect;
-use October\Rain\Exception\ApplicationException;
 use October\Rain\Router\Helper as RouterHelper;
-use October\Rain\Support\Facades\Flash;
-use October\Rain\Support\Str;
+use ApplicationException;
+use Exception;
 
 /**
  * Adds features for working with backend forms. This behavior
--- a/modules/backend/behaviors/FormController.php
+++ b/modules/backend/behaviors/FormController.php
@@ -303,7 +303,7 @@ class FormController extends ControllerBehavior
      * @param int $recordId Record identifier
      * @param string $context Form context
      * @return \Illuminate\Http\RedirectResponse|void
-     * @throws \October\Rain\Exception\ApplicationException
+     * @throws \October\Rain\Exception\ApplicationException if the provided recordId is not found
      */
     public function update_onSave($recordId = null, $context = null)
     {
@@ -340,8 +340,8 @@ class FormController extends ControllerBehavior
      *
      * @param int $recordId Record identifier
      * @return \Illuminate\Http\RedirectResponse|void
-     * @throws \October\Rain\Exception\ApplicationException
-     * @throws Exception
+     * @throws \October\Rain\Exception\ApplicationException if the provided recordId is not found
+     * @throws Exception if there is no primary key on the model
      */
     public function update_onDelete($recordId = null)
     {
@@ -408,7 +408,7 @@ class FormController extends ControllerBehavior
      * @see \Backend\Widgets\Form
      * @param array $options Render options
      * @return string Rendered HTML for the form.
-     * @throws \October\Rain\Exception\ApplicationException
+     * @throws \October\Rain\Exception\ApplicationException if the Form Widget isn't set
      */
     public function formRender($options = [])
     {
@@ -560,7 +560,7 @@ class FormController extends ControllerBehavior
      *     <?= $this->formRenderPreview() ?>
      *
      * @return string The form HTML markup.
-     * @throws \October\Rain\Exception\ApplicationException
+     * @throws \October\Rain\Exception\ApplicationException if the Form Widget isn't set
      */
     public function formRenderPreview()
     {
@@ -589,7 +589,7 @@ class FormController extends ControllerBehavior
      *     <?= $this->formRenderOutsideFields() ?>
      *
      * @return string HTML markup
-     * @throws \October\Rain\Exception\ApplicationException
+     * @throws \October\Rain\Exception\ApplicationException if the Form Widget isn't set
      */
     public function formRenderOutsideFields()
     {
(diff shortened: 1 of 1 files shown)


modules/backend/behaviors/FormController.php
```diff
@@ -1,17 +1,16 @@
...
+use Backend\Facades\Backend;
 use Exception;
+use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Facades\Event;
+use Illuminate\Support\Facades\Lang;
+use Illuminate\Support\Facades\Redirect;
+use October\Rain\Exception\ApplicationException;
+use October\Rain\Router\Helper as RouterHelper;
+use October\Rain\Support\Facades\Flash;
+use October\Rain\Support\Str;
```
LukeTowers: We will not be accepting these namespace changes, if you want to be able to click through to the source classes in your IDE then use https://octobercms.com/plugin/flynsarmy-idehelper. Fully specifying the namespaces to facades completely defeats the point and makes the application more brittle. An excellent example of this is the fact that Laravel removed the Input facade and we added it back, but any code that was directly referencing the Laravel version is now going to be broken in the L6 upgrade. 

@bennothommo in relation to that, do you think we can setup a class alias for the old facade to improve compatibility?

jacobdekeizer:
@LukeTowers Updated my pr

modules/backend/behaviors/FormController.php
```diff
@@ -763,8 +771,10 @@ public function formFindModelObject($recordId)
         /*
          * Prepare query and find model record
          */
+        /** @var \October\Rain\Database\Builder $query */
         $query = $model->newQuery();
         $this->controller->formExtendQuery($query);
+        /** @var \October\Rain\Database\Model $result */
```
LukeTowers: What is the point of these comments?

modules/backend/behaviors/FormController.php
```diff
@@ -8,7 +8,6 @@
 use Redirect;
 use Backend;
 use Backend\Classes\ControllerBehavior;
-use October\Rain\Html\Helper as HtmlHelper;
```
LukeTowers: Why did you remove this?

jacobdekeizer: Because it is unused

modules/backend/behaviors/FormController.php
jacobdekeizer: Because it is an illuminate function which returns as type hint an illuminate model instead of the correct october model

jacobdekeizer: With this docblock the typehint is restored to the correct model

bennothommo:
@jacobdekeizer I think the reasoning for the `Model` typehint was because the `FormController` also supports Halcyon models (eg. `October\Rain\Halcyon\Model`), so I think the model typehints would need to reflect that those methods and properties can take both.

If you want a real challenge, you could propose an interface/contract for the Rain library that covers both and use that as the type-hint.

jacobdekeizer:
@bennothommo I think it's best to add both Models as typehint for now.

jacobdekeizer:
@bennothommo I've added the Halcyon model as typehint

modules/backend/behaviors/FormController.php
```diff
@@ -339,7 +339,9 @@ public function update_onSave($recordId = null, $context = null)
      * `formAfterDelete`.
      *
      * @param int $recordId Record identifier
-     * @return mixed
+     * @return \Illuminate\Http\RedirectResponse|void
+     * @throws \October\Rain\Exception\ApplicationException
+     * @throws Exception
```
bennothommo: @jacobdekeizer I'm leaning towards these being redundant on all methods in this class. They would make sense if we have multiple exception classes for individual pieces of the system, but we don't - we generally use `ApplicationException` and `SystemException` and that's it, so it can be a general assumption that a method *may* throw one of those exceptions at any given point.

Would you have any problems with removing them from your changes?

bennothommo:
@jacobdekeizer Just out of curiousity, are these doc hints being generated by an automated system, or are you manually entering them in?

modules/backend/behaviors/FormController.php
jacobdekeizer: Why should we make an assumption when we can typehint that this method throws these exceptions. Phpstorm will warn you when you don't catch an exception, if we remove this phpstorm will not warn you for uncaught exceptions. So I'm against removing them, because it helps the developer with what he can expect from these methods.

jacobdekeizer:
@bennothommo Phpstorm can autogenerate these docblocks. In my opinion all octobercms maintainers should use phpstorm, because it finds many typehint errors and other coding errors. I could make another x amount of pull requests to fix a lot of files, but I don't have the time to do that. If the maintainers use phpstorm they could see all these typehint problems themselves in advance, instead of someone else fixing it with a pull request. You could also look into tools like phpstan and mess dectector for better code.

modules/backend/behaviors/FormController.php
bennothommo: @jacobdekeizer I recognise that PHPStorm can analyse this and give hints, but on the outset, if I'm reading that, those two lines in particular tell me absolutely nothing. It doesn't give me any indication as to *why* those exceptions might be thrown, especially when you view the code itself in which you find that the exception wouldn't be thrown by any of the method's code, and that PHPStorm found that the exception was instead potentially thrown by a method 3 or 4 levels down in the call stack.

It's also presumptuous in some cases, as there's potentially many more types of exceptions that can be thrown - especially when you bring plugins into the mix.

bennothommo:
@jacobdekeizer I appreciate the work that you have put into this so far. Perhaps for now, if its your intention to clean up a few more classes, maybe stick with the obvious cases of type hints that are incorrect - it's my hope that we can use actual parameter type hints and return types in the future, which will make most of the docblock changes redundant.

modules/backend/behaviors/FormController.php
jacobdekeizer: That's why we have to typehint all exceptions, so we (as developers) aren't surprised with undocumented exceptions.

LukeTowers: @jacobdekeizer the point @bennothommo is trying to make is that at some point it's not possible to typehint all possible exceptions that could be thrown, especially when you start throwing event handling into the mix.

bennothommo: @jacobdekeizer If you add context as to why those exceptions might be thrown (as per [@throws](https://docs.phpdoc.org/latest/references/phpdoc/tags/throws.html) documentation), it might make it more credible to keep these lines.

jacobdekeizer:
@bennothommo I've added descriptions to the `@throws` exception docblocks.

modules/backend/behaviors/FormController.php
bennothommo: @jacobdekeizer I get your reasoning, but why does this need to be specified when normally you would not interact with these variables at all besides through the `formExtendQuery` and `formExtendModel` methods which are already correctly type-hinted?

jacobdekeizer: @bennothommo You're right, these docblocks can be removed.

bennothommo:
@jacobdekeizer I'm happy to merge this, however, I'll check with @daftspunk if this is going to have any conflicts with the [API Docs](https://octobercms.com/docs/api) (mainly in terms of the multiple model type hints)

Issue #4969, opened 2020-03-04
Remove the 'Show the stacktrace' button in event logs


ayumi-cloud:
Please can the admins consider removing the `Show the stacktrace` button in event logs!

> It's totally not needed and is a real pain in the neck having to press this button everytime! It's a completely pointless extra step!

See this screenshot as an example:

![image](https://user-images.githubusercontent.com/57409060/75904740-ad316b80-5e3b-11ea-99d1-d126a6f9df93.png)

We try to look at the reason for doing this step and think it's maybe to safe guard mobile devices from loading a web page with a big height? Either way, it's totally not needed for a developer working on a computer system!

Here's a proposal re-design:

![image](https://user-images.githubusercontent.com/57409060/75905241-714ad600-5e3c-11ea-8d18-eb8fe4cb47a6.png)

What we have done is basically three things:

1. Removed `Show the stacktrace` button in event logs! This is a pointless step and instead just display the full stacktrace (that's why we have clicked on the web page in the first place - from the event log list page).

2. This removes a flash of unstyled content (FOUC) error that always shows to users! Because October CMS first displays the full stack trace and then switches to this pointless button user experience - we see this everytime, we use the event logs and it's not a nice user experience!

3. We added a the `return to event log` button to the top and bottom of the screen - this is to help people on mobile devices having to scroll with their fingers to return to the previous list page! Note we are going to add some basic logic and say if the web page height is bigger than the screen's device size then display the top `return to event log` button - otherwise don't display it as the bottom `return to event log` button will be displaying on the webpage! e.g.

- Long stack traces show bottom and top buttons.
- Short stack traces show bottom button only.

Lastly, I now open this issue up for discussion and hopefully other people find the current layout annoying as well.

LukeTowers:
Feel free to submit a PR that we can review.

daftspunk:
This is by design, sorry. For non-technical users they only need to be concerned with the error. It helps when instructing clients to read the logs and it focuses their attention on the right place. We went out of our way to build this button for this reason

ayumi-cloud:
@daftspunk would it be ok to have an option in one of the `config` files, to turn this feature `on/off` for **advanced** users?

bennothommo:
@ayumi-cloud as a suggestion, you can make a plugin that removes the button and shows the stacktrace by default.

PR #4973, opened 2020-03-08
Remove the 'Show the stacktrace' button in event logs

ayumi-cloud:
## Explanation

The event logs section has been updated in two parts:

1. Show full error message in the list view now.

2. Show the full stack trace in the form view now.

These two things stream line the event logs section.

### 1. Type correction

- Minor typo correction from a previous pr (changed 100 to 500), see here: https://octobercms.com/support/article/cl-465

I quote:

> The Event Log list now shows the entire first line of a logged error message, up to 500 characters maximum, in order to provide more context of errors in the list.

### 2. Remove the button step

- Removed the `Show the stacktrace` button in event logs - this is not needed.

Github issue found here: https://github.com/octobercms/october/issues/4969


ayumi-cloud:
~Need to rebase this.

daftspunk:
This is by design, sorry. For non-technical users they only need to be concerned with the error. It helps when instructing clients to read the logs and it focuses their attention on the right place. We went out of our way to build this button for this reason

PR #4936, opened 2020-02-11
Add configuration for throttling authentication attempts

mjauvin:
The backend AuthManager behavior is currently hard-coded.

This adds config/auth.php with the following settings to adjust default behavior:
- useThrottle
- attemptLimit
-  suspensionTime

PR #4974, opened 2020-03-09
Add configuration for throttling authentication attemptsAuth config

mjauvin:
The backend AuthManager behavior is currently hard-coded.

This adds config/auth.php with the following settings under 'throttle' section to adjust default behavior:

- enabled
- attemptLimit
- suspensionTime
￼
Replaces #4936

### DIFF
--- a/config/auth.php
+++ b/config/auth.php
@@ -1,0 +1,39 @@
+<?php
+
+return [
+
+    'throttle' => [
+        /*
+        |--------------------------------------------------------------------------
+        | Enable throttling of Backend authentication attempts
+        |--------------------------------------------------------------------------
+        |
+        | If set to true, users will be given a limited number of attempts to sign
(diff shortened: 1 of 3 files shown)


mjauvin:
I've set debug mode so that one can see what's happening.

https://84c7aac7f588.octodock.com/backend

user/pass = admin

bennothommo:
@mjauvin I've tested this on your Octodock instance, and it works great. There's only one *very minor* edge case that is not covered - raising the attempt limit while someone is suspended for too many attempts.

For example, if the attempt limit is 5, and someone attempts to login 5 times and fails, that user is suspended. If I then raise the attempt limit to 6, that person still cannot log in due to the suspension. I don't think we should really do anything about it at this stage, but I'm just making a note of it anyway.

PR #4978, opened 2020-03-12
Allow dot-notation for specifying external parameters for a component

bennothommo:
This is a minor change to allow users to specify deeper-level external parameters for component parameters.

My case was providing a "content" array to a partial:

```
{% partial 'content/blog-list' content=content %}
```

But being unable to use a singular key in that array for component parameters, eg. the below doesn't currently work without this fix:

```
[blogPosts]
pageNumber = "{{ :page }}"
categoryFilter = "{{ content.category }}"
postsPerPage = "{{ content.postsPerPage }}"
noPostsMessage = "No posts found"
sortOrder = "published_at desc"
categoryPage = 404
postPage = "{{ content.postPage }}"
```

### DIFF
--- a/modules/cms/classes/Controller.php
+++ b/modules/cms/classes/Controller.php
@@ -1538,7 +1538,7 @@ class Controller
                     $newPropertyValue = $routerParameters[$routeParamName] ?? null;
                 }
                 else {
-                    $newPropertyValue = $parameters[$paramName] ?? null;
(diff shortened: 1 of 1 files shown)


bennothommo:
I don't see this being a huge drama to merge - it's backwards compatible, only affects externally provided parameters (ie. through partial parameters), and keeps the same syntax consistency as using variables in Twig.

PR #4982, opened 2020-03-13
add getPluginUrl() method

mjauvin:
Instead of manually specifying the URL to link to resources for this plugin, we can have it auto-generated.

So if your plugin holds a new css asset, instead of the following call:
```
...addCss('/plugins/author/plugin/assets/css/myCssFile.css');
```

You can now use:
```
...addCss($this->getPluginUrl() . '/assets/css/myCssFile.css');
```

### DIFF
--- a/modules/system/classes/PluginBase.php
+++ b/modules/system/classes/PluginBase.php
@@ -6,6 +6,7 @@ use ReflectionClass;
 use SystemException;
 use Yaml;
 use Backend;
+use Config;
 
 /**
  * Plugin base class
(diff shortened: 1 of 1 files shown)


daftspunk:
Why is this useful to have a dynamic source when the location is static?

mjauvin:
The location is not necessarily static. What if you change the plugins path in cms config? It can also be useful in plugin extending other plugins (in a loop for example).

daftspunk:
I see. We should look at using an approach similar to `File::symbolizePath` so the outcome appears like this:

```
...addCss('$/author/plugin/assets/css/myCssFile.css');
```

Integrated directly into `AssetMaker::getAssetPath`

mjauvin:
that's a very good idea

mjauvin:
@daftspunk Although what I implemented is to point to resources local to your plugin for external controllers/models you extend, so what you suggest wouldn't work.

daftspunk:
We don't want to allow extension this via this method, it is too complex and creates a black box scenario in order to trace missing or replaced assets. Will be happy to look at a PR with what I suggested, or even show the code you want and we can find an alternative way. It's hard to accept empty features with nothing to support them

PR #4986, opened 2020-03-16
Allow quotes to be correctly handled by october:env

### LINKED ISSUE #4979: Database connection lost after running command php artisan october:env if mysql password has quotes in database.php file
- OctoberCMS Build: 464 
- PHP Version:  7.2.24
- Database Engine: mysql (InnoDB) 
- Plugins Installed: october.demo-v1.0.1

### Description:
Hello,

I get a mysql connection error after running command php artisan october:env ; everything worked fine before. I realized it was because I used a quote caracter (" or ') in the password. Password was correctly written in file config/database.php but quote caracter was removed in .env after running command artisan october:env.

The password was generated by ISP and I can't change it. Of course, I just could change the password in the new generated .env file ; but there is a risk that some people spend a lot of time looking for the solution.

I installed OctoberCMS via command line and composer. Password for mysql database was entered during normal CLI installation.

### Steps To Reproduce:

1. Install OctoberCMS via composer command `composer create-project october/october myoctober`
2. Enter a password for mysql database during normal database questions. Enter for example `3$Zvt?2dF.~"fZ*`
3. Finalize installation and check that everything works fine (frontend and backend)
4. Execute command `php artisan october:env` for creating dot env configuration file
5. Refresh backend end and confirm that there is now a mysql error (missing database)
6. Open the new created .env file and check that quote caracter is missing for mysql password : 3$Zvt?2dF.~fZ*

Issue is something similar to #3054 and #2361.

### Something else ?

Yes ! Thanks a lot for your work and this amazing CMS !

bennothommo:
Fixes #4979.

This fix will apply quotes around string environment variables which contain either a single, or double, quote as well as any variables with a hash symbol - escaping any double-quotes encountered. When `artisan october:env` is run, this should correctly transfer all configuration values from the config files to the `.env` file.

Includes a new unit test for the `october:env` command - **needs https://github.com/octobercms/library/pull/471 to be merged before the test will pass.**

### DIFF
--- a/modules/system/console/OctoberEnv.php
+++ b/modules/system/console/OctoberEnv.php
@@ -203,7 +203,7 @@ class OctoberEnv extends Command
     private function saveEnvSettings($key, $value)
     {
         if (! $this->envKeyExists($key)) {
-            $line = sprintf("%s=%s\n", $key, $this->stripQuotes($value));
+            $line = sprintf("%s=%s\n", $key, $value);
 
             if ($this->config == 'database' && $key != 'DB_CONNECTION') {
                 $this->writeDbEnvSettings($line);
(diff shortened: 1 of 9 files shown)


bennothommo:
I've run a few manual tests, and this is covered by unit tests, so it should be safe to merge.

PR #4989, opened 2020-03-17
Fix docbocks ListController behaviour.

jacobdekeizer:
Add correct parameter, return type and exception docblocks for ListController behaviour.

I made `listOverrideColumnValue` return `string|void` like it is in the current codebase. In pr #4972 @bennothommo talked about strict typing for the future, so we could change those return `string|void` to `string|null`. This allows us to add a strict return type `: ?string` in the future. It would not lead to any breaking changes now, but in the future it will when strict return types are added.

Methods that can be changed from `string|void` to `string|null` in this behaviour are:
- `listInjectRowClass`
- `listOverrideColumnValue`
- `listOverrideHeaderValue`

I would like to get your thoughts on this.

### DIFF
--- a/modules/backend/behaviors/ListController.php
+++ b/modules/backend/behaviors/ListController.php
@@ -204,7 +204,7 @@ class ListController extends ControllerBehavior
         if (isset($listConfig->toolbar)) {
             $toolbarConfig = $this->makeConfig($listConfig->toolbar);
             $toolbarConfig->alias = $widget->alias . 'Toolbar';
-            $toolbarWidget = $this->makeWidget('Backend\Widgets\Toolbar', $toolbarConfig);
+            $toolbarWidget = $this->makeWidget(\Backend\Widgets\Toolbar::class, $toolbarConfig);
             $toolbarWidget->bindToController();
             $toolbarWidget->cssClasses[] = 'list-header';
 
@@ -237,7 +237,7 @@ class ListController extends ControllerBehavior
 
             $filterConfig = $this->makeConfig($listConfig->filter);
             $filterConfig->alias = $widget->alias . 'Filter';
-            $filterWidget = $this->makeWidget('Backend\Widgets\Filter', $filterConfig);
+            $filterWidget = $this->makeWidget(\Backend\Widgets\Filter::class, $filterConfig);
             $filterWidget->bindToController();
 
             /*
@@ -564,8 +564,7 @@ class ListController extends ControllerBehavior
     public static function extendListColumns($callback)
     {
         $calledClass = self::getCalledExtensionClass();
-        Event::listen('backend.list.extendColumns', function ($widget) use ($calledClass, $callback) {
-            /** @var \Backend\Widgets\Lists $widget */
+        Event::listen('backend.list.extendColumns', function (\Backend\Widgets\Lists $widget) use ($calledClass, $callback) {
             if (!is_a($widget->getController(), $calledClass)) {
                 return;
             }
@@ -581,8 +580,7 @@ class ListController extends ControllerBehavior
     public static function extendListFilterScopes($callback)
     {
         $calledClass = self::getCalledExtensionClass();
-        Event::listen('backend.filter.extendScopes', function ($widget) use ($calledClass, $callback) {
-            /** @var \Backend\Widgets\Lists $widget */
+        Event::listen('backend.filter.extendScopes', function (\Backend\Widgets\Filter $widget) use ($calledClass, $callback) {
             if (!is_a($widget->getController(), $calledClass)) {
                 return;
             }
--- a/modules/backend/behaviors/ListController.php
+++ b/modules/backend/behaviors/ListController.php
@@ -42,17 +42,17 @@ class ListController extends ControllerBehavior
     protected $listConfig = [];
 
     /**
-     * @var \Backend\Classes\WidgetBase Reference to the list widget object.
+     * @var \Backend\Classes\WidgetBase[] Reference to the list widget object.
      */
     protected $listWidgets = [];
 
     /**
-     * @var \Backend\Classes\WidgetBase Reference to the toolbar widget objects.
+     * @var \Backend\Classes\WidgetBase[] Reference to the toolbar widget objects.
      */
     protected $toolbarWidgets = [];
 
     /**
-     * @var \Backend\Classes\WidgetBase Reference to the filter widget objects.
+     * @var \Backend\Classes\WidgetBase[] Reference to the filter widget objects.
      */
     protected $filterWidgets = [];
 
@@ -114,7 +114,7 @@ class ListController extends ControllerBehavior
 
     /**
      * Prepare the widgets used by this action
-     * @return void
+     * @return \Backend\Widgets\Lists
      */
     public function makeList($definition = null)
     {
@@ -165,7 +165,8 @@ class ListController extends ControllerBehavior
         /*
          * List Widget with extensibility
          */
-        $widget = $this->makeWidget('Backend\Widgets\Lists', $columnConfig);
+        /** @var \Backend\Widgets\Lists $widget */
+        $widget = $this->makeWidget(\Backend\Widgets\Lists::class, $columnConfig);
 
         $widget->bindEvent('list.extendColumns', function () use ($widget) {
             $this->controller->listExtendColumns($widget);
@@ -286,6 +287,7 @@ class ListController extends ControllerBehavior
     /**
      * Bulk delete records.
      * @return void
+     * @throws \October\Rain\Exception\ApplicationException when the parent definition is missing.
      */
     public function index_onDelete()
     {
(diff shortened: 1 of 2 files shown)


modules/backend/behaviors/ListController.php
```diff
@@ -473,8 +476,9 @@ public function listFilterExtendScopes($host)
 
     /**
      * Controller override: Extend supplied model
-     * @param Model $model
-     * @return Model
+     * @param \October\Rain\Database\Model|\October\Rain\Halcyon\Model $model
+     * @param string|null $definition
+     * @return \October\Rain\Database\Model|\October\Rain\Halcyon\Model
```
LukeTowers: @jacobdekeizer I don't think that the Lists widget actually supports Halcyon models given the deep integration with the Eloquent query builder

modules/backend/behaviors/ListController.php
```diff
@@ -558,14 +565,15 @@ public static function extendListColumns($callback)
     {
         $calledClass = self::getCalledExtensionClass();
         Event::listen('backend.list.extendColumns', function ($widget) use ($calledClass, $callback) {
+            /** @var \Backend\Widgets\Lists $widget */
```
LukeTowers: @bennothommo did we decide to allow these inline comments? I'm not sure why they're worth having?

bennothommo: @LukeTowers I think the intention is to show the expected variable returned from methods which may return several types of classes, like the `makeWidget` method does. It's mainly for providing autocompletion through these types of methods. I don't mind keeping them, but I'd hope their use is very minimal.

LukeTowers: hmmm, not really a fan. It's pretty clear from the event itself what widget is being returned. @jacobdekeizer thoughts?

jacobdekeizer: @LukeTowers @bennotjommo It's indeed for autocompletion. It also shows the developer witch widget is returned. We could also change it to `function (\Backend\Widget\Lists $widget) ...`. Is that allowed?

LukeTowers: @jacobdekeizer hmm, while that would prevent extensibility in terms of being able to inject any other widget in there instead that seems unlikely to be required and I do prefer that in terms of code clarity so I would say go ahead and give it a try and verify that it still works properly.

jacobdekeizer: @LukeTowers I've changed it and verified that it still works.

modules/backend/behaviors/ListController.php
```diff
@@ -165,7 +165,8 @@ public function makeList($definition = null)
         /*
          * List Widget with extensibility
          */
-        $widget = $this->makeWidget('Backend\Widgets\Lists', $columnConfig);
+        /** @var \Backend\Widgets\Lists $widget */
```
LukeTowers: Why is this required if the class reference is used to make the widget?

jacobdekeizer: Because the makeWidget method has as return typehint WidgetBase which doesn't have the methods that are used in the code below it. If we remove this we get many method does not exist errors by a decent IDE.

LukeTowers: There's a fine line between what the IDE says and common sense by the developer. There's very little reason in the first place for someone to be digging around in this method, and the makeList method already tells the IDE what widget is being returned.

jacobdekeizer: In my opinion good readable codebase should be the goal of every project. This includes the readability by the ide. Why would you want a file with 10 lines of errors by the ide or analyzer tool like phpstan. Phpstan also reads these kind of docblocks if I recall correctly. Coming from laravel I find it somewhat shocking to look into the octoberms codebase. So I'm trying to help you guys fixing some things up by making the code readable for an ide and a person.

LukeTowers: I'm going to defer to @daftspunk's opinion on this one, I personally don't like having `@var` comments littered all over the codebase since they provide little benefit in my eyes. I appreciate your effort in cleaning things up a bit though and your patience as we decide how we want to handle these sort of additions.

Klaasie: >  There's very little reason in the first place for someone to be digging around in this method

My personal opinion:
I find this odd. We should be encouraging people to get to know the system to such a degree that they end up digging around this method (any method like that really). It means they're looking to understand OC as a system. For a project that (I think) would like to have more contributions from the community it'd be great if they understand most of the system.

Having that said, allowing these kind of docblock lines will help most of the major IDE's understand exactly what the code is doing and enable functionality like jump to definition. It would help people understand the system at a faster pace.

list widget setSearchTerm

As an example, this is what my IDE currently knows about the method `setSearchTerm` (which is pretty much just listing any class having the method). While we could tell the IDE it's the first one.

daftspunk: Is there no way to train the makeWidget method to have a return type of the first argument?

jacobdekeizer: @daftspunk nog that I know of, since we don't have generics in PHP.

daftspunk: If you change it from returning WidgetBase to a generic object, will it stop the errors?

The proposed solution is a WET monkey patch

jacobdekeizer: @daftspunk No that wouldn't fix the errors. It's good that makeWidget returns WidgetBase, so I wouldn't change that.

daftspunk: It's not returning a WidgetBase. It's returning a generic object that inherits WidgetBase. Unfortunately the IDE is not smart enough to figure this out. I wonder how it would handle magic methods?

The issue is not everyone uses an IDE and this line of code makes no sense, and cannot be verified to be correct, so we will constantly face this problem of "incorrectness". You must treat the problem and not the symptom. Someh

jacobdekeizer: @daftspunk Phpstorm isn't a fan of magic method either 😄 . I don't know how you want me to fix it. If we remove the docblock my IDE warns about this, and not the accessed methods I talked about earlier (sorry for that).
afbeelding
Changing the return type of this method to WidgetBase also feels wrong, then I'm leaning more towards leaving this docblock out and accept the warning of Phpstorm.

daftspunk: I understand. The IDE helpers can be  overly strict which can stifle innovation; the things that make PHP special and October possible in the first place

My preference would be to make the return value generic and sacrifice the intellisense, so we don't keep revisiting this issue. But, since you are leading this initiative as an IDE user, we can follow your preference

It is a shame but hopefully the IDE will improve to support more flexibility in dynamic languages, or maybe we will think of a workable solution in the meantime, one that addresses the problem at its source

jacobdekeizer: @daftspunk what do you think about changing string|void to string|null as I talked about in the pull request comment.

jacobdekeizer: @daftspunk another possibility is to change the return type of `\Backend\Widgets\Lists` to `\Backend\Classes\WidgetBase|\Backend\Widgets\Lists`. Then we havo no errors in the code

LukeTowers: @jacobdekeizer that's worse, because now we have to start listing every single possible implementation of WidgetBase.

jacobdekeizer: @LukeTowers @daftspunk I'm still against removing @var tbh. Why can't we help the ide out so we have good autocompletion. I saw in pr #4960 someone also used @var notation to help the ide out. I don't think it's bad practice, it's the nature of php that we have to do this.

daftspunk: As discussed, the method returns `object|\Backend\Classes\WidgetBase`. We want to resolve IDE errors as first priority, autocomplete is secondary

jacobdekeizer: The @var notation fixes both issues. We fix the ide errors and it returns the correct type.

daftspunk: > The issue is not everyone uses an IDE and this line of code makes no sense, and cannot be verified to be correct, so we will constantly face this problem of "incorrectness". You must treat the cause and not the symptom

jacobdekeizer: @daftspunk if someone doesn't use an ide, then all the code is dus plain text. What difference does that one line then make? Instead all the ide users get good typehinting and autocompletion. We can return WidgetBase, but in my opinion adding `@var` and returning the  Lists object would be so much better and the only right thing to do. But if you guys are so against that one line then I will remove it.

daftspunk: Exactly, it is an opinion that says "Use PHPStorm/IDE". It's not just one line. It's a modification of existing architecture where that line must be introduced everywhere. Even in your own Pull Request, it is used some places and not others, how to verify this? How to maintain this?

Thank you for your understanding

daftspunk:
Thanks @jacobdekeizer!

PR #4990, opened 2020-03-17
add imagePaste Froala Editor option config

mjauvin:
This setting enables/disables pasting images into the RichEditor. The default has been set to allow pasting images, maintaining existing behavior.

This will allow disabling image pasting when using Rainlab.Translate on RichEditor fields to prevent pasting images bigger than the mediumtext database field size in rainlab_translate_attributes table.


bennothommo:
@mjauvin Personally, I'd be more keen to see [my suggestion](https://github.com/rainlab/translate-plugin/issues/362#issuecomment-493821086) implemented, as that would be a net benefit for everyone, not just Translate plugin users.

mjauvin:
@bennothommo how exactly do you propose to do this? The Froala settings you mention do not allow for this.

bennothommo:
@mjauvin We can probably hijack [this event](https://froala.com/wysiwyg-editor/docs/events/#image.beforePasteUpload) to do the upload and add the image in.

mjauvin:
@bennothommo Now THAT is interesting... I'll see what I can do with this, thanks.

mjauvin:
Regarless of what we end-up doing in Rainlab.Translate with the image pasting issue, I feel this PR should still be merged since it gives more flexibility to the RichEditor formwidget.

Thanks.

LukeTowers:
@mjauvin could you submit a PR to the docs repo documenting the use of that option?

mjauvin:
> @mjauvin could you submit a PR to the docs repo documenting the use of that option?

@LukeTowers None of the other Froala options are documented for the RichEditor... where do you want me to add this? It's merely a partial variable defined in prepareVars() method as is `useMediaManager`

mjauvin:
@LukeTowers What's preventing a merge of this PR?

LukeTowers:
@mjauvin I just tested this by going to https://google.com, right clicking the image, selecting "Copy Image" and pasting it into the richeditor field in static pages. The image was automatically uploaded to the server, it was not stored as a data URL.

Please provide replication instructions for the issue (taking into account my recent changes to the upload process: https://github.com/octobercms/october/pull/5026?

mjauvin:
@LukeTowers this is just a setting to *prevent* image pasting, this does NOT convert a pasted image to an uploaded image in media manager...

LukeTowers:
@mjauvin I know, but I'm saying I think it's useless, since in my testing pasting an image uploaded it as a file in the code we have right now; it didn't turn it into a data encoded image. What steps are required to trigger that behavior for testing?

mjauvin:
@LukeTowers if you use the "test#4990" container on octodock labs, you'll be able to test this.

Just change the "$allowImagePaste" value to test this.

Note: I've merged latest develop branch in my PR & also included the compiled build-plugins-min.js file in my PR

mjauvin:
You can use the Pages plugin to create a static page and try pasting in it.

LukeTowers:
@mjauvin where are you copying images from? I copied an image from the browser using "Copy image" and pasted it, when I pasted the image it was automatically uploaded. It was not turned into a data URI, which is the original reported problem.

mjauvin:
@LukeTowers did you modify the $allowImagePaste?

I copied from Google's search page.

mjauvin:
I just changed the $allowImagePaste to `false` and tested it again, it's working fine.

I left if with `false` so you can see for yourself.

LukeTowers:
@mjauvin what's the link? I can't see your containers in octodock.

mjauvin:
Sorry @LukeTowers I thought we could see each other's instances...

https://b85f07511403.octodock.com/backend

LukeTowers:
@mjauvin your code works yes. But you still haven't responded to my initial comment.

The original issue wants a solution to deal with images that are getting pasted **and turning into embedded data URIs** since that means that the image is now embedded in the database column used, which is bad.

However, my testing has revealed that pasting an image **does not** turn it into a data URL, it instead uploads the image as if the user had downloaded the image and then used the image uploader like normal to insert it, thus invaliding the original complaint. Am I missing something?

mjauvin:
> @mjauvin your code works yes. But you still haven't responded to my initial comment.
> 
> The original issue wants a solution to deal with images that are getting pasted **and turning into embedded data URIs** since that means that the image is now embedded in the database column used, which is bad.
>
> However, my testing has revealed that pasting an image **does not** turn it into a data URL, it instead uploads the image as if the user had downloaded the image and then used the image uploader like normal to insert it, thus invaliding the original complaint. Am I missing something?

@LukeTowers You're definitely right, I did not re-test the issue since it was first opened. Did your last changes to the RichEditor change something that would have fixed the issue?

mjauvin:
@LukeTowers I went ahead and tested this without your RichEditor changes and it's the same... so I guess we could close the original issue, I just cannot reproduce.

mjauvin:
I would still want to merge this PR though, as that can be usefull for a implementing (or extending) a formwidget which you do not want to allow pasting images, for example.

LukeTowers:
@mjauvin you should already be able to do that yourself in a plugin by extending the Froala defaults just like you would to add a new plugin / button to Froala

Issue #4956, opened 2020-02-26
Speedup bleeding edge updates releases


ayumi-cloud:
The reason to open this issue up is because I would like to request a change in the way the October team handles `Milestones` in this repo.

I read a statement from daftspunk in Slack saying releases should be every 1-2 weeks.

If we look at the current `Milestone v1.0.465` see here: https://github.com/octobercms/october/milestone/12

It currently says the following info: `36 open and 28 closed`.

### The problem

- The problem is pretty simple, the current method doesn't work and it's **_very slow_**!

- There is a **_long delay_** between each release to complete most of the Milestone.

- Another issue is `testing`.

- My finished pull requests are now sitting there for many weeks! They should have been added to the `bleeding edge updates` section! Instead they will continue to sit there for more weeks until the **_Milestone is full_**.

- Many people are working on projects for clients and discover a issue or bug in October and raise an issue and then submit a pr. It affects their relationship with their clients, having created a pr and it sits there! When they need the pr added to the cms via bleeding edge updates to continue work on their projects.

### Solution

I want to request the following solution it's pretty simple!

Let's break up the Milestone's into groups of max `10 pull requests`.

They can get released into the `bleeding edge updates` quickly every 1-2 weeks.

### Testing

In real life you can't expect every release to contain no bugs or errors - that's not how the real world works!

The pull requests sitting in the Milestone with the labels `Testing needed` have been checked by the admins and there is a really small chance of an error! There is no need to have them sitting around waiting for even more testing before adding them to the `bleeding edge updates` and then having another round of testing!

![image](https://user-images.githubusercontent.com/57409060/75362812-303d4980-58b1-11ea-9e65-42efeec0053c.png)

All `testing needed` labels should not just sit there and do nothing! Remove this pointless step and 
 push them out to the `bleeding edge updates` and let's get the October community to test them in the wild.

### Grey area

Having Milestone pull requests just sit there with `testing needed` labels is not working, no one can be bothered with this step.

Forcing these pull requests into the Bleeding updates will get people more pro-active into the testing process and help ease the work-load for the admins.

### Bleeding updates vs stable updates

Bleeding updates should allow a small set of errors! This way the community can give feedback and increase the speed of the stable updates!

Currently, Bleeding updates are being treated like Stable updates and trying to remove every single error! This is an overkill and a bit ocd.

### Discussion and action

1. Limit `Milestone` pull requests to 10.

[...]

PR #4991, opened 2020-03-18
Clarification of description of using 'bleeding edge updates' in 'production' mode

ayumi-cloud:
As per discussion here: https://github.com/octobercms/october/issues/4956

### DIFF
--- a/config/cms.php
+++ b/config/cms.php
@@ -21,7 +21,7 @@ return [
     | If you are developing with October, it is important to have the latest
     | code base. Set this value to 'true' to tell the platform to download
     | and use the development copies of core files and plugins. This feature
-    | is not intended for use on production mode websites.
+    | is not intended for use on production websites.
     |
     */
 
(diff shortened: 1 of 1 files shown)


config/cms.php
```diff
@@ -20,7 +20,8 @@
     |
     | If you are developing with October, it is important to have the latest
     | code base. Set this value to 'true' to tell the platform to download
-    | and use the development copies of core files and plugins.
+    | and use the development copies of core files and plugins. This feature
+    | is not intended for use on production mode websites.
```
bennothommo: @ayumi-cloud
```suggestion
    | is not intended for use on production websites.
```

LukeTowers:
I disagree with this. I would prefer that more people used edgeUpdates in production, at least if they have development resources to deal with a problem if one comes up. We strive to make even the pending builds as stable as possible, so I wouldn't necessarily agree with warning people away from using edge builds.

Now, if you're wanting to use the develop branch in production, you should probably be pretty comfortable with October (I myself have all of my client sites running the latest develop branch every time I work on them), but edgeUpdates should be safe for most people with any development resources on hand.

PR #4992, opened 2020-03-18
Add adaptive popup size options to Storm UI docs

LarBearrr:
Popups can be "adaptive" meaning they fill the width and/or height of the screen. This PR adds this information to the docs to make it better known.

This is useful when popups may contain a form, or some other large amount of information that the current listed sizes will not accomodate.


LukeTowers:
Huh, neat. Didn't realize this was an option. Could you post a screenshot of what that does?

LarBearrr:
I stumbled across it making a custom media manager form widget. I'll get some screens tomorrow 😄

LarBearrr:
Here's a screen shot of an adaptive popup:
image

I'm not using the adaptive height option here, but it would fill the height of the screen if it was set. 

The best example of this is the Media Manager popup when used as a form widget.

LukeTowers:
Thanks @LarBearrr!

Issue #4996, opened 2020-03-23
dependsOn isn't triggered on relationManager create and update pivot table with ManyToMany Relationship


yfktn:
- OctoberCMS Build: ### 464
- PHP Version: PHP 7.1.3
- Database Engine: MariaDB
- Plugins Installed: General

### Description:
When I create or update value of pivot table with many to many relationManager, existing fields that depends on the value doesn't updated because filterFields() doesn't called.

### Steps To Reproduce:
Suppose field:
```
total_price:
   label: 'Total Price'
   span: auto
   type: number
   readOnly: true
   dependsOn:
      - items_bought
items_bought:
   path: $/panatau/persewaan/controllers/booking/_load_items.htm
   span: full
   context:
      - update
      - preview
    type: partial
```

I fix it in **modules\backend\behaviors\relationcontroller\partials\\_pivot_form.htm**, in **Form::ajax** I add property **data-request-success** to triggered **oc.relationBehavior.changed**.

PR #4997, opened 2020-03-23
filterFields triggered with Add or Change with pivot

yfktn:
When any pivot value change or add with relationManager, the filterFields will triggered too, fix for #4996


mjauvin:
That actually solves a similar problem I was having with one of my form with a Pivot relation.

Nice work @yfktn !

mjauvin:
I tested it and it works as "advertized" :)

yfktn:
Glad to hear that @mjauvin , I tested it in my production server too :)

PR #5000, opened 2020-03-25
Add support for html attributes in the section formfield

Klaasie:
This PR will modify the section formfield to add html attributes to its main html tag.

This will be particulary usefull for adding extra classes or data attributes to be handled by custom javascript.

### DIFF
--- a/modules/backend/widgets/form/partials/_field_section.htm
+++ b/modules/backend/widgets/form/partials/_field_section.htm
@@ -1,7 +1,7 @@
 <!-- Section -->
 <div class="field-section">
     <?php if ($field->label): ?>
-        <h4><?= e(trans($field->label)) ?></h4>
+        <h4 <?= $field->getAttributes() ?>><?= e(trans($field->label)) ?></h4>
     <?php endif ?>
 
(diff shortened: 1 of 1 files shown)


LukeTowers:
@Klaasie can you provide an example of what you'd use this for on the section field? The section field is really meant to be a "section title", so I can't really think of any attributes that would actually serve a useful purpose.

Klaasie:
Similar to the PR https://github.com/octobercms/october/pull/5003 I was also checking this form field.

For the attributes definition I can think of a (global) usecase however.
Similar to how OC handles intialization of javascript objects through data tags I like to do the same.
By consistently applying the attributes definition within form fields I will be able to initialise custom javascript objects through yaml definitions.

I don't have a particular usecase to the sections form field though.

I will be making more PRs in the future where the attributes will be applied to all form fields. The PRs of today are just how far I've gotten today :)

LukeTowers:
@Klaasie if we don't have a specific use case for this field lets leave it alone for now then.

Klaasie:
@LukeTowers Fair enough!

PR #5002, opened 2020-03-25
Colorpicker improvements

Klaasie:
This PR adds disabled / readonly support to the colorpicker formwidget.
Additionally it adds support for html attributes as defined in form field yaml file.

### DIFF
--- a/modules/backend/formwidgets/ColorPicker.php
+++ b/modules/backend/formwidgets/ColorPicker.php
@@ -43,6 +43,16 @@ class ColorPicker extends FormWidgetBase
      */
     public $showAlpha = false;
 
+    /**
+     * @var bool If true, the color picker is set to read-only mode
+     */
+    public $readOnly = false;
+
+    /**
+     * @var bool If true, the color picker is set to disabled mode
+     */
+    public $disabled = false;
+
     //
     // Object properties
     //
@@ -61,6 +71,8 @@ class ColorPicker extends FormWidgetBase
             'availableColors',
             'allowEmpty',
             'showAlpha',
+            'readOnly',
+            'disabled',
         ]);
     }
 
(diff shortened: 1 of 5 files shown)


LukeTowers:
@Klaasie could you provide a before / after screenshot for this one as well?

Klaasie:
Before:
Schermafbeelding 2020-03-26 om 09 07 18
Schermafbeelding 2020-03-26 om 09 07 37

After:
Schermafbeelding 2020-03-26 om 09 08 05
Schermafbeelding 2020-03-26 om 09 08 17

Click listeners have been been disabled and the spectrum disabled option has been set (https://bgrins.github.io/spectrum/#options-disabled).

LukeTowers:
@Klaasie What do you think about changing the disabled / readonly version to only show the colour that was selected instead of all of them?

Klaasie:
@LukeTowers Yeah, that could be done. What do you think should be shown when no color is selected?

LukeTowers:
@Klaasie hmm, excellent question. Perhaps a box with the slash through it? @bennothommo any ideas?

bennothommo:
@Klaasie @LukeTowers I think the color picker has an "empty" state - you can see it by changing the "availableColors" parameter to an empty array. If I recall correctly, a box with a slash through it is generally used to denote an "invisible" colour (if you allow alpha control).

Klaasie:
It does. I suppose instead of showing all colors disabled we could just show the spectrum one in that case with its initial empty state, disabled.

See https://bgrins.github.io/spectrum/#toc25 for more information.

LukeTowers:
Sounds good to me @Klaasie

Klaasie:
@LukeTowers @bennothommo 

I've made changes per your recommendations. I ran into some issues with the default spectrum styling so I applied some custom styling instead to the color picker box. See the screenshots attached.

Colorpicker empty disabled/readonly:
colorpicker_empty_disabled

Colorpicker filled disabled/readonly:
colorpicker_filled_disabled

Let me know what you think :)

LukeTowers:
@Klaasie I don't think the disabled state of a colorpicker with a color selected should include the "x", if you remove that it should be fine.

PR #5003, opened 2020-03-25
Modify password field to support placeholder attribute

Klaasie:
This PR adds support for the placeholder tag in the password field.

### DIFF
--- a/modules/backend/widgets/form/partials/_field_password.htm
+++ b/modules/backend/widgets/form/partials/_field_password.htm
@@ -7,6 +7,7 @@
         name="<?= $field->getName() ?>"
         id="<?= $field->getId() ?>"
         value=""
+        placeholder="<?= e(trans($field->placeholder)) ?>"
         class="form-control"
         autocomplete="off"
         <?= $field->hasAttribute('maxlength') ? '' : 'maxlength="255"' ?>


LukeTowers:
Best practice seems to be to avoid placeholders as much as possible, especially in password fields. Could you provide an example of how you're wanting to use this?

https://ux.stackexchange.com/questions/85050/what-is-the-best-practice-for-password-field-placeholders

Klaasie:
I don't have a particular usecase for this PR. I've been testing most of the form field options to check their implementation global form field options (https://octobercms.com/docs/backend/forms#form-field-options) and noticed that the password field was missing the placeholder.

Personally I think as a user / developer I'd still like the opportunity to add the placeholder to the password field if I wish to do so (despite what best practices might tell me :) ). For inspiration on the matter I like to look at the "giants" to see how they work and some of the bigger ones (Facebook, LinkedIn and a couple of dutch e-commerce websites) all seem to use the placeholder attribute. Google does a combination where the placeholder gets transitioned to a label.

LukeTowers:
@Klaasie Could you provide examples of the placeholder use by those sites (except for the use where a placeholder turns into a label, there's a lot of information out there that explains why that's a terrible UX practice).

Klaasie:
@LukeTowers Sure! These are a couple of examples.

Facebook (Actually they implement both ways, login uses labels instead.)
image

LinkedIn:
image

Bol.com (popular dutch e-commerce website)
image

AliExpress
image

I suppose ultimately it comes down to preference anyway. If the October team feels they don't want to allow a placeholder in a password field then this PR can be discarded. Again, personally, since the field allows the attribute to be placed I don't think we should limit the user to actually implement it.

LukeTowers:
@Klaasie fair enough, just wanted to play devil's advocate.

PR #5004, opened 2020-03-25
Markdown formwidget improvements

Klaasie:
This PR adds readonly and disabled support to the markdown editor formwidget.
Additionally it adds support for setting html attributes through the form fields yaml.

_The .less file for the markdown editor was not included in the asset compiler. I've added it so that it's in line with other formwidgets._

### DIFF
--- a/modules/backend/ServiceProvider.php
+++ b/modules/backend/ServiceProvider.php
@@ -78,6 +78,7 @@ class ServiceProvider extends ModuleServiceProvider
             $combiner->registerBundle('~/modules/backend/formwidgets/richeditor/assets/js/build-plugins.js');
             $combiner->registerBundle('~/modules/backend/formwidgets/colorpicker/assets/less/colorpicker.less');
             $combiner->registerBundle('~/modules/backend/formwidgets/permissioneditor/assets/less/permissioneditor.less');
+            $combiner->registerBundle('~/modules/backend/formwidgets/markdowneditor/assets/less/markdowneditor.less');
 
             /*
              * Rich Editor is protected by DRM
--- a/modules/backend/formwidgets/MarkdownEditor.php
+++ b/modules/backend/formwidgets/MarkdownEditor.php
@@ -27,6 +27,16 @@ class MarkdownEditor extends FormWidgetBase
      */
     public $safe = false;
 
+    /**
+     * @var bool If true, the editor is set to read-only mode
+     */
+    public $readOnly = false;
+
+    /**
+     * @var bool If true, the editor is set to read-only mode
+     */
+    public $disabled = false;
+
     //
     // Object properties
     //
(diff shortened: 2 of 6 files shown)


LukeTowers:
@Klaasie could you add before / after screenshots of this change?

Klaasie:
@LukeTowers I've made no changes to the default state of the editor. I've modified the javascript to disable click events on the toolbar and set the ace editor to readonly. Added a `disabled` class that counters the hover states and added a `cursor: not-allowed` to several classes to give the visual feedback that this editor can't be clicked.

(if you'd still like a before and after screenshot I can make one later when I'm back on my macbook. Windows won't allow me to make screenshots with the cursor visible.)

LukeTowers:
Screenshot would be helpful @Klaasie, thanks!

Klaasie:
Before:
Schermafbeelding 2020-03-26 om 09 01 00
Schermafbeelding 2020-03-26 om 09 01 10

You can see, the cursor is a text icon and the toolbar buttons have hover and tooltip states.

After:
Schermafbeelding 2020-03-26 om 08 57 52
Schermafbeelding 2020-03-26 om 08 58 03

The cursor shows that this field is disabled, hovers are not showing and tooltips are disabled.

LukeTowers:
Can we just hide the toolbar entirely when it's disabled / readonly?

Klaasie:
@LukeTowers Yeah can do. See the screenshot below how that'd look.

image

Would you prefer this over blocking any actions within the toolbar through javascript?

LukeTowers:
hmm, I almost prefer just rendering the content by itself. What does the richeditor look like in disabled / readonly mode?

Klaasie:
image

This is what the richeditor looks like disabled. Hover effects on the toolbar still work but clicking them does nothing. So does clicking on the content. The cursor does not change, so there's no visual feedback there.

LukeTowers:
@Klaasie thanks for looking into that. Can we go with the original suggestion of showing the buttons but just make the colours grey and more muted like the richeditor example?

Klaasie:
@LukeTowers I've modified the colors to be the same as the richeditor.
There is still two differences compared to the richeditor:
1. I've counteracted hover states so you don't get the visual feedback that the button can be clicked.
2. I've kept the `cursor: not-allowed` styling just like browsers will do by default on disabled fields.

I personally find those 2 differences make sense from a "disabled" point of view. Once I get to looking into the richeditor I can look into making that toolbar work the same as the markdown toolbar does now.

Additionally I muted the text color as well and I've removed the cursor line indicator that you could still see in my screenshots 2 days ago.

Attached are 2 screenshots of my latest changes.

markdowneditor_empty_disabled
markdowneditor_filled_disabled

LukeTowers:
@Klaasie I would prefer the content section does not get muted, but I like the toolbar changes (and the cursor in the content field so long as you actually can't type anything into it)

PR #5006, opened 2020-03-26
Allow add or remove field on form.refreshFields event

RomainMazB:
This allow the developper to add or remove fields naturally using addFields or removeField on the backend.form.refreshFields event.

A use case is to display some additionnal fields pre-filled(using `default` field properties) with another model data (which is my actual case).

Here is an example of  use:
```php
// plugins/romainmazb/realestate/controllers/Lands.php
Event::listen('backend.form.refreshFields', function ($formWidget, $allFields) {
    $formWidget->removeField('tile_front');
    $formWidget->removeField('tile_back');
    $formWidget->addFields([
        'my_field1' => [
            'label'   => 'My Field 1',
            'comment' => 'This is a custom field I have added.',
            'default' => 'Some custom value'
        ],
    ]);
    $formWidget->addFields([
        'my_field2' => [
            'label'   => 'My Field 2',
            'comment' => 'This is a custom field I have added.',
            'default' => 'Some other value'
        ],
    ]);
});
```
And here is the result which can't be done without this changes since the actual source code only trigger makePartial on the field who are linked with dependsOn and don't even look if the $this->allFields changed:
ezgif com-optimize

Hope it will be helpful to the community.

### DIFF
--- a/modules/backend/widgets/Form.php
+++ b/modules/backend/widgets/Form.php
@@ -383,6 +383,7 @@ class Form extends WidgetBase
         $this->setFormValues($saveData);
         $this->prepareVars();
 
+        $fieldsBeforeRefresh = $this->allFields;
         /**
          * @event backend.form.refreshFields
          * Called when the form is refreshed, giving the opportunity to modify the form fields
@@ -402,6 +403,13 @@ class Form extends WidgetBase
          */
         $this->fireSystemEvent('backend.form.refreshFields', [$this->allFields]);
 
+        $fieldsStructureModifications = array_merge(array_diff_key($fieldsBeforeRefresh, $this->allFields), array_diff_key($this->allFields, $fieldsBeforeRefresh));
+        /**
+         * Does any fields has been removed or added?
+         */
+        $formStructureModified = (bool)$fieldsStructureModifications;
+        
+
         /*
          * If an array of fields is supplied, update specified fields individually.
          */
@@ -420,7 +428,7 @@ class Form extends WidgetBase
         /*
          * Update the whole form
          */
-        if (empty($result)) {
+        if (empty($result) || $formStructureModified) {
             $result = ['#'.$this->getId() => $this->makePartial('form')];
         }
 


RomainMazB:
And here is my finished use case, I added a switch box to "disable" the whole form refresh which can be annoying in some case just to show what is possible.
Here is an example of  use:
```php
// plugins/romainmazb/realestate/controllers/Lands.php
public function __construct()
{
    parent::__construct();
    BackendMenu::setContext('RomainMazB.RealEstate', 'main-menu', 'lands-menu');

    Event::listen('backend.form.refreshFields', function ($formWidget, $allFields) {
        // Disable refreshing with a simple switch box
        if( ! input('Land._activate_refresh_description'))
            return;

        $selected_descriptions = $this->getDescriptionForSelected();
        if(count($selected_descriptions)) {
            $fields = [];
            foreach($selected_descriptions as $house_model) {
                $fields["house_model///" . $house_model['id']] = [
                    'label'   => $house_model['name'],
                    'type' => 'Backend\FormWidgets\RichEditor',
                    'size' => 'huge',
                    'default' => $house_model['description'],
                    'cssClass' => 'tw-font-black'
                ];
            }
        } else {
            $fields["no_description"] = ['type' => 'section', 'comment' => 'Aucun modèle sélectionné'];
        }

        $formWidget->addFields($fields);
    });
}

public function getDescriptionForSelected() {

    $selected = input('Land._selected_house_models', []);

    if(empty($selected))
        return [];

    return HouseModel::select(['id', 'name', 'description'])
                                    ->whereIn('id', $selected)
                                    ->get()
                                    ->toArray();
}
```
And here is the result:
ezgif com-video-to-gif

mjauvin:
@RomainMazB is there a cleaner way to detect changes in fields after the event is fired?

mjauvin:
Like: 
```
$updated = $fieldsBeforeRefresh != $this->allFields;
```

RomainMazB:
I didn't find anyone, I know the `array_merge(array_diff_key(),array_diff_key())` is really ugly and perform loop where in 99% case you will not need but it's the afaik the best solution to be sure to detect any key changes in two array:

array_diff_key only detects the keys who's not present in the second array, so we are forced to perform it twice to intersect the results with `array_merge`

I can have done that with `array_intersect_key` which catches the keys who are matching, but still, we are forced then to "inverse" the selection with another loop, not so better.

I spent two hours to search another way to do just that line. Open to any idea who can solved that, I will be the first who'll enjoy it :)

I can create a function to be more eyefriendly but that doesn't change the fact it's ugly :]

The problem with the `== ` comparator is that any change in the array such as a value, a required parameter or any other changes would be triggered as a change.

I can also add a parameter to the formWidget to perform this loop just when needed and mentioned by the developer.

Something like:
```php
/**
* @var bool Watch for fields changes on refresh event.
*/
public $structureMetamorph = false;

// ...

$fieldsBeforeRefresh = $this->allFields;
$this->fireSystemEvent('backend.form.refreshFields', [$this->allFields]);
if($this->structureMetamorph) {
    $fieldsStructureModifications = array_merge(array_diff_key($fieldsBeforeRefresh, $this->allFields), array_diff_key($this->allFields, $fieldsBeforeRefresh));
}
        
/**
 * Does any fields has been removed or added?
 */
$formStructureModified = (bool)$fieldsStructureModifications ?? false;
```
then from the controller we could activate the metamorph behavior:
```php
Event::listen('backend.form.refreshFields', function ($formWidget, $allFields) {
    $formWidget->structureMetamorph = true;
    $formWidget->addFields();
    $formWidget->removeField('someField');
}

**Comment by mjauvin** [comment]
Try the test I provided above, seems to work for me...

**Comment by RomainMazB** [comment]
I [have done a test](http://sandbox.onlinephpfunctions.com/code/080829ed05b6de1bf1b3d315661fc92740d08baa), look at itt

This would happen when just a field parameter have changed(which is why the event "backend.form.refreshFields" is for).

**Comment by mjauvin** [comment]
@RomainMazB don't you want the form to update if you change fields values in the `backend.form.refreshFields` event handler?

**Comment by RomainMazB** [comment]
I do, but that behavior is already handled with the previous event:
```php
$dataHolder = (object) ['data' => $saveData];
$this->fireSystemEvent('backend.form.beforeRefresh', [$dataHolder]);
$saveData = $dataHolder->data;
```

RomainMazB:
Just a little bump with more informations.

I tried to do without modifying the file directly and try to override/extend it properly.

The first step obviously were to override the beforeRefresh method to implement the behavior I need. So I've created a custom FormWidget:
```php
<?php

namespace RomainMazB\RealEstate\Classes;

use Backend\Widgets\Form;

class FormWidget extends Form
{
    public function onRefresh()
    {

        // ... Useless lines for the demonstration

        $fieldsBeforeRefresh = $this->allFields;
        $this->fireSystemEvent('backend.form.refreshFields', [$this->allFields]);

        $fieldsStructureModifications = array_merge(
            array_diff_key($fieldsBeforeRefresh, $this->allFields),
            array_diff_key($this->allFields, $fieldsBeforeRefresh)
        );

        // ... Useless lines for the demonstration

        if (empty($result) || $formStructureModified) {
            $result = ['#'.$this->getId() => $this->makePartial('form')];
        }

        // ... Useless lines for the demonstration
    }
}
```

But since the original FormWidget is a "top-level one", I faced up an issue because the instantation of the `$formWidget` [is hard-coded in `\Backend\Behaviors\FormController`](https://github.com/octobercms/october/blob/master/modules/backend/behaviors/FormController.php#L145). So I had to override this method too:
```php
<?php


namespace RomainMazB\RealEstate\Classes;

use Backend\Behaviors\FormController as BaseFormController;

class FormController extends BaseFormController
{
    public function initForm($model, $context = null)
    {

        // ... Useless lines for the demonstration
        
        $this->formWidget = $this->makeWidget('RomainMazB\RealEstate\Classes\FormWidget', $config);

        // ... Useless lines for the demonstration
    }
}
```
Then for sure, I faced another issue because the form widget partials were not found. So I had to copy the whole [`october/modules/backend/widgets/form/partials/`](https://github.com/octobercms/october/tree/master/modules/backend/widgets/form/partials) into `/plugins/romainmazb/realestate/classes/formwidget/partials`

Now I get it to work, but I lose the updates for some really important class/partials from october.

Hope this pull request will be merged, let me know if you want me to make more modification, maybe the addition of the structureMetamorph or whatever named propertie discussed at the end of [this comment](https://github.com/octobercms/october/pull/5006#issuecomment-604487054) to not trigger this behavior when not needed (the default would be false).

If not merged, this comment will stand as a How-To :)

PR #5010, opened 2020-03-29
MediaFinder modifications

Klaasie:
This PR modifies two things:

1. The form field attributes (as set by the fields.yaml) are added to the file, image and preview divs. (_I've included it in the preview version since the preview version is used for the disabled state as well_)
2. I've added the formfield readOnly value to the check if the preview mode should be rendered. By doing this the readOnly tag is now supported as well and works the same as the disabled state.

Consequences:
The formfield attributes array will also contained "disabled" and "readonly" attributes if they are set through their corresponding form field options. When `disabled="disabled"` is added to the preview  `span` the css will make it looklike a disabled form field (see before and after images).

Personally I don't think this change is a bad thing, but I thought I should mention.

Before:
image

After:
image

### DIFF
--- a/modules/backend/formwidgets/mediafinder/partials/_mediafinder.htm
+++ b/modules/backend/formwidgets/mediafinder/partials/_mediafinder.htm
@@ -1,6 +1,6 @@
 <?php if ($this->previewMode && !$value): ?>
 
-    <span class="form-control"><?= e(trans('backend::lang.form.preview_no_media_message')) ?></span>
+    <span class="form-control" disabled="disabled"><?= e(trans('backend::lang.form.preview_no_media_message')) ?></span>
 
 <?php else: ?>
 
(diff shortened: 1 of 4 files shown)


modules/backend/formwidgets/mediafinder/partials/_mediafinder.htm
```diff
@@ -1,6 +1,6 @@
 <?php if ($this->previewMode && !$value): ?>
 
-    <span class="form-control"><?= e(trans('backend::lang.form.preview_no_media_message')) ?></span>
+    <span class="form-control" <?= $field->getAttributes() ?>><?= e(trans('backend::lang.form.preview_no_media_message')) ?></span>
```
LukeTowers: Does any other previewMode partial file in October include `$field->getAttributes()`? My concern would be if they don't this would be unexpected behaviour for devs who might be using the attributes property to include data that they wouldn't include in preview mode if they knew that was happening.

Klaasie: Good question. I've quickly checked all form widgets and some formfields and none of them do so. I'll remove it!

LukeTowers:
@Klaasie could you retain the disabled="disabled" attribute in the readonly version of the field?

Klaasie:
@LukeTowers done!

LukeTowers:
@Klaasie what happens if the widget is in previewMode but has a value?

Klaasie:
@LukeTowers See the image below. I've made no changes to that. It renders the field with the specific mode partial and hides all actions (like pick a new file or delete current).


image

PR #5013, opened 2020-03-29
allow bypassing model saving

mjauvin:
If formBeforeSave() or formBefore{Create,Update}() return content, this content is returned. This bypasses the model saving code that follows.

This is very usefull when needing very custom model validation that also need to refresh the form fields when certain conditions are met.


LukeTowers:
This is potentially a breaking change if anyone was incorrectly returning a value from those methods before. Please provide more information on what exactly your use case is.

PR #5014, opened 2020-03-30
Update Controller.php

LeMaX10:
Correction of the answer.

In the latest versions, when using Response::make Ajax responses to some redirect requests are issued as text/html.

Example:
```php
public function onSwitchLocale() {
...
return Redirect::to($pageUrl);
...
}
```

Response text/html:
```bash
HTTP/1.0 200 OK
Cache-Control: no-cache, private
Content-Type:  application/json
Date:          Mon, 30 Mar 2020 01:32:34 GMT

{"X_OCTOBER_REDIRECT":"http:\/\/localhost:8080\/en"}
```

print screen: https://a.radikal.ru/a18/2003/53/6af6e9756a13.png

### DIFF
--- a/modules/cms/classes/Controller.php
+++ b/modules/cms/classes/Controller.php
@@ -777,7 +777,7 @@ class Controller
                     return $result;
                 }
 
-                return Response::make($responseContents, $this->statusCode);
+                return $this->makeResponse($responseContents);
             }
             catch (ValidationException $ex) {
                 /*


bennothommo:
Sorry @LeMaX10, but we cannot accept this PR at this stage. The unit tests break with this change, and you have also removed the HTTP response code attribute which is needed for correct AJAX responses. Your screenshot looks more like an issue with the server itself.

If you revise this fix, please feel free to submit it as a new PR.

LukeTowers:
@LeMaX10 the AJAX framework specifically converts Redirect responses to that response format you provided, and then the redirect is performed on the client side. If you are not using framework.js then you will need to replicate that behaviour on your client.

PR #5015, opened 2020-03-30
ImportExportBehavior: Add iconv fallback support for import

Rike-cz:
There are still many ERPs that are able to export CSV only with Windows-1250 encoding in our country, so I made proposal of simple solution for other possible encodings that are not supported by mbstring lib.


LukeTowers:
Related: https://github.com/php-mime-mail-parser/php-mime-mail-parser/pull/213. @Rike-cz do you think you could add a unit test for both those branches (properly converting from a format mb_convert_encoding and properly converting from Windows-1250).

PR #5018, opened 2020-03-31
Added auth manager strings

SaifurRahmanMohsin:
This commit is added to support the PR made to octobercms/library#483.


LukeTowers:
This has been proposed before (see octobercms/library#318) and will not be accepted for the following reasons:

1. This translation should happen in the backend module layer, not the core authentication library
2. Specific reasons for authentication failure are **only** provided while the application is in debug mode and **only** for debugging purposes, they should **never** be used in production. Giving specific reasons for authentication failure in production only makes attackers lives easier.

PR #5025, opened 2020-04-02
Fix the richeditor image uploading issue

fansaien:
The richeditor.js sends the X_OCTOBER_MEDIA_MANAGER_QUICK_UPLOAD when uploading the image: https://github.com/octobercms/october/blob/da4f1705ca8c9fa93a0d56f4158737574116be18/modules/backend/formwidgets/richeditor/assets/js/richeditor.js#L160

The `MediaManager->checkUploadPostback()` can save the file_data.

So, we need to bind the `MediaManager` to the controller.

### DIFF
--- a/modules/backend/classes/Controller.php
+++ b/modules/backend/classes/Controller.php
@@ -153,9 +153,14 @@ class Controller extends ControllerBase
         /*
          * Media Manager widget is available on all back-end pages
          */
-        if ($this->user && $this->user->hasAccess('media.*')) {
-            $manager = new MediaManager($this, 'ocmediamanager');
-            $manager->bindToController();
+        if ($this->user) {
+            //RichEditor posts X_OCTOBER_MEDIA_MANAGER_QUICK_UPLOAD when upload the image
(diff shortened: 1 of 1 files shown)


mjauvin:
Can you be more specific about what issue you're talking about?

LukeTowers:
@mjauvin he means that users without permission to access the media manager (i.e. `media.*`) can't upload images in Froala because it relies on the MediaManager which isn't bound to the controller unless the user has the permission to access it.

We won't merge this as it represents a security vulnerability, since as soon as the MediaManager is added to the controller all of its methods are available without restriction, meaning that someone without permissions to manage the media library could still do so by crafting requests that just included `X_OCTOBER_MEDIA_MANAGER_QUICK_UPLOAD` would allow the rest of the widget to be controlled.

However, we could implement the upload logic in the RichEditor FormWidget directly instead to specifically handle richeditor uploads.

LukeTowers:
@fansaien https://github.com/octobercms/october/pull/5026 should fix it, I will be syncing the composer packages sometime next week so feel free to apply those changes manually in the meantime.

fansaien:
Thank you, I will apply that

PR #5030, opened 2020-04-06
Install any required plugins with theme through "theme:install"

bennothommo:
This change will make the `theme:install` Artisan command also install any plugins that are required by the theme, shortcutting the bootstrapping or deploying of a site with a theme.

This works by determining if any plugins are required by the theme, and if so - and these plugins are missing from the current install, the command will run subsequent `plugin:install` commands to install the missing plugins.

### DIFF
--- a/modules/system/console/ThemeInstall.php
+++ b/modules/system/console/ThemeInstall.php
@@ -1,12 +1,15 @@
 <?php namespace System\Console;
 
-use File;
 use Cms\Classes\Theme;
 use Cms\Classes\ThemeManager;
-use System\Classes\UpdateManager;
+use Exception;
+use File;
+use Yaml;
+use Artisan;
 use Illuminate\Console\Command;
 use Symfony\Component\Console\Input\InputArgument;
-use Exception;
+use System\Classes\PluginManager;
+use System\Classes\UpdateManager;
 
 /**
  * Console command to install a new theme.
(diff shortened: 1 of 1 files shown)


LukeTowers:
@bennothommo can you look at https://github.com/octobercms/october/blob/develop/modules/system/controllers/Updates.php#L874-L905 and see if there's anyway we can consolidate these two locations into using the same code?

bennothommo:
@LukeTowers it might be tricky - the Updates controller is using `UpdateManager` which is querying the October marketplace for most info, and checks dependencies *before* the theme is downloaded. Whereas, the console command downloads the theme and *then* checks dependencies.

I suppose an argument could be made to bring the command more inline with how the controller does it, but the controller is reliant on the AJAX framework, so we'd have to replicate the "steps" in the command.

bennothommo:
We could potentially refactor the `UpdateManager` to handle a lot more of the functionality that has been added to the controller, and perhaps introduce some sort of `output` feed that can work with either the AJAX framework, or a CLI output, but such a task is a) a lot of effort and, b) would likely require approval from @daftspunk as this is very much veering into the realm of affecting the marketplace functionality.

LukeTowers:
@bennothommo what aspects of the marketplace functionality would be affected by that suggestion?

bennothommo:
@LukeTowers none per se, it wouldn't affect the marketplace directly - it's more that it's so tied to the marketplace API that I'd imagine he might reservations of the functionality being substantially changed.

LukeTowers:
Would the functionality be changed or just the implementation of it?

bennothommo:
It's more an implementation change, I'd say.

LukeTowers:
I think it should be fine. I'd personally prefer that the logic was centralized in the UpdateManager class and the controller / console were just accessing the same core business logic to do their jobs.

bennothommo:
Closing as we'll implement this differently.

PR #5031, opened 2020-04-07
Unsuspend a backend user from user update screen.

ashleighsims:
Give the ability to easily "unsuspend" another backend user when they have been suspended, via a button in the administration panel. This has been added since the addition of the "unsuspend" button in the users plugin. 

backend-user-unsuspend


mjauvin:
@ashleighsims why did you close this? I think it's a good addition.

ashleighsims:
@mjauvin I closed this due to making the commits from the wrong Github account. Need to change that over.

Issue #4819, opened 2019-12-10
UI issues on Firefox


gergo85:
OctoberCMS: 460 (and 459, 458, ...)
Firefox: 71 (and 70, 69, ...)

1. There is a dotted border around the tab title. **FIXED**
![issue1](https://user-images.githubusercontent.com/2959112/70525582-fda07200-1b47-11ea-9b70-e9025fead2ce.JPG)

2. There is a dotted border around the menu, when it is active. **FIXED**
![issue2](https://user-images.githubusercontent.com/2959112/70525588-02652600-1b48-11ea-94b9-87176525667e.JPG)

3. There are two scroll bars, when the page content is long.
![issue3](https://user-images.githubusercontent.com/2959112/70525598-0729da00-1b48-11ea-9eac-fe95ef098e22.JPG)

4. There is an ugly default tooltip, when I move the cursor above the tab or menu title. **FIXED**
![issue4](https://user-images.githubusercontent.com/2959112/70525607-0abd6100-1b48-11ea-8196-e2661b29205c.jpg)
and
![issue5](https://user-images.githubusercontent.com/2959112/70525616-0ee97e80-1b48-11ea-8a46-37f5c3f963cd.jpg)

5. The arrow of source editor bounces, when I move the cursor above it.
![issue6](https://user-images.githubusercontent.com/2959112/70525748-4d7f3900-1b48-11ea-90b7-c34227c7dcb5.JPG)

ayumi-cloud:
I fixed points 1 and 2 and the pr's were merged, I just checked and now all the work I have done is not merged? 

See here part of the work:

https://github.com/octobercms/october/blob/0666e959e6f97ed9a22a33904c4a4cc2affd9fa3/modules/system/assets/ui/vendor/focus-visible/focus-visible.js

https://github.com/octobercms/october/blob/0666e959e6f97ed9a22a33904c4a4cc2affd9fa3/modules/backend/assets/less/layout/mainmenu.less#L14-L31

ayumi-cloud:
I fixed points 1 and 2 and the pr's were merged, I just checked and now all the work I have done is not merged? 

See here part of the work:

https://github.com/octobercms/october/blob/0666e959e6f97ed9a22a33904c4a4cc2affd9fa3/modules/system/assets/ui/vendor/focus-visible/focus-visible.js

and

[...]

PR #4965, opened 2020-03-03
Remove dotted line focus ring in firefox and have a standard one across browsers

ayumi-cloud:
See github issue: https://github.com/octobercms/october/issues/4964

Notes: In v465 we added a blue focus ring to remove the black focus ring in Google Chrome (as Chrome changed their default). The blue focus ring was decided to work in all cross-browsers. Therefore no need the `second` dotted line focus ring (which only displays in Firefox).

PR #5035, opened 2020-04-09
Reset outline to default - roll back #4895

### LINKED ISSUE #4964: Firefox styling error in v465
Running some tests in several browsers with the latest update and come across this error in v465 in Firefox.

See screenshot:

image

When we adjust css and remove the following code lines:

```css
a:focus {
    outline: thin dotted;
    outline: 5px auto -webkit-focus-ring-color;
    outline-offset: -2px;
}
```

Location: https://github.com/octobercms/october/blob/aff120fcec390309e93eae8f195f253623a5ea12/modules/system/assets/ui/storm.css#L93

and

https://github.com/octobercms/october/blob/aff120fcec390309e93eae8f195f253623a5ea12/modules/system/assets/ui/less/site.reset.less#L64-L66

We get the following result:

image

> The 2nd focus ring (white dotted line) has been removed.

Long story short - permission to remove this offending css code?

bennothommo:
Replaces #4965. Rolls back #4895. Fixes #4964 and some of #4819.

The previous attempts to add a stylised "focus" ring, while with the best intentions, did not take into account how "outline" normally works for people using visual aids. Most high contrast software will effect the outline if its available (ie. make it bolder or more prevalant), thus the "outline" property is paramount to maintaining accessibility.

The previous changes also present issues with elements using box-shadows already, and again, on high contrast, box-shadows are no longer rendered.

This change will bring back the outline property for focus, but with an addition to allow a blue highlight for Firefox, would should keep it in "parity" with WebKit.

You can see some of the links below to research why keeping outline is the best for accessibility.

- https://a11yproject.com/posts/never-remove-css-outlines/
- http://www.outlinenone.com
- https://stackoverflow.com/questions/52589391/css-box-shadow-vs-outline

In addition, there's a small change to the main navigation where items are now outlined as well. Since there is no "focus" style change on these items, and the nav is super important to highlight on focus, I've added the outline back in for the nav.

### DIFF
--- a/modules/backend/assets/css/october.css
+++ b/modules/backend/assets/css/october.css
@@ -664,7 +664,7 @@ nav#layout-mainmenu a {text-decoration:none}
 nav#layout-mainmenu a:focus {background:transparent}
 nav#layout-mainmenu ul {margin:0;padding:0;list-style:none;float:left;white-space:nowrap;overflow:hidden}
 nav#layout-mainmenu ul li {color:rgba(255,255,255,0.6);display:inline-block;vertical-align:top;position:relative;margin-right:30px}
-nav#layout-mainmenu ul li a {display:inline-block;font-size:14px;color:inherit;outline:none;padding:14px 0 10px}
+nav#layout-mainmenu ul li a {display:inline-block;font-size:14px;color:inherit;padding:14px 0 10px}
 nav#layout-mainmenu ul li a:hover {background-color:transparent}
 nav#layout-mainmenu ul li a:active,
 nav#layout-mainmenu ul li a:focus {text-decoration:none;color:rgba(255,255,255,0.6)}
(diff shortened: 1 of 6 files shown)


mjauvin:
Thanks @bennothommo, that was so annoying...

LukeTowers:
@bennothommo is this ready to go? Really appreciate you dealing with this, there were a lot of complaints.

LukeTowers:
Also this should fix #4819 correct?

LukeTowers:
And #4964?

Samuell1:
@bennothommo Does this fixes issue with focus appearing even on elements when you click with mouse?

bennothommo:
@LukeTowers yep - it's ready to go. It does fix #4964, and some of #4819, although @gergo85 mentioned some other issues in there that this won't fix.

bennothommo:
@Samuell1 I believe the outline will still appear when you click on the mouse - it's whatever behaviour is in-built with the browsers.

Samuell1:
@bennothommo maybe it be fixed by this small JS that fb uses too? https://stackoverflow.com/a/51093815

bennothommo:
@Samuell1 as @LukeTowers said, we don't want any more JS dependencies unless they are critical. The outline shows on click because some people with visual impairments still use a mouse, and may still be able to identify key elements, while still using the focus indicator to navigate to other areas of the page.

PR #5038, opened 2020-04-12
Allow translations of default column values

4nik3t:
Currently the default column values does not support localization. This change will allow us to translate them.

### DIFF
--- a/modules/backend/widgets/Lists.php
+++ b/modules/backend/widgets/Lists.php
@@ -1077,7 +1077,7 @@ class Lists extends WidgetBase
          * Apply default value.
          */
         if ($value === '' || $value === null) {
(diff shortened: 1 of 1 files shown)


modules/backend/widgets/Lists.php
```diff
@@ -1077,7 +1077,7 @@ public function getColumnValue($record, $column)
          * Apply default value.
          */
         if ($value === '' || $value === null) {
```
LukeTowers, line 1079: Change the conditional to `if (($value === '' || is_null($value)) && !empty($column->defaults)) {`

4nik3t: @LukeTowers done

PR #5039, opened 2020-04-14
Don't persist failed search terms

tobias-kuendig:
If a column is defined as searchable, that is not actually searchable (like a partial) and a user filters a list using the searchbox, an Exception is thrown.

Unfortunately, this errorenous search term has already been persisted to the session so on subsequent page loads, the user is presented with an Exception with no way to resolve it themselves.

This PR catches any exception that happens during the search and resets any persisted search terms so that the user can continue working after a page load.

The exception itself will still be thrown.

To reproduce the issue this PR solves:

* Set a partial column in any list as `searchable: true` 
* Submit a search query
* Hit F5

### DIFF
--- a/modules/backend/widgets/Search.php
+++ b/modules/backend/widgets/Search.php
@@ -128,7 +128,14 @@ class Search extends WidgetBase
          * Trigger class event, merge results as viewable array
          */
         $params = func_get_args();
-        $result = $this->fireEvent('search.submit', [$params]);
+        try {
+            $result = $this->fireEvent('search.submit', [$params]);
+        } catch (\Throwable $e) {
+            // Remove the search term from the session if the search has failed.
+            $this->setActiveTerm('');
+            throw $e;
+        }
+
         if ($result && is_array($result)) {
             return call_user_func_array('array_merge', $result);
         }


LukeTowers:
This basically just completely disables the search if a developer has misconfigured the columns. I would prefer we better detected that case and just chose to ignore poorly configured columns where they'd already triggered an issue.

LukeTowers:
While also maybe logging an warning for developers to review.

tobias-kuendig:
This PR does not change the old behaviour. It only makes sure that the user does not end up in an infinite exception loop.

Since the exception is re-thrown it will end up in the logs.

mjauvin:
Looks good to me, what problem do you see @LukeTowers ?

LukeTowers:
If we're going to be addressing this issue I would prefer that we addressed it properly rather than just hiding it. While this does prevent the user from getting stuck in a loop, at least that loop would inform the user & potentially the developers that something was wrong. If we make this change as is, then the user would no longer get stuck in the loop, but the search wouldn't work either. The user is probably less inclined to report that as an issue to a developer and thus it's less likely to be properly dealt with.

In my ideal world we detect what column is not actually supported to be searchable, and then we blacklist it in a cache so that the user's search can continue like normal.

tobias-kuendig:
A popup with the error message still appears and is logged to the system.log.

The system behaves exactly the same with this PR until you reload the page, then it just keeps working instead of remaining in a broken state. 

I do see, however, that we could handle this more elegantly.

I will take another look at this problem tomorrow.

LukeTowers:
Thanks @tobias-kuendig, I appreciate it!

tobias-kuendig:
There is a [`getSearchableColumns` method](https://github.com/octobercms/october/blob/baca3c8adf7f5ba1832a254979b373e3ae7016b5/modules/backend/widgets/Lists.php#L1445) we could change to ignore "unsearchable" column types, like partials:

```diff
    protected function getSearchableColumns()
    {
        $columns = $this->getColumns();
        $searchable = [];

        foreach ($columns as $column) {
            if (!$column->searchable) {
                continue;
            }
+            if (in_array($column->type, ['partial'])) {
+                continue;
+            }

            $searchable[] = $column;
        }

        return $searchable;
    }
```

However, this won't fix the problem where I have specified a column as searchable, that no longer exists in the databse. The user will end up in the same Exception loop as before. 

```yaml
    removed_from_the_db:
        type: text
        sortable: true
        searchable: true # this will trigger an Exception
```

Another option is to [delay the persisting of the search term to the session](https://github.com/octobercms/october/blob/baca3c8adf7f5ba1832a254979b373e3ae7016b5/modules/backend/widgets/Search.php#L125) until the query has happened.

In this case we need to change the `getActiveTerm` method to not use the session value by default. Otherwise [the actual search](https://github.com/octobercms/october/blob/baca3c8adf7f5ba1832a254979b373e3ae7016b5/modules/backend/behaviors/ListController.php#L215) no longer works.

```diff
    /**
     * Search field has been submitted.
     */
    public function onSubmit()
    {
        /*
         * Save or reset search term in session
         */
-        $this->setActiveTerm(post($this->getName()));
+        $term = post($this->getName());
+        $this->activeTerm = $term;

        /*
         * Trigger class event, merge results as viewable array
         */
        $params = func_get_args();
        $result = $this->fireEvent('search.submit', [$params]);

        if ($result && is_array($result)) {
+            $this->setActiveTerm($term);
            return call_user_func_array('array_merge', $result);
        }
    }

    /**
     * Returns an active search term for this widget instance.
     */
    public function getActiveTerm()
    {
+        if ($this->activeTerm) {
+            return $this->activeTerm;
+        }

        return $this->activeTerm = $this->getSession('term', '');
    }

    /**
     * Sets an active search term for this widget instance.
     */
    public function setActiveTerm($term)
    {
        if (strlen($term)) {
            $this->putSession('term', $term);
        } else {
            $this->resetSession();
        }

        $this->activeTerm = $term;
    }
```

The result of this change is the same as the original change proposed in this PR:

* Exception is still thrown
* The search is broken anyways
* The user can reload the page and continue working

LukeTowers:
@tobias-kuendig is it possible to parse the exception that gets thrown to identify the offending column and then cache it as invalid and try to execute the search again?

tobias-kuendig:
I guess this is possible, but I really wonder if the complexity is warranted.

An example error message:

> SQLSTATE[42S22]: Column not found: 1054 Unknown column 'offline_relationtranslation_entry.unsortable' in 'where clause' (SQL: select `offline_relationtranslation_entry`.* from `offline_relationtranslation_entry` where (((lower(offline_relationtranslation_entry.name) LIKE %xx%) or (lower(offline_relationtranslation_entry.unsortable) LIKE %xx%))) order by `id` desc)" on line 664 of /<dir>/vendor/laravel/framework/src/Illuminate/Database/Connection.php

This raises the following questions:

* What if the user uses Postgres oder SQLite?
* What if multiple columns are wrong?

Let's not forget that the error we are trying to catch here is an actual configuration error by the developer. In my opinion, this is an error that needs fixing and should be visible to the end-user. I'd rather have the end-user see an error message than just not getting the search result that contains the search query because we tried to handle this error gracefully. The original intent of this PR was to prevent a broken backend for the end-user due to the search query being persisted to the session. 

I'd like to propose the original change from this PR (as it has the least impact) or the one described [in the comment above](https://github.com/octobercms/october/pull/5039#issuecomment-613852563) which is a bit more elegant than just re-throwing the original exception.

LukeTowers:
Fair enough @tobias-kuendig, this does solve the painful issue of a page becoming persistently broken after a search when the search is misconfigured while also retaining the visible warning to the developer.

LukeTowers:
Just a note, partial columns can technically be searchable because there's no rule saying that they can't line up with a DB column, it's just a more custom way of displaying data.

PR #5040, opened 2020-04-15
Added Serbian localization.

jocic:
Localization for **Serbian** language has been added based on the existing English one, more specifically:
1. CMS
2. System
3. Backend


LukeTowers:
Awesome, thanks!

PR #5042, opened 2020-04-17
Allow blob files naming in AJAX request

aurelien-roy:
This change adds an opportunity to set the filename of BLOB files in AJAX requests (this is useful for instance for setting the file extension).

Usage:

```js
// blob_file is a Blob object.
// We set a filename property, which the framework will read to set the filename.
blob_file.filename = "avatar.png"

$(e).request('onBlobUpload', {
   data: {"avatar": blob_file},
   files: true
})

```

This is a non-breaking change, because "filename" is not a property of Blob objects. We create it specifically for this use case.


LukeTowers:
Could you create a PR to the docs as well documenting this?

PR #5046, opened 2020-04-19
Get file and folder metadata for media items using a single network call if possible

### LINKED ISSUE #5045: Initialization of library items of remote storages is really slow
- OctoberCMS Build: 1.0.465
- PHP Version: 7.2
- Database Engine: mysql
- Plugins Installed: _not relevant_

### Description
When opening the media library on an empty cache, the page loads really slow and even timeouts. This is only the case for me when storing media externally, in my case on Google Cloud Storage.  

### Steps To Reproduce
Set up a remote storage and use it for media. Run `artisan cache:clear` and open the media library.

### Proposed fix
By using `$this->getStorageDisk()->listContents($fullFolderPath)` in the MediaLibrary class, we can get the timestamp and size information of all files and folders in 1 network call instead of (in my case) 311 calls. Not only does this really speed up the process, it can also be a lot cheaper if your storage provider makes you pay per call to their API.

The only catch: the [`listContents`](https://flysystem.thephpleague.com/v1/docs/usage/filesystem-api/#list-contents) function returns an array of arrays and the only _required_ metadata is path info and file type. Whether  timestamp and size information is returned is dependant on the adapter implementation. This means that, depending on the used adapter, these extra network calls may still be necessary. Looking at a couple of adapter implementations, all of them seem to provide the required information though:

| Adapter                                                                                                                                                   | Size?  | Timestamp   |
|-----------------------------------------------------------------------------------------------------------------------------------------------------------|--------|-------------|
| [thephpleague/flysystem-aws-s3-v3](https://github.com/thephpleague/flysystem-aws-s3-v3/blob/master/src/AwsS3Adapter.php#L267)                             | `size` | `timestamp` |
| [Superbalist/flysystem-google-cloud-storage](https://github.com/Superbalist/flysystem-google-cloud-storage/blob/master/src/GoogleStorageAdapter.php#L359) | `size` | `timestamp` |
| [thephpleague/flysystem-rackspace](https://github.com/thephpleague/flysystem-rackspace/blob/master/src/RackspaceAdapter.php#L248)                         | `size` | `timestamp` |

#### Before
- Page load: 47.3s
- Network calls: 311

Screenshot

#### After
- Page load: 3.9s
- Network calls: 14

Screenshot

DieterHolvoet:
Fixes #5045.


mjauvin:
Please add unit tests in `tests/unit/system/classes/MediaLibrary.php` to validate these changes.

LukeTowers:
Awesome @DieterHolvoet! Are you able to provide some unit testing for this change just to make sure the library is still working fine with the regular local storage option with and without this change?

DieterHolvoet:
Of course, I'll try to do that the coming week.

DieterHolvoet:
I added some basic test coverage for MediaLibrary::listFolderContents. I don't have much experience with testing, so let me know if this does not look right.

LukeTowers:
Seems fine to me. @bennothommo can you review? Once we merge this we'll have to make sure we modify the test when merging it into the L6 branch

tests/unit/system/classes/MediaLibraryTest.php
```diff
@@ -62,4 +68,60 @@ public function testValidPathsOnValidatePath($path)
...
+        $contents = MediaLibrary::instance()->listFolderContents();
+        $this->assertNotEmpty($contents, 'Media library item is not discovered');
+
+        $item = reset($contents);
+        $this->assertAttributeEquals('file', 'type', $item, 'Media library item does not have the right type');
+        $this->assertAttributeEquals('/text.txt', 'path', $item, 'Media library item does not have the right path');
+        $this->assertAttributeNotEmpty('lastModified', $item, 'Media library item last modified is empty');
+        $this->assertAttributeNotEmpty('size', $item, 'Media library item size is empty');
+
+        $this->removeMedia();
```
bennothommo, line 85: @DieterHolvoet you will need to make this part of the `tearDown` process in the test class - otherwise, if the test fails, the test files that you had copied over will still be stored on disk.

tests/unit/system/classes/MediaLibraryTest.php
```diff
@@ -62,4 +68,60 @@ public function testValidPathsOnValidatePath($path)
...
+    }
+
+    protected function removeMedia()
+    {
+        foreach (glob(storage_path('app/media/*')) as $file) {
+            unlink($file);
+        }
+
+        rmdir(storage_path('app/media'));
+        rmdir(storage_path('app'));
```
bennothommo, line 125: @DieterHolvoet just to clarify, this media removal all happens within your temp storage folder, yes? It doesn't have any effect on the actual media folder?

modules/system/classes/MediaLibrary.php
```diff
@@ -665,17 +669,19 @@ protected function scanFolderContents($fullFolderPath)
...
+                $type = MediaLibraryItem::TYPE_FOLDER;
+                $key = 'folders';
             }
-        }
 
-        $folders = $this->getStorageDisk()->directories($fullFolderPath);
-        foreach ($folders as $folder) {
-            if ($libraryItem = $this->initLibraryItem($folder, MediaLibraryItem::TYPE_FOLDER)) {
-                $result['folders'][] = $libraryItem;
+            if ($libraryItem = $this->initLibraryItem($content, $type)) {
```
bennothommo, line 683: @DieterHolvoet I usually prefer variable assignments to happen before they are used as a condition, as it allows tools like Xdebug to review the variable before it jumps around in the condition statement. (I realise this wasn't your doing - but let's fix it anyways).

```suggestion
            $libraryItem = $this->initLibraryItem($content, $type);
            if (!is_null($libraryItem)) {
```

tests/unit/system/classes/MediaLibraryTest.php
DieterHolvoet: Yes, the storage path is changed by calling
```php
$this->app->useStoragePath(base_path('storage/temp'));
```

tests/unit/system/classes/MediaLibraryTest.php
DieterHolvoet: Yes, the storage path is changed by calling
```php
$this->app->useStoragePath(base_path('storage/temp'));
```
so there's no effect on the usual media folder.

tests/unit/system/classes/MediaLibraryTest.php
DieterHolvoet: This causes errors in the path validation tests. Also, the files will be copied and deleted 26 times because of those path validation tests, which seems a little wasteful. Is it an option to move this new test to a separate class?

bennothommo: @DieterHolvoet Alright, well then the `setupStorage` method should be run only in this test case, and not as part of the `setUp` callback. It's not needed for those other tests.

I'd prefer your new test to be kept in this class as it is testing a part of the MediaLibrary class.

tests/unit/system/classes/MediaLibraryTest.php
DieterHolvoet: I added an extra test that makes sure files are only deleted when the storage path is changed.

tests/unit/system/classes/MediaLibraryTest.php
DieterHolvoet: @bennothommo Did the changes, let me know what you think.

bennothommo: Thanks @DieterHolvoet - looks fine to me.

bennothommo:
@github-actions run

<details>

```js
(async () => {
  // Get issue/PR number

  // Add Testing Needed label
    owner: context.repo.owner,
    repo: context.repo.repo,
    issue_number: number,
    labels: ['Testing Needed'],
  });

  // Comment body
    + '\n\n'
    + '\n\n'
    + '```\n'
    + '```';

  // Add comment
})();
```

LukeTowers:
@bennothommo this is all good to go, no breaking changes just a performance improvement?

bennothommo:
@LukeTowers I haven't tested it yet. I don't really have a remote storage instance to test it on. It looks solid in theory, but just wanted to make sure it works before merging.

LukeTowers:
@DieterHolvoet Great job! Tested this and it works fine, and there is a noticeable performance increase for remote storage.

PR #5047, opened 2020-04-19
Update check for '.git' when using 'php artisan october:util git pull'

caloggs:
When using the `php artisan october:util git pull` command it only checks if `.git` is a directory in the plugin or theme folder, however if plugins have been added as git submodules `.git` is a file not a directory so the command was not running `git pull`. This PR just changes the check to account for files or directories.

### DIFF
--- a/modules/system/console/OctoberUtil.php
+++ b/modules/system/console/OctoberUtil.php
@@ -313,7 +313,7 @@ class OctoberUtil extends Command
     {
         foreach (File::directories(plugins_path()) as $authorDir) {
             foreach (File::directories($authorDir) as $pluginDir) {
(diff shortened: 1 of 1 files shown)


LukeTowers:
Can I see a before / after screenshot of the terminal output?

caloggs:
@LukeTowers Sure thing.

Before:
before

After:
after

LukeTowers:
Does `::exists()` check specifically for a file or does it handle the regular .git directories properly as well?

caloggs:
It checks for a file or directory, the `::exists()` method is a wrapper for `file_exists` (https://www.php.net/manual/en/function.file-exists.php) so it will handle both

PR #5051, opened 2020-04-25
Add possibility to render record list inside side panel

kubamarkiewicz:
**What**
Add possibility to render record list (using List controller behavior) inside backend side-panel.

**Why**
Sometimes it may be useful to place the record list inside the side-panel. This would allow to create interface similar to the one that we know from CMS module, where the page list is inside the side-panel, next to the form, always present in index, create and update actions.
The idea is to be able to use List controller behavior with all its configuration options to define and render the list inside the side-panel.

**How**
The list would be rendered inside side-panel by placing the following code inside the template files (index.htm, create.htm, update.htm, preview.htm):
```
<?php Block::put('sidepanel') ?>
    <?= $this->listRender() ?>
<?php Block::endPut() ?>
```
To simplify the list, list header will be always hidden inside the side-panel. 
In config_list.yaml it is possible to simplify the list more using the following options:
```
recordsPerPage: 0 - will hide the list footer
showCheckboxes: false
```
The final result looks like this:

Screenshot 2020-04-25 at 15 07 20

I have provided en example in test plugin:
https://github.com/octoberrain/test-plugin/pull/86


LukeTowers:
This fits better as a custom Widget that you can provide via a plugin than something built into the core.

bennothommo:
It definitely needs to be optional - there's a lot of functionality in the Lists widget that I don't see being available if it were placed in the side panel.

PR #5053, opened 2020-04-28
Clear input after creating tag in taglist

rubenvanerk:
When using the taglist field type, the input isn't cleared after a tag has been created. This problem is described here: https://github.com/select2/select2/issues/4698
This change fixes that issue.


LukeTowers:
Can you post a recording of the behaviour before and after?

rubenvanerk:
Broken:
broken
Fixed:
fixed

bennothommo:
Tested, and confirmed working. Thanks @rubenvanerk.

Issue #5054, opened 2020-04-28
Stay logged in checkbox appears twice on backend login form. 


mjauvin:
When cms.backendForceRemember is set to null, the "stay logged in" checkbox appears twice (see attached screenshot)

![Screenshot_20200428-092351](https://user-images.githubusercontent.com/2013630/80492959-b79b4d80-8932-11ea-99c5-df7ec4644f56.jpg)

mjauvin:
@LukeTowers this does not fix the problem because on mobile, it's overriden by the following media query:
```
@media (max-width: 768px)
body.outer .layout >.layout-row >.layout-cell .outer-form-container .horizontal-form input {
    display: block;
    width: 100% !important;
    margin-bottom: 20px;
}
```

LukeTowers:
It fixed it when I tested it, but that was in chrome's inspector so the CSS was probably slightly prioritized. Just add an !important and PR it.

PR #5057, opened 2020-04-29
add input tag id to css selector

mjauvin:
prevent css rule to be overriden by a more generic css rule


mjauvin:
improve on #5054

PR #5067, opened 2020-05-08
render the css classes on the filter widget

marcogrueter:
Just a quick one:  render the css classes from the FilterWidget on the control.

### DIFF
--- a/modules/backend/widgets/filter/partials/_filter.htm
+++ b/modules/backend/widgets/filter/partials/_filter.htm
@@ -1,5 +1,5 @@
 <div
-    class="control-filter"
+    class="control-filter <?= $cssClasses ?>"
     data-control="filterwidget"
     data-options-handler="<?= $this->getEventHandler('onFilterGetOptions') ?>"
     data-update-handler="<?= $this->getEventHandler('onFilterUpdate') ?>">


LukeTowers:
Can you provide some more background information on this? What is your use case?

marcogrueter:
I simply sent this PR because the filter widget has a public property $cssClasses, but it doesn't get rendered on the control. Looked like an omission to me.

I discovered this because I wanted to visually hide a filter on a backend form with a relation management widget.

In our web portal we got participants for events (belongsToMany) and each event has sessions they can book (many with pivot data).
I got the participants for specific including their booked sessions in the backend form of the event and wanted to be able to filter the participants list by their sessions, so we can send out individual invitations, generate certificates and various exports in the same view.
I did that with scopes and filter options on the respective models and the "dependsOn: events" option for the sessions.

For UX reasons I didn't want the user to be able to change the event in the filter, since we already are in the "event context". The most solid and simplest solution seemed to be to just hide the filter and preset the selected event in the controllers "relationExtendViewFilterWidget" function, instead of figuring out how to get the controller and form model inside the models scope and/or "getXXXOptions" with scopes functions and check if we are even in the right context. And as a bonus, I can use the same filter config for the event form and the general participants list.

That was more work than the PR itself :-)

LukeTowers:
@marcogrueter can you make a PR to the docs explaining how to use / set this property?

marcogrueter:
Well, I can certainly try.

This would probably go to https://octobercms.com/docs/backend/lists#extend-filter-scopes, something like a second table with public properties (or property, since it's the only public property that should be touched), right?

Or should I go the long way and create "addCssClass" and "removeCssClass" functions and add the description to the public methods table?

LukeTowers:
@marcogrueter you might be able to set it in the config_filter.yaml file as a top level property:

```yaml
cssClasses: 'example'
scopes:
    etc:
```

marcogrueter:
@LukeTowers no, doesn't work. At least not from the get go.

The property would have to be added to the `fillFromConfig` list in the `init()` function of Filter.php. And then I guess line 94 in Filter.php would have to be changed to a string concatenation instead of an implode, since none of the other yaml configs use array for the cssClasses (that I'm aware of).

But it is a possibility if you want me to go in that direction.

LukeTowers:
@marcogrueter thanks for looking into that. I think just adding a `$filter->cssClasses = array_merge($filter->cssClasses, ['my', 'array', 'of', 'classes']);` above the `$filter->addScopes()` call should be enough.

PR #5071, opened 2020-05-14
Trigger inspector dependency check when values are remotely loaded for dropdown properties

### LINKED ISSUE #4947: Inconsistent and bugged component dropdown behavior
- OctoberCMS Build: ### 464
- PHP Version: 7.4.2

### Description:
A component has two dropdown properties: *foo* and *bar* with *bar* depending on *foo*.

* If *foo* retrieves its options through a `get*Options` method, the dropdown will appear blank by default when opening the component (See *Options Method* component below).
* If *foo* sets its options within the `defineProperties` property array using the `options` key, the dropdown will have its first option populated by default (See *Options Array* component below).
* If *foo* sets its options within the `defineProperties` property array using the `options` key, the *bar* dropdown will fail to populate even though an option is selected in *foo*.
  * If you change *foo* option then change it back, *bar* will populate correctly
  * If you close and re-open the component, *bar* will populate correctly
  * If *foo* only has 1 option and *bar* is set to `required`, there is no way to close and re-open the component and your only option is to reload the page (See *Single Options Method* component below).

### Steps To Reproduce:
* `git clone https://github.com/Flynsarmy/oc-badcompdep-plugin.git` to */plugins/flynsarmy/badcompdep*
* `php artisan plugin:refresh Flynsarmy.Badcompdep`
* Open any page in Backend - CMS
* Add this plugins *Options Method* and *Options Array* and *Single Option Method* components to the page and witness the inconsistent (and in the case of *Single Options Method* broken) behavior.

bennothommo:
Fixes #4947.

This PR is a set of fixes to the Inspector widget that do the following:

- Adds a new base method `findDependentProperties` that returns all properties that depend on a specified property.
- Triggers a dependency check when a dropdown property which loads its values remotely has finished retrieving its values, which may include a default property being selected.
- Adds a synonym for the `placeholder` attribute, `emptyOption`, which is the name used in the Form widgets. The `placeholder` attribute takes precedence.


LukeTowers:
LGTM, although I haven't tested it

bennothommo:
@LukeTowers I'm just gonna merge this. I've done a fair bit of testing on it and it does seem to resolve @Flynsarmy's original issue. There's enough checks around it that if something were to fail, it should fail gracefully.

LukeTowers:
Perfect @bennothommo, could you add a blurb to the 466 update notes for the changes?

bennothommo:
Done @LukeTowers :)

PR #5080, opened 2020-05-20
Allow datepicker to submit only date or time values

### LINKED ISSUE #5072: Backend datepicker widget always sends datetime strings to the server
The [datepicker](https://octobercms.com/docs/api/backend/formwidgets/datepicker) form widget can be configured with 3 modes, `datetime`, `date`, or `time`. However if using either `date`, or `time`, a full timestamp string is always sent to the server.

image

The timestamp is also converted to UTC, so it is not necessarily even the correct value. For example, if I am using a date-only picker, and I select today's date, the server might get a timestamp value that's yesterday, today, or tomorrow, depending on my timezone offset. Likewise when using the time-only picker, it will be modified by my current timezone offset.

I would expect that if `time` mode is used, only the time would be sent to the server, or if `date` is used, only the date would be sent to the server, and in both of these cases, the timezone offset should not modify them.

Note, when using the datepicker in datetime mode, it makes perfect sense to convert it to UTC since this is how it should be stored in a database. It doesn't make sense for a single date (which should be stored in a `DATE` column) or single time.

multiwebinc:
Proposed solution for #5072. 

This modifies the behavior of the backend datepicker widget. If the widget's `mode` option is set to `date`, **only** the date value will be submitted to the server. If the mode is set to `time`, **only** the time value will be submitted to the server. With these two modes, the values will no longer be modified by the user's timezone offset, and the `ignoreTimezone` option will now only affect values if the `datetime` mode is used.


bennothommo:
On hold until after build 466 is released.

msimkunas:
Are there any plans to merge this sometime in the future?

LukeTowers:
@msimkunas yes, we've just been busy with a bunch of other stuff

bennothommo:
@multiwebinc did you have any thoughts on making the widget enable `ignoreTimezone` automatically when the `date` or `time` modes are used? I know that MySQL does ignore them for `TIME` fields, but I'm not so sure about `DATE` fields, so enabling it automatically should remove all doubt.

multiwebinc:
@bennothommo To be honest, I've been working on other things lately so I don't really have this code fresh on my mind right now and I made this PR several months ago. Feel free to make any modifications you feel would improve it.

I would assume MySQL stores only the actual date in the `DATE` fields since the timezone would be totally redundant without the actual time (just like it is for the time without the date). However I can't vouch for other DBMS, like SQLite, for example, which doesn't have a dedicated date-only column (or even a true dedicated datetime column for that matter, since dates are either stored as strings or numeric Julian day numbers or Unix time).

bennothommo:
I have tested this and found that there is a bug with time-only fields where the field comes up as "Invalid date" after a refresh - ostensibly because of the seconds being stored and returned by the database, even though they aren't in the defined format for those fields. The bug doesn't exhibit with `develop` currently. Will have to investigate further.

daftspunk:
Sorry for the delay in response to this.

This is a complicated issue and unfortunately, the proposed solution only serves to increase that complexity by returning different values. Laravel will always return a complete date and so this widget adheres to that convention, as it has always done.

Sincerely, thanks for your efforts on this attempt. An alternative fix to the original problem has been implemented in 2.0

PR #5087, opened 2020-05-23
Implement improved "set build" utility

bennothommo:
Replaces #4615.

This change implements an improved `php artisan october:util set build` utility that, instead of checking the October CMS server for the latest build (which may not even be the installed version), it checks the module files against a source manifest kept in the October Meta repo (https://raw.githubusercontent.com/octoberrain/meta/master/manifest/builds.json).

This check allows us to accurately determine the build based on the module files in the October CMS installation, and can even detect versions if the module files are modified (except in the cases of extreme modification) or if the installation is missing one or more of the core modules installed by default.

The `set build` utility now prints out the following:

image

When a modified build is detected, this is also indicated in the Updates & Plugins area in the Settings:

Screenshot from 2020-05-24 00-44-28

An additional utility has been implemented, `php artisan october:util build manifest`, which will build the manifest JSON file in order to provide the maintainers with a way of generating this manifest file as required (ideally, we'll do this through GitHub Actions when a tag is generated on the main repo).


modules/system/classes/FileManifest.php
```diff
@@ -0,0 +1,200 @@
+<?php namespace System\Classes;
+
+use ApplicationException;
+use Config;
+
+/**
+ * Stores the file manifest for this October CMS installation.
+ *
+ * This manifest is a file checksum of all files within this October CMS installation. When comapred to the source
```
LukeTowers:
```suggestion
 * This manifest is a file checksum of all files within this October CMS installation. When compared to the source
```

modules/system/classes/FileManifest.php
```diff
@@ -0,0 +1,200 @@
...
+        }, $modules);
+    }
+
+    /**
+     * Finds all files within the path.
+     *
+     * @param string $basePath The base path to look for files within.
+     * @return array
+     */
+    protected function findFiles(string $basePath)
```
LukeTowers, line 152: What are your thoughts on using the Halcyon FileDatasource to do the path iteration?

modules/system/classes/SourceManifest.php
```diff
@@ -0,0 +1,343 @@
+<?php namespace System\Classes;
+
+use ApplicationException;
+use Config;
+use System\Classes\FileManifest;
```
LukeTowers: This is in the same namespace so you can reference it directly without importing it.

modules/system/classes/FileManifest.php
bennothommo: @LukeTowers doesn't FileDatasource only support a single level of subdirectories?

LukeTowers: @bennothommo the FileDatasource class directly doesn't care about the nesting level, it's merely the Halcyon Model & Builder that implement the check for nesting. See the example code I pasted in https://github.com/octobercms/october/pull/4615#issuecomment-611696247, it returned all paths.

bennothommo: @LukeTowers the [`select`](https://github.com/octobercms/library/blob/develop/src/Halcyon/Datasource/FileDatasource.php#L112) method of the FileDatasource class does only support a single layer though. That method's the only reason I'd use FileDatasource, as it would save a lot of the conditional code I originally had.

LukeTowers: @bennothommo the selectOne method supports arbitrary depth AFAIK though

bennothommo: @LukeTowers I was meaning using the `select` method to get all paths that match a certain extension through the `extensions` option. That would allow me to get rid of [this block of code](https://github.com/octobercms/october/blob/4363d129a7b124a58e91b67446e617606cbbac9f/modules/system/classes/FileManifest.php#L165-L171), at least. `selectOne` would only get a single file.

LukeTowers: Ah, I see. Why are you only looking at those specific file extensions? Why not include all of them?

bennothommo: @LukeTowers mainly to keep the filesize of the manifest down. I figured we'd very unlikely push out a build that just makes changes to the Markdown files, for example.

bennothommo: That being said, I probably should include `htm` and `html` files though :P

LukeTowers: We wouldn't push out a build no, but I personally think it's important to know if there are any changes period. Using markdown as an example, when it comes to the rainlab docs plugin, it could be useful to know if the docs files have been modified locally, etc.

Are you able to test the filesize difference between the filtering you have right now (including htm & html) and not doing any filtering period?

bennothommo: @LukeTowers It's ~409kb with the specified extensions (plus `htm` files), and ~496kb with everything. So I suppose it's not a huge drama including them all.

LukeTowers: Yeah, let's go with including them all then.

bennothommo: Done :)

modules/system/classes/FileManifest.php
```diff
@@ -0,0 +1,171 @@
...
+    public function getFiles()
+    {
+        if (count($this->files)) {
+            return $this->files;
+        }
+
+        $files = [];
+
+        foreach ($this->modules as $module) {
+            $path = $this->root . DIRECTORY_SEPARATOR . 'modules' . DIRECTORY_SEPARATOR . $module;
```
LukeTowers: Since DIRECTORY_SEPERATOR doesn't actually provide any real-world benefit (see https://dev.to/c33s/always-use--as-directory-seperator-in-php-43l7) I'd prefer that we just standardized on forward slashes in paths always because then it's easier to read and more consistent.

modules/system/classes/FileManifest.php
```diff
@@ -0,0 +1,171 @@
...
+        }
+
+        $modules = [];
+        foreach ($this->modules as $module) {
+            $modules[$module] = '';
+        }
+
+        foreach ($this->files as $path => $hash) {
+            // Determine module
+            $module = explode(DIRECTORY_SEPARATOR, $path)[2];
```
LukeTowers: DIRECTORY_SEPERATOR might be required here, but if you are able to test on a windows machine for sure, that would be helpful.

modules/system/console/OctoberUtil.php
```diff
@@ -120,21 +125,135 @@ protected function utilSetBuild()
...
-        try {
-            $build = UpdateManager::instance()->setBuildNumberManually();
-            $this->comment('*** October sets build: '.$build);
+        $this->comment('*** Detecting October CMS build...');
+
+        $build = UpdateManager::instance()->setBuildNumberManually();
+
+        if (is_null($build)) {
+            $this->error('Unable to detect your build of October CMS.');
+            exit(0);
```
LukeTowers: Is exit(0) recommended in CLI commands or is there a more laravel-y way of doing that? I ask because we need to avoid exit() / die() in the web code in order to support tools like swoole

samuelpatro: @LukeTowers i think returning should stop it no?

bennothommo: I suppose that instance can be changed to a return, but the `exit(1)` calls elsewhere need to remain in order for the CLI command to actually end in an error state. I don't think Laravel has a method that kills the program like that.

Isn't Swoole a web request handler, not CLI?

LukeTowers: It is, was just checking if there was something similar in Laravel, like returning an error code instead of exiting. This specific use case should't actually be a problem for tools like Swoole, just like to be thinking about it in the back of our heads.

modules/system/classes/FileManifest.php
bennothommo: @LukeTowers AFAIK, Laravel still use it pretty extensively in file handling. I gotta assume there's a reason for that.

I read the post you linked but that seemed to only make the argument that these paths *might* end up in URLs where they will encode incorrectly. That wouldn't occur with this command.

samuelpatro: @bennothommo i tested it and it doesnt work in windows looks like

bennothommo: @Samuell1 Interesting - what does the path look like on Windows with `DIRECTORY_SEPARATOR`?

bennothommo:
@LukeTowers would you have any objections to moving this out of the `october:util` command into its own command `october:build`? `october:util` is becoming quite large, and it would make sense to have this new command perhaps locked as a privileged action.

LukeTowers:
@bennothommo my objection to that would be that we can't force people to update their composer files to reference october:build set instead of october:util set build, and the util command is still less than 500 lines total despite everything it does already. I don't necessarily disagree with the approach though, some of the util commands should probably be split out at some point.

What do you mean by having this command locked as a privileged action?

bennothommo:
@LukeTowers the Composer point is true, although we could have a stub for the `october:util set build` command, to run this new command.

Making it privileged would be like the `october:up` and `october:update` command - preventing plugins from instantiating, so that the process can run without interference (except from those plugins using `$privileged`). I was going to suggest the same for `october:env`, to mitigate an issue we had reported to us recently with the Image Resizer plugin, but I'll post that as a separate PR.

LukeTowers:
Sure, lets go ahead with that then

Samuell1:
@bennothommo I was going to suggest that it will be cool if every util cmd that is used every day should be separated. Same for purge or compile commands like `october:purge` and `october:compile`.

bennothommo:
@Samuell1 We can certainly look at that at a later stage, although I believe most of them are in there to try and "hide" advanced commands that only people developing October CMS really need to use. `october:util set build` is used by a lot of people however (albeit unknowingly, since it's run by Composer mostly), so I feel it can be split off.

modules/system/classes/FileManifest.php
bennothommo: @Samuell1 @LukeTowers FileDatasource just uses forward-slashes, so I'll remove the `DIRECTORY_SEPARATOR` constant. If it becomes a problem, we'll revisit it.

modules/system/console/OctoberUtil.php
bennothommo: @LukeTowers I've changed that instance to a `return`.

bennothommo:
@LukeTowers This is ready for a final review.

LukeTowers:
Also do you think we should be scanning the library folder too?

Samuell1:
@LukeTowers i think it should be called `october:version` returns version and automatically sets it. I think that will be better then naming it build because it can mean to "build" stuff

LukeTowers:
@Samuell1 that's why I think it makes the most sense to just keep it as `set build` because that's exactly what it's doing, setting the build number.

Samuell1:
@LukeTowers yes but still set build doesnt make sense because its setting version and returning it or telling you your version is modified etc

bennothommo:
@LukeTowers 

> Also do you think we should be scanning the library folder too?

I felt the library is "separate" enough that it wasn't really warranted. I wanted to keep the scope of tracked files as small as possible, whilst still maintaining accuracy.

> after reading over this again I think it makes sense for october:manifest to be a separate command but october:build doesn't really make sense from a naming perspective and the actual implementation of it is small enough that I think it makes sense for it to stay in the util class.

The reasoning for it being a separate command was so I could make it "privileged", thereby preventing plugins from instantiating. This would allow the `build` command to work even if, say, a plugin was failing to boot.

bennothommo:
Happy to change the term - I think `version` might be a good term as that is the common term used for CLI programs, ie. `php --version`.

LukeTowers:
Fair enough @bennothommo, I would be fine with `october:version`

bennothommo:
@LukeTowers this is ready to go. We can either include it in 467 or 468.

LukeTowers:
@bennothommo why is the line break conversion required?

bennothommo:
@LukeTowers for people on Windows. @Samuell1 tested it out for me and found that sometimes Windows changes the line breaks for giggles, even though the content of the file itself hasn't changed.

LukeTowers:
@bennothommo can you wrap the code in a check for running on windows instead then? Nobody should really be hosting on windows platforms to begin with, just developing.

Samuell1:
@LukeTowers i wasnt running windows server, i was just testing it locally. It looks like editor is forcing line breaks different thats why Ben forced it to be same per every system to prevent any issues.

LukeTowers:
What editor @Samuell1? We don't care if someone opens a core file and then saves it and changes the line endings, if they do that then yeah, the file has been modified. Most people shouldn't be opening and saving changes made to a core file.

bennothommo:
@LukeTowers The change was to head off any reports of people using Composer on Windows and finding that the `october:version` call we do there reports that the version of October could not be determined. Both @Samuell1 and I found that there's actually very little difference in performance, so I'm going to change my comment indicating otherwise.

LukeTowers:
@bennothommo if that's the case, why have it as a separate parameter? I would say either run it every time, or only run it when `PHP_EOL === "\r\n"`.

Samuell1:
@LukeTowers but i wasnt saving/editing the file but file was already in different linebreaks.

modules/system/console/OctoberVersion.php
```diff
@@ -0,0 +1,46 @@
...
+            $this->error('Unable to detect your build of October CMS.');
+            return;
+        }
+
+        if ($build['modified']) {
+            $this->info('*** Detected a modified version of October CMS build ' . $build['build']);
+        } else {
+            $this->info('*** Detected October CMS build ' . $build['build']);
+        }
+    }
```
samuelpatro:
```suggestion
    public function handle()
    {
        $this->comment('*** Detecting October CMS build...');
        // // Skip setting the build number if no database is detected to set it within
        if (!App::hasDatabase()) {
            $this->comment('No database detected - skipping setting the build number.');
            $build = UpdateManager::instance()->getBuildNumberManually();
        } else {
            $build = UpdateManager::instance()->setBuildNumberManually();
        }
        if (is_null($build)) {
            $this->error('Unable to detect your build of October CMS.');
            return;
        }
        if ($build['modified']) {
            $this->info('*** Detected a modified version of October CMS build ' . $build['build']);
        } else {
            $this->info('*** Detected October CMS build ' . $build['build']);
        }
    }
```

modules/system/classes/UpdateManager.php
```diff
@@ -363,22 +363,21 @@ public function uninstall()
     }
```
samuelpatro, line 359:
```suggestion
    /**
     * Determines build number from source manifest.
     *
     * @return void
     */
    public function getBuildNumberManually()
    {
        $source = new SourceManifest();
        $manifest = new FileManifest(null, null, true);
        // Find build by comparing with source manifest
        return $source->compare($manifest);
    }
```

modules/system/classes/UpdateManager.php
```diff
@@ -363,22 +363,21 @@ public function uninstall()
...
-     * Asks the gateway for the lastest build number and stores it.
+     * Determines build number from source manifest.
+     *
      * @return void
      */
     public function setBuildNumberManually()
     {
-        $postData = [];
+        $source = new SourceManifest();
+        $manifest = new FileManifest(null, null, true);
```
samuelpatro, line 374:
```suggestion

```

modules/system/classes/UpdateManager.php
```diff
@@ -363,22 +363,21 @@ public function uninstall()
...
 
-        if (Config::get('cms.edgeUpdates', false)) {
-            $postData['edge'] = 1;
-        }
+        // Find build by comparing with source manifest
+        $build = $source->compare($manifest);
```
samuelpatro:
```suggestion
        $build = $this->getBuildNumberManually();
```

Samuell1 (commented):
This allows to detect version even if database isnt used.

bennothommo:
@LukeTowers this is ready for final review. I've implemented @Samuell1's changes, and removed the line break handling as it affected image files and other files where line breaks may be more important. We'll just have to treat different line breaks as October CMS being modified too.

Samuell1:
@bennothommo Can be linebreaks check added when PR is made to make it simple?

bennothommo:
@Samuell1 I'm thinking it might be best to consider line breaks being changed in the core files as a "modification". If someone's OS is changing line breaks on the fly, it may have unintended consequences.

LukeTowers:
@bennothommo were you able to get the tests working on your windows setup?

bennothommo:
@LukeTowers should be fixed now - `hash_file` doesn't handle line breaks, so a different hash occurs for Windows and Linux on the same file. I'm just going to convert line breaks to Unix for the manifest and for comparison purposes.

LukeTowers:
@bennothommo looks good to me! Feel free to merge into L6 and update the manifest files if appropriate.

LukeTowers:
@bennothommo Woohoo! Excited to have it in the core :)

bennothommo:
@LukeTowers same - the real test awaits for when 469 is tagged though - them new installs better show up as unmodified 469 :D

LukeTowers:
@bennothommo one final request, could we make the error message on the console instead say `Unable to detect your build of October CMS. Closest match was Build 468 with a -0.45% similarity.` We might want to adjust those numbers to have it make more sense or just communicate the closeness value differently, perhaps in terms of the number of changed / added / removed files.

LukeTowers:
Another thing we could consider is to just always return the closest build but include the number of changed / modified / removed files so that people can judge for themselves how close it is to reality.

Samuell1:
@LukeTowers i like that idea to shoe what files doesnt match maybe use cmd tag to return all files that doesnt match or are changed

bennothommo:
@LukeTowers @Samuell1 I've added a `--changes` argument to `october:version` to get the list of changed files as part of https://github.com/octobercms/october/commit/5bceda94960288e5403459eea7b101e3a4d788da. I've also adjusted the command to show a version regardless of if we're confident or not - the message is `We could not accurately determine your October CMS build due to the number of modifications. The closest detected build is October CMS build xxx`.

LukeTowers:
Absolutely wicked @bennothommo! Love it!

PR #5088, opened 2020-05-24
Fix client side validation

### LINKED ISSUE #5076: Validate backend forms client-side breaks current backend forms
I upgraded to the latest version and this pr broke my current forms in the backend: #4804.

I have a backend form field with a `dependsOn` on another form field. The depends on form field only gets updated when all the required fields are filled in, but not all required fields are filled in when you are in the middle of creating a new record.

For example my fields.yaml:

```yaml
    name:
        label: oms.core::lang.shop.name
        required: true
        type: text 
        span: auto

    code:
        label: oms.core::lang.shop.shop_code
        type: dropdown
        span: auto
        emptyOption: oms.core::lang.shop.select_shop_code

    _shop_fields:
        type: partial
        dependsOn: code
```

When I change the dropdown without filling in the name field, the `_shop_fields` isn't updating and the html 5 required field validation pops up. You can also see that the `_shop_fields` loading spinner is stuck, because the request isn't executed because of this pr.

image

LukeTowers:
Adds `data-request-validate` option to only execute client side validation when explicitly requested. Fixes #5076. Related: #4804.

### DIFF
--- a/modules/backend/controllers/preferences/index.htm
+++ b/modules/backend/controllers/preferences/index.htm
@@ -11,6 +11,7 @@
                 <button
                     type="submit"
                     data-request="onSave"
+                    data-request-validate
                     data-request-data="redirect:0"
                     data-hotkey="ctrl+s, cmd+s"
                     data-load-indicator="<?= e(trans('backend::lang.form.saving')) ?>"
--- a/modules/backend/controllers/usergroups/create.htm
+++ b/modules/backend/controllers/usergroups/create.htm
@@ -19,14 +19,16 @@
                 <button
                     type="submit"
                     data-request="onSave"
+                    data-request-validate
                     data-hotkey="ctrl+s, cmd+s"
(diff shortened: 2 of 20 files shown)


bennothommo:
@LukeTowers looks good to me - definitely a lot better as an "opt-in" function.

alxy:
Do you think the name is a good idea? The same is used to trigger the [ajax validation feature](https://octobercms.com/docs/ajax/extras#ajax-validation) on forms, now it does a completely different thing if applied to the submit buttons if is see that correctly?

LukeTowers:
@alxy good catch! Thanks for picking up on that, it's used in the framework extras file. I knew I had seen it somewhere before :(

LukeTowers:
@alxy fixed it in https://github.com/octobercms/october/commit/5c3ba9a28a9e14fbb47d1bf3dc5d60d9d8b1f901

PR #5100, opened 2020-05-31
Create a dashboard link in demo theme

### LINKED ISSUE #5098: Difficulties finding the backend
I struggled to find the backend UI. Here's what I did:

1. curl -s https://octobercms.com/api/installer | php
2. php artisan october:install
3. Opened it in browser (using Valet service)
4. See an 'October Demo' site; w/ 3 pages

But I could not find the backend admin UI; and I saw no indication of one (e.g., no login) in the demo.

IMHO this should be simpler/clearer to find. I ended up guessing by appending '/backend' to the URL and finding it while typing this issue. I tried '/admin' initially.

Proposed solution: reference (link to) it from the October Demo; and also clarify its location in the documentation. Perhaps the latter already exists as I'm only 10 minutes into this; but figured this feedback is potentially useful anyways.

RomainMazB:
Fix #5098

### DIFF
--- a/themes/demo/layouts/default.htm
+++ b/themes/demo/layouts/default.htm
@@ -1,5 +1,10 @@
 description = "Default layout"
 ==
+    function onInit()
+    {
+        $this['backend_url'] = Backend::url('/');
+    }
+==
 <!DOCTYPE html>
 <html>
     <head>
@@ -19,7 +24,7 @@ description = "Default layout"
 
         <!-- Header -->
         <header id="layout-header">
-            {% partial 'site/header' %}
+            {% partial 'site/header' backend_url=backend_url %}
         </header>
 
         <!-- Content -->
--- a/themes/demo/layouts/default.htm
+++ b/themes/demo/layouts/default.htm
@@ -1,10 +1,5 @@
 description = "Default layout"
 ==
-    function onInit()
-    {
-        $this['backend_url'] = Backend::url('/');
-    }
-==
 <!DOCTYPE html>
 <html>
     <head>
@@ -24,7 +19,7 @@ description = "Default layout"
 
         <!-- Header -->
         <header id="layout-header">
-            {% partial 'site/header' backend_url=backend_url %}
+            {% partial 'site/header' %}
         </header>
 
         <!-- Content -->
--- a/themes/demo/partials/site/header.htm
+++ b/themes/demo/partials/site/header.htm
@@ -16,6 +16,7 @@
                 <li class="{% if this.page.id == 'home' %}active{% endif %}"><a href="{{ 'home'|page }}">Basic concepts</a></li>
                 <li class="{% if this.page.id == 'ajax' %}active{% endif %}"><a href="{{ 'ajax'|page }}">AJAX framework</a></li>
                 <li class="{% if this.page.id == 'plugins' %}active{% endif %}"><a href="{{ 'plugins'|page }}">Plugin components</a></li>
+                <li><a href="{{ backend_url }}">{{ 'backend::lang.dashboard.menu_label'|trans }}</a></li>
             </ul>
         </div>
     </div>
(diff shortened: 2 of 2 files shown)


themes/demo/partials/site/header.htm
```diff
@@ -16,6 +16,7 @@
                 <li class="{% if this.page.id == 'home' %}active{% endif %}"><a href="{{ 'home'|page }}">Basic concepts</a></li>
                 <li class="{% if this.page.id == 'ajax' %}active{% endif %}"><a href="{{ 'ajax'|page }}">AJAX framework</a></li>
                 <li class="{% if this.page.id == 'plugins' %}active{% endif %}"><a href="{{ 'plugins'|page }}">Plugin components</a></li>
+                <li><a href="{{ backend_url }}">{{ 'backend::lang.dashboard.menu_label'|trans }}</a></li>
```
LukeTowers, line 19:
```suggestion
                <li><a href="{{ backend_url }}" target="_blank">Backend</a></li>
```

No need to translate it, nothing else in the demo theme is and it would only apply to the backend locale, which is always going to be english for new installations anyways.

themes/demo/layouts/default.htm
```diff
@@ -19,7 +24,7 @@
 
         <!-- Header -->
         <header id="layout-header">
-            {% partial 'site/header' %}
+            {% partial 'site/header' backend_url=backend_url %}
```
LukeTowers, line 27: You can just implement this in the partial itself, see https://octobercms.com/docs/cms/partials#partial-life-cycle

themes/demo/partials/site/header.htm
RomainMazB: You're right.

The French guy (who translate everything).

themes/demo/layouts/default.htm
RomainMazB: Hmmm. I tried this before implement it in the layout, but I surely made a crap because it just displayed the php code block as raw text.

Will modify my crap.

RomainMazB:
I've moved the php block to the partial.

themes/demo/partials/site/header.htm
```diff
@@ -1,3 +1,9 @@
+==
+    function onStart()
+    {
+        $this['backend_url'] = Backend::url('/');
+    }
```
LukeTowers, line 5:
```suggestion
function onStart()
{
    $this['backend_url'] = Backend::url('/');
}
```

bennothommo:
@RomainMazB @LukeTowers one thing to consider - the term "Backend" is a less-common term used for what a lot of people consider to be the "Administration Panel". Given that the link currently just says "Backend", it might not be glaringly obvious on someone's first-time use. I wonder if the link instead says "Sign in to Backend" - which should be enough of a hint that this is a privileged area?

LukeTowers:
@bennothommo good point, although "Sign in to Backend" is not quite accurate because it's not taking you to the login page. What about "View Backend"?

RomainMazB:
In fact in a fresh install, it does redirect to the login page since the session isn't initialized yet.
Then the user knows where the link goes.

Since the demo theme is a "short-lived theme", the link text should definitely target the really new user. In my project it lives maybe 5 seconds until I delete it to install my bootstrap theme.

I've changed the link to what @bennothommo said it should be. Feel free to purpose anything else. I stay tuned.

bennothommo:
@LukeTowers re: "View Backend" - might be a bit of a euphemism :P

themes/demo/partials/site/header.htm
```diff
@@ -1,3 +1,9 @@
+==
+function onStart()
+{
+    $this['backend_url'] = Backend::url('/');
```
bennothommo, line 4:
```suggestion
    $this['backendUrl'] = Backend::url('/');
```

themes/demo/partials/site/header.htm
```diff
@@ -16,6 +22,7 @@
                 <li class="{% if this.page.id == 'home' %}active{% endif %}"><a href="{{ 'home'|page }}">Basic concepts</a></li>
                 <li class="{% if this.page.id == 'ajax' %}active{% endif %}"><a href="{{ 'ajax'|page }}">AJAX framework</a></li>
                 <li class="{% if this.page.id == 'plugins' %}active{% endif %}"><a href="{{ 'plugins'|page }}">Plugin components</a></li>
+                <li><a href="{{ backend_url }}" target="_blank">Sign in to Backend</a></li>
```
bennothommo, line 25:
```suggestion
                <li><a href="{{ backendUrl }}" target="backend">Sign in to the Backend</a></li>
```

PR #5109, opened 2020-06-05
Send button value with AJAX requests

### LINKED ISSUE #5107: Send button value in AJAX requests
I have the following HTML for one of my components:

``` html
<form data-request="onAjaxRequest">
    <button type="submit" name="direction" value="forward">Continue</button>
    <button type="submit" name="direction" value="back">Go back</button>
</form>
```

In the component's `onAjaxRequest() method, I do:

``` php
public function onAjaxRequest()
{
    var_dump(Input::all());
    exit;
}
```
This outputs an empty array:

```
array(0) {
}
```

According to the [`<button>` spec](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button):

> **value**
Defines the value associated with the button’s name when it’s submitted with the form data. This value is passed to the server in params when the form is submitted.

multiwebinc:
Fixes #5107

This changes the behavior of the AJAX framework to act more like a standard HTML `<form>` element. 

For regular, non-AJAX form submissions, if the form is submitted by clicking on a specific submit button, the name/value attributes of that button are sent along with the form data. If the form is submitted without clicking on a submit button (e.g. by pushing enter on an `<input type="text">` field), the name/value pair of the **first** submit button to appear in the DOM are sent. In either case, if there is no `name` attribute, then nothing is added to the form data that sent. 

This PR mimics this behavior for AJAX requests.

### DIFF
--- a/modules/system/assets/js/framework.js
+++ b/modules/system/assets/js/framework.js
@@ -127,7 +127,9 @@ if (window.jQuery.request !== undefined) {
         }
         else {
             var submitButtonValue,
-                ae = $(document.activeElement)
+                ae = $(document.activeElement),
+                submits,
+                submitValue
 
             /*
              * A specific submit button was clicked, so add that value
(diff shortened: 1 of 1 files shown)


mjauvin:
Why is that useful exactly?

multiwebinc:
@mjauvin This allows the server to perform different actions depending on which button was pressed.

mjauvin:
You can still do that by passing data with data-request-data

LukeTowers:
@multiwebinc @mjauvin has a good point, why not just use data-request-data on the button itself?

multiwebinc:
@LukeTowers @mjauvin Good point. I wasn't aware of that attribute. Then again, I'm probably not alone. Allowing the AJAX framework to behave more like default HTML forms I wouldn't think would be a bad thing.

LukeTowers:
Perhaps not, but it is a change in existing behaviour. I have a WIP changing the way request data is compiled right now (supporting pulling in parent request data as well) so we can revisit this suggestion once that's done.

LukeTowers:
@multiwebinc is this still something you're wanting to push through?

bennothommo:
I personally feel this is something that is not needed at this stage. I worry about the amount of auditing we'd have to do to entire codebase to ensure that no values are being sent with buttons throughout the CMS.

We've implemented a standard that extra data is sent through the `data-request-data` attribute, so I think it's best we stick with that.

multiwebinc:
@bennothommo Sounds good to me!

PR #5112, opened 2020-06-06
Unbind event handlers upon destruction of the form widget

### LINKED ISSUE #5113: JS form widgets are not properly disposed of in the relation manager
- OctoberCMS Build: 466

### Steps To Reproduce:

1. Create an instance of a relation manager that manages a form with some field dependencies in it.
2. Click the "Create" button in the relation manager to open up a popup with the form.
3. Trigger a reload of a dependant field by modifying the value of the field it depends on. A single `onRefresh` request is sent to the backend. Close the popup.
4. Click the "Create" button again and trigger a reload of a dependant field. **Two** `onRefresh` requests are sent to the backend instead of one.
5. Repeating steps 2-3 will [bind](https://github.com/octobercms/october/blob/52cdfda93564723de684a0951c190d532c8f2dc7/modules/backend/widgets/form/assets/js/october.form.js#L126) an additional `onRefresh` handler each time the form is constructed which is not unregistered when closing the popup.

The solution is to properly dispose of the form instance by calling its `dispose()` method before closing the popup.

This bug may be present in other widgets as well. I have only tried to reproduce it in the relation manager and my own custom form widgets which load a popup form via AJAX.

PR #5112 is related to this issue because it introduces some fixes to the disposal logic of the form widget. An additional fix/PR would be required in order to fix the issue described above.

msimkunas:
This PR unbinds various event handlers of the form widget when an instance of it is destroyed.

The current implementation does not perform proper clean up in the `dispose` method. This results in lingering event handlers when the form widget is destroyed.

The problem is especially apparent in use cases where a form widget is created and destroyed multiple times without reloading the page, e.g. when loading the form inside a popup via AJAX. If the form has any field dependencies, the JS handler `onRefreshDependants` which is responsible for refreshing those fields will get registered each time the form is constructed and it will **not** be unregistered when the form object is disposed of.

Calling the `dispose()` method in order to manually destroy the form widget [sets `this.$el` to `null`](https://github.com/octobercms/october/blob/develop/modules/backend/widgets/form/assets/js/october.form.js#L48) but it does not unbind the refresh handlers. Because of this they are executed when the form is rebuilt and a field is refreshed. This results in the following error:
```
TypeError: form is null
```
...at `october.form.js:88:28`, which breaks the refresh handlers altogether. From what I gather, this happens because the old handlers are attempting to [access `this.$el`](https://github.com/octobercms/october/blob/develop/modules/backend/widgets/form/assets/js/october.form.js#L88). However, since it has already been `null`'ed when calling `dispose()`, they get no reference to the element and a `TypeError` is returned.

This PR fixes this by properly unbinding the event handlers when the form widget is destroyed with the `dispose()` method.

### DIFF
--- a/modules/backend/widgets/form/assets/js/october.form.js
+++ b/modules/backend/widgets/form/assets/js/october.form.js
@@ -160,7 +160,7 @@
             })
         })
 
-        return fieldMap;
+        return fieldMap
     }
 
     /*
--- a/modules/backend/widgets/form/assets/js/october.form.js
+++ b/modules/backend/widgets/form/assets/js/october.form.js
@@ -129,7 +129,7 @@
         var fieldMap = this._getDependants()
 
         $.each(fieldMap, function(fieldName, toRefresh) {
-            $(document).off('change.oc.formwidget', '[data-field-name="' + fieldName + '"]');
+            $(document).off('change.oc.formwidget', '[data-field-name="' + fieldName + '"]')
         })
     }
 
--- a/modules/backend/widgets/form/assets/js/october.form.js
+++ b/modules/backend/widgets/form/assets/js/october.form.js
@@ -242,7 +242,7 @@
     FormWidget.prototype.unbindLazyTabs = function() {
         var tabControl = $('[data-control=tab]', this.$el)
 
-        $('.nav-tabs', tabControl).off('click', '.tab-lazy [data-toggle="tab"]');
+        $('.nav-tabs', tabControl).off('click', '.tab-lazy [data-toggle="tab"]')
     }
 
     /*
--- a/modules/backend/widgets/form/assets/js/october.form.js
+++ b/modules/backend/widgets/form/assets/js/october.form.js
@@ -39,6 +39,7 @@
 
         this.$el.on('oc.triggerOn.afterUpdate', this.proxy(this.toggleEmptyTabs))
         this.$el.one('dispose-control', this.proxy(this.dispose))
+        this.$form.closest('.modal').on('hidden.oc.popup', this.proxy(this.dispose))
     }
 
     FormWidget.prototype.dispose = function() {
--- a/modules/backend/widgets/form/assets/js/october.form.js
+++ b/modules/backend/widgets/form/assets/js/october.form.js
@@ -43,10 +43,10 @@
     }
 
     FormWidget.prototype.dispose = function() {
-        this.unbindDependants();
-        this.unbindCheckboxList();
-        this.unbindLazyTabs();
-        this.unbindCollapsibleSections();
+        this.unbindDependants()
+        this.unbindCheckboxList()
+        this.unbindLazyTabs()
+        this.unbindCollapsibleSections()
 
         this.$el.off('dispose-control', this.proxy(this.dispose))
         this.$el.removeData('oc.formwidget')
--- a/modules/backend/widgets/form/assets/js/october.form.js
+++ b/modules/backend/widgets/form/assets/js/october.form.js
@@ -85,8 +85,8 @@
      * Unbind checkboxlist handlers
      */
     FormWidget.prototype.unbindCheckboxList = function() {
-        this.$el.off('click', '[data-field-checkboxlist-all]');
-        this.$el.off('click', '[data-field-checkboxlist-none]');
+        this.$el.off('click', '[data-field-checkboxlist-all]')
+        this.$el.off('click', '[data-field-checkboxlist-none]')
     }
 
     /*
--- a/modules/backend/widgets/form/assets/js/october.form.js
+++ b/modules/backend/widgets/form/assets/js/october.form.js
@@ -112,7 +112,7 @@
             fieldMap = this._getDependants()
 
         /*
-         * When a master is updated, refresh its slaves
+         * When a field is updated, refresh its dependents
          */
         $.each(fieldMap, function(fieldName, toRefresh) {
             $(document).on('change.oc.formwidget',
--- a/modules/backend/widgets/form/assets/js/october.form.js
+++ b/modules/backend/widgets/form/assets/js/october.form.js
@@ -118,7 +118,7 @@
             $(document).on('change.oc.formwidget',
                 '[data-field-name="' + fieldName + '"]',
                 $.proxy(self.onRefreshDependants, self, fieldName, toRefresh)
-            );
+            )
         })
     }
 
--- a/modules/backend/widgets/form/assets/js/october.form.js
+++ b/modules/backend/widgets/form/assets/js/october.form.js
@@ -308,7 +308,7 @@
      * Unbinds collapsible section handlers
      */
     FormWidget.prototype.unbindCollapsibleSections = function() {
-        $('.section-field[data-field-collapsible]', this.$form).off('click');
+        $('.section-field[data-field-collapsible]', this.$form).off('click')
     }
 
     FormWidget.DEFAULTS = {
--- a/modules/backend/widgets/form/assets/js/october.form.js
+++ b/modules/backend/widgets/form/assets/js/october.form.js
@@ -42,6 +42,11 @@
     }
 
     FormWidget.prototype.dispose = function() {
+        this.unbindDependants();
+        this.unbindCheckboxList();
+        this.unbindLazyTabs();
+        this.unbindCollapsibleSections();
+
         this.$el.off('dispose-control', this.proxy(this.dispose))
         this.$el.removeData('oc.formwidget')
 
@@ -76,6 +81,14 @@
     }
 
     /*
+     * Unbind checkboxlist handlers
+     */
+    FormWidget.prototype.unbindCheckboxList = function() {
+        this.$el.off('click', '[data-field-checkboxlist-all]');
+        this.$el.off('click', '[data-field-checkboxlist-none]');
+    }
+
+    /*
      * Get all fields elements that belong to this form, nested form
      * fields are removed from this collection.
      */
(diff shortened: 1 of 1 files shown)


msimkunas:
@LukeTowers @bennothommo I've updated the problem description in my original post to more accurately reflect what is going on.

msimkunas:
I've documented a possible fix for #5113 (see the issue for details). Should I add it as a commit here or post it as a separate PR?

LukeTowers:
@msimkunas add it as a commit here for us to review.

modules/backend/widgets/form/assets/js/october.form.js
```diff
@@ -39,9 +39,15 @@
...
         this.$el.on('oc.triggerOn.afterUpdate', this.proxy(this.toggleEmptyTabs))
         this.$el.one('dispose-control', this.proxy(this.dispose))
+        this.$form.closest('.modal').on('hidden.oc.popup', this.proxy(this.dispose))
     }
 
     FormWidget.prototype.dispose = function() {
+        this.unbindDependants();
+        this.unbindCheckboxList();
+        this.unbindLazyTabs();
+        this.unbindCollapsibleSections();
```
bennothommo, line 49: @msimkunas We don't use semicolons in our JS code at the moment (even though I'd love to). Let's remove them for consistency.
```suggestion
        this.unbindDependants()
        this.unbindCheckboxList()
        this.unbindLazyTabs()
        this.unbindCollapsibleSections()
```

modules/backend/widgets/form/assets/js/october.form.js
```diff
@@ -75,6 +81,14 @@
 
     }
 
+    /*
+     * Unbind checkboxlist handlers
+     */
+    FormWidget.prototype.unbindCheckboxList = function() {
+        this.$el.off('click', '[data-field-checkboxlist-all]');
+        this.$el.off('click', '[data-field-checkboxlist-none]');
```
bennothommo, line 89:
```suggestion
        this.$el.off('click', '[data-field-checkboxlist-all]')
        this.$el.off('click', '[data-field-checkboxlist-none]')
```

modules/backend/widgets/form/assets/js/october.form.js
```diff
@@ -94,13 +108,40 @@
      * Bind dependant fields
      */
     FormWidget.prototype.bindDependants = function() {
+        var self = this,
+            fieldMap = this._getDependants()
 
+        /*
+         * When a master is updated, refresh its slaves
```
bennothommo, line 115: @msimkunas Just to make it more clear what is happening (and ward off the PC crowd)
```suggestion
         * When a field is updated, refresh its dependents
```

modules/backend/widgets/form/assets/js/october.form.js
```diff
@@ -94,13 +108,40 @@
...
+    }
+
+    /*
+     * Dispose of the dependant field handlers
+     */
+    FormWidget.prototype.unbindDependants = function() {
+        var fieldMap = this._getDependants()
+
+        $.each(fieldMap, function(fieldName, toRefresh) {
+            $(document).off('change.oc.formwidget', '[data-field-name="' + fieldName + '"]');
```
bennothommo, line 132:
```suggestion
            $(document).off('change.oc.formwidget', '[data-field-name="' + fieldName + '"]')
```

modules/backend/widgets/form/assets/js/october.form.js
```diff
@@ -119,15 +160,7 @@
...
-        /*
-         * When a master is updated, refresh its slaves
-         */
-        $.each(fieldMap, function(fieldName, toRefresh) {
-            $(document).on('change.oc.formwidget',
-                '[data-field-name="' + fieldName + '"]',
-                $.proxy(self.onRefreshDependants, self, fieldName, toRefresh)
-            );
-        })
+        return fieldMap;
```
bennothommo, line 163:
```suggestion
        return fieldMap
```

modules/backend/widgets/form/assets/js/october.form.js
```diff
@@ -203,6 +236,15 @@
         }
     }
 
+    /*
+     * Unbind the lazy tab handlers
+     */
+    FormWidget.prototype.unbindLazyTabs = function() {
+        var tabControl = $('[data-control=tab]', this.$el)
+
+        $('.nav-tabs', tabControl).off('click', '.tab-lazy [data-toggle="tab"]');
```
bennothommo, line 245:
```suggestion
        $('.nav-tabs', tabControl).off('click', '.tab-lazy [data-toggle="tab"]')
```

modules/backend/widgets/form/assets/js/october.form.js
```diff
@@ -262,6 +304,13 @@
             .nextUntil('.section-field').hide()
     }
 
+    /*
+     * Unbinds collapsible section handlers
+     */
+    FormWidget.prototype.unbindCollapsibleSections = function() {
+        $('.section-field[data-field-collapsible]', this.$form).off('click');
```
bennothommo, line 311:
```suggestion
        $('.section-field[data-field-collapsible]', this.$form).off('click')
```

bennothommo (changes requested):
@msimkunas In addition to the below, I've noticed the form JS uses the term `dependants` for variables and method names. Could you correct the spelling where applicable to `dependents`.

msimkunas:
> @msimkunas In addition to the below, I've noticed the form JS uses the term `dependants` for variables and method names. Could you correct the spelling where applicable to `dependents`.

Sure, will do in a moment. I simply reused the same term found in `bindDependants` in order to maintain consistency. I guess I should rename that as well.

modules/backend/widgets/form/assets/js/october.form.js
msimkunas: This should also be updated in the existing code as well: https://github.com/octobercms/october/blob/develop/modules/backend/widgets/form/assets/js/october.form.js#L106-L131

modules/backend/widgets/form/assets/js/october.form.js
```diff
@@ -94,13 +108,40 @@
...
+            fieldMap = this._getDependants()
+
+        /*
+         * When a field is updated, refresh its dependents
+         */
+        $.each(fieldMap, function(fieldName, toRefresh) {
+            $(document).on('change.oc.formwidget',
+                '[data-field-name="' + fieldName + '"]',
+                $.proxy(self.onRefreshDependants, self, fieldName, toRefresh)
+            );
```
msimkunas, line 121:
```suggestion
            )
```

bennothommo:
@msimkunas On second thoughts, let's not change the term. I've just been informed by a colleague that "dependant" is actually valid - https://www.merriam-webster.com/words-at-play/spelling-variants-dependent-vs-dependant.

bennothommo (approved):

bennothommo:
@github-actions run

<details>

```js
(async () => {
  // Get issue/PR number

  // Add Testing Needed label
    owner: context.repo.owner,
    repo: context.repo.repo,
    issue_number: number,
    labels: ['Testing Needed'],
  });

  // Comment body
    + '\n\n'
    + '\n\n'
    + '```\n'
    + '```';

  // Add comment
})();
```

msimkunas:
> @msimkunas On second thoughts, let's not change the term. I've just been informed by a colleague that "dependant" is actually valid - https://www.merriam-webster.com/words-at-play/spelling-variants-dependent-vs-dependant.

I was about to mention that at first but then figured that it may not be worth the energy to discuss. :) Anyway, thanks for reviewing this!

bennothommo:
Apparently I've been spelling it wrong my entire life :P

msimkunas:
I've noticed that the `getFieldElements` method does not return fields within lazily loaded tabs which is why no refresh handlers are bound to those fields. This makes it impossible to `dependOn` such fields in the form field configuration.

This may be a bit out of scope for this PR so I guess I'll just make a new one if/when I come up with a solution.

bennothommo:
@msimkunas The docs on https://octobercms.com haven't been updated yet, but our latest docs do indicate the lazy tabs and triggers/dependsOn do not mix: https://github.com/octobercms/docs/blob/master/backend-forms.md#tab-options - so we've indemnified ourselves for now.

I think it's best to keep the scope of this PR lean for now, and we'll revisit the lazy tabs trigger incompatibility after the L6 update is out.

mjauvin:
@msimkunas would you be able to provide a simple PR to the [October/test-plugin](https://github.com/octoberrain/test-plugin) in order to test this? I could setup an Octodock instance with this PR and the one to the test-plugin merged in so we can easily test this...

Thanks

msimkunas:
@mjauvin I'll try to post a PR as soon as I have time.

LukeTowers:
@msimkunas are you still going to be able to put up a PR to the test plugin so that we can more easily test this?

msimkunas:
@LukeTowers Yes, I am. Sorry for the delay, I've been pretty busy lately.

modules/backend/widgets/form/assets/js/october.form.js
```diff
@@ -39,9 +39,15 @@
 
         this.$el.on('oc.triggerOn.afterUpdate', this.proxy(this.toggleEmptyTabs))
         this.$el.one('dispose-control', this.proxy(this.dispose))
+        this.$form.closest('.modal').on('hidden.oc.popup', this.proxy(this.dispose))
```
mplodowski, line 42: If you have recordfinder field type in relation popup and have some fields that depends on it, than this line causes to break dependsOn functionality. If you select something from recordfinder it will not refresh other fields.

msimkunas: Thanks for letting me know. It seems like we need to bind the hidden popup event on a different form element. I'll look into this.

daftspunk:
Thanks @msimkunas!

Issue #5097, opened 2020-05-28
Potential vulnerability in jQuery


mrgswift:
Since this is already a public security advisory and a jQuery vulnerability, I am posting it as a public issue.

**Summary:**

[**CVE-2020-11022**](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-11022)
In jQuery versions greater than or equal to 1.2 and before 3.5.0, passing HTML from untrusted sources - even after sanitizing it - to one of jQuery's DOM manipulation methods (i.e. .html(), .append(), and others) may execute untrusted code. This problem is patched in jQuery 3.5.0

**Problem:**

The current stable version of OctoberCMS uses jQuery 3.4.0 which is vulnerable.  

**Solution:**

The code-base should either be updated to support jQuery 3.5.0 or higher, or **jQuery.htmlPrefilter** should be overridden as a workaround - discussed here:
[jQuery GitHub security advisory](https://github.com/jquery/jquery/security/advisories/GHSA-gxr4-xjj5-5px2)

From link above - workaround:
`jQuery.htmlPrefilter = function( html ) {
	return html;
};`

LukeTowers:
@mrgswift I realize that this is already public but I would prefer to have these reports come through privately first. We deal with all security reports very quickly so I prefer to discuss them back and forth over email before public issues are submitted when they may not be applicable.

mrgswift:
Understood.  From now on anything related to security will be discussed privately.

LukeTowers:
Thanks, I appreciate it!

LukeTowers:
Fixed by https://github.com/octobercms/october/commit/5c7ba9fbe9f2b596b2f0e3436ee06b91b97e5892, thanks for the report @mrgswift!

@bennothommo is there a way we can tell GitHub what our JS dependencies are so that we can get notified when they release security advisories?

bennothommo:
@LukeTowers we have to define the specific versions we are running in the `package.json` file I've already set up for JS testing - https://github.com/octobercms/october/network/dependencies. At the moment, it see us as running as `^ 3.4.1`, which means we wouldn't have got the advisory as we're saying we can accept versions above that.

LukeTowers:
@bennothommo we should probably switch that to be more of a reporting feature than an actual package management file, i.e. specify the exact package versions we use

bennothommo:
@LukeTowers yep, agreed, as long as the tests still work. I'll take a look soon.

[...]

Issue #4665, opened 2019-10-04
Icons pages not showing icon list  


Nvironmental:
Hi Guys, 

I noticed something weird, on this page [October Icons](https://octobercms.com/docs/ui/icon) the icon list is missing. It was available till like 2 days ago :/ 

Am I missing something ?

w20k:
Check your adblocker:
<img width="882" alt="Screenshot 2019-10-04 at 20 30 54" src="https://user-images.githubusercontent.com/1053320/66227485-e8d9d500-e6e5-11e9-9a0a-f0492a5a797b.png">

mlieshout:
The icon's won't show in latest Firefox (69.0.2):

<img width="889" alt="Firefox 69 0 2 Screenshot 2019-10-05 at 13 26 16" src="https://user-images.githubusercontent.com/994055/66254253-ffc70880-e773-11e9-9199-84935ab6de32.png">

<img width="1247" alt="Screenshot 2019-10-05 at 13 30 31" src="https://user-images.githubusercontent.com/994055/66254270-57fe0a80-e774-11e9-902e-269d95aa8fb3.png">

bennothommo:
@mlieshout It might've been a temporary error. They are appearing just fine for me, albeit on 69.0.1:

![image](https://user-images.githubusercontent.com/15900351/66255275-679d4380-e7b4-11e9-98c4-19f564ae498a.png)

w20k:
Works for me as well.
<img width="955" alt="Screenshot 2019-10-05 at 17 01 11" src="https://user-images.githubusercontent.com/1053320/66255934-d917ca80-e791-11e9-8fd5-89f5467da4a4.png">
<img width="973" alt="Screenshot 2019-10-05 at 17 01 43" src="https://user-images.githubusercontent.com/1053320/66255935-d9b06100-e791-11e9-8beb-ae08b2663d16.png">

mlieshout:
It's a min-height problem of the iframe. When I change it from 500px to 800px, a part is shown. When I look into the frame source, the whole page is loaded.
Same thing happens in Chrome v77.0.3865.90 
I'm on MacOS 10.14.6 btw.

<img width="1364" alt="Screenshot 2019-10-05 at 17 12 01" src="https://user-images.githubusercontent.com/994055/66256950-935c0180-e793-11e9-8453-7a851370e723.png">

w20k:
[...]

PR #5115, opened 2020-06-07
Update jQuery 3.5.1 and jQuery Migrate 3.3.0

summercms:
@LukeTowers

This pr fixes issue: #5097

The reason I'm tagging Luke is because he has upgraded jQuery to 3.4.0 (which doesn't contain the security fix) - see here for his commit: 5c7ba9f

This PR adds jQuery 3.5.1 and jQuery Migrate 3.3.0

Any issues tag me and I will sort it out.

---

jQuery upgrade guide here: https://jquery.com/upgrade-guide/3.5/


LukeTowers:
@ayumi-cloud my commit does contain the fix, I added it in manually even though it's still running 3.4. Didn't upgrade the version because I didn't have the time to test to make sure 3.5 still works

ayumi-cloud:
@LukeTowers no problem, I just saw the commit and wondered what was going on and thought I'd just help things out and push this pr - as I know you guys are busy!

I was thinking the backend wouldn't matter with this CVE-2020-11022. The only place that maybe an issue might be the `framework` files on the front end. If someone wants to test any issues there that would be great.

Lastly, I been using 3.5.1 for a few weeks in October and seen no issues at all.

LukeTowers:
@ayumi-cloud I think you're missing the update to the jquery standalone file too, could you add that?

ayumi-cloud:
Sorry my bad - done.

mjauvin:
I created an Octodock instance with this PR merged in for testing (user/pass = admin/admin as usual):

https://5ec7a038898f.octodock.com/backend
https://5ec7a038898f.octodock.com/

summercms:
Next build jQuery 3.6.0 - 73% complete, see here: https://github.com/jquery/jquery/milestone/22

Also much later on in life jQuery 4.0.0, see here: https://github.com/jquery/jquery/milestone/7

mjauvin:
ok, but let's not get ahead of ourselves... what about testing this one first?!

LukeTowers:
Meh, I'm not sure I see the value in sticking with the absolute latest version of jQuery. Sure, we should update it every once in a while but what's the actual tangible benefit of updating it right now?

summercms:
@LukeTowers Security fixes is the reason for the upgrade from 3.5.0 to 3.5.1 and many features and bug fixes for upgrade from 3.4.0 to 3.5.0, see below an outline for you.

### Security Fix 1

The main change in this release is a security fix, and it’s possible you will need to change your own code to adapt. Here’s why: jQuery used a regex in its jQuery.htmlPrefilter method to ensure that all closing tags were XHTML-compliant when passed to methods. For example, this prefilter ensured that a call like `jQuery("<div class='hot' />")` is actually converted to` jQuery("<div class='hot'></div>")`. Recently, an issue was reported that demonstrated the regex could introduce a cross-site scripting (XSS) vulnerability.

The HTML parser in jQuery <=3.4.1 usually did the right thing, but there were edge cases where parsing would have unintended consequences. The jQuery team agreed it was necessary to fix this in a minor release, even though some code relies on the previous behaviour and may break. The `jQuery.htmlPrefilter` function does not use a regex in 3.5.0 and passes the string through unchanged.

If you absolutely need the old behaviour, using the latest version of the jQuery migrate plugin provides a function to restore the old `jQuery.htmlPrefilter`. After including the plugin you can call `jQuery.UNSAFE_restoreLegacyHtmlPrefilter()` and jQuery will again ensure XHTML-compliant closing tags.

However, to sanitize user input properly, we also recommend using `dompurify` with the `SAFE_FOR_JQUERY` option to sanitize HTML from a user. If you don’t need the old behaviour, but would still like to sanitize HTML from a user, `dompurify` should be used without the `SAFE_FOR_JQUERY` option, starting in jQuery 3.5.0. For more details, please see the 3.5 Upgrade Guide.

### Security Fix 2

jQuery 3.5.0 included fixes for two security issues in jQuery’s DOM manipulation methods, as in `.html()`, `.append()` and the others. Security advisories for both of these issues have been published on GitHub. While we provided all of the details on the first issue in the jQuery 3.5.0 blog post, we did not provide all of the details on the second and would like to do that in this post.

The second issue was very similar to the first. It was an XSS vulnerability that had to do with passing `<option>` elements to jQuery’s DOM manipulation methods. Essentially, we’re using a regex to wrap <option> elements with `<select>` elements to ensure those elements get parsed correctly in old IE (IE <= 9 replaces any `<option>` tags with their contents when inserted outside of a `<select`> element).

Our fix is to only apply this code where it is needed. Fortunately, because of the different parsing behaviour in IE9, we can keep the fix in IE9 without exposing it to the same vulnerability as other browsers. Please upgrade when you get a chance to avoid these vulnerabilities.

## Changelog: 3.5.1

### Build

- Test on Node.js 14, stop testing on Node.js 8 & 13 (205dd134)
- Enable reportUnusedDisableDirectives in ESLint (b21d6710)
- Updating the 3.x-stable version to 3.5.1-pre. (898784ab)

### Data

- Make the data object a regular object again (#4665, 65e90984)

### Docs

- Fix typos (ea3766c0)
- Update the link to the jsdom repository (f7fed7ee)
- Use https for hyperlinks in README (763dd3fe)

### Tests

- Workaround failures in recent XSS tests in iOS 8 – 12 (ea2d0d50)
- Add tests for recently fixed manipulation XSS issues (58a8e879)
- Cleanup `window` & `document` handlers in a new event test (c1c0598d)
- Fix flakiness in the “jQuery.ajax() – JSONP – Same Domain” test (46ba70c5)

## Changelog: 3.5.0

### Ajax

- Do not execute scripts for unsuccessful HTTP responses (#4250, #4655, da3dd85b)
- Overwrite s.contentType with content-type header value, if any (#4119, 065143c2)
- Deprecate AJAX event aliases, inline event/alias into deprecated (7a3cf9c0)

### Build

- Resolve Travis config warnings (7506c9ca)
- Enable ESLint one-var rule for var declarations in browser code (0fdfdd82)
- Test the no-Sizzle build on Travis (362075ae)
- Update .mailmap & AUTHORS.txt (19f2dcba)
- Tests: Fix custom build tests, verify on Travis; name Travis jobs (d525ae34)
- Lint the minified jQuery file as well (#3075, 37df5cdf)
- Make Karma work in AMD mode (46c284b1)
- Create a `grunt custom:slim` alias for the Slim build (4cbdc745)
- Run tests on Travis only on browsers defined in the config (471b0043)
- Run tests on Firefox ESR as well (0a73b94a)
- Run tests on Node.js 13 in addition to 8, 10 & 12 (64c1fcc1)
- Drop workarounds for Node.js 6 in Gruntfile.js (9f4204ec)
- Run tests on Travis on FirefoxHeadless as well (ad3c2efa)
- Require strict mode in Node.js scripts via ESLint (ac2da4e6)
- Support jquery-release –dry-run flag (c7a5e1bd)
- Stop copying src/core.js to dist on release (#4489, 279d2e97)
- ESLint: forbid unused function parameters (d7e13f12)
- Fix the regex parsing AMD var-modules (#4389) (36b59c96)

### Core

- Ajax: Align nonce & global with master, fix an AMD issue (22bf701f)
- Fire iframe script in its context, add doc param in globalEval (#4518, 3dedc3f2)
- Deprecate jQuery.trim (#4363, 56e73e0c)
- Use Array.prototype.flat where supported (#4320, 2f666c1d)
- Implement .even() & .odd() to replace POS :even & :odd (409cbda7)

### CSS

- Workaround buggy getComputedStyle on table rows in IE/Edge (#4490, 6d31477a)

### Data

- Event:Manipulation: Prevent collisions with Object.prototype (#3256, 413ff796)

### Docs

- Update links to EdgeHTML issues to go through Web Archive (d72faced)
- Convert link to Homebrew from HTTP to HTTPS (ff5a43eb)

### Effect

- Fix a unnecessary conditional statement in .stop() (#4374, 30f5c6c3)

### Event

- Use only one focusin/out handler per matching window & document (#4652, 9e15d6b4)
- Only attach events to objects that accept data – for real (#4397, f36f6abb)

### Manipulation

- Skip the select wrapper for option elements
- Make jQuery.htmlPrefilter an identity function (1d61fd94)

### Offset

- Send px-ed strings to .css() (57038fae)

### Selector

- Update Sizzle from 2.3.4 to 2.3.5 (#4424, #4435, #4441, #4453, 04bf577e)
- Make selector-native’s isXMLDoc recognize HTML-embedded SVG (2d9d6d5b)

### Tests

- Blacklist one focusin test in IE (1a4f10dd)
- Pass a number of necessary done() calls to assert.async() (5ea844f6)
- Make the support tests pass on Firefox 4x/5x/60 (f0d5ec62)
- Skip a “width/height on a table row with phantom borders” test in Firefox (c79e1d5f)
- Don’t test synchronous XHR on unload in Chrome (c5b48c8c)
- Fix offset fractions tests in Chrome for Android (0c67da4b)
- Move Android user agent detection above iOS, put Safari last (6276cb2e)
- Make support tests accept Safari 13 & newer (8167327f)
- update npo.js and include unminified source instead (3654bc83)

### Traversing

- Fix contents() on object elements with children in IE (90f78b9a)
- Fix contents() on object elements with children (#4384, 42badf34)

summercms:
Then the future jQuery 3.6.0 fixes the Safari browser bug doing this:

56849420-33178880-68f4-11e9-920f-63966de3f023

(To give you some idea of the reasons behind the jQuery version updates).

LukeTowers:
Let's just wait for 3.6 then

mjauvin:
@LukeTowers you're saying this PR should be closed in favor of future update to jquery v3.6.0 ?

LukeTowers:
Yeah, @ayumi-cloud just resubmit with all the details with 3.6 is ready and review it then.

summercms:
Opening a new pr as something has come up in regards to jQuery - will create a new pr. I pm'ed Luke in Slack about it.

PR #5117, opened 2020-06-08
Explain to people how deal with security issues

ayumi-cloud:
I've got an issue with how this repo deals with security issues!

A person opens an issue and says it's a security issue. Then an admin tells the person they should not open an issue and they should tell them in private. The problem I have with this is that people are not mind readers and should be told before they try to open an issue.

This pr tries to fix this problem and be a bit more open to the community and let them understand the process a bit better.

### Before

image

### After

image

### DIFF
--- a/.github/ISSUE_TEMPLATE/1_BUG_REPORT.md
+++ b/.github/ISSUE_TEMPLATE/1_BUG_REPORT.md
@@ -1,6 +1,6 @@
 ---
 name: "🐛 Bug Report"
-about: 'Report a general OctoberCMS issue "FOR SECURITY ISSUES PLEASE CONTACT THE ADMINS IN PRIVATE AND DONT OPEN AN ISSUE HERE"'
+about: 'Report a general OctoberCMS issue'
 labels: 'Status: Review Needed, Type: Unconfirmed Bug'
 ---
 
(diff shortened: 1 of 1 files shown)


ayumi-cloud:
@LukeTowers Please can I get your feedback on this.

RomainMazB:
Just to add an information, I've found this yesterday which seems to be accurate to add to the discussion:
[SECURITY.md](https://github.com/octobercms/october/blob/develop/SECURITY.md)

ayumi-cloud:
@RomainMazB Thanks for your input.

In a perfect world yah, people would read that. But in truth no one can be bothered to hunt deep into the repo and waste time finding that info.

So a quick note to stop people makes more sense to me.

Then can sort issues like this (before they happen):

image

p.s. hahaha I see your updated comment.

ayumi-cloud:
Updated commit based on @RomainMazB comment, see screenshot:

image

LukeTowers:
Ummmmmm. It's right there in every single one of your screenshots. At the bottom. It literally says "Report a security vulnerability". Not sure how much more clear it can be made. There's also the security tab. And the security policy document.

ayumi-cloud:
I guess even clearer to stop this:-

image

RomainMazB:
@ayumi-cloud I guess I didn't found the security.md before because I never had to report a security issue.

That being said, I haven't seen either the link @LukeTowers mentioned :no_mouth: (maybe just an icon could help: it's the only one without an icon and placed an the end, it looks like a "other categories link").

I guess the senior dev I'm not will investigate further and knows it's not a good idea to post it publicly.

LukeTowers:
Unfortunately the bottom one is added by GitHub automatically and AFAIK we can't add an icon to it.

bennothommo:
I have to concur with @LukeTowers - we've taken all reasonable and standard steps to make the security policy visible. Most security researchers would know GitHub's setup with security policies by now, and know where to look.

PR #5122, opened 2020-06-09
Allow default checked state for all checkboxes in checkboxlist

RomainMazB:
This PR add the `default` option to the checkboxlist.
The option works the same as in a basic checkbox and change the state of all the checkboxes in the list.

```yaml
selected_primary_mixable_house_models:
    label: Choisissez les modèles de maisons à mixer
    type: checkboxlist
    default: true
```
render something like:
image

Not a huge update but could be useful in some cases...

If the PR is merged, I will update the docs to add the option.

### DIFF
--- a/modules/backend/widgets/form/partials/_field_checkboxlist.htm
+++ b/modules/backend/widgets/form/partials/_field_checkboxlist.htm
@@ -4,7 +4,7 @@
     $isScrollable = count($fieldOptions) > 10;
     $readOnly = $this->previewMode || $field->readOnly || $field->disabled;
     $quickselectEnabled = $field->getConfig('quickselect', $isScrollable);
-    $defaultState = $field->getConfig('default', false)
+    $defaultState = (bool) $field->getConfig('default', false)
 ?>
 <!-- Checkbox List -->
 <?php if ($readOnly && $field->value): ?>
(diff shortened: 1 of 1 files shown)


LukeTowers:
@RomainMazB isn't this already supported by providing an array of keys that are selected by default to the default property?

LukeTowers:
Type casting won't help @RomainMazB, arrays can be cast to boolean.

RomainMazB:
To be honest I didn't try but my case is more vicious because I don't know the keys in a many to many relationship.

LukeTowers:
@RomainMazB you could always modify the default property dynamically with all the keys.

RomainMazB:
Sure for the array but what woukd happen if we use this?
```yaml
default: 'string'
```

Sure I could have used dynamically set default value in the controller event as an example but beside all the available options in others field types and widgets, this one feel missing.

RomainMazB:
Closing this PR since @alxy made me realize it would make a breaking change.
The right way to do is the dynamic way:
```php
Event::listen('backend.form.extendFieldsBefore', function ($formWidget) {
    if ($formWidget->getContext() !== 'create') {
        return;
    }
    
    // Generate an array of id to check all checkboxes
    $formWidget->data->your_field_name = HouseModel::get()->map(fn($model) => (string)$model->id)->toArray();
}
```

LukeTowers:
@RomainMazB feel free to add that to October Tricks :)

Issue #3305, opened 2017-12-18
Richeditor 'Code view' strips 'form' tag even though it's in the 'Allowed tags' list


ranmichael:
##### Expected behavior
Html `<form>` tag and its attributes are saved.

##### Actual behavior
Html `<form>` and `</form>` tags are automatically stripped

##### Reproduce steps
- Edit a page using StaticPages plug-in
- In the richeditor, go to 'code view'
- Try to save a valid form, for example:
```
<form>
  First name:<br>
  <input type="text" name="firstname"><br>
  Last name:<br>
  <input type="text" name="lastname">
</form>

```
- Hit save (or ctrl+s)

Notice that the opening and closing `<form>` tags (and attirbutes, if present) disappear, while the `<input>` tags are successfully saved.

##### October build
Tested both on builds 419 and 430.

alxy:
The allowed tags is for the codeeditor, not the richeditor widget. I would like to question. wether it is a good idea to allow arbitrary HTML tags in the richeditor. Can't you use a snippet for that purpose?

ranmichael:
Thanks @alxy for your attention.

> The allowed tags section is for the codeeditor
Exactly. I am using the 'code view' accessible via the RichEditor.
What am I missing?

> Can't you use a snippet for that purpose?
Actually, I first tried to use a snippet for this, but there is an issue with rendering snippets in Static Pages:
Snippets work in the built-in 'Content' Richeditor or 'simply' added RichEditors.

But, when I add a RichEditor in a Repeater, there is a parsing problem of snippets.
I get:
`<figure data-snippet="basicContactForm">&nbsp;</figure>`
Instead of the actual form.

So adding the actual `<form>` code which isn't the best practice was a workaround (which failed, hence the bug)
To achieve this, I found another workaround - I added a text field in the repeater called 'formName' and I inject the partial with the input of 'formName' from the user...
Also ugly. But works.

Any thoughts?

LukeTowers:
@ranmichael The real bug here is with the snippet support, not the richeditor. The codeeditor is an entirely separate formwidget from the richeditor, it is not the "code view" for the richeditor. The richeditor should not (and will not) be allowing `<form>` tags

ranmichael:
Thanks @LukeTowers for the clarifications.
I figured it might be stripped on purpose - but in that case it is weird that form `<input>` tags are not stripped... no? Is that the bug then?

On a side note,
[...]

PR #5126, opened 2020-06-11
Fix Froala strip empty tags issue

### LINKED ISSUE #2108: New rich editor (Froala) strips out empty <i> tag
I'm using the new rich editor (build 345) and included the "italics" tag "i" at the allowed empty tags section, which is accessible via the settings page, but my setting is ignored. 

The following:
`<div class="item"><i class="right triangle icon"></i> Item 1</div>`
always becomes
`<div class="item">Item 1</div>`

Is this intended (if so can I avoid this issue) or a bug?

RomainMazB:
This PR fix octobercms/support#30 and octobercms/october#3305 .

Three problems were found:
1. No default list were provided to [`htmlAllowedEmptyTags`](https://froala.com/wysiwyg-editor/docs/options/#htmlAllowedEmptyTags)
2. Inversion between the October's shortname and the Froala option name: [allowEmptyTags should be htmlAllowedEmptyTags and vice versa](https://github.com/octobercms/october/blob/develop/modules/backend/formwidgets/richeditor/assets/js/richeditor.js#L137)
3. Even if `script` and `style` tags were added to the `htmlAllowedEmptyTags` option, the `removeTags` [will remove them](https://froala.com/wysiwyg-editor/docs/options/#htmlRemoveTags)

The first point didn't cause any issue but discussed with @LukeTowers and decided to provide a default list for usual empty tags.

The second one was actually the problem for the two opened issue: octobercms/october#3305 and octobercms/support#30:
- octobercms/support#30 happened because the Froala default was active: `['textarea', 'a', 'iframe', 'object', 'video', 'style', 'script', '.fa', '.fr-emoticon', '.fr-inner', 'path', 'line', 'hr']` so the `i` tag was erased
- octobercms/october#3305 happened because of the typo: the option was never overriden even if we provide other default list.

The third point is just to prevent `script` and `style` tag to be erased since we've decided to put in the default list.

Enjoy :)


LukeTowers:
Existing behaviour would be to remove style and script so we need to retain that behaviour.

RomainMazB:
I'm reviewing the default behavior after my daily brain reboot and view Froala put script and style tag into both allowed empty tags and remove tags... I will investigate further to understand what is really the default behavior. My tests yesterday pointed out they were erased before my modifications, but in that case, why Froala put them into allowed blank tag...

I'll be back with the answer.

RomainMazB:
Ok, here the is answer with jsfiddle examples:
- [Default is to remove script tag](https://jsfiddle.net/RomainMazB/naqbjy08/) even if they are presents in `htmlAllowedTags`
- [Just putting base tag](https://jsfiddle.net/RomainMazB/ohmd98zw/) into `htmlRemovedTags` allows `script` and `style` to stay alive.
- [And here the behavior we want](https://jsfiddle.net/RomainMazB/ohjx8mqr/) (I guess): Allow the `i` tags and keep the default Froala behavior for both `htmlAllowedTags` and `htmlRemoveTags`. I've just added `i` to the default Froala list of tags.

Just thumb up to valid my guess and I'll modify my commit. Or comment to let me know what you want me to change.

RomainMazB:
@LukeTowers This one should be the good one.

Users who have modified default settings will need to reset it or add the the modification manually because I modified the EditorSettings instead of include the default directly into the js file.

More cleaner that way.

igor-tv:
I tested the problem associated with removing allowed empty tags. This works on a clean project. But when I replaced the files on an existing project, my toolbar in the editor stopped showing up. This happens after replacing the build-plugins-min.js file.

Снимок экрана 2020-06-15 в 02 23 53

CURRENT BUILD 465

Installed Plugins:
RainLab.Sitemap
RainLab.Translate
RainLab.Pages
Vdlp.Redirect
RainLab.Location
RainLab.Blog
DigitreadMedia.MediaManager
Vdomah.TranslateTabs
Indikator.Backend
PopcornPHP.RedirectToHTTPS
Utopigs.Seo
Edps.Push
OFFLINE.GDPR
Responsiv.Currency
Indikator.DevTools
RainLab.Builder
Igor.Lessons
October.Test
PanaKour.Translate
Bedard.BlogTags
RainLab.User
Flynsarmy.SocialLogin
RainLab.UserPlus
AnandPatel.SeoExtension
Lutpy.Komen
VojtaSvoboda.TwigExtensions
Bedard.Debugbar
Octobro.Deploy
Excodus.TranslateExtended
Martin.Adminer
PMK.Redirect
Mey.Breadcrumbs
OFFLINE.Mall
October.Demo
October.Drivers
Mirjan.Vklogin
Renatio.DynamicPDF
WebVPF.DarkBackend
OFFLINE.Speedy
RLuders.Cors
RainLab.GoogleAnalytics
Vdlp.Phast
OFFLINE.ResponsiveImages
Flynsarmy.IdeHelper
ToughDeveloper.ImageResizer
Lovata.Toolbox
VojtaSvoboda.UserImportExport
RainLab.MailChimp
AnandPatel.WysiwygEditors

LukeTowers:
@igor-tv what errors are in your console? Did you make sure to clear your browser cache?

igor-tv:
Yes, I checked it without a cache. Errors in my console:
```
Uncaught DOMException: Failed to execute 'querySelectorAll' on 'Element': '*:empty:not(area):not(base):not(br):not(col):not(embed):not(hr):not(img):not(input):not(keygen):not(link):not(menuitem):not(meta):not(param):not(source):not(track):not(wbr):not(textarea):not(a):not(iframe):not(object):not(video):not(style):not(script):not(span):not(i):not(i.fa):not(i.fas):not(i.far):not(i.fab):not(address):not(article):not(aside):not(audio):not(blockquote):not(canvas):not(details):not(dd):not(div):not(dl):not(dt):not(fieldset):not(figcaption):not(figure):not(footer):not(form):not(h1):not(h2):not(h3):not(h4):not(h5):not(h6):not(header):not(hgroup):not(hr):not(li):not(main):not(nav):not(noscript):not(ol):not(output):not(p):not(pre):not(section):not(table):not(tbody):not(td):not(tfoot):not(th):not(thead):not(tr):not(ul):not(video):not(.fr-marker)' is not a valid selector.
    at cleanEmptyTags (http://site.test/modules/backend/formwidgets/richeditor/assets/js/build-min.js?v465:482:32)
    at _normalize (http://site.test/modules/backend/formwidgets/richeditor/assets/js/build-min.js?v465:527:54)
    at Object.set (http://site.test/modules/backend/formwidgets/richeditor/assets/js/build-min.js?v465:542:136)
    at Object._init (http://site.test/modules/backend/formwidgets/richeditor/assets/js/build-min.js?v465:680:98)
    at FE.load (http://site.test/modules/backend/formwidgets/richeditor/assets/js/build-min.js?v465:25:292)
    at FE.<anonymous> (http://site.test/modules/backend/formwidgets/richeditor/assets/js/build-min.js?v465:16:11)
    at HTMLTextAreaElement.i (http://site.test/modules/backend/assets/js/vendor/jquery.min.js?v=465:2:87491)
    at HTMLTextAreaElement.dispatch (http://site.test/modules/backend/assets/js/vendor/jquery.min.js?v=465:2:42523)
    at HTMLTextAreaElement.v.handle (http://site.test/modules/backend/assets/js/vendor/jquery.min.js?v=465:2:40524)
    at Object.trigger (http://site.test/modules/backend/assets/js/vendor/jquery.min.js?v=465:2:70633)
```

RomainMazB:
@luketowers using a validator, it seems the part which fails is this one:
```
:not(i):not(i.fa):not(i.fas):not(i.far):not(i.fab)
```
The validator judge this like "overqualified" because it begins by the `i` tag so the classes targeting are useless.

@igor-tv don't you have set some custom rules in the rich editor settings to avoid FA icons to not disappear?

igor-tv:
> @igor-tv don't you have set some custom rules in the rich editor settings to avoid FA icons to not disappear?

Yes, I had a rule for allowed empty tags. 
`i.fa, i.fas, i.far, i.fab,`
When I removed it, everything worked as it should.

RomainMazB:
The `i` tag is now allowed by default and you should provide if needed only a tag, without class selector into this settings.

LukeTowers:
@RomainMazB if class selectors aren't supported, can we have them automatically removed by this code? I don't want an update to break anyone's existing installs.

RomainMazB:
I'll do some tests to fully understand the problem, classes is may be supported but not for the tags which are already globally named:
- `div.myClass` is maybe supported
- `div div.myClass` is not

When confirmed, I'll provide a fix for this eventual misuse:
If a tag with a class is inserted, ensure that it's not already called globally.

LukeTowers:
@RomainMazB I feel like the third option of keeping both but separating them is probably the least damaging option.

RomainMazB:
`i.fa` will be splitted to `i, .fa` and only `.fa` will be added since `i` is already in the allowed list
It supports multiple classes, such as `i.fa.fas` will be splitted to `i, .fa, .fas`
And even without a tag: `.fa.fas.fa-2x` will be splitted to `.fa, .fas, .fas-2x`

Tested on 466

modules/backend/formwidgets/richeditor/assets/js/richeditor.js
```diff
@@ -95,46 +95,57 @@
...
         froalaOptions.toolbarButtonsMD = froalaOptions.toolbarButtons
         froalaOptions.toolbarButtonsSM = froalaOptions.toolbarButtons
         froalaOptions.toolbarButtonsXS = froalaOptions.toolbarButtons
 
-        if (this.options.htmlAllowedEmptyTags) {
-            froalaOptions.allowEmptyTags = this.options.htmlAllowedEmptyTags.split(/[\s,]+/)
+        if (this.options.allowEmptyTags) {
+            // Filter the tags list to split eventual tag.class and add them if necessary
+            let filteredTagList = [];
+            this.options.allowEmptyTags.split(/[\s,]+/).forEach(tagString => {
```
LukeTowers, line 139: is .forEach supported in IE11?

modules/backend/formwidgets/richeditor/assets/js/richeditor.js
```diff
@@ -95,46 +95,57 @@
...
+                let i = 0;
+                tagString.split('.').forEach(tag => {
+                    if (tag !== null && ! filteredTagList.includes(tag)) {
```
LukeTowers, line 142:
```suggestion
                    if (tag !== null && !filteredTagList.includes(tag)) {
```

modules/backend/formwidgets/richeditor/assets/js/richeditor.js
bennothommo: @LukeTowers according to MDN, it is: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach#Browser_compatibility

bennothommo: But `let` isn't properly supported.

bennothommo:
@RomainMazB I've refactored your code a bit to make the following changes:

- I've changed the allowEmptyTags section in the JS to use ES5 code - your PR was using ES6 code that is not supported by IE11.
- I've implemented the defaults into the JS, as the defaults specified in the `EditorSetting` model were the defaults used by Froala - as this PR intends to change the Froala defaults, we'll need to override the Froala options regardless.

bennothommo:
I've tested this and confirmed it working - @RomainMazB, if you're happy with my changes, I'm happy to merge this.

RomainMazB:
@bennothommo You can merge ;)

PR #5129, opened 2020-06-16
Updated Polish translation

tomaszstrojny:
It turned out there are pretty much Polish strings that haven't been translated yet. There they are.

I used `RainLab.Builder` to achieve this and it removed all unneeded whitespaces and changed `"` to `'` but despite the change, everything seems fine.


LukeTowers:
@tomaszstrojny please revert the formatting changes you made, it makes the git history a lot more complicated to understand what's going on and this PR specifically is a lot hard to actually review.

tomaszstrojny:
Yea, I guess it's harder to understand but this is the formatting used in the `en` translation and I think it's the thing that has to be done sooner or later. This PR introduces about 60 new strings which finding without something like `RainLab.Translate` localization feature is a real pain in the ass.

Take a look at those changes one more time, please. It's harder but still pretty easy to review.

LukeTowers:
@bennothommo if you're fine with merging these changes into 467 go for it

bennothommo:
@LukeTowers I'd say this falls within the realm of bug fixes, so I'll merge :)

bennothommo:
Regarding the spacing changes, I know we've likely discussed it before, but I'd personally vote for them to be up close to the arrows in the standard PSR-2 fashion, as opposed to formatted like a column - it's less arbitrary, and doesn't require the entire column to be shifted if we decide to use a larger term.

LukeTowers:
@bennothommo from a stylistic and readability point of view I prefer the spaces aligned for visual clarity, but given how frequently the language files change and how many entries there are, and how often new keys could be added I think for this instance it does indeed make sense to have the values up close to the keys for better maintainability. I'm not going to do that for everything in the code base though, just these language files 😄

Issue #5084, opened 2020-05-21
Relationfinder on pivot form - A widget with class name has not been bound to the controller


empower-josh:
- OctoberCMS Build: 465
- PHP Version:  7.2.24
- Database Engine: MySQL

### Description:
When using recordfinder on a third model from within a pivot form on two many-to-many models, the initial list will populate, and if you select a record, it will work as expected. However if you try to search or move between pages of results it fails:

```
"A widget with class name 'relationAssigneesManagePivotFormPivotLocationIdSearch' has not been bound to the controller" on line 605 of /home/joshm/Sites/sandbox/public/modules/backend/classes/Controller.php
```

From @LukeTowers via Slack:
>The fix has to be in JS, it needs to send the same information that any other AJAX request triggered by a relationcontroller widget sends in order to have the relationcontroller properly initialize the recordpicker widget in order to respond to the ajax request.

>The pivot update form will only be initialized when it is needed, so the code needs to know that you need it, and that's done by sending the _relation property in the AJAX request

### Steps To Reproduce:
Configure recordpicker on a pivot form like:

Pivot form was configured like:
```
pivot[weight]:
        label: Weight
        type: number
        default: 1
    pivot[notifications]:
        label: Notification Addresses
        type: text
    pivot[active]:
        label: Active
        type: switch
        default: 1
    pivot[location_id]:
        label: Location
        type: recordfinder
        list: ~/plugins/empowertools/empowercontact/models/location/columns.yaml
        keyFrom: id
        nameFrom: name
        useRelation: false
        searchScope: likeName
        modelClass: EmpowerTools\EmpowerContact\Models\Location
```

Note, this failed regardless of searchScope defined (That was my attempt to overcome.)

LukeTowers:
@bennothommo this should be really similar to the work you recently did on (filter scopes? dependsOn? either way, same root problem of the relation data not being sent with the popup's requests given it's location in the DOM). Could you take a look at fixing it?

LukeTowers:
@mjauvin this would be related to our discussion in the Discord channel regarding the pivot form not sending the parent form data with every AJAX request.

bennothommo:
I have tried a few different ways to replicate your problem, but have been unable to do so, even when emulating a setup that I interpreted was close to your example.

[...]

Issue #4925, opened 2020-02-02
Relation controller update form - Parent model data isn't in Request


danielbidala:
I have two models: Order and Orderitems. I add Orderitems to Order with relation manager. When I add new orderitem I get all field's value of parent Order in the Request. 

For example, in my Orderitem model's FilterFields method I can use:
`Request::input('Order.order_type')`

But if I update an orderitem with the relation manager, the parent model (Order) data is not in the Request.

Create Form Request:
![create_request](https://user-images.githubusercontent.com/16636699/73615599-9237c580-4609-11ea-9c4e-d5e513505e1d.png)

Update Form Request:
![update_request](https://user-images.githubusercontent.com/16636699/73615602-a54a9580-4609-11ea-8e9d-011b16a64bd0.png)

What is the reason for this? Can we send the parent data with the request always?

(my October CMS version = 464)

LukeTowers:
Github issues really isn't the place for questions like this. Please use one of the many other avenues available instead of submitting new issues for your questions.

Forum: https://octobercms.com/forum
StackOverflow: https://stackoverflow.com/questions/tagged/octobercms

To answer your question however, the parent record information is required to create the related record, but only the record's own identifier is required after it has been created. What are you trying to do with that information? I can guarantee you there's a better way to get access to it then pulling it directly from the request.

danielbidala:
@LukeTowers, yes I know this isn't the best place for questions but on the forum there is not too much activity. (There is tons of unanswered questions)

My use case: I need the parent record's data in the filterfield method of the child form **also when the parent record isn't saved yet**. I have to manipulate the child form based on the parent's data. Thanks for all the help in advance.

vortechron:
@danielbidala if i remember correctly, i was able to access relation data by using _session_key.

something like this:
` $this->formWidget->model->some_relationship()->withDeferred(post('_session_key'))->get()`

danielbidala:
@vortechron thanks for help but i think you can't query the unsaved parent model.

@LukeTowers what if in `modules\backend\behaviors\relationcontroller\assets\js\october.relation.js`
we add somthing like:

```
        function serializeParentForm(form) {
            var parentFormData = [];
            $.each($(form).serializeArray(), function() {
                if(this.name != '_session_key' && this.name != '_token') {
                        parentFormData[this.name] = this.value;
                }
            });
            return parentFormData;
        }
```
and than extend `clickViewListRecord` function in same file:

```
[...]

Issue #2727, opened 2017-03-01
using file upload widget in a popup


salehasadi:
##### Expected behavior
( upload and attach files to model as we use fileupload in create or update pages)

##### Actual behavior
(I render a form from yaml config file in a popup window every thing works correctly but as i choose a file to upload it upload the file but when i click on delete or edit file the system will throw " a widget with class name 'dashboardFiles' has not been bound to the controller" and form does not send the file to save method)

this is my stack trace data:
 

> exception 'October\Rain\Exception\SystemException' with message 'A widget with class name 'dashboardFiles' has not been bound to the controller' in /opt/lampp/htdocs/imis-lawyer/modules/backend/classes/Controller.php:517
Stack trace:
#0 /opt/lampp/htdocs/imis-lawyer/modules/backend/classes/Controller.php(423): Backend\Classes\Controller->runAjaxHandler('dashboardFiles:...')
#1 /opt/lampp/htdocs/imis-lawyer/modules/backend/classes/Controller.php(233): Backend\Classes\Controller->execAjaxHandlers()
#2 /opt/lampp/htdocs/imis-lawyer/modules/backend/classes/BackendController.php(106): Backend\Classes\Controller->run('index', Array)
#3 [internal function]: Backend\Classes\BackendController->run('imis/processman...')
#4 /opt/lampp/htdocs/imis-lawyer/vendor/laravel/framework/src/Illuminate/Routing/Controller.php(256): call_user_func_array(Array, Array)
#5 /opt/lampp/htdocs/imis-lawyer/vendor/laravel/framework/src/Illuminate/Routing/ControllerDispatcher.php(164): Illuminate\Routing\Controller->callAction('run', Array)
#6 /opt/lampp/htdocs/imis-lawyer/vendor/laravel/framework/src/Illuminate/Routing/ControllerDispatcher.php(112): Illuminate\Routing\ControllerDispatcher->call(Object(Backend\Classes\BackendController), Object(Illuminate\Routing\Route), 'run')
#7 [internal function]: Illuminate\Routing\ControllerDispatcher->Illuminate\Routing\{closure}(Object(Illuminate\Http\Request))
#8 /opt/lampp/htdocs/imis-lawyer/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(139): call_user_func(Object(Closure), Object(Illuminate\Http\Request))
[...]

PR #5136, opened 2020-06-24
Implement support for sending parent form data when making AJAX requests in a popup

LukeTowers:
This should resolve a lot of issues that occur because popups are not nested under form elements and thus don't include the data that was included in the request that initialized / displayed the popup in the first place.

October.Test PR: https://github.com/octoberrain/test-plugin/pull/87

Related issues:
- #5084 
- #4925 
- #2727 (Somewhat related, not entirely though)
- https://github.com/octoberrain/test-plugin/issues/73 (Somewhat related)
- #5447 (may be related)


modules/system/assets/js/framework.js
```diff
@@ -106,9 +108,50 @@ if (window.jQuery.request !== undefined) {
...
+                            parentForms.push($parentForm)
+                            findParentForms($parentForm)
+                        }
+                    }
+                }
+            }
+        findParentForms($form)
+        parentForms.reverse().forEach(function ($formEl) {
+            $.extend(parentFormData, serializeFormToObj($formEl))
+        })
```
bennothommo, line 144: @LukeTowers maybe instead of `$formEl`, it could be `$childForm` just to make it clear what we're dealing with in this loop.
```suggestion
        parentForms.reverse().forEach(function ($childForm) {
            $.extend(parentFormData, serializeFormToObj($childForm))
        })
```

bennothommo:
@LukeTowers looks good so far, just suggested a couple of minor tweaks.

modules/system/assets/js/framework.js
LukeTowers: I feel like $parentForm is more accurate, I'll change up the variables a bit

PR #5139, opened 2020-06-25
Allow to override a shared partial from the theme using only one file

RomainMazB:
For now the only way to override a shared partial is to create a directory named with the component alias and create a file in it.

This is fine but an annoying thing happen when you use multiple components which use the same shared partial (it's definitely the goal of shared partial!). And it's worst if you use the same components multiple times in the same page using aliases!

This modification allow to directly override the shared partial using the path relative to the plugin in your theme folder with only one file:

`/themes/plugins/author-name/components/partials/file.htm`
will override:
`/plugins/author-name/components/partials/file.htm`

This is useful when using a shared partial multiple times, with many components alias.

As an example, my theme folder and how I figured out this functionnality was missing:
image
and all the aliases are not presents yet in the theme folder, I still needed to add it more than 5 times... Annoying and useless ;/

Internally, each `tile.htm` file only contains a call for a unique partial:
`{% partial 'widgets/tile.htm' %}` to modify only one file if I need to modify the view but the theme folder is still unnecessary huge.

This functionality allows to do this:
image

You still can override individually a component alias, the priority is:
- Components alias folder in theme directory
- Shared folder in theme directory (Using the path I used in example)
- Component folder in plugin directory
- Shared folder in plugin directory

The path can be more friendly if needed.

It could be something like:
`/themes/components/author-name/plugin-name/file-name.htm`

To go further, this functionality can be extended to override a component (non-shared) partial regardless to its alias:

```
[myComponent componentA]
[myComponent componentB]
[myComponent componentC]
```
We should be able to use the same partial located in `/themes/my-theme/partial/mycomponent/default.htm` without the need to create three directory (`componenta`, `componentb` and `componentc`) containing three exact same file including a fourth partial file to have only one html markup view file.

The creation of the file `/themes/my-theme/partial/componenta/default.htm` should still take the priority over the previous file path (and over the eventual shared partial if this PR is merged) if the developer wants to.

### DIFF
--- a/modules/cms/classes/ComponentPartial.php
+++ b/modules/cms/classes/ComponentPartial.php
@@ -104,23 +104,31 @@ class ComponentPartial extends Extendable implements CmsObjectContract
         $partial = Partial::loadCached($theme, strtolower($component->alias) . '/' . $fileName);
 
         if ($partial === null) {
-            $pluginPath = 'plugins' . str_replace(
+            $partial = Partial::loadCached($theme, $component->alias . '/' . $fileName);
+        }
+
+        if ($partial === null) {
+            $partial = Partial::loadCached($theme, strtolower($component->name) . '/' . $fileName);
+        }
+
+        if ($partial === null) {
+            $partial = Partial::loadCached($theme, $component->name . '/' . $fileName);
+        }
+
+        if ($partial === null) {
+            $pluginPath = 'components-shared' . str_replace(
                 plugins_path(),
                 '',
                 $component->getPath()
             ) . '/' . $fileName;
 
             $sharedPath = str_replace(
-                [strtolower($component->alias), strtolower($component->name)],
-                'partials',
+                ['/components/' . strtolower($component->alias), '/components/' . strtolower($component->name) . '/'],
+                '',
                 $pluginPath
             );
 
             $partial = Partial::loadCached($theme, $sharedPath);
-        }
-
-        if ($partial === null) {
-            $partial = Partial::loadCached($theme, $component->alias . '/' . $fileName);
         }
 
         return $partial;
--- a/modules/cms/classes/ComponentPartial.php
+++ b/modules/cms/classes/ComponentPartial.php
@@ -104,6 +104,22 @@ class ComponentPartial extends Extendable implements CmsObjectContract
         $partial = Partial::loadCached($theme, strtolower($component->alias) . '/' . $fileName);
 
         if ($partial === null) {
+            $pluginPath = 'plugins' . str_replace(
+                plugins_path(),
+                '',
+                $component->getPath()
+            ) . '/' . $fileName;
+
+            $sharedPath = str_replace(
+                [strtolower($component->alias), strtolower($component->name)],
+                'partials',
+                $pluginPath
+            );
+
+            $partial = Partial::loadCached($theme, $sharedPath);
+        }
+
+        if ($partial === null) {
             $partial = Partial::loadCached($theme, $component->alias . '/' . $fileName);
         }
 
--- a/modules/cms/classes/ComponentPartial.php
+++ b/modules/cms/classes/ComponentPartial.php
@@ -6,6 +6,7 @@ use Cms\Contracts\CmsObject as CmsObjectContract;
 use Cms\Helpers\File as FileHelper;
 use October\Rain\Extension\Extendable;
 use ApplicationException;
+use October\Rain\Support\Facades\Str;
 
 /**
  * The CMS component partial class. These objects are read-only.
@@ -108,15 +109,29 @@ class ComponentPartial extends Extendable implements CmsObjectContract
         }
 
         if ($partial === null) {
-            $partial = Partial::loadCached($theme, strtolower($component->name) . '/' . $fileName);
+            $classPath = ComponentManager::instance()->resolve($component->name);
+            $filePath = str_replace(['Components\\', '\\'], ['', '/'], substr($classPath, 1));
+
+            if ($component->name !== $component->alias) {
+                $pathParts = explode('/', $filePath);
+                $filePath = str_replace(array_pop($pathParts), $component->alias, $filePath);
+            }
+            $filePath = strtolower($filePath);
+            $partial = Partial::loadCached($theme, $filePath . '/' . $fileName);
+        }
+
+        if ($partial === null) {
+            $pathParts = explode('/', $filePath);
+            $filePath = str_replace(array_pop($pathParts), '', $filePath);
+            $partial = Partial::loadCached($theme, $filePath . $fileName);
         }
 
         if ($partial === null) {
             $pluginPath = 'components-shared' . str_replace(
-                plugins_path(),
-                '',
-                $component->getPath()
-            ) . '/' . $fileName;
+                    plugins_path(),
+                    '',
+                    $component->getPath()
+                ) . '/' . $fileName;
 
             $sharedPath = str_replace(
                 ['/components/' . strtolower($component->alias), '/components/' . strtolower($component->name) . '/'],
--- a/modules/cms/classes/ComponentPartial.php
+++ b/modules/cms/classes/ComponentPartial.php
@@ -128,10 +128,10 @@ class ComponentPartial extends Extendable implements CmsObjectContract
 
         if ($partial === null) {
             $pluginPath = 'components-shared' . str_replace(
-                    plugins_path(),
-                    '',
-                    $component->getPath()
-                ) . '/' . $fileName;
+                plugins_path(),
+                '',
+                $component->getPath()
+            ) . '/' . $fileName;
 
             $sharedPath = str_replace(
                 ['/components/' . strtolower($component->alias), '/components/' . strtolower($component->name) . '/'],
--- a/modules/cms/classes/ComponentPartial.php
+++ b/modules/cms/classes/ComponentPartial.php
@@ -112,10 +112,6 @@ class ComponentPartial extends Extendable implements CmsObjectContract
         }
 
         if ($partial === null) {
-            $partial = Partial::loadCached($theme, $component->name . '/' . $fileName);
-        }
-
-        if ($partial === null) {
             $pluginPath = 'components-shared' . str_replace(
                 plugins_path(),
                 '',


LukeTowers:
@RomainMazB assuming all those aliased components are powered by the same component behind the scenes, why can't you put the override in `partials/$defaultAliasOfCommonComponent/tile.htm`?

RomainMazB:
@LukeTowers it's not every time the same components ( I have 4 component using a same time shared partial), and each components are used multiple times on the same page. An example would be to display with the `landsGrid` on the same page to display:
- the lands near to the one currently displayed (+/- 10 km)
- the lands on the country
- and 8 randomly picked lands in all the available lands

And even if it was all the same components (in this case I shouldn't use a shared partial), the actual behavior is only related to the component's alias used by the developer

https://github.com/RomainMazB/october/blob/64a92cc27c278d0239736401830085b11572bbab/modules/cms/classes/ComponentPartial.php#L104

modules/cms/classes/ComponentPartial.php
```diff
@@ -103,6 +103,22 @@ public static function loadOverrideCached($theme, $component, $fileName)
...
+            ) . '/' . $fileName;
+
+            $sharedPath = str_replace(
+                [strtolower($component->alias), strtolower($component->name)],
+                'partials',
+                $pluginPath
+            );
+
+            $partial = Partial::loadCached($theme, $sharedPath);
+        }
```
LukeTowers, line 143: This should be below the next statement

LukeTowers:

RomainMazB:
Review fixed.
I've modified the path to be more friendly:
`themes/my-theme/partials/components-shared/author-name/plugin-name/file-name.htm`

I've added the ability to override a partial globally for a component, relatively to its name with this path:
`themes/my-theme/partials/component-name/file-name.htm`

### An example to fully understand what is the priority.
Assuming I'm working with two different components: `componentA` and `componentB` sharing a same `partial-file.htm` and a page using this components and aliases:
```
[componentA]
[componentA aliasComponentA]
[componentB]
```
If I create this structure of file in my theme's partials directory:
```
componentA
|___partial-file.htm
aliasComponentA
|___partial-file.htm
components-shared
|___romainmazb
    |____myplugin
         |____partial-file.htm
```
- `aliasComponentA` will use the `aliasComponentA/partial-file.htm` => *Default behavior, still have the priority*
- `componentA` will use the `componentA/partial-file.htm` => *A component named directory take precedence over the components-shared directory*
- `componentB` will use the `components-shared/romainmazb/myplugin/partial-file.htm` => *A file in components-shared directory takes precedence over the default plugin partial*
- And if I haven't created the `components-shared/romainmazb/myplugin/partial-file.htm` file, `componentB` would have use the default plugin file located in `/plugins/romainmazb/myplugin/components/partials/partial-file.htm`

modules/cms/classes/ComponentPartial.php
```diff
@@ -107,6 +107,30 @@ public static function loadOverrideCached($theme, $component, $fileName)
             $partial = Partial::loadCached($theme, $component->alias . '/' . $fileName);
         }
 
+        if ($partial === null) {
+            $partial = Partial::loadCached($theme, strtolower($component->name) . '/' . $fileName);
+        }
+
+        if ($partial === null) {
+            $partial = Partial::loadCached($theme, $component->name . '/' . $fileName);
+        }
```
LukeTowers, line 127: You don't need this section because it wasn't supported previously so there's no reason to support case sensitive names

LukeTowers:
It may be worth investigating how the backend partials handle this (it uses the ViewMaker trait which has the concept of adding viewPaths to investigate for a given file dynamically), and exposing some of that flexibility to plugins to define their end points. After reviewing this further we actually can't globally start allowing component names as a fallback view path behind their aliases because that could expose existing sites to having a partial that was meant for a specific instance of a component suddenly start being used for other more specific instances of that component as well. Exposing it as a point of extension allows us to shift that burden of verifying that it doesn't break existing sites to plugin developers instead.

RomainMazB:
> It may be worth investigating how the backend partials handle this
I will investigate how the backend partials are handled.

> we actually can't globally start allowing component names as a fallback view path behind their aliases because that could expose existing sites to having a partial that was meant for a specific instance of a component suddenly start being used for other more specific instances of that component as well

As you said in your review: this wasn't supported before this changes so the only existing sites which can be exposed are the ones which made a mistake trying to override partial and tried to use the component name to override. Actually this few websites (if there is one) aren't overriding any partial and yes doing this modification will make the things they tried suddenly work.

The priority before this changes was basically:
- If a directory named with the component alias exists, use it
- Else use the default plugin's partial

I've just add some variants between the two possibilities, without modifying the priority order and without taking any precedence over the component alias directory.

LukeTowers:
> As you said in your review: this wasn't supported before this changes so the only existing sites which can be exposed are the ones which made a mistake trying to override partial and tried to use the component name to override. Actually this few websites (if there is one) aren't overriding any partial and yes doing this modification will make the things they tried suddenly work.

This is incorrect. Imagine you have component `easyForms`. You then have a page that uses it with no alias: `[easyForms]`. You also have another page that uses it with an alias `[easyForms contactForm]`. You then need to modify the partials used by the first instance, so you put them in `partials/easyForms`. With this change, you would now have `contactForm` being overridden by those same partials.

RomainMazB:
> Imagine you have component `easyForms`. You then have a page that uses it with no alias: `[easyForms]`. You also have another page that uses it with an alias `[easyForms contactForm]`. You then need to modify the partials used by the first instance, so you put them in `partials/easyForms`. With this change, you would now have `contactForm` being overridden by those same partials.

This is correct ;)

It's too bad to lose this ability though... If we need to use on the same page `[easyForms]` and `[easyForms contactForm]` with the same customized partial: we need to create 2 directories with 2 exact same files. (Or IMO a best structure use 3 files: 2 which call the same one to only modify one file).

Two possibilities:
- Rollback this changes (but keep the `component-shared` possibility)
- Using a specific directory structure to not break anything, such as `/themes/my-themes/partials/components/author-name/plugin-name/component-name/file-name.htm` . Not beautiful but still useful.

bennothommo:
@RomainMazB @LukeTowers I think this might be a duplicate of https://github.com/octobercms/october/pull/4677.

RomainMazB:
@bennothomo
Yes it is for the default partial using the component's name.

The shared partial override (which was the beginning of all this conversation) is not a duplicate.

RomainMazB:
@LukeTowers
need your thoughts about that to end up with this PR.

LukeTowers:
I'll review again after we release the L6 upgrade

RomainMazB:
I'm fine with it, L6 is the priority :+1:

RomainMazB:
@luketowers do you have more time now so we can close or merge this one? 

To quickly remind you. This one add the possibility to load a component shared partial presents in the theme folder from a directory named by the component's name.

The benefit appears when you use aliases: if you want to override the partial with the same markup, you don't need to create 3 exact same files to override the 3 aliased component.

You just need to create a component named folder and override it with only one file for all the aliased components.

If a folder named by the component's alias exists, it still take the priority.

No breaking change here, just a more convenient way to handle component's partials override.

LukeTowers:
@RomainMazB there's a lot of stuff to re-read through here to assess this, could you make a PR to the docs repo documenting this change and then I can use that to review it?

modules/cms/classes/ComponentPartial.php
```diff
@@ -107,6 +107,26 @@ public static function loadOverrideCached($theme, $component, $fileName)
             $partial = Partial::loadCached($theme, $component->alias . '/' . $fileName);
         }
 
+        if ($partial === null) {
+            $partial = Partial::loadCached($theme, strtolower($component->name) . '/' . $fileName);
```
Rike-cz: You are using `$component->name` directly, but isn't this attribute something like `Acme\Plugin\Components\Component`? I tried to deal with the same fallback [this way](https://github.com/octobercms/october/pull/4677/files#diff-be1899dc8e821d54db6ebb064918f0540328a1f3d9b02f0473259f873b23a938R111).

RomainMazB: I've just tested, `$component->name` just return the component name as expected:
```
[demoTodo demoTodoAlias]
```
returns `demoTodoAlias`

In most cases it should work but yeah, you are right: [according to the doc](https://octobercms.com/docs/cms/components#aliases), to avoid an eventual conflict, it could be a full class name including the namespace :+1:  :
```
[AuthorName\PluginName\Components\DemoTodo demoTodoAlias]
```
returns `AuthorName\PluginName\Components\DemoTodo`.

In those cases, your way seems to be the way, but I also see one case where your way will break: using the full path of a component means we are dealing with two plugins components which have the same name.

Knowing that it means that your way would try to load the partials of the two components from the same folder:
```
[AuthorOneName\MyAwesomePlugin\Components\DemoTodo awesomeDemoTodo]
[AuthorTwoName\PluginTwoName\Components\DemoTodo]
```
both of these statement will try to look into the `demotodo` folder, am I right?

What's your thoughts about the best of our two worlds?
```
strtolower(str_replace(['Components\\', '\\'], ['', '/'], $component->name))
```
would returns:
- `authorname/pluginname/demotodo` when used with `[AuthorName\PluginName\Components\DemoTodo demoTodoAlias]`
- `demotodo` when used with `[DemoTodo demoTodoAlias]`

So we can - *using the previous example where there was two DemoTodo components from two differents plugins* - load partials from:
- authoronename/myawesomeplugin/demotodo
- authortwoname/plugintwoname/demotodo

Do you agree with all my points @Rike-cz ?

Rike-cz: @RomainMazB I am still not sure if `$component->name` should be anything else than `AuthorName\PluginName\Components\DemoTodo`. How did you achieve that? If I want to find component's short name, which component got when registered (`PluginName::registerComponents()`), I have to call ComponentManager to ask him. And this step still missing here, so it looks like all my components will fall through theme partials overriding process (by component's registered short code) to their default templates (or to `themes/themename/partials/authoronename/pluginaname/demotodo` in your PR). Is good idea to ignore component's registered short code and send them automatically to `authoronename/pluginaname/demotodo` subfolder?

My previous PR was handy for me and reusing theme's partial design for many snippets of the same component, but it might not be handy for everyone :-)

RomainMazB: I've committed some new change, joined by [a new repository](https://github.com/RomainMazB/partials-override) so everyone can test easily the modifications.

I'll be happy to continue the discussion on the new commit ;)

RomainMazB:
With the last commit, the files are loaded with this priority (unchanged from the start of the PR, just a reminder):
- From an alias directory (normal behaviour)
- From a component directory (normal behaviour)
- From a full path "author/plugin/alias(or name)" (new behaviour)
- From a full "author/plugin" (new behaviour)
- From the plugin itself

To test it, just create a new october project, download and unzip the archive from [this repository](https://github.com/RomainMazB/partials-override) and switch from the two page of the theme.

For those who don't have the time, here are the pages, and you can see the files structure on the repository.
As you can see:
[On the home page](https://github.com/RomainMazB/partials-override/blob/main/themes/demo/pages/home.htm), even if these files are present, they don't take precedence over the actually known behaviour:
- [/themes/demo/partials/author/partialsplugin/mycomponent/shared.htm](https://github.com/RomainMazB/partials-override/blob/main/themes/demo/partials/author/partialsplugin/mycomponent/shared.htm)
- [/themes/demo/partials/author/partialsplugin/shared.htm](https://github.com/RomainMazB/partials-override/blob/main/themes/demo/partials/author/partialsplugin/shared.htm)
  This is due to the fact that the basic behaviour is kept and the files are loaded from [myComponent, myComponentAlias and myComponentFullPathAlias](https://github.com/RomainMazB/partials-override/tree/main/themes/demo/partials)
  image

[On the new behaviours page](https://github.com/RomainMazB/partials-override/blob/main/themes/demo/pages/new.htm), we can now load the **shared partials** from these folders:
- [/themes/demo/partials/author/partialsplugin/](https://github.com/RomainMazB/partials-override/tree/main/themes/demo/partials/author/partialsplugin)
- [/themes/demo/partials/author/partialsplugin/mysecondcomponentalias/](https://github.com/RomainMazB/partials-override/tree/main/themes/demo/partials/author/partialsplugin/mysecondcomponentalias)

This new behaviours as showed below allow us to only create one shared partial file into the theme to override all the component and alias who use this shared file from the plugin. And this default behaviour can be overridden by a component alias or name specific folder.

Without this PR changes, you would have to create 5 or 6 exact same files in each alias-named directory, to override the shared partial.

image

daftspunk:
We want to discourage people from overriding partials this way and in general. Component markup was only ever intended to be a use case example that you copy into your theme completely. Overriding is something to be done rarely

Issue #3202, opened 2017-10-26
Filter type "group" does not work when 500+ options are available


ChVuagniaux:
The group filter dosen't work when the Model table defined in `modelClass` have to many row. 

For big application, like a shop by example, it's important to can filter fields with a model that have many records. Example : filter all commands by client from a list of 5000 clients.

##### Actual behavior
1. When open the filter the system do a query for load all records (`select * from table;`) and display them in the list. This behavior is very nice when you have only few items but query a big tables is a big time consumer
2. When the filter is valided, all records are send in the POST request to the server : when you have several thousand of records you have many chances to exceed the PHP max_input_vars limit.

##### Expected behavior
1. Two scope option for can choose **a limit** and an other for can completly **disable** the initial dropdown (before starting to search) can be very useful and lets the AJAX autocomplete do her job (still with a limit).
2. Only send active records to the server on the POST request don't think that is useful to have available fields provided here.

##### Reproduce steps
1. Use this PR https://github.com/daftspunk/oc-test-plugin/pull/44 of the octobrcms test plugin 
2. Run `php artisan test:seed-posts` for seed the table
3. Go to http://octobercms/backend/october/test/posts comment tab & try to filter by post

##### October build
426

LukeTowers:
@ChVuagniaux look at the `recordsPerPage` config option for the list behavior: http://octobercms.com/docs/backend/lists#configuring-list

ChVuagniaux:
Hi @LukeTowers `recordsPerPage` affect the number of item in the list, any impact on the dropdown.

LukeTowers:
Ah, I see. Do you have any suggestions for how this could be resolved?

ChVuagniaux:
@LukeTowers

For part 1 - I suggest to add two params for the list filters `recordsToShow` & `hideIntialDropDown` (names are not very nice but you see what I mean).
```yaml
scopes:
    post:
        label: Post
        modelClass: October\Test\Models\Post
        conditions: id in (:filtered)
        nameFrom: name
        recordsToShow: 10 # Only show 10 items in the list (initial and ajax filtered) -> add limit to select
        
        hideIntialDropDown: true # No show dropdown before starting to type 
              # avoid to wait that the query was done before open the dropdown
              # and no confusions abount only 10 items listed despite the fact that there are 2k in DB
```

For part 2 - Edit JS of dropdown for no longer send availables items on filter request

LukeTowers:
[...]

PR #5141, opened 2020-06-27
Fix: Filter type "group" does not work when 500+ options are available

Klaasie:
In relation to: https://github.com/octobercms/october/issues/3202.

This PR addresses the issue where the filter type group stops working due to surpassing the `max_input_vars` amount by preventing the available option types from being posted when filtering.
This is achieved by splitting the available and active options into 2 seperate variables in the javascript.

I've tested this code with all possible filter types to ensure they keep working.
I have set up a testing area in the october-test plugin here: https://github.com/Klaasie/test-plugin/tree/fix/filter
The users list view page has all the possible filters available for testing. A seeder has been made available that will seed the roles table with enough records to test the fix.

_P.S. I wasn't sure about including the minified version the changes, I vaguely remember I need to ignore that as it will be minified on release anyway. If you need it I can commit it._


bennothommo:
@Klaasie does this theoretically only raise the limit from 500 to 1000, ie. if you have 500 available options, and 501 active options, 1 of those active options will be dropped? You could potentially send the data as a JSON object - we did a similar fix for the Static Menu data in the Static Pages plugin, which was also affected by `max_input_vars` (https://github.com/rainlab/pages-plugin/pull/439/files)

Klaasie:
@bennothommo That's a clever solution to the problem. I've modified the code to post the values as a JSON string and handle the data properly at the php side. This is on top of the modifications that avoid the available options from being posted. I personally don't think they should be in the post data, none of the other filter options do.

LukeTowers:
@Klaasie Can you recompile the js too?

Klaasie:
@LukeTowers Done!

bennothommo:
@github-actions run

<details>

```js
(async () => {
  // Get issue/PR number

  // Add Testing Needed label
    owner: context.repo.owner,
    repo: context.repo.repo,
    issue_number: number,
    labels: ['Testing Needed'],
  });

  // Comment body
    + '\n\n'
    + '\n\n'
    + '```\n'
    + '```';

  // Add comment
})();
```

bennothommo:
@Klaasie I've given this a test and it's working well. There's only one bug I encountered - when I use the "Clear" button with the group filter, it no longer allows me to apply that filter again. When I click the filter once more afterwards, and try and make changes to the filter, the "Apply" and "Clear" buttons no longer show up.

I've double checked the behaviour without your changes and those buttons appear, so it looks like there's something in your changes that prevents the buttons from appearing if you press the "Clear" button.

I'll message you the access details to a test instance on Slack.

Klaasie:
@bennothommo 

The reason this broke is because available and selected options are now split in 2 different variables. Previously the one variable with both available and selected options would resolve to null so when the popover was opened it'd retrieve the data (through ajax) again and set all variables correctly. Now the available options would still be there while selected would resolve to null causing `if (!this.scopeValues[this.activeScopeName])` (line 285) to exit the function early.

To fix this issue I've modified the javascript to behave it would previously. Meaning hitting the "clear" button will clear both the selected and available options. When opening the popover data will be retrieved again (through ajax) and the popover will function properly again.

bennothommo:
Great work @Klaasie. That fixed it.

Tested, and confirmed working.

danielbidala:
@bennothommo This fix doesn't seem to work with build 468 and even with 469. I have about 1800 filter item (related record) for group filter. When I open group filter, I see the first 500 record. When I start type to search filter item (which is not in the first 500) I get the search results but I can't apply the filter (the apply button doesn't appear).

LukeTowers:
@danielbidala did you try selecting the option and then clicking out of the filter box?

danielbidala:
@LukeTowers Yes I've tried, the selected filter takes no effect:

Megrendelések _ Bright CRM

bennothommo:
@danielbidala Make sure you don't have a cached version of the backend JS files. Try running `php artisan october:util compile js`, then flush your browser cache, and try again.

msimkunas:
@danielbidala I'm experiencing the exact same issue on build 469. I have more than 500 items in the filter and initial testing suggests that 500 is the cut-off point after which the remaining items are not processed correctly. Selecting those items doesn't make the filter buttons appear and the active filter count is not incremented.

msimkunas:
I believe that the issue must be related to the following limit of 500 records: https://github.com/octobercms/october/blob/517c588ef7224941af761ba4df4f329dd0a442db/modules/backend/widgets/Filter.php#L385-L395

msimkunas:
Related: #3202

PR #5192, opened 2020-06-29
Fix: Rollback plugins

gitlog:
Plugin's rollback fail: rollback functions is calling in the direct order like migrations. But rollback order must be reversed.

### DIFF
--- a/modules/system/classes/UpdateManager.php
+++ b/modules/system/classes/UpdateManager.php
@@ -327,7 +327,7 @@ class UpdateManager
         /*
          * Rollback plugins
          */
-        $plugins = $this->pluginManager->getPlugins();
+        $plugins = array_reverse($this->pluginManager->getPlugins());
         foreach ($plugins as $name => $plugin) {
             $this->rollbackPlugin($name);
         }


bennothommo:
@gitlog Could you provide further explanation as to what error this change is intending to fix?

As I understand it, the plugins are loaded alphabetically, so reversing the order would likely have no greater chance of success than keeping it in the order it is currently.

LukeTowers:
@bennothommo if getPlugins() takes into account dependencies then this could make sense since you'd want the dependent plugins to be rolled back before the plugin that they depend on rolls back.

gitlog:
Plugins aren't loaded alphabetically. They load with given dependencies - if some plugin was extentened, child plugin will be load after parent. It rule works for execute migrations. Now, when we need to rollback plugins, parent plugin is rollbacked firstly too, child is next.
My commit is changes (to reverse) sort order for plugins, when they is rolled back

4nik3t:
@gitlog  is this what I understand. I have a plugin which extends the User plugin. Now If I run php artisan plugin:refresh Rainlab.User it refreshes the user plugin but breaks my plugin. Now (after this pr) what will happen is that it will first rollback my plugin and then rollback user plugin and then again  migrate the user  plugin and then my plugin?

LukeTowers:
@gitlog could you add some unit testing for this fix?

mjauvin:
I confirm that the plugins are now being loaded in least dependant first order:
```
    public function loadPlugins()
    {   
        $this->plugins = [];

        /**
         * Locate all plugins and binds them to the container
         */
        foreach ($this->getPluginNamespaces() as $namespace => $path) {
            $this->loadPlugin($namespace, $path);
        }

        $this->sortDependencies();
                 
        return $this->plugins;
    }
```
```
    /**
     * Sorts a collection of plugins, in the order that they should be actioned,
     * according to their given dependencies. Least dependent come first.
     *
     * @return array Array of sorted plugin identifiers and instantiated classes ['Author.Plugin' => PluginBase]
     * @throws SystemException If a possible circular dependency is detected
     */
    protected function sortDependencies()
```

mjauvin:
This has not always been the case, though.

LukeTowers:
This looks like it only fixes the uninstall method, it doesn't fix anywhere else that would be affected like `plugin:rollback` or `plugin:refresh`. Please provide reproduction instructions for the original issue and test against those commands to see if other changes are required and then we can revisit this.

mjauvin:
please keep this open

mjauvin:
> This looks like it only fixes the uninstall method, it doesn't fix anywhere else that would be affected like `plugin:rollback` or `plugin:refresh`. Please provide reproduction instructions for the original issue and test against those commands to see if other changes are required and then we can revisit this.

@LukeTowers you're right, but rollback/refresh actions on a single plugin do not currently take dependencies into account anyhow, am I right?

LukeTowers:
@mjauvin I still need to know the reproduction instructions for the original issue to understand how this fix improves October

daftspunk:
This is a no brainer. Install A-Z, uninstall Z-A. Good catch @gitlog

PR #5199, opened 2020-07-07
[FileUpload] Ability to automatically delete an uploaded file when detached from its parent.

goldmont:


LukeTowers:
@bennothommo what are your thoughts on this? I almost feel like this should have been the default behaviour up to this point anyways.

bennothommo:
Only concern I'd have is if the removal is deferred until the form is saved. Deleting the file before the removal is confirmed might cause problems if the person ends up cancelling the form save.

modules/backend/formwidgets/FileUpload.php
```diff
@@ -337,6 +343,9 @@ public function onRemoveAttachment()
         $fileModel = $this->getRelationModel();
         if (($fileId = post('file_id')) && ($file = $fileModel::find($fileId))) {
             $this->getRelationObject()->remove($file, $this->sessionKey);
+            if ($this->deleteOnDetach) {
+                $file->delete();
+            }
```
LukeTowers, line 348: @goldmont As mentioned by @bennothommo this probably has to be implemented in the base File model itself to be run after detaching is committed by the deferred binding process.

bennothommo:
@goldmont Are you still interesting in finishing this off?

goldmont:
@bennothommo Hi, currently I'm not working on it. Won't you implement this feature?

bennothommo:
@LukeTowers on looking at this again, I think @goldmont has this in the right location. I think all we need is some tests to determine that the deferred binding works and doesn't remove the file until the deletion is fully committed. I'll try and work out a test case for it.

bennothommo:
@goldmont @LukeTowers I've just had another look through this. Unless I'm mistaken, this change won't be necessary at all. All that is needed is for `delete` to be defined in the relation config for the file, ie.

```php
public $attachOne = [
    'file' => [
        'System\Models\File',
        'delete' => true,
    ]
];
```

This already accounts for deferred binding, in that the [removal of the relation](https://github.com/octobercms/library/blob/develop/src/Database/Relations/AttachOneOrMany.php#L214-L216) will delete the model, which in turn will delete the file. The removal of the relation will only be triggered when deferred binding is committed, ie. after the form is saved.

goldmont:
@bennothommo I'll try! Thanks.

PR #5206, opened 2020-07-09
Fix "use statement with non-compound names has no effect"

LukeTowers:
The linked issue is complaining about commented out use statements still being parsed, but I've actually run into the problem of it breaking when non-compound names are used before since it's not obvious


LukeTowers:
Can't actually use this solution since it would break all existing PHP code blocks that are using root namespace classes (facades mostly) without importing them via `use` first. Will submit a different approach.

PR #5209, opened 2020-07-14
Reveal sensitive field if part of an AJAX request

bennothommo:
The sensitive field currently does not work with AJAX requests correctly if it is unrevealed - it sends the hidden placeholder instead of the real value with AJAX requests.

This change allows the sensitive field to intercept AJAX requests on elements within the same form, and reveals the field. Once revealed, it then replays the AJAX request on the original element, which should then pick up the sensitive field's true value in the form data.


LukeTowers:
@bennothommo could you provide an example of why you need the data present in an AJAX request?

bennothommo:
Closing in favour of #5212.

PR #5212, opened 2020-07-15
Provide an accessor for the form widget of the Settings controller

bennothommo:
Replacement for #5209.

The Settings controller does not currently have any method to get the form widget or its data through normal means as the form widget is protected and does not use the normal FormController behavior.

This prevents extensions to the controller from being able to manipulate the form.

This adds a `formGetWidget` method (similar to the FormController behavior) that allows people to access the widget. As the form widget in the Settings controller is dependent on the URL to determine which plugin or module is being configured, a method to auto-guess the current model has also been added.

### DIFF
--- a/modules/system/controllers/Settings.php
+++ b/modules/system/controllers/Settings.php
@@ -219,7 +219,7 @@ class Settings extends Controller
     }
 
     /**
-     * Auto-guesses the requested setting item by URL segments from the Request object.
+     * Guesses the requested setting item by URL segments from the Request object.
      *
      * @return array
      */
--- a/modules/system/controllers/Settings.php
+++ b/modules/system/controllers/Settings.php
@@ -219,7 +219,7 @@ class Settings extends Controller
     }
 
     /**
-     * Guesses the requested setting item by URL segments from the Request object.
+     * Guesses the requested setting item from the current URL segments provided by the Request object.
      *
      * @return array
      */
--- a/modules/system/controllers/Settings.php
+++ b/modules/system/controllers/Settings.php
@@ -1,5 +1,7 @@
 <?php namespace System\Controllers;
 
+use Config;
+use Request;
 use Lang;
 use Flash;
 use Backend;
@@ -140,6 +142,22 @@ class Settings extends Controller
     }
 
     /**
+     * Returns the form widget used by this behavior.
+     *
+     * @return \Backend\Widgets\Form
+     */
+    public function formGetWidget()
+    {
+        if (is_null($this->formWidget)) {
+            $item = $this->findSettingItem();
+            $model = $this->createModel($item);
+            $this->initWidgets($model);
+        }
+
+        return $this->formWidget;
+    }
+
+    /**
      * Prepare the widgets used by this action
      * Model $model
      */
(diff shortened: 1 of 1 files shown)


modules/system/controllers/Settings.php
```diff
@@ -187,4 +217,23 @@ protected function findSettingItem($author, $plugin, $code)
 
         return $item;
     }
+
+    /**
+     * Auto-guesses the requested setting item by URL segments from the Request object.
+     *
+     * @return array
+     */
+    protected function autoGuessSettingItem()
```
LukeTowers, line 226:
```suggestion
    protected function guessSettingItem()
```
Adding "auto" to this method name seems redundant

modules/system/controllers/Settings.php
```diff
@@ -169,10 +187,22 @@ protected function createModel($item)
...
+     * @param string|null $plugin
+     * @param string|null $code
+     *
+     * @return array
      */
-    protected function findSettingItem($author, $plugin, $code)
+    protected function findSettingItem($author = null, $plugin = null, $code = null)
     {
+        if (is_null($author) || is_null($plugin)) {
+            [$author, $plugin, $code] = $this->autoGuessSettingItem();
```
LukeTowers, line 203: What version of PHP does this auto expanding require?

modules/system/controllers/Settings.php
```diff
@@ -187,4 +217,23 @@ protected function findSettingItem($author, $plugin, $code)
...
+    /**
+     * Auto-guesses the requested setting item by URL segments from the Request object.
+     *
+     * @return array
+     */
+    protected function autoGuessSettingItem()
+    {
+        $segments = Request::segments();
+
+        if (!empty(Config::get('cms.backendUri', 'backend'))) {
```
LukeTowers, line 230: Have you tested it with both `/backend` and `/`?

modules/system/controllers/Settings.php
```diff
@@ -1,5 +1,7 @@
 <?php namespace System\Controllers;
 
+use Config;
+use Request;
```
LukeTowers, line 4: It's a minor quibble I know, but I like having the imports ordered by lengt

modules/system/controllers/Settings.php
bennothommo: True - changed

modules/system/controllers/Settings.php
bennothommo: PHP 7.1 introduced it, so it should be right for L6.

modules/system/controllers/Settings.php
bennothommo: Changed.

modules/system/controllers/Settings.php
```diff
@@ -187,4 +217,23 @@ protected function findSettingItem($author, $plugin, $code)
 
         return $item;
     }
+
+    /**
+     * Auto-guesses the requested setting item by URL segments from the Request object.
```
LukeTowers, line 222:
```suggestion
     * Guesses the requested setting item by URL segments from the Request object.
```

modules/system/controllers/Settings.php
```diff
@@ -187,4 +217,23 @@ protected function findSettingItem($author, $plugin, $code)
 
         return $item;
     }
+
+    /**
+     * Guesses the requested setting item by URL segments from the Request object.
```
LukeTowers, line 222:
```suggestion
     * Guesses the requested setting item from the current URL segments provided by the Request object.
```

modules/system/controllers/Settings.php
bennothommo: Yep, confirmed it working with `/` as well.

LukeTowers:
LGTM!

LukeTowers:
Add this to the Laravel 6 wip release notes in the meta repo please too when merging

PR #5215, opened 2020-07-16
Enable nested repeaters inside an initial repeater group

b8ne:
This update allows a repeater group to include additional 'nested repeaters'.
Previously the repeater widget would directly add 'fields'.  
This update checks the field type of the group element first.  If it is a repeater type then the entire config is loaded.

### DIFF
--- a/modules/backend/formwidgets/Repeater.php
+++ b/modules/backend/formwidgets/Repeater.php
@@ -450,10 +450,10 @@ class Repeater extends FormWidgetBase
 
         if ($group && isset($group->{$code}['type']) && $group->{$code}['type'] === 'repeater') {
             $fields = [
-                $code => $group->{$code}
+                $code => $group->{$code},
             ];
         } else {
-            $fields = array_get($this->groupDefinitions, $code.'.fields');
+            $fields = array_get($this->groupDefinitions, $code . '.fields');
         }
 
         if (!$fields) {
--- a/modules/backend/formwidgets/Repeater.php
+++ b/modules/backend/formwidgets/Repeater.php
@@ -442,7 +442,19 @@ class Repeater extends FormWidgetBase
             return null;
         }
 
-        $fields = array_get($this->groupDefinitions, $code.'.fields');
+        // Bring the config in to use all attributes in repeater
+        $group = $this->getConfig('groups', []);
+        if (is_string($group)) {
+            $group = $this->makeConfig($group);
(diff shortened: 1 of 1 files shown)


modules/backend/formwidgets/Repeater.php
```diff
@@ -442,7 +442,19 @@ protected function getGroupFormFieldConfig($code)
...
-        $fields = array_get($this->groupDefinitions, $code.'.fields');
+        // Bring the config in to use all attributes in repeater
+        $group = $this->getConfig('groups', []);
+        if (is_string($group)) {
+            $group = $this->makeConfig($group);
+        }
+
+        if ($group && isset($group->{$code}['type']) && $group->{$code}['type'] === 'repeater') {
+            $fields = [
+                $code => $group->{$code}
```
bennothommo: We are moving towards having trailing commas on all arrays in our code style now, to simplify Git diffs.
```suggestion
                $code => $group->{$code},
```

modules/backend/formwidgets/Repeater.php
```diff
@@ -442,7 +442,19 @@ protected function getGroupFormFieldConfig($code)
...
+            ];
+        } else {
+            $fields = array_get($this->groupDefinitions, $code.'.fields');
```
bennothommo: Our code style suggests having spaces between concatenation parts.
```suggestion
            $fields = array_get($this->groupDefinitions, $code . '.fields');
```

bennothommo (changes requested):
@b8ne sorry for the delay in reviewing this. Just a couple of code style tweaks for you to action.

modules/backend/formwidgets/Repeater.php
b8ne: @bennothommo should be resolved now.

bennothommo:
Nice one @b8ne - just need to test it first, then it's good to merge.

bennothommo:
@b8ne I've just tested this, and I've found that irrespective of whether this change is applied or not, a nested repeater inside a repeater group is loaded just fine.

Is there are particular config you are using that is exhibiting the issue that this PR intends to resolve?

b8ne:
@bennothommo yeah confirmed.  Not sure where but another update may have fixed this.  Previously, nested repeaters would load as a blank element because there was no group to attach the fields to. Closing for now.

PR #5217, opened 2020-07-21
Fix updating HasOne relation

alvaro-canepa:
`$this->viewModel` doesn't have `$saveData` attributes, it will be reflected after `$this->viewWidget->setFormValues($saveData);`, but in `$this->viewWidget->model`.

To save related model with user data, must use `$model` var inside of `$this->viewWidget`.

### DIFF
--- a/modules/backend/behaviors/RelationController.php
+++ b/modules/backend/behaviors/RelationController.php
@@ -1157,10 +1157,8 @@ class RelationController extends ControllerBehavior
             }
         }
         elseif ($this->viewMode == 'single') {
-            $this->viewModel = $this->manageWidget->model;
-
             $this->viewWidget->setFormValues($saveData);
-            $this->viewModel->save(null, $this->manageWidget->getSessionKey());
+            $this->viewWidget->model->save(null, $this->manageWidget->getSessionKey());
         }
 
         return $this->relationRefresh();
--- a/modules/backend/behaviors/RelationController.php
+++ b/modules/backend/behaviors/RelationController.php
@@ -1157,8 +1157,10 @@ class RelationController extends ControllerBehavior
             }
         }
         elseif ($this->viewMode == 'single') {
+            $this->viewModel = $this->viewWidget->model;
+
             $this->viewWidget->setFormValues($saveData);
-            $this->viewWidget->model->save(null, $this->manageWidget->getSessionKey());
+            $this->viewModel->save(null, $this->manageWidget->getSessionKey());
         }
 
         return $this->relationRefresh();


modules/backend/behaviors/RelationController.php
```diff
@@ -1157,10 +1157,8 @@ public function onRelationManageUpdate()
             }
         }
         elseif ($this->viewMode == 'single') {
-            $this->viewModel = $this->manageWidget->model;
```
LukeTowers, line 1160: Why would you just change this line to `$this->viewWidget->model`?

alvaro-canepa: You right, I made many tests and this was the latest working code. Let me change it.

LukeTowers:
@alvaro-canepa can you provide a screen recording and steps to reproduce the original bug?

alvaro-canepa:
## Steps to reproduce bug
Follow [documentation](https://octobercms.com/docs/backend/relations#has-one) steps to add backend relation form for HasOne.

### 1.- Create Vendor model with "user" relation
```php
public $belongsTo = [
        'user' => [User::class]
    ];
```

### 2.- Extends User Model (`Lovata\Buddies\User`), adding "vendor" relation
```php
User::extend(
    function ($obModel) {
        /** @var User $obModel */
        $obModel->hasOne['vendor'] = [Vendor::class];
    }
);
```

### 3.- Extends Users controller, adding vendor relation config
```yaml
vendor:
    label: 'planetadeleste.vendorsshopaholic::lang.label.name'
    manage:
        form: $/planetadeleste/vendorsshopaholic/models/vendor/fields.yaml
        list: $/planetadeleste/vendorsshopaholic/models/vendor/columns.yaml
    view:
        form: $/planetadeleste/vendorsshopaholic/models/vendor/fields.yaml
        toolbarButtons: update|delete
```

#### Vendor form fields
```yaml
fields:
    commission:
        label: 'planetadeleste.vendorsshopaholic::lang.label.commission'
        span: left
```

Screen

From backend, go to update an User record, go to "Vendor" tab and click to "Update Vendor" button, when records doesn't exists, it will to created and attribute `commission` is saved. On this case, the method `\Backend\Behaviors\RelationController::onRelationManageCreate` was used.

After "vendor" relation was created, click again on "Update Vendor" button and change the value of `comission` and click Save,  only update the preview form, but the new value was not saved. 
`\Backend\Behaviors\RelationController::onRelationManageUpdate` method was exec now.

Thanks!

bennothommo:
Confirmed the bug in the Test plugin:

- Go to Playground > People.
- Click "Manage Phones".
- Edit or create a phone and change the view to "Relation Controller", which uses a singular link to a Person. (I had to add _ajax_options.htm into the `controller/phones` directory for this to work, will fix this in a PR to the test plugin).

I've also confirmed this fix does work, but I'm still trying to figure out the *why*.

bennothommo:
@LukeTowers on diving into this a bit deeper, this change effectively reverts https://github.com/octobercms/october/commit/6e25c5a8d0834aa083455f9f2b7c75b9bda7ff1c#diff-63fc4fc41aa256f8d5bcfe7a37245ecf, which would mean that saving the translations would no longer work. The change suggested by @alvaro-canepa brings it back to the state it was before that commit - the line is effectively not needed since the view widget is already using its own model instance, and the commit posted changed the view widget's model to the manage widget's model.

I've also realised that the above commit breaks creating a new record through the relation controller if a previous model instance already exists. You can do this by following my instructions above and saving a link with the record, then trying to create a new record when a link exists.

Given this, I feel it might be best to roll back https://github.com/octobercms/october/commit/6e25c5a8d0834aa083455f9f2b7c75b9bda7ff1c and try and approach @tobias-kuendig's fix in a different manner.

LukeTowers:
@bennothommo are you sure about
> the view widget is already using its own model instance, and the commit posted changed the view widget's model to the manage widget's model.

The code prior to the posted commit was `$model = $this->viewModel` not `$model = $this->viewWidget->model`

bennothommo:
@LukeTowers this was the code previous to 6e25c5a - https://github.com/octobercms/october/blob/cd86c62b94f4a5dee140bac50e3878c592bfcf10/modules/backend/behaviors/RelationController.php#L1143-L1162.

There was no definition for `$this->viewModel` in that method, which means it was using the model defined for the view widget.

LukeTowers:
Hmm, it's late so I'll have to review this more in depth tomorrow and see if I can figure it out.

bennothommo:
@LukeTowers Just to throw another curve ball into the mix, the fix in https://github.com/octobercms/october/commit/6e25c5a8d0834aa083455f9f2b7c75b9bda7ff1c *does* in fact work if you have the Translate plugin enabled. So this particular bug only exhibits if Translate is unavailable.

Pindagus:
> @LukeTowers on diving into this a bit deeper, this change effectively reverts [6e25c5a#diff-63fc4fc41aa256f8d5bcfe7a37245ecf](https://github.com/octobercms/october/commit/6e25c5a8d0834aa083455f9f2b7c75b9bda7ff1c#diff-63fc4fc41aa256f8d5bcfe7a37245ecf), which would mean that saving the translations would no longer work. The change suggested by @alvaro-canepa brings it back to the state it was before that commit - the line is effectively not needed since the view widget is already using its own model instance, and the commit posted changed the view widget's model to the manage widget's model.
> 
> I've also realised that the above commit breaks creating a new record through the relation controller if a previous model instance already exists (in the instance of a Test plugin, I get validation errors saying the Name field is empty, when it is not). You can do this by following my instructions above and saving a link with the record, then trying to create a new record when a link exists.
> 
> Given this, I feel it might be best to roll back [6e25c5a](https://github.com/octobercms/october/commit/6e25c5a8d0834aa083455f9f2b7c75b9bda7ff1c) and try and approach @tobias-kuendig's fix in a different manner.

@bennothommo I agree it might be best to roll back [6e25c5a](https://github.com/octobercms/october/commit/6e25c5a8d0834aa083455f9f2b7c75b9bda7ff1c). Right now I'm actively avoiding using this since creating a new record through the relation controller is broken. Any progress on the roll back?

LukeTowers:
Should be fixed by https://github.com/octobercms/october/commit/5a843bc90986a29ec0375697490af9864ecbd188; please let me know if there's still any issues.

PR #5225, opened 2020-07-27
Improved use case extraction for CMS objects.

### LINKED ISSUE #5224: Does the PHP code section of theme templates support traits?
This seems to be a strange problem, because an old project uses this framework, so I am learning to use it, but I am now encountering a very serious problem.

When I was developing the theme, I found that my development tool (PHPSTORM) was not well supported because this file was too complicated. It had declaration blocks, PHP code blocks, and twig template blocks. If I set the file type `.htm` of phpstorm to `twig`, it can format my twig code normally, but this will ruin my PHP code block, because his format is all messed up, so I To deal with it in an elegant way is to define a `trait` object Template, and then override the onStart method in this trait. I can even group more things here, like the following.

```php
title = "Demonstration"
url = "/test"
layout = "default"
==
// Please note here
use October\Demo\Contract\Template;
==
<div>
    hello ,{{ name }}
</div>

```

However, I found that things are not so wonderful, because my use statement became the following after compilation.

```php
<?php

use October\Demo\Contract\Template;

class Cms5f1da03a53c34871546283_073c680da975bfac81380a3d810f93aaClass extends Cms\Classes\PageCode
{

}
```

This is very bad. In my expectation, it should be included in the class, but now, the use statement loses its meaning.

So, can it become like this?

```php
<?php

class Cms5f1da03a53c34871546283_073c680da975bfac81380a3d810f93aaClass extends Cms\Classes\PageCode
{
    use October\Demo\Contract\Template;

}
```
In this way he will work very well, and I can also write beautiful code in my template, as well as perfect code hints and highlights.

I have noticed that it seems that such support cannot be provided temporarily when parsing reorganized files.

- modules/cms/classes/CodeParser.php:122

```php
    protected function rebuild($path)
    {
        $uniqueName = str_replace('.', '', uniqid('', true)).'_'.md5(mt_rand());
        $className = 'Cms'.$uniqueName.'Class';

        $body = $this->object->code;
        $body = preg_replace('/^\s*function/m', 'public function', $body);

        $codeNamespaces = [];
        $pattern = '/(use\s+[a-z0-9_\\\\]+(\s+as\s+[a-z0-9_]+)?;\n?)/mi';
        preg_match_all($pattern, $body, $namespaces);
        $body = preg_replace($pattern, '', $body);

        $parentClass = $this->object->getCodeClassParent();
        if ($parentClass !== null) {
            $parentClass = ' extends '.$parentClass;
        }

        $fileContents = '<?php '.PHP_EOL;
        foreach ($namespaces[0] as $namespace) {
            $fileContents .= $namespace;
        }

        $fileContents .= 'class '.$className.$parentClass.PHP_EOL;
        $fileContents .= '{'.PHP_EOL;
        $fileContents .= $body.PHP_EOL;
        $fileContents .= '}'.PHP_EOL;

        $this->validate($fileContents);

        $this->makeDirectorySafe(dirname($path));

        $this->writeContentSafe($path, $fileContents);

        return $className;
    }
```

Or here, can you judge the use statement ending with `Trait` and put it in the class.

Or, is there any other more elegant way to deal with php code blocks in template files?

bennothommo:
- Prevents commented-out use cases from being inserted into code. (for example, use cases were still being added for instances found in `/** comment */` blocks)
- Inserts traits into correct location.
- Handles grouped use imports.

Fixes #5224.

Test cases have been added for these new handlers.

### DIFF
--- a/modules/cms/classes/CodeParser.php
+++ b/modules/cms/classes/CodeParser.php
@@ -4,6 +4,7 @@ use File;
 use Lang;
 use Cache;
 use Config;
+use ReflectionClass;
 use SystemException;
 
 /**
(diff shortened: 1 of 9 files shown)


WonlyCms:
There are some problems. The code that [removes the comment](https://github.com/octobercms/october/blob/3a2f9dbeec10366a5fdbd65d2e38a63bf6f46d17/modules/cms/classes/CodeParser.php#L133-L134) may be too violent and cause accidents. Even adding `^` before the regular expression may cause accidents. It is recommended not to deal with it, or use `php-parser` for processing.

```php

<?php

$str = 'http://example.com';


$str = <<<DOC
<img src="
//img.example.com/girl.jpg
" />
DOC;




$result = preg_replace('/^\/\/.*?$/m','',$str);

var_dump($result);

```

bennothommo:
We've discussed the ramifications of this change internally and won't be going ahead with this PR at this stage.

PR #5229, opened 2020-08-03
Alternative check for asset pathing

bennothommo:
Replaces https://github.com/octobercms/october/commit/b1954e92586f33a345068d56b957d7e366a5ba71.

Instead of using a temporary file to allow checks with `realpath()` to succeed, this allow the `getFilePath()` method to do a real path check on the parent directory of a file if the file does not exist.

This will allow new asset files to still be saved correctly, whilst still preventing paths outside of the assets directory to be used.


bennothommo:
@LukeTowers One thing to consider with this fix as well as your previous commit is that, previously, people were able to create a new file in a subdirectory that may not exist, like so:

image

This code block below would create the subdirectory first then try to save the file:

https://github.com/octobercms/october/blob/74059fc61bd415d9335dae8f34d3de2ba3ba5b7d/modules/cms/classes/Asset.php#L178-L187

Your commit throws an exception in this case, as `touch()` will not create missing directories, and mine fails because the `getFilePath()` method will return false due to it not being able to resolve the parent directory either. So either we'll have to have a warning or something requiring people to create the directories first, or find an alternative method.

tests/unit/cms/classes/AssetTest.php
```diff
@@ -0,0 +1,133 @@
...
+class AssetTest extends TestCase
+{
+    public function testLoad()
+    {
+        $theme = Theme::load('test');
+
+        // Forward compatible assertions
+        // @TODO: Use only `assertStringContainsString` after L6 upgrade
+
+        if (method_exists($this, 'assertStringContainsString')) {
```
LukeTowers, line 15: You can just target the L6 branch for this PR, we'll probably merge it in this week.

LukeTowers:
@bennothommo excellent point re: previously being able to create new directories. We definitely need to make sure we support that use case.

tests/unit/cms/classes/AssetTest.php
```diff
@@ -0,0 +1,133 @@
...
+            Asset::load($theme, 'js/invalid.js')
+        );
+
+        // Check that we cannot break out of assets directory
+        $this->assertNull(
+            Asset::load($theme, '../../../../js/helpers/fakeDom.js')
+        );
+        $this->assertNull(
+            Asset::load($theme, '../content/html-content.htm')
+        );
```
LukeTowers, line 64: Can you also add `../../../config/database.php` to the tests?

LukeTowers:
@bennothommo The way I see it we should probably simplify this a bit. I don't think this method should care if the location exists or not, it should just build the path, with some validation to prevent malicious paths from being returned (additionally we should implement the changes in https://github.com/octobercms/october/blob/develop/modules/cms/classes/CmsObject.php#L224 too)

We could do that in one of two ways as I see it:
- Create a helper function that is the same as realpath() except it doesn't require the path to exist, it just fully resolves the provided path
- Explicitly ban `../` and `./` from the path, like is done here: https://github.com/octobercms/october/blob/develop/modules/cms/widgets/AssetList.php#L482

bennothommo:
@LukeTowers this is ready for final review.

PR #5230, opened 2020-08-03
'enableCsrfProtection' => false

tommychyzh:
Invalid security token fix


LukeTowers:
That is not a "fix", that is disabling a security feature.

Fix your sessions and then you won't have problems with CSRF anymore.

PR #5257, opened 2020-09-01
Fix error when tab location could not be found

jacobdekeizer:
I had a situation where the tab location couldn't be found and an exception is thrown. When the exception was thrown, the richeditor in de rainlab pages plugin didn't work anymore, see below:
image

I added a try catch to solve this issue.

### DIFF
--- a/modules/system/assets/ui/js/tab.js
+++ b/modules/system/assets/ui/js/tab.js
@@ -89,7 +89,10 @@
         this.updateClasses()
 
         if (location.hash && this.$tabsContainer.is('[data-linkable]')) {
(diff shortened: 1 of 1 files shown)


LukeTowers:
Fixed in https://github.com/octobercms/october/commit/bf241baee748a4af542cdb55317a48b23e617e94. While I appreciate the interest in contributing to the framework, when you're writing code for a project that will be used by tens of thousands of people around the world you have to always make sure you identify and resolve the root cause of issues instead of just applying band-aids to hide the symptoms since that will only result in headaches and harder bugs to solve further down the road.

PR #5262, opened 2020-09-05
let app name configured from env

panakour:
This is let you more simple to automate deployments

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -28,7 +28,7 @@ return [
     | any other location as required by the application or its packages.
     */
 
-    'name' => 'October CMS',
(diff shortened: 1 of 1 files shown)


bennothommo:
@panakour sorry, but we can't merge this. The config files are kept the way they are for people who don't use environment variables.

If you are automating deployments, you can always make `php artisan october:env` part of your deployment process to take advantage of environment variables.

LukeTowers:
Personally I wouldn't recommend using `october:env` as a part of your deployment process, you should just run it once during your development and then just use .env files in your deployment process.

PR #5263, opened 2020-09-06
Improve zh-tw translation

LeoCantThinkOfAName:
### Some adjustments and improvements for zh-tw translation.
- Add some missing entries in the original zh-tw translate file.
- Better and more nature terms that fits Taiwanese user.
- Create client.php translate file, and recompile js.


LeoCantThinkOfAName:
Gonna fix some typo.

PR #5264, opened 2020-09-06
Reopen: Improve zh-tw translate (fix typo from last time)

LeoCantThinkOfAName:
### Some adjustments and improvements for zh-tw translation.

- Add some missing entries in the original zh-tw translate file.
- Better and more nature terms that fits Taiwanese user.
- Create client.php translate file, and recompile js.


LukeTowers:
For future reference you don't have to close and re-open a PR to modify it, you can just continue committing to the branch that you made the PR from (in this case `develop` on your fork).

modules/backend/lang/zh-tw/lang.php
```diff
@@ -297,7 +304,10 @@
         'email' => 'Email'
     ],
     'filter' => [
-      'all' => '全部'
+        'all' => '全部',
+        'options_method_not_exists' => "The model class :model must define a method :method() returning options for the ':filter' filter.",
```
LukeTowers: Ideally there shouldn't be untranslated strings in language files

modules/cms/lang/zh-tw/lang.php
```diff
@@ -168,8 +171,8 @@
         'content' => '內容',
         'hidden' => '隱藏',
         'hidden_comment' => '隱藏頁面只能被登錄的後台使用者訪問.',
-        'enter_fullscreen' => '進入全屏模式',
-        'exit_fullscreen' => '退出全屏模式'
+        'enter_fullscreen' => '進入全螢幕模式',
+        'exit_fullscreen' => '退出全螢幕模式'
```
LukeTowers: Please make sure trailing commas are always used in language files, this makes it easier for future enhancements to be made

modules/system/lang/zh-tw/client.php
```diff
@@ -0,0 +1,102 @@
...
+        ],
+        'editor' => [
+            'title' => 'Source code editor',
+            'description' => 'Your operating system should be configured to listen to one of these URL schemes.',
+            'openWith' => 'Open with',
+            'remember_choice' => 'Remember selected option for this session',
+            'open' => 'Open',
+            'cancel' => 'Cancel',
+        ],
+    ],
```
LukeTowers: Untranslated strings shouldn't be needed in the client.php file either, I believe it should fallback to the application's fallback locale

PR #5265, opened 2020-09-06
app name configured from env

panakour:
In continues for our conversation on #5262. This will let you change app name from env after run php artisan `october:env`

### DIFF
--- a/modules/system/console/OctoberEnv.php
+++ b/modules/system/console/OctoberEnv.php
@@ -355,6 +355,7 @@ class OctoberEnv extends Command
         return [
             'app' => [
                 'APP_DEBUG' => 'debug',
+                'APP_NAME' => 'name',
                 'APP_URL' => 'url',
                 'APP_KEY' => 'key',
             ],


LukeTowers:
I've tested this and it works, however I'm not sure we want to encourage people to set the app name in their .env files by default, most of the time the Application Name is static for the entire application, regardless of what environment it's deployed to. Could you provide more information as to why you want it included in the environment file?

panakour:
@LukeTowers I used to have diferent app name for each environment. This is help you seperate each app with its own name. For example if you test the staging environment sending an email that uses `{{ appName }}` it will help you regognize it.

Also even laravel have the app name by default in the env file:
https://github.com/laravel/laravel/blob/dc5817bbf2552b48d93bde99688d56e99b7a7708/config/app.php#L16
https://github.com/laravel/laravel/blob/dc5817bbf2552b48d93bde99688d56e99b7a7708/.env.example#L1

LukeTowers:
Hmm, it's a tough call. I can see the value in naming your environments different for distinguishing emails etc, but I'm also not sure about it, even after reading through the Laravel PR that added it: https://github.com/laravel/laravel/pull/4204.

@bennothommo what's your thoughts on this?

bennothommo:
@LukeTowers @panakour it seems that in this instance here, as well as in the Laravel PR that @LukeTowers mentioned, the only reason that people seem to bring up having a different app name is for identifying a different environment - so my question is, why not just make the name of the environment (eg. `APP_ENV`) available for mail templates, if it isn't already?

panakour:
@bennothommo Thank you for your thoughts, it help me rethink the way to solve my issue. 
The APP_ENV is one solution. If you think in general the APP_NAME is not needed to configure from env vars feel free to close this.

PR #5266, opened 2020-09-06
disable Rocket Loader for all assets added with addJs() method

### LINKED ISSUE #536: Translate plugin doesn't show message in "Update messages" section
I can see languages and translations in database. 
Screen Shot 2020-01-09 at 12 42 56 AM

mjauvin:
fixes https://github.com/rainlab/translate-plugin/issues/536


LukeTowers:
Separate plugin used instead, see https://github.com/rainlab/translate-plugin/issues/536#issuecomment-688002175

PR #5273, opened 2020-09-10
facades docblocks

SergeyKasyanov:


bennothommo:
@serega-kasyanow Would you mind adjusting these docblocks to be consistent with the Facades in the library? (eg. https://github.com/octobercms/library/blob/develop/src/Support/Facades/Config.php#L5-L29)

- Remove the `@package` definition, and comment above it.
- Add a `@see` tag at the bottom which links to the class that this Facade is covering.

SergeyKasyanov:
@bennothommo Done. Also I removed `@see` from `getFacadeAccessor()` for same reason.

bennothommo:
@serega-kasyanow Perfect - thanks heaps!

PR #5281, opened 2020-09-19
Check for any other flash messages before flashing a success message.

aniket-magadum:
Inside of my settings model, I am listening to the save event and I am making a API Call to a server and setting some flash messages in the session.

```php
public function afterSave()
{
   App::make('zoho_client');
   try {
       $oAuthClient = ZohoOAuth::getClientInstance();
       $oAuthClient->generateAccessToken(self::get('grant_token'));
       Flash::success('Access Token Generated Successfully');
    } catch (\Throwable $th) {
       Flash::error($th->getMessage());
    }
}
```
The problem here is that neither of this flash messages get displayed as there is another message [here](https://github.com/octobercms/october/blob/15ca68c22df788397b2e84a9abfa5ecbc579b3ff/modules/system/controllers/Settings.php#L107) which is overriding the earlier messages.

So I have just checked if there are any existing messages then do not flash the success message. Please let me know if there is any other way to achieve this.

### DIFF
--- a/modules/system/controllers/Settings.php
+++ b/modules/system/controllers/Settings.php
@@ -104,8 +104,9 @@ class Settings extends Controller
         }
         $model->save(null, $this->formWidget->getSessionKey());
 
-        Flash::success(Lang::get('system::lang.settings.update_success', ['name' => Lang::get($item->label)]));
-
+        if(!Flash::check()){
+            Flash::success(Lang::get('system::lang.settings.update_success', ['name' => Lang::get($item->label)]));
+        }
         /*
          * Handle redirect
          */
--- a/modules/system/controllers/Settings.php
+++ b/modules/system/controllers/Settings.php
@@ -104,7 +104,7 @@ class Settings extends Controller
         }
         $model->save(null, $this->formWidget->getSessionKey());
 
-        if(!Flash::check()){
+        if (!Flash::check()) {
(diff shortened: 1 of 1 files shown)


LukeTowers:
@4nik3t why are you not just throwing an ApplicationException? That way the user will see a popup with the exception message that they have time to read and close on their own before it disappears like a flash message. Generally within October we recommend avoiding Flash::error() because it's not a good way of displaying error messages to the end user.

aniket-magadum:
@LukeTowers Your point to throw an ApplicationException instead Flash::error() is fair enough. But what if I want to show a different success message or for that matter also shows a Flash::info() or Flash::warning() message?

LukeTowers:
@4nik3t I can't think of any good reason why you would be needing to show a different flash message in the save action.

aniket-magadum:
@LukeTowers my use case for this that when the user saves the setting I want to display a feedback message to the user if there is any issue with the token he has entered. Anyways I can do it another way and this issue can be closed if not serving any purpose.

LukeTowers:
@4nik3t For that case I would just have it only inform the user if something went wrong, if the record saves successfully they could just assume that everything worked fine. I'm not necessarily opposed to this if a valid use case can be found, but I'll close it for now.

PR #5284, opened 2020-09-25
Removal of custom fallback image feature

### LINKED ISSUE #5150: Allow custom fallback image for missing images
The rain library currently uses [a base64 string](https://github.com/octobercms/library/blob/30e610d871622df754f028bc566a412a3727f14a/src/Database/Attach/BrokenImage.php#L26) that translates to this image. It would be good to allow upload of custom image/base64 encoded string like say a company logo so that when BrokenImage appears on my website it shows the company logo instead. Is there any other way to do it? I thought maybe it would be a good idea to make a plugin for this but I cannot see how I could manipulate the october library code from plugin to allow such custom image.

mohsin:
Added a form field under the customize backend page to allow addition of a custom fallback image when an image is missing. Support code to fix #5150.

### DIFF
--- a/modules/backend/models/BrandSetting.php
+++ b/modules/backend/models/BrandSetting.php
@@ -41,7 +41,8 @@ class BrandSetting extends Model
 
     public $attachOne = [
         'favicon' => \System\Models\File::class,
-        'logo' => \System\Models\File::class
+        'logo' => \System\Models\File::class,
+        'fallback_image' => \System\Models\File::class,
     ];
 
     /**
@@ -113,6 +114,17 @@ class BrandSetting extends Model
         }
 
         return self::getDefaultLogo() ?: null;
+    }
+
+    public static function getFallbackImage()
+    {
+        $settings = self::instance();
+
+        if ($settings->fallback_image) {
+            return $settings->fallback_image->getDiskPath();
+        }
+
+        return null;
     }
 
     public static function renderCss()
--- a/modules/backend/models/brandsetting/fields.yaml
+++ b/modules/backend/models/brandsetting/fields.yaml
@@ -39,6 +39,15 @@ tabs:
             span: right
             fileTypes: jpg,jpeg,bmp,png,webp,gif,svg,ico
 
+        fallback_image:
+            label: backend::lang.branding.fallback_image
+            type: fileupload
+            commentAbove: backend::lang.branding.fallback_image_description
+            mode: image
+            tab: backend::lang.branding.brand
+            span: right
+            fileTypes: jpg,jpeg,bmp,png,webp,gif,svg,ico
+
         primary_color:
             label: backend::lang.branding.primary_color
             type: colorpicker
--- a/modules/backend/lang/en/lang.php
+++ b/modules/backend/lang/en/lang.php
@@ -450,6 +450,8 @@ return [
         'logo_description' => 'Upload a custom logo to use in the back-end.',
         'favicon' => 'Favicon',
         'favicon_description' => 'Upload a custom favicon to use in the back-end',
+        'fallback_image' => 'Fallback Image',
+        'fallback_image_description' => 'Upload a custom fallback image to show as a broken image',
         'app_name' => 'App Name',
         'app_name_description' => 'This name is shown in the title area of the back-end.',
         'app_tagline' => 'App Tagline',


modules/backend/models/brandsetting/fields.yaml
```diff
@@ -39,6 +39,15 @@ tabs:
...
             fileTypes: jpg,jpeg,bmp,png,webp,gif,svg,ico
 
+        fallback_image:
+            label: backend::lang.branding.fallback_image
+            type: fileupload
+            commentAbove: backend::lang.branding.fallback_image_description
+            mode: image
+            tab: backend::lang.branding.brand
+            span: right
+            fileTypes: jpg,jpeg,bmp,png,webp,gif,svg,ico
```
LukeTowers, line 49: ico is not a supported general image type in browsers AFAIK. Even bmp would be pushing it a bit.

modules/backend/models/BrandSetting.php
```diff
@@ -115,6 +116,17 @@ public static function getLogo()
         return self::getDefaultLogo() ?: null;
     }
 
+    public static function getFallbackImage()
+    {
+        $settings = self::instance();
+
+        if ($settings->fallback_image) {
+            return $settings->fallback_image->getDiskPath();
```
LukeTowers, line 124: This is too tightly coupled to the local storage driver, especially in coordination with the storage_path('app' . $filePath) from the related PR. Additionally, this should be able to be configured with `brand.fallbackImagePath` just like the logo and favicon can

LukeTowers:
In addition to the other comments, broken images are currently only used for File models that fail to be resized with the provided configuration, they are not used for any other type of image resizing done in the core, and I'm not sure I want them to be. Could you provide some examples as to why you would prefer broken images to appear rather than an actual broken image in the browser?

modules/backend/models/brandsetting/fields.yaml
SaifurRahmanMohsin: Alright, I can remove them both.

mohsin: As per [the MDN web docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#Supported_image_formats), all these formats mentioned are supported by "Chrome, Edge, Firefox, Internet Explorer, Opera, Safari", the only exception being webp which is not supported by Internet Explorer and Safari. But since you don't want them, I can remove them off thro' a commit.

LukeTowers: .ico should only be used for favicons, it's not a great format for general images.

modules/backend/models/BrandSetting.php
mohsin: Making it configurable with brand.fallbackImagePath is a matter of including a getDefaultFallbackImage() method in BrandSettings.php similar to those of the logo and favicon. I can add this as a commit. But regarding the removal of the storage_path, I’m uncertain what can be done. What do you suggest?

SaifurRahmanMohsin:
@LukeTowers I honestly cannot think of specific use cases where I’d want custom broken images to appear in the frontend. But that is true even about them appearing in the backend i.e. without this OctoberCMS base64 BrokenImage svg code, the browser does render it’s own broken image too and should be sufficient enough yet it was OctoberCMS that offered this custom broken image and so far clients I’ve build for using OctoberCMS have liked the appearance over the typical broken image by the browser. Ofcourse, there have even been requests on whether this can be changed, which was the original reason for even raising the original issue.

I imagine that this same website-consistent broken image rather than browser-dependent broken image can be implemented throughout the frontend code i.e. a custom broken image if a user wants, and I can think of ways to make this—probably adding code to the |theme or |media twig filters to serve a custom broken image for any files that end with specific image-format extensions but those files do not exist in the file system, making the broken images uniform throughout the browser and user-specified than browser-specific. But again, this is upto OctoberCMS whether it wants to have such a feature or not.

mohsin:
@LukeTowers I honestly cannot think of specific use cases where I’d want custom broken images to appear in the frontend. But that is true even about them appearing in the backend i.e. without this OctoberCMS base64 BrokenImage svg code, the browser does render it’s own broken image too and should be sufficient enough yet it was OctoberCMS that offered this special red and white broken image and so far clients I’ve build for using OctoberCMS have liked the appearance over the typical broken image by the browser. Ofcourse, there have even been a request on whether this can be changed, which was the reason for even raising the original issue.

I imagine that this same website-consistent broken image rather than browser-dependent broken image can be implemented throughout the frontend code i.e. a custom broken image if a user wants, and I can think of ways to make this—probably adding code to the |theme or |media twig filters to serve a custom broken image for any files that end with specific image-format extensions but those files do not exist in the file system, making the broken images uniform throughout the browser and user-specified than browser-specific. But again, this is upto OctoberCMS whether it wants to have such a feature or not.

modules/backend/models/BrandSetting.php
LukeTowers: You should be returning the File object, not a path, and then calling methods on the File object to get its content through the flysystem abstraction

LukeTowers:
Personally I'm not a fan of the broken image, the only real purpose it serves is to stop the getThumb method from constantly trying to resize an image with a configuration that just won't work because it triggers an exception; since the getThumb method will always try to resize the image if the resized image doesn't exist.

@bennothommo @mjauvin @jaxwilko @robinbonnes do any of you have any thoughts on the broken image?

Samuell1:
@LukeTowers I think broken image should return 404 with maybe custom msg that image is broken, because if broken image gets cached in browsers or CDNs then its a big issue.

jaxwilko:
@LukeTowers I'm with @Samuell1 that a missing image should throw a 404. 

It would be nice if that was consistent across the frontend and backend. 

For the backend we could use the img onerror attribute or equivilent in jQuery and just set the src to the broken image (maybe add the original path to a data-original-src for debugging). 

Doing that would retain the nice backend feature while having a simpler and consistent image service.

modules/backend/models/brandsetting/fields.yaml
bennothommo: `.bmp` files should be dropped as well, I can't think of any uses for it.

bennothommo:
@LukeTowers I think the broken image functionality should be relegated to a plugin. I agree with @Samuell1 and @jaxwilko that they should, by default, just throw a 404. While I get that they were done to make it obvious that an image was missing, the default fallback image generally looks out of place and I tend to use our Image Resizer plugin to overwrite the fallback image to something that matches the brand of a particular site.

mohsin:
@LukeTowers I agree with @bennothommo that it makes better sense to have this in a separate plugin, so that users may opt-in to use custom broken image. Shall I go ahead with removing this fallback image code completely from even the backend?

As for the plugin, I can either make one thro' my own author page or you can choose to have it hosted via RainLab's github--you decide and let me know.

LukeTowers:
I'm going to close this PR, I'm not sure there's anything actionable here.

bennothommo:
@SaifurRahmanMohsin in regards to your question about removing the fallback image code, feel free to go ahead with a PR to remove it, and we'll review.

Samuell1:
@LukeTowers @bennothommo @SaifurRahmanMohsin I had yesterday issue when fallback image got cached on CDN and created fallback image on S3 and i needed to reupload whole image because system was thinking it still exists, but when i was checking in backend there was correct image for source. Not sure what happend but i think this functionality should be removed to prevent caching issues like i said before.

Issue #5065, opened 2020-05-05
Mail Templates Settings page can throw a fatal error


Flynsarmy:
- OctoberCMS Build: 465
- PHP Version: 7.4.5
- Database Engine: mysql 15.1

### Description:
If a plugin registers a mail partial, the partial is modified in Backend - Settings - Mail Templates, and the plugin is then disabled or removed, the mail templates page becomes unable to load, throwing the fatal exception
> Unable to find a registered partial with code: some.mail.partial

![Screen Shot 2020-05-05 at 3 10 46 pm](https://user-images.githubusercontent.com/334808/81036498-aa5a0200-8ee2-11ea-8bab-f42807e17f79.png)

Stack Trace:
|#|Called Code|Document|Line|
|--- |--- |--- |--- |
|77|System\Models\MailPartial->fillFromCode()|~/modules/system/Models/MailPartial.php|48|
|76|System\Models\MailPartial->afterFetch()|~/vendor/october/rain/src/Database/Model.php|168|
|75|October\Rain\Database\Model->October\Rain\Database\{closure}(…)|||
|74|call_user_func_array(…)|~/vendor/october/rain/src/Events/Dispatcher.php|233|
|73|October\Rain\Events\Dispatcher->dispatch(…)|~/vendor/october/rain/src/Events/Dispatcher.php|197|
|72|October\Rain\Events\Dispatcher->fire(…)|~/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Concerns/HasEvents.php|148|
|71|Illuminate\Database\Eloquent\Model->fireModelEvent(…)|~/vendor/october/rain/src/Database/Model.php|444|
|70|October\Rain\Database\Model->newFromBuilder(…)|~/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Builder.php|260|
|69|Illuminate\Database\Eloquent\Builder->Illuminate\Database\Eloquent\{closure}(…)|||
|68|array_map(…)|~/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Builder.php|259|
|67|Illuminate\Database\Eloquent\Builder->hydrate(…)|~/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Model.php|1477|
|66|Illuminate\Database\Eloquent\Model->__call(…)|~/vendor/october/rain/src/Extension/ExtendableTrait.php|428|
[...]

PR #5305, opened 2020-10-08
Purge orphaned mail templates/layouts/partials. 

mjauvin:
Purge Mail Templates when plugins are removed / disabled.

This prevents the following exception to be thrown when a mail templates registered by a plugin are no longer active:
```
if (! isset($this->hints[$segments[0]])) {
   throw new InvalidArgumentException("No hint path defined for [{$segments[0]}].");
}
```
ref. vendor/laravel/framework/src/Illuminate/View/FileViewFinder.php line 112

In order to replicate the problem, do the following:

1. Install a plugin that registers mail templates (e.g. RainLab.User).
2. Install RainLab.Translate.
3. Go to `Settings->Mail->Mail templates` page (this actually add the registered mail templates to the database).
4. Go to `Settings->System->Updates & Plugins` and disable the plugin installed in step1.
5. Go to `Settings->Translate->Translate messages` page.
6. Click on `Scan for messages`
7. Click on `Begin scan`

You should see the Exception mentioned above being thrown.

### DIFF
--- a/modules/system/classes/PluginManager.php
+++ b/modules/system/classes/PluginManager.php
@@ -632,6 +632,18 @@ class PluginManager
         if ($pluginObj = $this->findByIdentifier($code)) {
             $pluginObj->disabled = true;
         }
+
+        /**
+         * @event system.plugins.afterDisable
+         * Provides an opportunity to take actions after a plugin has been disabled.
+         *
+         * Example usage:
+         *
+         *     Event::listen('system.plugins.afterDisable', function ((String) $pluginCode) {
+         *         trace_log('Plugin ' . $pluginCode . ' has been disabled.');
+         *     });
+         *
+         */
         Event::fire('system.plugins.afterDisable', [$code]);
 
         return true;
@@ -663,6 +675,18 @@ class PluginManager
         if ($pluginObj = $this->findByIdentifier($code)) {
             $pluginObj->disabled = false;
         }
+
+        /**
+         * @event system.plugins.afterEnable
+         * Provides an opportunity to take actions after a plugin has been enabled.
+         *
+         * Example usage:
+         *
+         *     Event::listen('system.plugins.afterEnable', function ((String) $pluginCode) {
+         *         trace_log('Plugin ' . $pluginCode . ' has been enabled.');
+         *     });
+         *
+         */
         Event::fire('system.plugins.afterEnable', [$code]);
 
         return true;
@@ -863,6 +887,17 @@ class PluginManager
         $code = $this->normalizeIdentifier($id);
         unset($this->plugins[$code]);
 
+        /**
+         * @event system.plugins.afterRemove
+         * Provides an opportunity to take actions after a plugin has been removed.
+         *
+         * Example usage:
+         *
+         *     Event::listen('system.plugins.afterRemove', function ((String) $pluginCode) {
+         *         trace_log('Plugin ' . $pluginCode . ' has been removed.');
+         *     });
+         *
+         */
         Event::fire('system.plugins.afterRemove', [$code]);
     }
 
--- a/modules/system/classes/PluginManager.php
+++ b/modules/system/classes/PluginManager.php
@@ -633,19 +633,6 @@ class PluginManager
             $pluginObj->disabled = true;
         }
 
-        /**
-         * @event system.plugins.afterDisable
-         * Provides an opportunity to take actions after a plugin has been disabled.
-         *
-         * Example usage:
-         *
-         *     Event::listen('system.plugins.afterDisable', function ((String) $pluginCode) {
-         *         trace_log('Plugin ' . $pluginCode . ' has been disabled.');
-         *     });
-         *
-         */
-        Event::fire('system.plugins.afterDisable', [$code]);
-
         return true;
     }
 
@@ -675,19 +662,6 @@ class PluginManager
         if ($pluginObj = $this->findByIdentifier($code)) {
             $pluginObj->disabled = false;
         }
-
-        /**
-         * @event system.plugins.afterEnable
-         * Provides an opportunity to take actions after a plugin has been enabled.
-         *
-         * Example usage:
-         *
-         *     Event::listen('system.plugins.afterEnable', function ((String) $pluginCode) {
-         *         trace_log('Plugin ' . $pluginCode . ' has been enabled.');
-         *     });
-         *
-         */
-        Event::fire('system.plugins.afterEnable', [$code]);
 
         return true;
     }
--- a/modules/system/classes/PluginManager.php
+++ b/modules/system/classes/PluginManager.php
@@ -857,7 +857,7 @@ class PluginManager
         }
 
         // actually remove the plugin from our internal container
-        unset($this->plugins[$id]);
+        unset($this->plugins[ $this->normalizeIdentifier($id) ]);
     }
 
     /**
--- a/modules/system/classes/PluginManager.php
+++ b/modules/system/classes/PluginManager.php
@@ -844,21 +844,20 @@ class PluginManager
      */
     public function deletePlugin($id)
     {
-        $code = $this->normalizeIdentifier($id);
-
         /*
          * Rollback plugin
          */
-        UpdateManager::instance()->rollbackPlugin($code);
+        UpdateManager::instance()->rollbackPlugin($id);
 
         /*
          * Delete from file system
          */
-        if ($pluginPath = self::instance()->getPluginPath($code)) {
+        if ($pluginPath = self::instance()->getPluginPath($id)) {
             File::deleteDirectory($pluginPath);
         }
 
         // actually remove the plugin from our internal container
+        $code = $this->normalizeIdentifier($id);
         unset($this->plugins[$code]);
     }
 
--- a/modules/system/classes/PluginManager.php
+++ b/modules/system/classes/PluginManager.php
@@ -867,7 +867,7 @@ class PluginManager
          *
          * Example usage:
          *
-         *     Event::listen('system.plugins.afterRemove', function ((String) $pluginCode) {
+         *     Event::listen('system.plugins.afterRemove', function ((string) $pluginCode) {
          *         trace_log('Plugin ' . $pluginCode . ' has been removed.');
          *     });
          *
--- a/modules/system/classes/PluginManager.php
+++ b/modules/system/classes/PluginManager.php
@@ -855,6 +855,9 @@ class PluginManager
         if ($pluginPath = self::instance()->getPluginPath($id)) {
             File::deleteDirectory($pluginPath);
         }
+
+        // actually remove the plugin from our internal container
+        unset($this->plugins[$id]);
     }
 
     /**
--- a/modules/system/classes/PluginManager.php
+++ b/modules/system/classes/PluginManager.php
@@ -4,6 +4,7 @@ use Db;
 use App;
 use Str;
 use Log;
+use Event;
 use File;
 use Lang;
 use View;
--- a/modules/system/classes/PluginManager.php
+++ b/modules/system/classes/PluginManager.php
@@ -631,6 +631,7 @@ class PluginManager
         if ($pluginObj = $this->findByIdentifier($code)) {
             $pluginObj->disabled = true;
         }
+        Event::fire('system.plugins.afterDisable', [$code]);
 
         return true;
     }
@@ -661,6 +662,7 @@ class PluginManager
         if ($pluginObj = $this->findByIdentifier($code)) {
             $pluginObj->disabled = false;
         }
+        Event::fire('system.plugins.afterEnable', [$code]);
 
         return true;
     }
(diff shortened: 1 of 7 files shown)


modules/system/classes/PluginManager.php
```diff
@@ -855,6 +855,9 @@ public function deletePlugin($id)
         if ($pluginPath = self::instance()->getPluginPath($id)) {
             File::deleteDirectory($pluginPath);
         }
+
+        // actually remove the plugin from our internal container
+        unset($this->plugins[$id]);
```
LukeTowers: Is the ID correctly normalized at this point?

mjauvin: not necessarily, because it can be called from outside the class... I'll add normalization to make sure.

modules/system/classes/PluginManager.php
```diff
@@ -855,6 +855,9 @@ public function deletePlugin($id)
         if ($pluginPath = self::instance()->getPluginPath($id)) {
             File::deleteDirectory($pluginPath);
         }
+
+        // actually remove the plugin from our internal container
+        unset($this->plugins[ $this->normalizeIdentifier($id) ]);
```
LukeTowers: Would be better to normalize it at the start of the method to pass through to all the other calls

mjauvin: I didn't want to modify the current behavior, it's tricky code...

mjauvin: Let's see what the unit tests say...

mjauvin: Went well... there are no tests for deletePlugin()! :rofl: 

I'll try to add some tests/

mjauvin: @LukeTowers I'm having trouble writing unit tests for deletePlugin() as it introduces calls to the UpdateManager class (which in turn uses the VersionManager class) and I'm getting errors like the following:
```
1) PluginManagerTest::testDeletePlugin 
Illuminate\Database\QueryException: SQLSTATE[HY000] [1045] Access denied for user 'root'@'localhost' (using password: NO) (SQL: select * from `system_plugin_history` where `code` = October.Sample order by `id` asc) 
```

Do you mind if I revert to normalizing the code only for the unset() call to make sure that does not introduce any bugs, and we can revisit using the normalized code for the other calls in a different PR?

LukeTowers: Sure, that's fine

mjauvin:
All done here.

modules/system/classes/PluginManager.php
```diff
@@ -855,6 +855,10 @@ public function deletePlugin($id)
         if ($pluginPath = self::instance()->getPluginPath($id)) {
             File::deleteDirectory($pluginPath);
         }
+
+        // actually remove the plugin from our internal container
```
LukeTowers:
```suggestion
        // Remove the plugin from the internal container
```

modules/system/controllers/Updates.php
```diff
@@ -877,6 +880,12 @@ public function onBulkAction()
             }
         }
 
+        if (in_array($bulkAction, ['disable', 'remove'])) {
+            // purge orphaned mail templates
+            MailTemplate::syncAll();
```
LukeTowers: Upon further review, I think what we should probably do is fire an event after a plugin is disabled or removed and then listen to that event from the System ServiceProvider to handle this syncing.

LukeTowers:
@mjauvin could you provide replication instructions for the original issue for testing purposes?

mjauvin:
> @mjauvin could you provide replication instructions for the original issue for testing purposes?

done @LukeTowers

modules/system/controllers/Updates.php
mjauvin: Any suggestion on the event name @LukeTowers ? `system.plugins.pluginRemoved` ?

Should we fire a different event if it's removed (`system.plugins.pluginRemoved`) or if it's disabled (`system.plugins.pluginDisabled`) ?

Any suggestion on where to listen for those events in the ServiceProvider?

mjauvin: Or maybe just `system.plugins.afterUpdate` with the action that was done (freeze, unfreeze, disable, enable, refresh, remove) ?

mjauvin:
@LukeTowers I pushed changes that do what you suggested, once we agree on the event name and functionality, I'll complete with the doc block comments to document the event.

modules/system/controllers/Updates.php
LukeTowers: @mjauvin `system.plugins.afterDisable` & `system.plugins.afterRemove`

mjauvin: @LukeTowers You don't like what I commited?

LukeTowers: Not really, it's way too coupled with the controller / view implementation. The events should be fired from the PluginManager inside of the appropriate methods so that they're fired even if someone is programatically removing or disabling plugins from outside of the controller context.

mjauvin:
yeah, I wanted to avoid firing the event multiple time if many plugins are removed/disabled in bulk, but you're right, it's safer if done within PluginManager.

bennothommo (approved):
LGTM!

mjauvin:
@bennothommo you have no problem with the following ?
```
            if (!$this->app->runningUnitTests()) {
                MailTemplate::syncAll();
            }
```
I couldn't find why there's an error in `MailTemplate::syncAll()` when running the unit tests, so I skipped this when running unittests...

modules/system/classes/PluginManager.php
```diff
@@ -631,6 +632,7 @@ public function disablePlugin($id, $isUser = false)
         if ($pluginObj = $this->findByIdentifier($code)) {
             $pluginObj->disabled = true;
         }
+        Event::fire('system.plugins.afterDisable', [$code]);
```
LukeTowers: @mjauvin almost forgot, we need the inline event docblocks on these

mjauvin: done @LukeTowers !

bennothommo:
@mjauvin The PluginManager tests do not use the `PluginTestCase` class, which means that the database stubbing does not occur. I've tested this locally with the database stubbing and it completely fails. I realise that the `PluginManager` is set up to work gracefully without a DB (as it is purely based on the files in the plugins directory) anyway.

Your attached listener to the new event calls `MailTemplate::syncAll()`, which does require the DB, effectively making it so that a DB is required with PluginManager. Instead of disabling the call when in unit testing, I would instead perhaps disable the call if a DB doesn't exist, doing the following:

```php
if (!$this->app->hasDatabase()) {
    MailTemplate::syncAll();
}
```

mjauvin:
@bennothommo thanks for the explanation, although I still fail to understand why the test is running fine if I only run the PluginManagerTest like this:
```
./vendor/bin/phpunit tests/unit/system/classes/PluginManagerTest.php
```
And this fails:
```
./vendor/bin/phpunit tests/unit/system/classes/
```
Oh, and I reversed the logic in your snippet above... :wink:

mjauvin:
I found a problem with the new implementation with the events fired by the PluginManager... the MailTemplate::syncAll() is being called in places it shouldn't (e.g. when loading the plugins when we click on "Manage plugins" in the Updates controller...

modules/system/models/PluginVersion.php
```diff
@@ -116,6 +117,40 @@ public function afterFetch()
         }
     }
 
+    /**
+     * After the model is saved
+     */
+    public function afterSave()
+    {
+        if ($this->getOriginal('is_disabled') === 0 && $this->is_disabled === 1) {
```
LukeTowers: These comparison checks should probably be loose type comparisons, just in case some databases don't return those values as integers but as actual booleans.
```suggestion
        if (!$this->getOriginal('is_disabled') && $this->is_disabled) {
```

modules/system/models/PluginVersion.php
```diff
@@ -116,6 +117,67 @@ public function afterFetch()
         }
     }
 
+    /**
+     * After the model is saved
+     */
+    public function afterSave()
+    {
+        if (!$this->getOriginal('is_disabled')  && $this->is_disabled) {
```
LukeTowers:
```suggestion
        if (!$this->getOriginal('is_disabled') && $this->is_disabled) {
```

modules/system/models/PluginVersion.php
```diff
@@ -116,6 +117,67 @@ public function afterFetch()
...
+             * Example usage:
+             *
+             *     Event::listen('system.plugins.afterEnable', function ((String) $pluginCode) {
+             *         trace_log('Plugin ' . $pluginCode . ' has been enabled.');
+             *     });
+             *
+             */
+            Event::fire('system.plugins.afterEnable', [$this->code]);
+        }
+        if (!$this->getOriginal('is_frozen')  && $this->is_frozen) {
```
LukeTowers:
```suggestion
        if (!$this->getOriginal('is_frozen') && $this->is_frozen) {
```

modules/system/classes/PluginManager.php
```diff
@@ -855,6 +856,23 @@ public function deletePlugin($id)
...
+        $code = $this->normalizeIdentifier($id);
+        unset($this->plugins[$code]);
+
+        /**
+         * @event system.plugins.afterRemove
+         * Provides an opportunity to take actions after a plugin has been removed.
+         *
+         * Example usage:
+         *
+         *     Event::listen('system.plugins.afterRemove', function ((String) $pluginCode) {
```
LukeTowers:
```suggestion
         *     Event::listen('system.plugins.afterRemove', function ((string) $pluginCode) {
```

modules/system/models/PluginVersion.php
```diff
@@ -116,6 +117,28 @@ public function afterFetch()
         }
     }
 
+    /**
+     * After the model is saved
+     */
+    public function afterSave()
+    {
+        // The plugin has been disabled.
```
LukeTowers:
```suggestion
        // Detect if the plugin has been disabled.
```

modules/system/models/PluginVersion.php
```diff
@@ -116,6 +117,28 @@ public function afterFetch()
         }
     }
 
+    /**
+     * After the model is saved
```
LukeTowers:
```suggestion
     * Fired after the model is saved
```

modules/system/models/PluginVersion.php
```diff
@@ -116,6 +117,28 @@ public function afterFetch()
...
+    {
+        // Detect if the plugin has been disabled.
+        if (!$this->getOriginal('is_disabled') && $this->is_disabled) {
+            /**
+             * @event system.plugins.afterDisable
+             * Provides an opportunity to take actions after a plugin has been disabled.
+             *
+             * Example usage:
+             *
+             *     Event::listen('system.plugins.afterDisable', function ((String) $pluginCode) {
```
LukeTowers:
```suggestion
             *     Event::listen('system.plugins.afterDisable', function ((string) $pluginCode) {
```

mjauvin:
Anything missing here @LukeTowers ?

LukeTowers:
@mjauvin is there a related issue for this?

Also would the system benefit from syncing mail templates on plugin install / enable?

mjauvin:
@LukeTowers I'll try to find issues related to this.

As for syncing at install/enable time, there is no need. Whenever the `Mail Templates` page get accessed, the sync will happen.

In the other cases (disable/remove), the "orphaned" templates were causing exception, which won't be the case when the plugin is installed/enabled.

mjauvin:
@LukeTowers 
Fixes: https://github.com/rainlab/translate-plugin/issues/513

LukeTowers:
@mjauvin I know you've spent a lot of time on this, but I've just realized this is for the Translate plugin:
> Go to Settings->Translate->Translate messages page.
> Click on Scan for messages
> Click on Begin scan

I thought this was related to https://github.com/octobercms/october/issues/5065. How come the translate plugin message scanner couldn't be fixed to ignore this exception and just skip over invalid templates?

mjauvin:
@LukeTowers this is triggered by the `MailTemplate::allTemplates()` call in this function:
```
    public function scanMailTemplatesForMessages()
    {   
        $messages = [];

        foreach (MailTemplate::allTemplates() as $mailTemplate) {
            $messages = array_merge($messages, $this->parseContent($mailTemplate->subject));
            $messages = array_merge($messages, $this->parseContent($mailTemplate->content_html));
        }

        Message::importMessages($messages);
    }
```
You want me to wrap the foreach with a try/catch to fix the issue? Isn't that a bit of a workaround? What about all other plugins that could call this method?

mjauvin:
@LukeTowers it cannot be fixed in RainLab.Translate, but can be fixed like this:
```
diff --git a/modules/system/models/MailTemplate.php b/modules/system/models/MailTemplate.php
index 9940b271..6ec9cceb 100644
--- a/modules/system/models/MailTemplate.php
+++ b/modules/system/models/MailTemplate.php
@@ -68,7 +68,11 @@ class MailTemplate extends Model
         $codes = array_keys(self::listAllTemplates());
 
         foreach ($codes as $code) {
-            $result[] = self::findOrMakeTemplate($code);
+            try {
+                $result[] = self::findOrMakeTemplate($code);
+            } catch (\InvalidArgumentException $e) {
+                // skip orphaned templates for disabled/removed plugins
+            }
         }
    
         return $result;
```

mjauvin:
Closed in favor of #5322

Issue #5301, opened 2020-10-08
Form setting for Input.trigger.js does not accept null


Rike-cz:
When in your form setting is something like this:
```
trigger:
    action: disable
    field: common_text_or_number_field
    condition: value[]
```
...everything works as expected. `""` value is compared with empty text field `""` and result is **true**.

But in this case:
```
trigger:
    action: disable
    field: dropdown_field
    condition: value[]
```
... `""` is compared with `null` returned by dropdown and result is **false**.

I did not find way how to impose `null` to form field condition setting.

### Possible solution
When Input.trigger.js `self.triggerConditionValue` contains `""`, add automatically `null` value too.

LukeTowers:
@Rike-cz do you think you could submit a PR to https://github.com/octobercms/october/blob/develop/modules/system/assets/ui/js/input.trigger.js#L67-L91 to handle dropdown null values better?

Rike-cz:
@LukeTowers I found very simple way to achieve what I want.

What is another problem - automatically changed dropdowns (via dependsOn) does not trigger any event. It's more complex I think, maybe for @bennothommo eyes and brain.
![trigger](https://user-images.githubusercontent.com/4247582/95553254-72fdfd00-0a0e-11eb-90c3-1b33929975fc.gif)

dennisgon:
hi, can you assign this to me? i
I want to try to solve the problem

bennothommo:
I think a `null` for a dropdown implies nothing is selected, which might be why it's set up to apply the filter that way (ie. I want the trigger to occur if nothing is selected). Does the same thing occur if you specifically include a blank string in the value - ie. `condition: value['']`? Not sure if that will work, but worth a go.

LukeTowers:
@Rike-cz can't you set the price field to automatically trigger based on the product option field as well as the option value field?

Rike-cz:
@LukeTowers If every _Product option_ has one or more _Option value_, yes, I can. But in marginal case with no option values for product option I will optain enabled _Price_ too. But it's a detail I think.

Rike-cz:
If @bennothommo would make some changes in dropdowns, I think this is not correct behavior too:
![loading](https://user-images.githubusercontent.com/4247582/95829678-d81b6080-0d36-11eb-81e7-eea58dcba816.gif)
Changing of value in dropdown fires loading indicators not only in his own dependents.

bennothommo:
[...]

PR #5306, opened 2020-10-09
Enabled null as condition value when empty in Input.trigger.js

Rike-cz:
#5301


mjauvin:
@Rike-cz I've setup an Octodock instance with October/test-plugin installed so you can setup a test-case for this so we can test this PR in a real situation.

Please let me know if you can make the required changes and we'll test this.

https://bf5598f85650.octodock.com/backend
Note: user/pass = admin/admin as usual

Rike-cz:
Please forget it. After upgrade to 1.1 it works as expected without my PR. I am sorry for making problems :-(

PR #5312, opened 2020-10-14
Change the max length of login and email

gergo85:
The maximum length of fields are 191 in the database.
db


LukeTowers:
You should probably take the logic from https://github.com/octobercms/october/blob/develop/modules/system/ServiceProvider.php#L586-L604 and create a new helper method `public static function getDbDefaultStringLength()` with the logic that identifies the default string length based on the current DB and the code in the linked method, and then dynamically set the max length of those strings based on the results of the new method.

PR #5325, opened 2020-10-22
Changing install plugins/themes buttons order

TimFoerster:
Change order to plugin -> theme, like the install producs page to reduce missclicks.

image

image


SebastiaanKloos:
I still click the wrong button all the time 🙈

TimFoerster:
I can feel the pain. For me it is like 95%

PR #5326, opened 2020-10-23
Add computed value columns in data-tables

bennothommo:
This PR adds a new `computed` type for columns in the data-table widget. The main purpose of this new column type is to provide "computed" values based on the rest of a row's data, the result of which is fetched by an AJAX request to the form's controller. Computed values are generally informational in nature, to provide further context or calculations for a given record.

In the example below, we use the Rating and Type columns to derive a "Danger Index" value, which is the result of a `onGetDangerComputedValue` request call to the form controller.

computed-values

This could also be used for things like totals or costings of an individual record.

### Usage

Following the configuration for [setting up data tables](https://octobercms.com/docs/backend/forms#widget-datatable), you can add a computed column by setting the type to `computed`.

A computed column accepts two additional options:
 - `handler` - Defines the AJAX handler that the result will be retrieved from. The AJAX handler will receive a `rowData` variable in POST which will contain an array - column names for keys and column values for values. If left blank, this will use a default callback of `onGet(column name in studly-caps)ComputedValue` - for example, for a `name` column, it would be `onGetNameComputedValue`.
- `dependsOn` - Defines the columns that will trigger an update when changed. It can be either a string for a single column, or array for multiple columns.

Computed columns will retrieve a value immediately on creation, if a value does not exist for that record.

### Example

https://github.com/octoberrain/test-plugin/pull/89

### Further tasks
 
- Exclude computed values from the final POST data for the data-table. (or make this optional?)
- Testing

### DIFF
--- a/modules/backend/ServiceProvider.php
+++ b/modules/backend/ServiceProvider.php
@@ -69,6 +69,7 @@ class ServiceProvider extends ModuleServiceProvider
             $combiner->registerBundle('~/modules/backend/assets/less/october.less');
             $combiner->registerBundle('~/modules/backend/assets/js/october.js');
             $combiner->registerBundle('~/modules/backend/widgets/table/assets/js/build.js');
+            $combiner->registerBundle('~/modules/backend/widgets/table/assets/less/table.less');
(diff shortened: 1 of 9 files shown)


mjauvin:
Nice one @bennothommo!

Issue #5300, opened 2020-10-07
Wrong var type


davodsaraei:
In this line, The `$config` variable defined as an array but used as an object

https://github.com/octobercms/october/blob/af6db51b7978aae874844e79e0103b0952d7b762/modules/backend/classes/ControllerBehavior.php#L26

LukeTowers:
@davodsaraei feel free to submit a PR, make the change here too please: https://github.com/octobercms/october/blob/af6db51b79/modules/system/traits/ConfigMaker.php#L30

SaidiAli:
@davodsaraei I think it's intended to be used as an object. The array is turned into an object in this method; https://github.com/octobercms/october/blob/af6db51b79/modules/system/traits/ConfigMaker.php#L30

davodsaraei:
@SaidiAli @LukeTowers It will be changed to object but I have another issue when I run the `composer install`.
https://github.com/octobercms/october/issues/5302#issuecomment-705536788

LukeTowers:
@davodsaraei see my comment on that.

PR #5330, opened 2020-10-24
Fix for #5300 - Wrong type hinting for config variable

shahlin:
There were 2 occurrences of incorrect type hinting for the variable `$config`. It was declared as an array while it was being used as an object. These are the 2 occurrences:

https://github.com/octobercms/october/blob/af6db51b7978aae874844e79e0103b0952d7b762/modules/backend/classes/ControllerBehavior.php#L26

https://github.com/octobercms/october/blob/af6db51b7978aae874844e79e0103b0952d7b762/modules/system/traits/ConfigMaker.php#L30

This PR is an attempt to fix issue #5300

### DIFF
--- a/modules/backend/classes/ControllerBehavior.php
+++ b/modules/backend/classes/ControllerBehavior.php
@@ -23,7 +23,7 @@ class ControllerBehavior extends ExtensionBase
     }
 
     /**
-     * @var array Supplied configuration.
+     * @var \stdClass Supplied configuration.
      */
     protected $config;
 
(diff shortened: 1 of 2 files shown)


LukeTowers:
@bennothommo we need to lock the test suite to running composer 1.0 until https://github.com/wikimedia/composer-merge-plugin/pull/189 is merged. Ideally we should probably help test that PR and get it merged in ASAP as composer 2.0 is now out.

bennothommo:
@LukeTowers I've updated the test suite for October and Rain library - I'll need to do the same for the RainLab plugins that have tests at some point later today.

LukeTowers:
Please resubmit if desired, not sure why the tests aren't liking this one.

PR #5334, opened 2020-10-26
Theme:sync path pattern fix

hackolein:
When synchronizing files to the database, there is a problem with matching the paths to be synchronized because regex does not work as expected.

I have executed the following command:

`artisan theme:sync --target=database --force --paths=layouts/,pages/,partials/`

But at the moment it also synchronizes paths that should not be synchronized. For example: `content/static-pages/`. Because here the pattern applies to `pages/`.

The fix is that the pattern starts at the beginning of the line.


LukeTowers:
To be perfectly honest, this functionality was supposed to match paths exactly, i.e. manual specification of individual files, not overall paths, but I'm fine having it be an undocumented goody that it works like that.

PR #5338, opened 2020-10-29
Italian translations improvement

ArchimedeSolutions:
- Improved translation in backend\lang\it\lang.php
- Added some missing entries in system\lang\it\client.php


bennothommo:
@ArchimedeSolutions would you mind re-adding the trailing commas on the last items in the arrays? We keep them as it makes it cleaner for code diff purposes.

ArchimedeSolutions:
Yes, no problem.. just re-added

PR #5344, opened 2020-11-03
Add support for defining quick actions in the Backend's main nav

bennothommo:
This PR adds support for defining quick actions in the main navigation in the Backend. Quick actions in this case are defined as the single-purpose icon links located to the left of the user menu - the only one currently available in an October CMS install is the link to preview the website.

image

Plugins now have the ability to define quick actions through a `registerQuickActions` method, which follows the same configuration as the `registerNavigation` method. It is still recommended and preferred that most plugin functionality be defined in their own main menu items, but this will allow a plugin to easily define a shortcut (or remove one).

```php
/**
 * Registers back-end quick actions for this plugin.
 *
 * @return array
 */
public function registerQuickActions()
{
    return [
        'help' => [
            'label' => 'rainlab.docs::lang.links.docsLink',
            'icon' => 'icon-question-circle',
            'url' => Backend::url('docs'),
        ],
    ];
}
```

As with the normal navigation, the quick actions can also be added or removed programatically through the `NavigationManager::registerCallback()` method, or through the `backend.menu.extendItems` event. The following methods will manage these functions:

- `addQuickActionItems`
- `removeQuickActionItem`
- `listQuickActionItems`

One additional change to this PR has been included - the Preview link that was previously hard-coded in this area has been converted to a quick action provided by the CMS module. If the CMS module is disabled or uninstalled, for instance with a contained application, the preview link is effectively removed.

### DIFF
--- a/modules/backend/classes/NavigationManager.php
+++ b/modules/backend/classes/NavigationManager.php
@@ -28,6 +28,11 @@ class NavigationManager
      */
     protected $items;
 
+    /**
+     * @var QuickActionItem[] List of registered quick actions.
+     */
+    protected $quickActions;
+
     protected $contextSidenavPartials = [];
 
     protected $contextOwner;
@@ -54,6 +59,9 @@ class NavigationManager
      */
     protected function loadItems()
     {
+        $this->items = [];
+        $this->quickActions = [];
+
         /*
          * Load module items
          */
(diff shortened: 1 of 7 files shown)


seanthepottingshed:
Beautiful.

modules/backend/classes/NavigationManager.php
```diff
@@ -183,10 +202,6 @@ public function registerCallback(callable $callback)
      */
     public function registerMenuItems($owner, array $definitions)
     {
-        if (!$this->items) {
-            $this->items = [];
-        }
```
LukeTowers, line 188: Why was this removed?

LukeTowers (changes requested):

modules/backend/classes/NavigationManager.php
bennothommo: @LukeTowers it's not needed, I don't believe. The `loadItems` method resets the `items` and `quickActions` properties to empty arrays, and that's fired before any callbacks or events are processed.

LukeTowers:
Awesome work @bennothommo! Are we going to document this or leave it undocumented because you really wouldn't want a ton of these added by various plugins?

seanthepottingshed:
Ha ha, shiny new thing - must use it!

bennothommo:
@LukeTowers let's leave it undocumented for now until we add some way of collapsing the quick actions or presenting them in a better fashion if people go absolutely crazy with them.

PR #5348, opened 2020-11-09
fix themes localization namespace

mjauvin:
Namespace was using "theme.*" instead of "themes.*"

Related: #4960


mjauvin:
There you go @LukeTowers.

Documentation was already referencing this properly.

PR #5349, opened 2020-11-10
Move theme localization to the boot process of the CMS Service Provider

bennothommo:
Follow-up to https://github.com/octobercms/october/pull/4960.

It appears that doing this in the "register" process can result, at times, in an exception to be thrown as the connection to the DB is not yet established.

image

This seemed to occur for me only on signing in to the Backend, and also seemed to be intermittent - sometimes it worked.

Moving it to the `boot` method in the CMS service provider makes it more stable.


mjauvin:
@bennothommo this still works for me when done in the boot method.

PR #5352, opened 2020-11-13
Add max files option to fileupload

RomainMazB:
Hi,
this PR just gives the ability to limit the number of fileupload using the native DropZone.js option directly from the yaml:
```yaml
photos:
    label: Photos
    commentAbove: Maximum 3 files
    type: fileupload
    maxFiles: 3
```
render this when you try to upload more than 3 files:
image
with an clear error message when you click on it:
image

I didn't find the way to translate the error message.
When switching to other language, the string `You can not upload any more files.` is not translated (I didn't expect it to of course). *If someone can tell me how to do* :+1: 

I'll update the doc when/if this PR is merged.

### DIFF
--- a/modules/backend/formwidgets/FileUpload.php
+++ b/modules/backend/formwidgets/FileUpload.php
@@ -65,6 +65,11 @@ class FileUpload extends FormWidgetBase
     public $maxFilesize;
 
     /**
+     * @var mixed Max files number.
+     */
+    public $maxFiles;
+
+    /**
      * @var array Options used for generating thumbnails.
      */
     public $thumbOptions = [
@@ -109,6 +114,7 @@ class FileUpload extends FormWidgetBase
             'imageHeight',
             'fileTypes',
             'maxFilesize',
+            'maxFiles',
             'mimeTypes',
             'thumbOptions',
             'useCaption',
@@ -159,6 +165,7 @@ class FileUpload extends FormWidgetBase
         $this->vars['cssDimensions'] = $this->getCssDimensions();
         $this->vars['cssBlockDimensions'] = $this->getCssDimensions('block');
         $this->vars['useCaption'] = $this->useCaption;
+        $this->vars['maxFiles'] = $this->maxFiles;
         $this->vars['prompt'] = $this->getPromptText();
     }
 
--- a/modules/backend/formwidgets/FileUpload.php
+++ b/modules/backend/formwidgets/FileUpload.php
@@ -65,7 +65,7 @@ class FileUpload extends FormWidgetBase
     public $maxFilesize;
 
     /**
-     * @var mixed Max files number.
+     * @var integer|null Max files number.
      */
     public $maxFiles;
 
--- a/modules/backend/formwidgets/FileUpload.php
+++ b/modules/backend/formwidgets/FileUpload.php
@@ -158,14 +158,14 @@ class FileUpload extends FormWidgetBase
         $this->vars['singleFile'] = $fileList->first();
         $this->vars['displayMode'] = $this->getDisplayMode();
         $this->vars['emptyIcon'] = $this->getConfig('emptyIcon', 'icon-upload');
-        $this->vars['imageHeight'] = $this->imageHeight;
-        $this->vars['imageWidth'] = $this->imageWidth;
+        $this->vars['imageHeight'] = (is_int($this->imageHeight)) ? $this->imageHeight : null;
+        $this->vars['imageWidth'] = (is_int($this->imageWidth)) ? $this->imageWidth : null;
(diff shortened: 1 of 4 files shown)


LukeTowers:
I feel like this is the case where it's better to perform this sort of validation by evaluating the data in beforeSave() and throwing an ApplicationException when failing. Thoughts?

RomainMazB:
That's not my thoughts:
performing this in `beforeSave` - *in addition to the fact that we re-invent a wheel which is already implemented in dropzone.js* - is not so beneficial for the UX imo because it will process all the files upload to the server before telling the user he can't do that.

Using directly the dropzone doesn't process the unnecessary file upload (as you probably guessed):
image

And I've not looked into that way but I feel like it would be more complex due to the fact that each files are processed independently.

This means we would have on each independent upload to look into the database to see how many file have already been uploaded and attached or deferred to the model. This would add another unnecessary load time.

And without actually storing the limit-exceeded files in the filesystem & database, we are not able to make a user-friendly feedback on which files has not been uploaded, we could just say "hmm, some file has not been uploaded, guess which ones :]".

And - *I may be totally wrong on this one* - because the file are processed independently, the multiple Ajax calls would return many error flash message which would be displayed. The files are queued before upload, so if we upload 20 files where the limit only allows 3, the user will annoyingly receive many untimely flash messages.

LukeTowers:
Fair enough, those are all good points. I would still recommend that it clearly states in the documentation that this limit should be validated serverside as well (maybe also provide an example) because I don't want people to expect that setting the maxFiles property will guarantee that they never will get any more files than that because all client side validation is trivially bypassed. That way developers can evaluate their needs and decide whether the limit is an absolute necessity which should be validated serverside or if it's more of a soft limit for UX that they can get away with not validating.

RomainMazB:
I'll write the doc with your rules, of course. And I will search for a way to validate the widget with a one shot validation, querying a files count from the database only on form submission.

The files will still be uploaded on the server,  it's still *"not safe"*, I agree with you on that. But the model won't be attached to the files so the files will be deleted when the deferred bindings that have not been committed are cleared.

I've tested more and more my PR and found the bug I've just fixed in the last commit:
Due to the fact that the already uploaded files was directly rendered with the page without passing through the javascript widget, the Dropzone files list was considered as empty. So the max files limits was reseated by a basic refresh on the form.

I've fixed it using "only" the javascript template:
1. Twig creates empty div containing the needed data
2. Dropzone walk through these divs:
    - Add the file to it files list so its internal counter is always sync to the reality
    - Generates and render the template for each file
    - Remove the data container div (optional, just to keep the DOM cleaner for eventual future DOM-manipulation)

When the limit is reached, the upload button is now disabled and doesn't trigger the browser's file explorer.
When a file is deleted and the limit is greater than the files still present in the file list, the button is reactivated.

I've tested all the frontend parts: popup, title & caption update, file deletion, caption deactivation, and all of that in photo and file mode, all works great now.

modules/backend/formwidgets/FileUpload.php
```diff
@@ -159,6 +165,7 @@ protected function prepareVars()
         $this->vars['cssDimensions'] = $this->getCssDimensions();
         $this->vars['cssBlockDimensions'] = $this->getCssDimensions('block');
         $this->vars['useCaption'] = $this->useCaption;
+        $this->vars['maxFiles'] = $this->maxFiles;
```
bennothommo: @RomainMazB we'll need to make sure this is an integer (or null)
```suggestion
        $this->vars['maxFiles'] = (is_int($this->maxFiles)) ? $this->maxFiles : null;
```

bennothommo (changes requested):

modules/backend/formwidgets/FileUpload.php
RomainMazB: I agree with you, that's better but why only maxFiles and not on maxFilesize, imageHeight or imageWidth?

bennothommo: @RomainMazB I only commented on your actual changes - if you want to add it on the other properties too, go right ahead :)

RomainMazB: Done, and added to all the numeric options :+1:

modules/backend/formwidgets/FileUpload.php
```diff
@@ -64,6 +64,11 @@ class FileUpload extends FormWidgetBase
      */
     public $maxFilesize;
 
+    /**
+     * @var mixed Max files number.
```
LukeTowers:
```suggestion
     * @var integer|null Max files number.
```

bennothommo:
Looks fine to me - needs to be tested first.

daftspunk:
Thanks @RomainMazB!

PR #5357, opened 2020-11-20
PHP 8 compatibility

bennothommo:
This is an open PR towards PHP 8 compatibility in October CMS.

It is not recommended to use October CMS in a PHP 8 environment at this time. Please stick to PHP 7.3 and 7.4, which are the currently supported PHP versions of October CMS that are still receiving updates from the PHP development team.

Rain Library PR: https://github.com/octobercms/library/pull/558

Big thanks to @tolkamsoft for their assistance with this so far.


ayumi-cloud:
Interesting! 

I seen you wrote `phpVersion: '8.0'` 

I've seen other repo's write this: `phpVersion: 'nightly'` instead, for php 8.x

@bennothommo When you free, could you tell me why all the pull requests are failing here? (I see the last few pr's are all failing with the same issue).

marcsoler:
PHP 8 is oficially out and as far as I understand, doctrine/dbal latest version (3.0) now supports PHP 8. [october/rain](octobercms/library) currently requires dbal 2.6.

bennothommo:
@ayumi-cloud `nightly` would be PHP 8.1 at this point, and given that GitHub Actions doesn't have an "allow failure" type of action, we'd probably constantly get failures if we tracked `nightly`. I'd rather us just focus our efforts on the supported versions, as opposed to chasing a white unicorn like `nightly` releases.

ayumi-cloud:
Yah good point about `8.1` didn't think of that.

`white unicorn releases` haha made me laugh.

bennothommo:
image

iotch:
Seems like PHP 8.0 tests are failing on `Doctrine\DBAL\Driver\PDOConnection::query()` however Doctrine DBAL is compatible with php 8 and just needs to be updated to v2.12.

bennothommo:
@LukeTowers this is ready for final review. Unit tests appear to be working with PHP 8, but I haven't yet tested October directly on PHP 8.

composer.json
```diff
@@ -51,8 +51,7 @@
         "fzaninotto/faker": "~1.9",
         "squizlabs/php_codesniffer": "3.*",
         "php-parallel-lint/php-parallel-lint": "^1.0",
-        "meyfa/phpunit-assert-gd": "^2.0.0",
```
LukeTowers, line 54: I'm assuming this is removed because it's included in the library test suite?

composer.json
```diff
@@ -82,10 +81,7 @@
         ]
     },
     "config": {
-        "preferred-install": "dist",
-        "platform": {
-            "php": "7.2.5"
```
LukeTowers, line 87: This is to prevent differences between the marketplace delivered version and the composer one, but perhaps it's finally been fixed on the marketplace side by @daftspunk?

modules/backend/classes/BackendController.php
```diff
@@ -111,6 +111,14 @@ public static function extend(Closure $callback)
         self::extendableExtendCallback($callback);
     }
 
+    /**
+     * @inheritDoc
+     */
+    public function callAction($method, $parameters)
+    {
+        return parent::callAction($method, array_values($parameters));
+    }
```
LukeTowers, line 120: Why was this required? Has Laravel not fixed it on their end?

modules/backend/classes/BackendController.php
```diff
@@ -210,7 +218,7 @@ protected function getRequestedController($url)
          * Look for a Plugin controller
          */
         if (count($params) >= 2) {
-            list($author, $plugin) = $params;
+            [$author, $plugin] = $params;
```
LukeTowers, line 221: Has list() been removed or is this just a coding style preference?

LukeTowers (changes requested):

composer.json
bennothommo: Correct :)

modules/backend/classes/BackendController.php
iotch: This is just coding style applied by IDE, sorry

Tolkam-PHP-Components: This is just coding style applied by IDE, sorry

modules/backend/classes/BackendController.php
Tolkam-PHP-Components: I believe this is due to Laravel is using array destructing and in php 8 array keys should match parameter names during destructing, so 0, 1, etc. are not valid parameters names in this case.

PR #5358, opened 2020-11-22
add "disabled" attribute to dropdown options

mjauvin:
```php
public function getStatusOptions()
{   
    return [
        'regular' => 'Regular, no icon',
        'disabled' => ['Disabled Example with folder icon', 'disabled' => true, 'icon-folder'],
        'regular-language-icon' => ['regular with language icon', 'icon-language'],
    ];
}
```
result:
Screenshot from 2020-11-21 23-43-57


LukeTowers:
Out of curiosity, why would you disable an option?

mjauvin:
@LukeTowers This can be usefull when you still want to let users know this IS an option but is currently unavailable to select.

LukeTowers:
@mjauvin could you give me a more concrete example of when you'd use that? The disabled attribute doesn't give any serverside protection against the value being selected anyways, so I'm unsure in what real world case this feature would actually be used.

Rike-cz:
Little off topic. Why is not icon declaration with key too, and why is not documented? I didn't find it in docs.

PR #5361, opened 2020-11-23
use new October\Rain Mail facade

mjauvin:


mjauvin:
I guess `vendor/october/rain/src/Support/Facades/Mail.php` won't be available until it's part of a release...

bennothommo:
@mjauvin I've gotta refactor the tests a bit due to the `DbTestCase` added to the library, at the same time, I'll change it around so that the `develop` branch of the library is included, so it should also pick up your Mail facade. I'll implement this fix afterwards.

PR #5363, opened 2020-11-25
Add ability to configure whether or not backend users must be activated before they can login

LeoCantThinkOfAName:
The original activation function was half implemented. So I add an extra config option, and wire it into `AuthManager` initialization phase.

This change may open some other possible to further implement new functions like register as backend user, or create an alternative way to block certain user beside deleting the record.

### DIFF
--- a/config/auth.php
+++ b/config/auth.php
@@ -36,4 +36,5 @@ return [
         'suspensionTime' => 15,
     ],
 
+    "requireActivation" => false,
 ];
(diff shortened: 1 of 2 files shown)


bennothommo:
@LeoCantThinkOfAName is there actually a way that Backend users can be activated though? If the intention is for a plugin to provide this functionality, then this new setting should form part of the plugin instead.

daftspunk:
This doesn't have a clear use case. Closing for now

Issue #2986, opened 2017-07-21
Repeater Index


shoguniphicus:
##### Expected behavior
Repeater's index should be the same whether it's record derived from database or from click "+".

##### Actual behavior
If there is already record on any repeater item, the _repeater_index always start with 0. However, if user click "+" to add a new repeater item on an empty list, the _repeat_index always start with 1.

##### Reproduce steps
Click on "+" on any repeater type form field, and check out the _repeater_index, it's always starting with 1. 

This might not be an issue for most that don't build custom javascript code for backend. But it's inconsistent that the index is different each time.

Right now, I changed the formwidget file at /modules/backend/formwidgets/Repeater.php at line 56, change to -1 and it's behaving as expected.

##### October build
419

LukeTowers:
Closing as it has been over a month since any activity on this occurred.

If you'd still like this fixed @shoguniphicus please make a PR with a fix.

mohsin:
@LukeTowers Is there any provision by which I can specify the label to show the line number in a repeater field?

```
lines:
   label: Line
   type: repeater
   style: accordion
   maxItems: 3
   form:
       fields:
           line:
               label: Line :index
```

LukeTowers:
@SaifurRahmanMohsin not sure exactly what you mean?

bennothommo:
@LukeTowers I think he means having some sort of placeholder that could be replaced by the repeater item's index. ie. Line 1 for the first repeater item, Line 2 for the second, etc.

@SaifurRahmanMohsin I don't think there's any provision for that currently, but sounds like a great idea for a PR.

mohsin:
Yep, exactly as @bennothommo said. I see there is a provision like this in [the RecordFinder](https://github.com/octobercms/october/blob/f8f5e6e02273c81488542e51be50b5baed7a8af2/modules/backend/formwidgets/RecordFinder.php#L223) where the prompt text can hold a placeholder that's replaced by a search icon. Was looking for something of that sort in repeater for it's index.

PR #5377, opened 2020-12-02
(feat) Repeater Index Placeholder

mohsin:
Added ability to specify an index in the repeater items label by specifying an :index placeholder in the string [as mentioned here](https://github.com/octobercms/october/issues/2986#issuecomment-737020996).

### DIFF
--- a/modules/backend/formwidgets/Repeater.php
+++ b/modules/backend/formwidgets/Repeater.php
@@ -300,6 +300,13 @@ class Repeater extends FormWidgetBase
         $config->alias = $this->alias . 'Form' . $index;
         $config->arrayName = $this->getFieldName().'['.$index.']';
         $config->isNested = true;
+        $config->fields = array_map(function ($item) use ($index) {
+            if (!empty($item['label'])) {
+                $item['label'] = str_replace(':index', $index + 1, e(trans($item['label'])));
+            }
+            return $item;
+        }, $config->fields);
+
         if (self::$onAddItemCalled || $this->minItems > 0) {
             $config->enableDefaults = true;
         }


LukeTowers:
Please provide a screenshot of this in action.

mohsin:
@LukeTowers This is the form config I used:
```
address_line:
    label: Address Line
    type: repeater
    form:
        fields:
            line:
                label: "Address Line #:index"
```
And the result:

Screenshot 2020-12-03 at 1 20 07 AM

LukeTowers:
Hmm, not super convinced this is worth adding, seems a bit too specific for this use case. Can you think of other use cases where this would be useful? Also, if we do add it I would prefer we add it through the translation system, so `trans($label, ['index' => $index])`

mohsin:
That's your call. I use this for my own plugin's repeater, and made a PR here only because @bennothommo said "sounds like a great idea for a PR". You guys can hash it out on whether it's worth adding to OctoberCMS or not.

LukeTowers:
@bennothommo can you think of some other instances where this would be useful?

summercms:
Why change the core code?

Two options

1. Use javascript to change the dom.

2. Create a custom widget that replaces Address Line text field and then add your custom code (recommended option).

bennothommo:
@LukeTowers I've had clients mention sometimes that they get lost on which repeater item they're editing if there's a big list of items in the repeater. An index would give them some clarity, even if it is arbitrary. It could also be used to populate values in the fields, almost like an autoincrement, however, this could be done on the Twig side too.

I admit this is probably situational, and as @ayumi-cloud mentioned, could be done through JS as well.

LukeTowers:
@bennothommo that use case is already covered by the `titleFrom` property for large repeaters. If someone else can come up with a valid use case for this that is more likely to be of benefit for a greater user base than the currently proposed use cases then I'm more than happy to re-open and look at this again. Otherwise I'm just going to close this for now. @SaifurRahmanMohsin thanks for taking the time to submit this!

PR #5379, opened 2020-12-03
ImageResizer fixes

mjauvin:
- missing "string" in width/height accepted params types
- fix config key names (only local disk was being used)


LukeTowers:
Also fixes https://github.com/octobercms/october/issues/5405

PR #5383, opened 2020-12-04
Take extension from options into account

mjauvin:
Without this, the temporary path given to the resizer does not contain the extension provided in $options['extension'] which prevents the resizer from converting to the new image type.


mjauvin:
After more thoughts, this will cause problems because the source/destination paths are the same in this case.

the `resizer->open()` and `resizer->save()` use the same path obtained from `getLocalTempPath()`

So this really needs to be fixed in Database/Attach/Resizer::save().

PR #5384, opened 2020-12-05
Add support for jsonable fields

scottbedard:
Our code editor form widget currently does not support jsonable fields. This seems like a pretty expected use case for the widget, so I've extended it to check if the field is jsonable. If so, the value will be encoded/decoded upon read/write.

I considered doing this commit a bit higher up in the `FormWidgetBase`, but decided against it because the use case for editing jsonable fields is pretty narrow, and I figured this would be easier to merge.

### Steps to reproduce

1. Create a model with a [jsonable field](https://octobercms.com/docs/database/model#property-jsonable)
2. Create a backend form with a [code editor form widget](https://octobercms.com/docs/backend/forms#widget-codeeditor) on the jsonable field

### Expected result

I should be able to use the code editor to modify the raw json.

### Actual result

Screen Shot 2020-12-04 at 11 52 35 PM

### Fixed result

Screen Shot 2020-12-05 at 12 02 01 AM

### DIFF
--- a/modules/backend/formwidgets/CodeEditor.php
+++ b/modules/backend/formwidgets/CodeEditor.php
@@ -215,4 +215,28 @@ class CodeEditor extends FormWidgetBase
         $this->displayIndentGuides = $preferences->editor_display_indent_guides;
         $this->showPrintMargin = $preferences->editor_show_print_margin;
     }
+
+    /**
+     * @inheritDoc
+     */
+    public function getLoadValue()
+    {
+        if ($this->model->isJsonable($this->fieldName)) {
+            return json_encode($this->formField->value, JSON_PRETTY_PRINT);
+        }
+
+        return parent::getLoadValue();
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function getSaveValue($value)
+    {
+        if ($this->model->isJsonable($this->fieldName)) {
+            return json_decode($value);
+        }
+
+        return parent::getSaveValue($value);
+    }
 }


LukeTowers:
Hmm, I think this might have to be an optional behaviour, disabled by default in order to be merged. I wouldn't think that it would be expected behaviour for a form field to just completely decide to bypass the jsonable property of a model instance without being asked to, and it's entirely possible that it existing code editor fields could be storing data in jsonable properties without wanting to interact with the JSON form of the property. Also, this somewhat opens a pandoras box when it comes to ensuring that the contents remain valid JSON lest you brick the model accidentally requiring manual editing of DB data in order to regain access to the model. @bennothommo any thoughts?

scottbedard:
@LukeTowers I completely agree, it would be the responsibility of the application to handle that supposedly valid json. My use case for it is for [complex configuration like this](http://next.speedcube.site/solve/gigaminx/config). I don't think calling `json_decode` on your servers is a pandoras box, but I wouldn't be stunned to hear that it was. It does probably need error handling built in though, I'll think on it a bit 🤔

bennothommo:
@LukeTowers @scottbedard I'm leaning towards this being something that should be provided by a plugin, not core. I would imagine you could make the code editor work in this fashion by using the events of the form widget to ensure the JSON data is correctly validated and handled.

JSON editing is something that I feel is outside the scope of trying to keep everything simple in October - it's mainly a scenario for developers.

LukeTowers:
@bennothommo I don't necessarily think it's outside the scope of keeping things simple, the codeeditor field itself is meant for developers, so it's not outlandish to want to have to edit JSON. I will say though that this should be achievable by using a proxy property however; you could easily do something like this:

```php
public function setJsonPropertyAsCodeAttribute($value)
{
    $this->json_property = json_decode($value)
}
public function getJsonPropertyAsCodeAttribute()
{
    return json_encode($this->json_property);
}
```
And then just set the codeeditor field to use `json_property_as_code` as the source. So for that reason I'll close this for now, unless @scottbedard has any further thoughts on this

bennothommo:
@LukeTowers what I mean is that it is very unlikely for the "average Joe" to edit a JSON file, and I feel the scope of JSON editing through the code editor is too narrow to warrant official support through the core, but most certainly could be done through a plugin if people really want it.

scottbedard:
Sounds good, thanks for the feedback guys 👍

PR #5386, opened 2020-12-07
Register plugin after installation

mjauvin:
Without this, the plugin is not registered and its migrations won't run properly if there are external dependencies.

### DIFF
--- a/modules/system/console/PluginInstall.php
+++ b/modules/system/console/PluginInstall.php
@@ -49,13 +49,17 @@ class PluginInstall extends Command
         $manager->extractPlugin($code, $hash);
 
         /*
-         * Migrate plugin
+         * Make sure plugin is registered
          */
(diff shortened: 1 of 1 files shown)


modules/system/console/PluginInstall.php
```diff
@@ -52,7 +52,10 @@ public function handle()
          * Migrate plugin
          */
         $this->output->writeln(sprintf('<info>Migrating plugin...</info>', $code));
-        PluginManager::instance()->loadPlugins();
+        $pluginManager = PluginManager::instance();
+        $pluginManager->loadPlugins();
+        $plugin = $pluginManager->findByIdentifier($code);
+        $pluginManager->registerPlugin($plugin, $code);
```
LukeTowers, line 57: Can you move this above the migrate plugin step and add its own comment "Ensure plugin is registered"?

mjauvin: done!

Issue #5393, opened 2020-12-09
Tabs produce a FOIT (Flash of Invisible Text)


summercms:
A few examples to explain the issue.

![image](https://user-images.githubusercontent.com/57409060/101645943-d7e5cb80-3a2e-11eb-8335-78b56306fedc.png)

Above: When opening tab one `directly` there is no error.

![image](https://user-images.githubusercontent.com/57409060/101646051-f64bc700-3a2e-11eb-96bd-319e7d851613.png)

Above: When opening tab two `directly`, October displays the whole contents of tab one before opening tab two! If there is a lot of data in tab one this can be shown for quite a long time. Especially when a developer has added a custom widget to tab one.

### Steps to reproduce this error

Open up a tab directly (excluding tab 1), e.g.

```
http://localhost/backend/acme/example/something/update/1#primarytab-tab2
```

### Actual behaviour

When I load a page in my browser I see the form field widgets being loaded above the tabs, then the tabs getting loaded, I see tab one being loaded first (being displayed for several hundred milliseconds) then tab two being loaded.

The more form fields being added to tab one the worse the problem gets.

### Expected behaviour

When I load a page in my browser I see the form field widgets being loaded above the tabs, then the tabs getting loaded and I only see tab two (no flash of form fields from tab one).

mjauvin:
The anchor is applied client-side, so I don't really see a way around this, unless you apply custom javascript.

ayumi-cloud:
In my browser I see the form field widgets being loaded above the tabs, then the tabs getting loaded, seeing tab one being loaded first (maybe several hundred milliseconds) then tab two being loaded.

Using a modern tab components from other libraries you don't see this error. I was hoping we could add the same technology to October to clean up the tabs.

ayumi-cloud:
In my browser I see the form field widgets being loaded above the tabs, then the tabs getting loaded, seeing tab one being loaded first (maybe several hundred milliseconds) then tab two being loaded.

Using a modern tab component from other libraries you don't see this error. I was hoping we could add the same technology to October to clean up the tabs.

p.s. Was surprised to see this error as the tabs are meant to be lazy loaded.

ayumi-cloud:
Going to play around with this issues.

I'm thinking of setting tab one's css to `display: none` on page load and then the javascript can display the tab requested tab and hopefully this gets rid of this annoying flash of undesired code and speed up loading?

ayumi-cloud:
Going to play around with this issue.

[...]

PR #5399, opened 2020-12-12
Set the `active` class to the tabs and handle `fragment` urls

summercms:
See github issue: https://github.com/octobercms/october/issues/5393

How this pr works:

1. CMS adds `active` class to tabs via php.
2. CMS adds `tabload` class to the tabs content panels (hides the tab contents) and not the tab navigation via php.
3. CMS checks to see if a user is loading a fragment url and if so then move the `active` class to the matched tab via js.
4. CMS then removes the `tabload` class (shows the tab contents) via js (final stage when everything has been processed).

This process removes 100% any flash of unwanted content from tab one!

image


modules/backend/widgets/form/assets/js/october.form.js
```diff
@@ -164,6 +165,47 @@
...
+            if (window.location.hash === $(this).attr('href')) {
+                // Remove all `active` classes
+                $('[href*="' + window.location.hash + '"]').closest('.nav-tabs').find('li').removeClass('active');
+                $('[href*="' + window.location.hash + '"]').closest('.control-tabs').find('.tab-content div.tab-pane').removeClass('active');
+        
+                // Add `active` class to matched tab
+                $('[href*="' + window.location.hash + '"]').parent('li').addClass('active');
+                $($(this).attr('data-target')).addClass('active');
+        
+            // Url doesnt contain fragment or fragement doesnt match
```
bennothommo:
```suggestion
            // URL doesn't contain fragment or fragment doesn't match
```

bennothommo:
@ayumi-cloud One change required:

Technically the primary and secondary tabs are separate entities. You will need to set the first tab of both if no hash is present. If a hash is present and does correspond to either a primary or a secondary tab, you will need to set the first tab of the *other* to active.

ayumi-cloud:
@bennothommo 

Thanks for feedback, have re-coded it to reflect your points raised, new code structure works like this now:

1. Check primary tabs exists and add `active` class to first tab.

2. Check secondary tabs exists and add `active` class to first tab.

3. The `fragment` url matches a tab `href` on the page, overwrite the `active` class for that tab.

ayumi-cloud:
@bennothommo 

Thanks for feedback, have re-coded it to reflect your points raised, new code structure works like this now:

1. Check primary tabs exist and add `active` class to first tab.

2. Check secondary tabs exist and add `active` class to first tab.

3. If the `fragment` url matches a tab `href` on the page then overwrite the `active` class for that tab.

Ok please re-review pr when you got some free time.

ayumi-cloud:
Going to add a comment about `performance` just in case people start looking at this issue in the future.

The old code did the following:

1. Add `active` class via php
2. Then javascript changes the `active` class at a later date

This pr does the following:

1. Removed the `active` class setting in php.
2. All the `active` class for the tabs are contained in a single javascript function (this should increase the performance by removing the loading of other scripts while processing the `active` class for the tabs).

What I mean by this, is the following:

1. Add `active` class via php
2. CMS then runs some other javascript code
3. Then javascript changes the `active` class

Point 2 can create a large gap in time - increasing the flash of the contents in tab one if a user has loaded the page from a `fragment` url pointing to the tab location. Removing this gap then reduces the unwanted flash of content from tab one.

- This comment is for people reading this issue in the future and gives an advantage of running all the `active` class code inside a single javascript function.

LukeTowers:
@ayumi-cloud please share a screen recording of the behaviour as it is currently and how it works with your changes.

ayumi-cloud:
@bennothommo 

6 successful and 2 failing checks.

Some error in the github actions.

summercms:
> please share a screen recording of the behaviour as it is currently and how it works with your changes.

### Current October behaviour



### Updated October behaviour (pr added)

Untitled

Note: Open the above gif's up to make them larger to view.

modules/backend/widgets/form/assets/js/october.form.js
```diff
@@ -164,6 +165,28 @@
...
+        $('.control-tabs .nav-tabs li a').each(function () {
+            // If the `fragment` url matches a tab `href` on the page then overwrite the `active` class for that tab.
+            if (window.location.hash === $(this).attr('href')) {
+                // Remove all `active` classes
+                $('[href*="' + window.location.hash + '"]').closest('.nav-tabs').find('li').removeClass('active');
+                $('[href*="' + window.location.hash + '"]').closest('.control-tabs').find('.tab-content div.tab-pane').removeClass('active');
+
+                // Add `active` class to matched tab
+                $('[href*="' + window.location.hash + '"]').parent('li').addClass('active');
+                $($(this).attr('data-target')).addClass('active');
```
LukeTowers, line 182: Does this still support nested tabs? I.e. if you have a page with a nestedform displayed in a tab, and then that nestform itself has tabs, does that still work?

staff members profile

ayumi-cloud:
If pr gets merged update code examples in doc's : https://octobercms.com/docs/ui/tab

(add `tabload`)

summercms (commented):
@LukeTowers 

This pr acts like an `add-on` feature and works in two parts.

### Part 1 (fragment url checking)

All 4 different tabs contain the `control-tabs` class. You can see the code here: https://octobercms.com/docs/ui/tab

So the code in this pr can check all of them to see if there is a fragment url match or not?

### Part 2 (displaying the tab content panels)

The cms is setup to only deal with **Primary** and **Secondary** tabs - see here: https://github.com/octobercms/october/blob/develop/modules/backend/widgets/form/partials/_form.htm

The pr adds the class `tabload` only to these two tabs.

### Nested Tabs

As the nested tabs are not under the code in part 2. They would not get the `tabload` class unless a developer added it in. 

Therefore there would be two options:

1. The `tabload` class not added - the cms would show a flash of unwanted styles (current setup).
2. The developer adds the `tabload` class - the cms would not show a flash of unwanted styles.

> It would be useful to mention this in the **Changelog**
>
> Let's say this pr gets merged, developers that have plugins with nested tabs should add the `tabload` class to their **tab panel contents** (not tab navigation) to remove the flash of unwanted styles, for example see commit: https://github.com/octobercms/october/pull/5399/commits/8254440b5246e9920897967527cfcdfc3f579a3e 
>
> Going to add a comment to the doc's to make this **stupid proof**! See here: https://github.com/octobercms/october/pull/5399/commits/06772714c32345ea4f4ccd568d8e9a2dd85d34d3

Your question is in regards to would this pr break nested tabs? (Is what I'm guessing here) The answer would be no this pr wouldn't break anything with nested tabs because this pr acts as an add-on and enhances what October is currently doing by 1. Dealing with fragment urls and 2. Removing the flash of unwanted styles.

I would like you to consider resolving this issue if possible please.

If there is any more questions, fire away - will answer them.

Otherwise please add the `testing` label.

(as I can't think of an issue this pr will cause).

ayumi-cloud:
Wait I forgot the doc's are in this repo and not in the octobercms/docs need to update this pr

modules/backend/widgets/form/assets/js/october.form.js
summercms: All 4 tabs contain `control-tabs` class and use the location `.control-tabs .nav-tabs li a` so the fragment url feature would work for nested tabs. This pr is an add-on and not a breaking change - so there shouldn't be any issues.

ayumi-cloud:
Notes: Only thing left for me to do is resubmit pr when github actions testing is fixed.

PR #5401, opened 2020-12-13
Honor cms.linkPolicy config in ImageResizer

mjauvin:


bennothommo (approved):

LukeTowers:
@mjauvin does this support being installed in a subfolder? Or do we require linkPolicy to be force in those cases?

mjauvin:
@LukeTowers I guess not, but wouldn't it be the case here as well?

ref. https://github.com/octobercms/october/blob/develop/modules/system/classes/MediaLibrary.php#L541-L552

I think we need to revisit the paths for the whole stack regarding subfolder installations, don't you think?

LukeTowers:
@mjauvin just checked the config file and I think we're covered for subfolder use because we clearly document how to make sure that works: https://github.com/octobercms/october/blob/develop/config/cms.php#L303-L306

modules/system/classes/ImageResizer.php
```diff
@@ -499,7 +499,13 @@ public function getResizerUrl()
         // Store the current configuration
         $this->storeConfig();
 
-        return Url::to("/resizer/$identifier/$resizedUrl");
+        $path = "/resizer/$identifier/$resizedUrl";
+
+        if (Config::get('cms.linkPolicy') === 'force') {
```
LukeTowers:
```suggestion
        if (Config::get('cms.linkPolicy', 'detect') === 'force') {
```

modules/system/classes/ImageResizer.php
```diff
@@ -523,7 +529,13 @@ public function getResizedUrl()
         // Ensure that a properly encoded URL is returned
         $segments = explode('/', $url);
         $lastSegment = array_pop($segments);
-        return implode('/', $segments) . '/' . rawurlencode(rawurldecode($lastSegment));
+        $path = implode('/', $segments) . '/' . rawurlencode(rawurldecode($lastSegment));
+
+        if (Config::get('cms.linkPolicy') === 'force') {
```
LukeTowers:
```suggestion
        if (Config::get('cms.linkPolicy', 'detect') === 'force') {
```

modules/system/classes/ImageResizer.php
```diff
@@ -499,7 +499,13 @@ public function getResizerUrl()
...
         $this->storeConfig();
 
-        return Url::to("/resizer/$identifier/$resizedUrl");
+        $path = "/resizer/$identifier/$resizedUrl";
+
+        if (Config::get('cms.linkPolicy', 'detect') === 'force') {
+            return Url::to($path);
+        }
+
+        return $path;
```
LukeTowers:
```suggestion
        $url = "/resizer/$identifier/$resizedUrl";

        if (Config::get('cms.linkPolicy', 'detect') === 'force') {
            $url = Url::to($url);
        }

        return $url;
```

modules/system/classes/ImageResizer.php
```diff
@@ -523,7 +529,13 @@ public function getResizedUrl()
...
         $segments = explode('/', $url);
         $lastSegment = array_pop($segments);
-        return implode('/', $segments) . '/' . rawurlencode(rawurldecode($lastSegment));
+        $path = implode('/', $segments) . '/' . rawurlencode(rawurldecode($lastSegment));
+
+        if (Config::get('cms.linkPolicy', 'detect') === 'force') {
+            return Url::to($path);
+        }
+
+        return $path;
```
LukeTowers:
```suggestion
        $url = implode('/', $segments) . '/' . rawurlencode(rawurldecode($lastSegment));

        if (Config::get('cms.linkPolicy', 'detect') === 'force') {
            $url = Url::to($url);
        }

        return $url;
```

LukeTowers:
@mjauvin could you add some unit tests for the linkPolicy settings?

mjauvin:
But the path here is static in the code to "/resizer/..." this is not configurable.

LukeTowers:
@mjauvin I'm not sure what you mean?

mjauvin:
there should be a config for the resizer path prefix, it's currently hard-coded to "/resizer"

LukeTowers:
I don't see why that should be made configurable, it's a core piece of functionality handled by October. We don't make the combiner URL configurable: https://github.com/octobercms/october/blob/develop/modules/system/routes.php#L15

mjauvin:
Ok, so even when in a sub-folder installation you're saying this won't cause problems? Fine with me.

mjauvin:
unit tests for the linkPolicy settings, do you mean for `getResizerUrl()` ?

mjauvin:
Odd, we got "Segmentation fault" for php7.2 & 7.3 tests...

LukeTowers:
And for getResizedUrl(), just to verify that those methods are generating URLs according to the linkPolicy set. Don't worry about the failing tests for now, seems like it's a github issue.

mjauvin:
tests added @LukeTowers !

mjauvin:
ping

LukeTowers:
What's wrong with the tests @mjauvin?

mjauvin:
@LukeTowers it's one of those transient unexplained failures.

LukeTowers:
@mjauvin can you push another commit (perhaps merging develop into this branch?). I don't seem to be able to re-run the checks manually

LukeTowers:
Thanks @mjauvin, good job!

PR #5402, opened 2020-12-13
Restore full Laravel Mailer behavior

### LINKED ISSUE #5124: Mail with two view files for html and plain text alternative isn't working properly
- OctoberCMS Build: 465
- PHP Version: 7.0.28

### Description:
When using the Mail fascade to send mail with an array of two view files as first argument (the one for html the other for plain text - as written in the docs)  isn't working properly. The first file is used for both parts in the mail. Using array keys or not doesn't matter.

The docs says:

> By default, the view given to the send method is assumed to contain HTML. However, by passing an array as the first argument to the send method, you may specify a plain text view to send in addition to the HTML view: 
> `Mail::send(['acme.blog::mail.html', 'acme.blog::mail.text'], $data, $callback);`

### Expected behavior
The mail message should contain a html-part from the html view file and a plain text alternative from the plain text file.

### Actual behavior
The same view file is used for both parts

### Steps To Reproduce:
Send a mail like this:

```
$data= ['msg' => 'Hello world'];
Mail::send(
    ['text' => 'acme.test::mails.text', 
     'html' => 'acme.test::mails.html'],  
    $data, function($message) { 
        $message->to('name@example.com'); 
        $message->subject('Test message');
    });
```
### Workaround
You can put both views (plain text and html) in the same view file.

mjauvin:
Fixes #5124

### DIFF
--- a/modules/system/classes/MailManager.php
+++ b/modules/system/classes/MailManager.php
@@ -186,7 +186,7 @@ class MailManager
         if (empty($swiftMessage->getSubject())) {
             if ($html) {
                 $message->subject(Twig::parse($html->subject, $data));
-            } else {
+            } else if ($text) {
                 $message->subject(Twig::parse($text->subject, $data));
             }
         }
@@ -209,7 +209,8 @@ class MailManager
          * Text content
          */
         if ($text) {
-            $message->addPart($this->renderTextTemplate($text, $data), 'text/plain');
+            $method = $html ? 'addPart' : 'setBody';
+            $message->{$method}($this->renderTextTemplate($text, $data), 'text/plain');
         }
 
 
--- a/modules/system/classes/MailManager.php
+++ b/modules/system/classes/MailManager.php
@@ -121,7 +121,7 @@ class MailManager
     {
         $html = $text = null;
 
-        if (isset($view)) {
+        if (!is_null($view)) {
             if (isset($this->templateCache[$view])) {
                 $html = $this->templateCache[$view];
             } else {
@@ -129,7 +129,7 @@ class MailManager
             }
         }
 
-        if (isset($plain)) {
+        if (!is_null($plain)) {
             if (isset($this->templateCache[$plain])) {
                 $text = $this->templateCache[$plain];
             } else {
@@ -138,13 +138,13 @@ class MailManager
         }
 
         // raw content will overwrite plain view content, as done in laravel
-        if (isset($raw)) {
+        if (!is_null($raw)) {
             $text = new MailTemplate;
             $text->fillFromContent($raw);
         }
 
         // bailout if we have no content
-        if (!isset($html) && !isset($text)) {
+        if (is_null($html) && is_null($text)) {
             return false;
         }
 
--- a/modules/system/classes/MailManager.php
+++ b/modules/system/classes/MailManager.php
@@ -117,7 +117,7 @@ class MailManager
      * @param array $data
      * @return bool
      */
-    public function addContent($message, $view, $plain, $raw, $data)
+    public function addContent($message, $view=null, $plain=null, $raw=null, $data=[])
     {
         $html = $text = null;
 
--- a/modules/system/classes/MailManager.php
+++ b/modules/system/classes/MailManager.php
@@ -119,8 +119,7 @@ class MailManager
      */
     public function addContent($message, $view, $plain, $raw, $data)
     {
-        $html = null;
-        $text = null;
+        $html = $text = null;
 
         if (isset($view)) {
             if (isset($this->templateCache[$view])) {
--- a/modules/system/classes/MailManager.php
+++ b/modules/system/classes/MailManager.php
@@ -111,9 +111,9 @@ class MailManager
      * Replaces both `addContentToMailer` and `addRawContentToMailer`
      *
      * @param \Illuminate\Mail\Message $message
-     * @param string $view
-     * @param string $plain
-     * @param string $raw
+     * @param string|null $view
+     * @param string|null $plain
+     * @param string|null $raw
      * @param array $data
      * @return bool
      */
@@ -157,12 +157,12 @@ class MailManager
      * Internal method used to share logic between `addContent`, `addRawContentToMailer` and `addContentToMailer`
      *
      * @param \Illuminate\Mail\Message $message
-     * @param MailTemplate $html
-     * @param MailTemplate $text
+     * @param MailTemplate|null $html
+     * @param MailTemplate|null $text
      * @param array $data
      * @return void
      */
-    protected function addContentToMailerInternal($message, $html, $text, $data)
+    protected function addContentToMailerInternal($message, $html = null, $text = null, array $data = [])
     {
         /*
          * Start twig transaction
--- a/modules/system/classes/MailManager.php
+++ b/modules/system/classes/MailManager.php
@@ -115,7 +115,7 @@ class MailManager
      * @param string $plain
      * @param string $raw
      * @param array $data
-     * @return void
+     * @return bool
      */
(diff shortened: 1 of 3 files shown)


LukeTowers:
@mjauvin can you add tests for the bug that this fixes?

mjauvin:
@LukeTowers of course, didn't get to that yet!

mjauvin:
@schmauch would you be able to test this PR?

bennothommo:
@mjauvin When this is ready for review, take this out of draft and I'll have a look :)

modules/system/classes/MailManager.php
```diff
@@ -93,21 +96,73 @@ public function addContentToMailer($message, $code, $data, $plainOnly = false)
...
-     * @param string $template
+     * @param string $view
+     * @param string $plain
+     * @param string $raw
      * @param array $data
-     * @param bool $plainOnly Add only plain text content to the message
      * @return void
      */
-    protected function addContentToMailerInternal($message, $template, $data, $plainOnly = false)
+    public function addContent($message, $view, $plain, $raw, $data)
```
bennothommo: @mjauvin I imagine most of these parameters will be optional, since not all types of content will be defined every time.
```suggestion
    public function addContent($message, $view = null, $plain = null, $raw = null, $data = [])
```

modules/system/classes/MailManager.php
```diff
@@ -93,21 +96,73 @@ public function addContentToMailer($message, $code, $data, $plainOnly = false)
...
     /**
-     * Internal method used to share logic between `addRawContentToMailer` and `addContentToMailer`
+     * Restore proper behavior in-line with Laravel Mailer
+     * Replaces both `addContentToMailer` and `addRawContentToMailer`
      *
      * @param \Illuminate\Mail\Message $message
-     * @param string $template
+     * @param string $view
+     * @param string $plain
+     * @param string $raw
```
bennothommo:
```suggestion
     * @param string|null $view
     * @param string|null $plain
     * @param string|null $raw
```

modules/system/classes/MailManager.php
```diff
@@ -93,21 +96,73 @@ public function addContentToMailer($message, $code, $data, $plainOnly = false)
...
      * @param array $data
-     * @param bool $plainOnly Add only plain text content to the message
      * @return void
```
bennothommo:
```suggestion
     * @return bool
```

modules/system/classes/MailManager.php
```diff
@@ -93,21 +96,73 @@ public function addContentToMailer($message, $code, $data, $plainOnly = false)
...
+        }
+
+        // raw content will overwrite plain view content, as done in laravel
+        if (isset($raw)) {
+            $text = new MailTemplate;
+            $text->fillFromContent($raw);
+        }
+
+        // bailout if we have no content
+        if (!($html || $text)) {
```
bennothommo: @mjauvin Need to do a proper null check here
```suggestion
        if (!isset($html) && !isset($text)) {
```

modules/system/classes/MailManager.php
```diff
@@ -93,21 +96,73 @@ public function addContentToMailer($message, $code, $data, $plainOnly = false)
...
+
+        return true;
+    }
+
+    /**
+     * Internal method used to share logic between `addContent`, `addRawContentToMailer` and `addContentToMailer`
+     *
+     * @param \Illuminate\Mail\Message $message
+     * @param MailTemplate $html
+     * @param MailTemplate $text
```
bennothommo:
```suggestion
     * @param MailTemplate|null $html
     * @param MailTemplate|null $text
```

modules/system/classes/MailManager.php
```diff
@@ -93,21 +96,73 @@ public function addContentToMailer($message, $code, $data, $plainOnly = false)
...
+     * @param array $data
+     * @return void
+     */
+    protected function addContentToMailerInternal($message, $html, $text, $data)
```
bennothommo:
```suggestion
    protected function addContentToMailerInternal($message, $html = null, $text = null, array $data = [])
```

bennothommo (changes requested):

modules/system/classes/MailManager.php
mjauvin: Actually, they're always provided, even if null.

bennothommo: @mjauvin If that were the case, we wouldn't need the `isset` checks throughout this method. We also can't discount that perhaps someone is using this method directly, as opposed to passing it through the event.

mjauvin:
can this also be added to 1.1.2 milestone?

mjauvin:
Anything preventing this to be merged?

modules/system/classes/MailManager.php
```diff
@@ -93,21 +96,73 @@ public function addContentToMailer($message, $code, $data, $plainOnly = false)
...
+     * @param string|null $plain
+     * @param string|null $raw
+     * @param array $data
+     * @return bool
+     */
+    public function addContent($message, $view, $plain, $raw, $data)
+    {
+        $html = $text = null;
+
+        if (isset($view)) {
```
LukeTowers: Shouldn't this be `!is_null()` instead of isset()?

mjauvin: @LukeTowers What's the advantage of using !is_null() ? They're both doing pretty much the same thing.

LukeTowers: It's more semantically correct. I don't personally agree with `isset()` returning false for variables that are set to null. It's within the realm of possibilities that that behaviour could be changed in a future version of PHP, but mostly it's for semantics. @bennothommo any thoughts?

mjauvin: I must say I agree with you, but it's used extensively in october code base and Laravel. I can change it if you prefer, I don't really mind.

LukeTowers: Yes please, ideally we make the change over for any other similar cases as we come across them.

modules/system/classes/MailManager.php
```diff
@@ -128,28 +183,35 @@ protected function addContentToMailerInternal($message, $template, $data, $plain
...
-            $html = $this->renderTemplate($template, $data);
-
-            $message->setBody($html, 'text/html');
+        /*
+         * HTML content
+         */
+        if ($html) {
+            $message->setBody($this->renderTemplate($html, $data), 'text/html');
+            if (!$text && $html->content_text) {
+                $text = $html;
```
LukeTowers, line 203: Shouldn't this be `$text = $html->content_text`?

mjauvin: @LukeTowers no, because `renderTextTemplate()` method will be called with this value and it's going to extract the content_text value.

modules/system/classes/MailManager.php
bennothommo: @LukeTowers @mjauvin I'm admittedly of the opposite position. I predominantly use `isset` because it also safely handles undefined variables - I don't believe `is_null` does the same. Note that when we use null-coalesce syntax (`??`), that's technically firing an `isset` too.

`null` traditionally means the *absence* of a value, which is why I agree that `isset` returns `false` for it.

mjauvin: @bennothommo that is true for accessing array values or object properties, but for direct variable access, isser isn't necessary.

bennothommo: @mjauvin How so? It's the same principle.

LukeTowers: @bennothommo in this case those variables will never not exist since they're required in the constructor

bennothommo: @LukeTowers that's fair enough in this case, but I'm against changing it wholesale.

PR #5408, opened 2020-12-15
Use setComponentContext everywhere

### LINKED ISSUE #5406: Nested component isn't detected by the Page->hasComponent method.
- OctoberCMS Build:471
- PHP Version:7.4

### Description:
When calling a component inside of antoher component, the page can't correctly detect the nested component.

### Steps To Reproduce:
I just tried to play around with Mall plugin which uses this techniques in all its components.
Example in its checkout component:
image

So It can easily be reproduced installing it and trying to call this on the checkout page:
```
Event::listen('cms.page.initComponents', function ($controller, $page, $layout) use ($componentManager) {
    dump($page->components);
    dump($page->hasComponent('shippingMethodSelector'));
    dump($page->hasComponent('OFFLINE\Mall\Components\ShippingMethodSelector'));
});
```
the result:
image

This is probably due to the CompoundObject context and may be related to an (very) old one issue: [#2619](https://github.com/octobercms/october/issues/2619)

mjauvin:


mjauvin:
@RomainMazB you're gonna be the alpha tester! ;)

daftspunk:
Why?

PR #5415, opened 2020-12-18
Add trusted hosts configuration value

bennothommo:
Related Rain PR: https://github.com/octobercms/library/pull/549

Adds support for defining trusted hosts in `config/app.php`

### DIFF
--- a/modules/backend/controllers/Auth.php
+++ b/modules/backend/controllers/Auth.php
@@ -13,6 +13,7 @@ use ApplicationException;
 use ValidationException;
 use Exception;
 use Config;
+use October\Rain\Foundation\Http\Middleware\CheckForTrustedHost;
 
 /**
  * Authentication controller
@@ -150,16 +151,18 @@ class Auth extends Controller
         // Force Trusted Host verification on password reset link generation
         // regardless of config to protect against host header poisoning
         $trustedHosts = Config::get('app.trustedHosts', false);
+
         if ($trustedHosts === false) {
-            $url = Config::get('app.url', null);
-            if (!empty($url)) {
-                // Explicitly only allow the APP_URL host and subdomains
-                Request::setTrustedHosts(['^(.+\.)?' . preg_quote(parse_url($url, PHP_URL_HOST)) . '$']);
+            $hosts = CheckForTrustedHost::processTrustedHosts(true);
+
+            if (count($hosts)) {
+                Request::setTrustedHosts($hosts);
+
                 // Trigger the host validation logic
                 Request::getHost();
             }
         }
-        
+
         $rules = [
             'login' => 'required|between:2,255'
         ];
--- a/modules/backend/controllers/Auth.php
+++ b/modules/backend/controllers/Auth.php
@@ -13,7 +13,6 @@ use ApplicationException;
 use ValidationException;
 use Exception;
 use Config;
-use October\Rain\Foundation\Http\Middleware\CheckForTrustedHost;
 
 /**
  * Authentication controller
@@ -148,21 +147,6 @@ class Auth extends Controller
      */
     public function restore_onSubmit()
     {
-        // Force Trusted Host verification on password reset link generation
-        // regardless of config to protect against host header poisoning
-        $trustedHosts = Config::get('app.trustedHosts', false);
-
-        if ($trustedHosts === false) {
-            $hosts = CheckForTrustedHost::processTrustedHosts(true);
-
-            if (count($hosts)) {
-                Request::setTrustedHosts($hosts);
-
-                // Trigger the host validation logic
-                Request::getHost();
-            }
-        }
-
         $rules = [
             'login' => 'required|between:2,255'
         ];
--- a/modules/backend/controllers/Auth.php
+++ b/modules/backend/controllers/Auth.php
@@ -13,6 +13,7 @@ use ApplicationException;
 use ValidationException;
 use Exception;
 use Config;
+use October\Rain\Foundation\Http\Middleware\CheckForTrustedHost;
 
 /**
  * Authentication controller
@@ -147,6 +148,21 @@ class Auth extends Controller
      */
     public function restore_onSubmit()
     {
+        // Force Trusted Host verification on password reset link generation
+        // regardless of config to protect against host header poisoning
+        $trustedHosts = Config::get('app.trustedHosts', false);
+
+        if ($trustedHosts === false) {
+            $hosts = CheckForTrustedHost::processTrustedHosts(true);
+
+            if (count($hosts)) {
+                Request::setTrustedHosts($hosts);
+
+                // Trigger the host validation logic
+                Request::getHost();
+            }
+        }
+
         $rules = [
             'login' => 'required|between:2,255'
         ];
--- a/modules/backend/controllers/Auth.php
+++ b/modules/backend/controllers/Auth.php
@@ -147,6 +147,19 @@ class Auth extends Controller
      */
     public function restore_onSubmit()
     {
+        // Force Trusted Host verification on password reset link generation
+        // regardless of config to protect against host header poisoning
+        $trustedHosts = Config::get('app.trustedHosts', false);
+        if ($trustedHosts === false) {
+            $url = Config::get('app.url', null);
+            if (!empty($url)) {
+                // Explicitly only allow the APP_URL host and subdomains
+                Request::setTrustedHosts(['^(.+\.)?' . preg_quote(parse_url($url, PHP_URL_HOST)) . '$']);
+                // Trigger the host validation logic
+                Request::getHost();
+            }
+        }
+        
         $rules = [
             'login' => 'required|between:2,255'
         ];
--- a/config/app.php
+++ b/config/app.php
@@ -45,39 +45,6 @@ return [
 
     /*
     |--------------------------------------------------------------------------
-    | Trusted hosts
-    |--------------------------------------------------------------------------
-    |
-    | You may specify valid hosts for your application as an array or boolean
-    | below. This helps prevent host header poisoning attacks.
-    |
-    | Possible values:
-    |  - `true`: Trust the host specified in app.url, as well as the "www"
-    |            subdomain, if applicable.
-    |  - `false`: Disable the trusted hosts feature.
-    |  - array: Defines the domains to be trusted hosts. Each item should be
-    |           a string defining a domain, IP address, or a regex pattern.
-    |
-    | Example of array values:
-    |
-    |    'trustedHosts' => [
-    |       'example.com',           // Matches just example.com
-    |       'www.example.com',       // Matches just www.example.com
-    |       '^(.+\.)?example\.com$', // Matches example.com and all subdomains
-    |       'https://example.com',   // Matches just example.com
-    |    ]
-    |
-    | If you do not wish to enable this feature, please leave this as `false`.
-    |
-    | Note: Even when set to `false`, this functionality is explicitly enabled
-    | on the Backend password reset flow for security reasons.
-    |
-    */
-
-    'trustedHosts' => true,
-
-    /*
-    |--------------------------------------------------------------------------
     | Application Timezone
     |--------------------------------------------------------------------------
     |
@@ -181,7 +148,7 @@ return [
     */
 
     'loadDiscoveredPackages' => false,
-
+    
     /*
     |--------------------------------------------------------------------------
     | Class Aliases
--- a/config/app.php
+++ b/config/app.php
@@ -49,7 +49,7 @@ return [
     |--------------------------------------------------------------------------
     |
     | You may specify valid hosts for your application as an array or boolean
-    | below. This will prevent host header poisoning in your application.
+    | below. This helps prevent host header poisoning attacks.
     |
     | Possible values:
     |  - true: Trust the host specified in app.url, as well as all subdomains.
@@ -57,11 +57,23 @@ return [
     |  - array: Defines the domains to be trusted hosts. Each item should be
     |           a string defining a domain, or a regex pattern.
     |
+    | Example of array values:
+    |
+    |    'trustedHosts' => [
+    |       'example.com',           // Matches just example.com
+    |       'www.example.com',       // Matches just www.example.com
+    |       '^(.+\.)?example\.com$', // Matches example.com and all subdomains
+    |       'https://example.com',   // Matches just example.com
+    |    ]
+    |
     | If you do not wish to enable this feature, please leave this as false.
+    |
+    | Note: Even when set to 'false', this functionality is explicitly enabled
+    | on the Backend password reset flow for security.
     |
     */
 
-    'trustedHosts' => false,
+    'trustedHosts' => true,
 
     /*
     |--------------------------------------------------------------------------
--- a/config/app.php
+++ b/config/app.php
@@ -45,6 +45,39 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Trusted hosts
+    |--------------------------------------------------------------------------
+    |
+    | You may specify valid hosts for your application as an array or boolean
+    | below. This helps prevent host header poisoning attacks.
+    |
+    | Possible values:
+    |  - `true`: Trust the host specified in app.url, as well as the "www"
+    |            subdomain, if applicable.
+    |  - `false`: Disable the trusted hosts feature.
+    |  - array: Defines the domains to be trusted hosts. Each item should be
+    |           a string defining a domain, IP address, or a regex pattern.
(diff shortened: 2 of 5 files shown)


modules/backend/controllers/Auth.php
```diff
@@ -147,6 +147,19 @@ public function restore()
      */
     public function restore_onSubmit()
     {
+        // Force Trusted Host verification on password reset link generation
+        // regardless of config to protect against host header poisoning
```
bennothommo: @LukeTowers I think I would prefer that we don't force it, on the basis that:

a) some people would be opting to configure their web server to protect against this kind of attack and would disable this feature in October CMS to get a small performance increase.
b) it might be a bit misleading to say that `app.trustedHosts` set to `false` disables the feature entirely, just to have it be overridden here regardless.

ayumi-cloud:
@bennothommo @LukeTowers 

I'm really against this pr! It opens up a huge security risk!

Let me give an example.

### First issue

Let's say a person installs October CMS on a shared hosting plan with cPanel. 

Firstly, the hosting company will install the standard code which creates a ton of useless subdomains! I'm pretty sure most users don't bother to go through their DNS settings and deleted most of these subdomains. 

I'm talking about common subdomains such as `mail.example.com` etc.

If you go to such a subdomain on a shared hosting plan it will display another persons website using the same server!

> So setting Trusted host on a shared server would actually trust all websites on that shared hosting!

### Second issue

Subdomain hijacking, because many users don't check their subdomains it's pretty common to do subdomain takeovers!

> So setting Trusted host would open the door for subdomain take overs!

---

Sorry but I really don't want to see this pr get merged - way too dangerous!

ayumi-cloud:
I give a quick example of an attack if this pr gets merged.

I went to octobercms website and grabbed the first partner website.

Put it in a dns scanner:

image

Many subdomains - mostly likely there's an issue.

image

Above can do a subdomain hijacking on `office dot vdlp dot nl`

image

Above redirects to another person's website on shared hosting.

---

That was my first example and took me less than 30 seconds. So if you merged this pr that website would be in trouble.

---

You would really make so many people's websites in trouble merging this pr.

modules/backend/controllers/Auth.php
al-sultani: @bennothommo For A we can't depend on the server configuration to solve a security issue this should be implemented in the logic of the software. For B you must do this in order to keep safe from host header poisoning. However, I agree with you about this one it sounds misleading to enable it when the user already disabled it and didn't set Trusted Hosts.

al-sultani:
@ayumi-cloud I understand your point, but the host header poisoning attack will be less exploitable since you can't easily get a subdomain for a host and it isn't that simple as you demonstrated. I suggest keeping the server_name only as a trusted host.

modules/backend/controllers/Auth.php
bennothommo: @ahussam There are certainly ways of preventing it at the server level - for example, through proxies.

bennothommo:
@ayumi-cloud I really don't understand your opposition to this PR.

The current situation with October CMS is *worse*, in that **any domain** can be entered into the Host header and poisoning the request. We have a very specific example to this that we cannot disclose, but that this PR intends to solve.

I urge you to consider this PR with a bit more thought towards the greater benefit before knee-jerking on one domain.

modules/backend/controllers/Auth.php
al-sultani: @bennothommo I know there is but as an open-source project, this should be fixed inside the project itself rather than depending on external components.

LukeTowers:
@ayumi-cloud additionally your point is moot if the user just sets up the hosts they want to allow explicitly instead of just leaving it for October to auto allow the app.url and all subdomains.

modules/backend/controllers/Auth.php
LukeTowers: @bennothommo personally I'm not super concerned about the performance impact of this, the middleware is loaded regardless, so the actual performance benefit is on the order of 5-20 lines of regular code doing not very much at all, i.e. pretty useless from an optimization standpoint. 

I'm not a huge fan of manually enabling it even when it's explicitly set to false, but it is documented (https://github.com/octobercms/october/pull/5415/files#diff-950397cc5ef29707b54d0c774bd94af85836bab1a27f2f86533d832d25caf224R71-R72). Perhaps we could say that 'null' is the default value and it means that everything is unprotected except for the backend password reset (or even the entire backend), and false is a value that can be used to disable everything.

ayumi-cloud:
What I'm saying is I have an issue with allowing users to setup `wildcards` for their trusted hosts, just allows subdomain hijacking attacks.

As for the host header poisoning attack just copy the same as what Django has with it's ALLOWED_HOSTS, see here:
https://docs.djangoproject.com/en/1.10/ref/settings/#std:setting-ALLOWED_HOSTS

Plus use the https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host

But without the `wildcards` then I would be happy!

ayumi-cloud:
What I'm saying is I have an issue with allowing users to setup `wildcards` for their trusted hosts, just allows subdomain hijacking attacks.

As for the host header poisoning attack just copy the same as what Django has with it's ALLOWED_HOSTS, see here:
https://docs.djangoproject.com/en/1.10/ref/settings/#std:setting-ALLOWED_HOSTS

Plus use the https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host

But without the `wildcards` then I would be happy!

e.g.

This is fine:

```
ALLOWED_HOSTS = ['localhost', '127.0.0.1', '111.222.333.444', 'mywebsite.com']
```

This is not fine:

```
ALLOWED_HOSTS = ['*']
```

or

```
ALLOWED_HOSTS = ['*.mywebsite.com']
```

p.s.

```
ALLOWED_HOSTS = ['localhost', '127.0.0.1']
```

Are only allowed when config `app.debug` is set to `true` and not allowed when set to `false`.

LukeTowers:
Wildcards not being allowed is a decision that individual developers should be making, not the platform or as whole.

summercms:
> Wildcards not being allowed is a decision that individual developers should be making, not the platform or as whole.

That's the whole issue I'm getting bothered about.

```
ALLOWED_HOSTS = ['*.octobercms.com'] // This no good
```

and

```
ALLOWED_HOSTS = ['www.octobercms.com', 'gateway.octobercms.com'] // This correct
```

The cms code should parse the hosts like this!

It's well documented over in the Django camp using wildcards is really bad - please learn from their mistakes!

LukeTowers:
@ayumi-cloud there are completely legitimate reasons to enable wildcard subdomains, yes doing that is more risky but despite that fact there are completely valid reasons for doing so. Are you able to link to some articles arguing your point?

summercms:
@LukeTowers Sure let me do a quick search (don't really want to argue to you guys - as I respect the hard work you, bennothommo  and all the admins do - to be honest).

Here's a very quick set of search results.

### Acunetix Recommendation not to use wildcards

image

### When Django Framework fixed (James Kettle wrote this blog)

image

> Note: The part saying **Exploiting wildcard whitelists**

Link: https://www.skeletonscribe.net/2013/05/practical-http-host-header-attacks.html

### Long article on Wildcards and Content Poisoning

Link: https://w00tsec.blogspot.com/2014/03/wilcard-dns-content-poisoning-xss-and.html

### Phishing with Wildcard DNS Attacks and Pharming

Link: https://info.phishlabs.com/blog/phishing-with-wildcard-dns-attacks-and-pharming

---

Do you really want me to go further and start posting `Exploiting wildcard whitelists` ??? If so I can go over to security researcher Troy Hunt or Google's Mike West websites and get some strong cases why it's a bad idea.

ayumi-cloud:
Another thing I would like to mention is this from the Django community is the following statement:

> The condition to be satisfied is that the host header  should match one of the values in `ALLOWED_HOSTS` or the `X-Forwarded-Host` if the `USE_X_FORWARDED_HOST` is enabled.

That's why I previously mentioned about: [X-Forwarded-Host](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host).

Personally I did test `X-Forwarded-Host` months ago with October and it did cause issues.

At the time I tested all these:

- CLIENT_IP
- FORWARDED
- FORWARDED_FOR
- FORWARDED_FOR_IP
- VIA
- X_FORWARDED
- X_FORWARDED_FOR
- HTTP_CLIENT_IP
- HTTP_FORWARDED
- HTT P_FORWARDED_FOR
- HTTP_FORWARDED_FOR_IP
- HTTP_PROXY_CONNECTION
- HTTP_VIA
- HTTP_X_FORWARDED
- HTTP_X_FORWARDED_FOR

Found two issues, can't remember what they were.

modules/backend/controllers/Auth.php
bennothommo: @LukeTowers agreed with the first point.

I'm still favoring a way of disabling it for the second point, but I'm happy to defer and give it a try in this fashion for now. If we get a heap of people getting caught out by it going forward, then I'll push for a change here. Regarding using `null`, I think `null` should probably follow the convention of other configuration options (ie. `false` on dev, `true` on prod).

bennothommo:
@ayumi-cloud I'd be willing to consider changing the `true` value to just represent the `app.url` (plus the `www` subdomain) as opposed to blindly accepting all subdomains. But if someone explicitly uses a regular expression to match many domains, I don't think we should be denying them this on a software level. They know the risks.

LukeTowers:
@bennothommo had the exact same comment typed up but forget to hit "comment" lol

@ayumi-cloud thanks for the additional information. I would say that there's a legitimate case to be made for not allowing through wildcard domains by default (i.e. when trustedHosts = true, only allow in app.url, don't auto allow subdomains) as well as providing a warning in the config file about using regex patterns to enable wildcard subdomains; but it's still something that devs should be able to do if they explicitly accept those risks.

modules/backend/controllers/Auth.php
LukeTowers: @bennothommo Would it be that useful for this particular config item to have the same null behavior as other items?

bennothommo: @LukeTowers it might be useful to have it off for dev, especially for environments where the URL is dynamic (like Octodock).

ayumi-cloud:
@bennothommo @LukeTowers 

I'm just going to write my thinking from my mind to paper. Hopefully this may help you guys to create this pr.

1. This pr is trying to harden against: **host header poisoning** and as I like to call it **Web cache poisoning**

```
GET /example HTTP/1.1
Host: example.com
Host: bad-stuff-here
```

Above: I have injected a duplicate `Host` headers.

2. The request line should be given precedence when routing the request but, in practice, this isn't always the case. You can potentially exploit these discrepancies in much the same way as duplicate `Host` headers.

```
GET https://example.com/ HTTP/1.1
Host: bad-stuff-here
```

3. You can also uncover quirky behaviour by indenting HTTP headers with a space character.

```
GET /example HTTP/1.1
 Host: bad-stuff-here
Host: example.com
```

Above: Notice I put a space in front of the second line.

4. You can sometimes use `X-Forwarded-Host` to inject your malicious input while circumventing any validation on the `Host` header itself.

```
GET /example HTTP/1.1
Host: example.com
X-Forwarded-Host: evil.com
```

5. Although `X-Forwarded-Host` is the de facto standard for this behaviour, you may come across other headers that serve a similar purpose, including:

```
X-Host
X-Forwarded-Server
X-HTTP-Host-Override
Forwarded
```

6. **Exploiting wildcard whitelists** the first time I learnt about **Web cache poisoning** was being shown from websites allowing wildcards!

```
ALLOWED_HOSTS = ['*.octobercms.com'] // This no good
```

and

```
ALLOWED_HOSTS = ['www.octobercms.com', 'gateway.octobercms.com'] // This correct
```

In the above example `*.octobercms.com` then allows a hacker to inject `evil.octobercms.com`, this has an advantage of being a **phishing attack** and **Web cache poisoning** at the same time!

The recommendation from many security websites say:

> What you can do is make SERVER_NAME trustworthy. This can be achieved under Apache (instructions) and Nginx (instructions) by creating a dummy vhost that catches all requests with unrecognized Host headers. It can also be done under **Nginx by specifying a non-wildcard SERVER_NAME**, and under **Apache by using a non-wildcard serverName** and **turning the UseCanonicalName directive on**. I'd recommend using both approaches wherever possible.

## Fixing this vulnerability

The Django framework patched this issue many years ago and I like their setup.

1. Use a config file to store `ALLOWED_HOSTS`

2. Allow users to add whitelist rules - but not allow them to add wildcards. These include: the root domain, exact subdomains and the decicated ip address of the server, for example:

```
ALLOWED_HOSTS = ['111.222.333.444', 'example.com', 'japanese.example.com']
```

3. The cms also checks to see if the website is running in `developer` or `production` mode. If the cms is in `production` mode and the user tries to add a developer location such as `localhost` or `127.0.0.1` the cms should block these in its validation rules and send the user a warning message. For example:

```
ALLOWED_HOSTS = ['localhost', 111.222.333.444', 'example.com', 'japanese.example.com']
```

and config file `environment.php` is set to:

```
'default' => 'production',
```

4. This security feature should be turned on and **there shouldn't be an option to turn it off**. By default the cms should set the following rule:
 
```
ALLOWED_HOSTS = ['example.com', 'www.example.com']
```

5. It is also important to check that you do not support additional headers that may be used to construct these attacks, in particular `X-Forwarded-Host`. Remember that these may be supported by default.

Several months ago we tested the `X-Forwarded-Host` in October and found that it was being used to allow the website to send it's `POST` requests in certain situations.

The cms should also have extra rules to address these headers.

6. An extra level of security which we coded in our Summer CMS prospal was controlling `Proxies` headers. Some of these headers can be used to perform **Web cache poisoning** attacks. A list of all the proxy headers we control in our cms update propsal are a follows:

```
CLIENT_IP
FORWARDED
FORWARDED_FOR
FORWARDED_FOR_IP
HTT P_FORWARDED_FOR
HTTP_CLIENT_IP
HTTP_FORWARDED
HTTP_FORWARDED_FOR_IP
HTTP_PC_REMOTE_ADDR
HTTP_PROXY_CONNECTION
HTTP_VIA
HTTP_X_FORWARDED
HTTP_X_FORWARDED_FOR
PROXY_CONNECTION
USERAGENT_VIA
VIA
X_FORWARDED
X_FORWARDED_FOR
XPROXY_CONNECTION
```

bennothommo:
@ayumi-cloud You make good points about the `X-Forwarded-Host` header, but I worry that you are conflating this issue with vulnerabilities that are introduced at the *server* level, for example, with malformed or duplicate headers, or other forms of host header that may be used.

The fix we are introducing is to combat a common security vulnerability and *complement* hardening at a server level, but given that October is compatible with several different hosting software, there is a degree of responsibility on server administrators to follow proper hardening procedures.

ayumi-cloud:
@bennothommo 

My comment was in two parts, the first was an overview of the attack and the second part ideas for this pr. Wasn't sure how much background you guys know on this attack so added it anyway, yah first half talks a lot about the server end. Please focus more on the `Fixing this vulnerability` section.

ayumi-cloud:
@bennothommo 

> You make good points about the X-Forwarded-Host header, but I worry that you are conflating this issue with vulnerabilities that are introduced at the server level, for example, with malformed or duplicate headers, or other forms of host header that may be used.
>
>The fix we are introducing is to combat a common security vulnerability and complement hardening at a server level, but given that October is compatible with several different hosting software, there is a degree of responsibility on server administrators to follow proper hardening procedures.

My comment was is two parts, the first was an overview of the attack and the second suggestions for this pr.

Since reading your comment I added a header example (point 9) which can override the server level instructions to point out that the cms must deal with these headers as well as setting up a trusted hosts configuration value. Because October is using application frameworks that allow them!

LukeTowers:
@ayumi-cloud none of those headers matter to October in the slightest. 

Host header poisoning protection is provided by Symfony (https://github.com/symfony/symfony/blob/5.x/src/Symfony/Component/HttpFoundation/Request.php#L1177-L1203), so we don't really care what headers they look at, because that is handled before reaching the protection offered by October. Of course, the trustedHosts protection does nothing for developers that are manually reading those headers, but there's nothing that can or should be done in that case as developers should be using Request::getHost() so that they are protected by these measures.

bennothommo:
Urgh, VSCode stuffed up this PR by merging in another branch automatically. I'm going to have to recreate it.

ayumi-cloud:
@LukeTowers Thanks for the link to Symfony - I keep going back to the php file and check to see what RFC standards they are using for each `Request::()` function, it's really cool they comment everything.

PR #5421, opened 2020-12-21
Properly scope dependent fields to their current form

### LINKED ISSUE #5311: Repeater dropdowns and triggering loading indicator
This bug manifests itself when using dependsOn inside form repeater.
Changing of value in dropdown fires loading indicators not only in his own dependents.
loading

bennothommo:
Fixes #5311

Previously, the event to update dependent fields was indiscriminate and would affect all forms, which would include repeater items and nested forms. This limits the scope of affected fields to within the form that is modified, so for example, if you modify a field within a repeater item, it will update the dependent fields only within that repeater item.

This is technically a change in behaviour, so I'll put this up for review first before merging.

### DIFF
--- a/modules/backend/widgets/form/assets/js/october.form.js
+++ b/modules/backend/widgets/form/assets/js/october.form.js
@@ -30,8 +30,9 @@
     FormWidget.prototype.init = function() {
 
         this.$form = this.$el.closest('form')
+        this.dependencyMap = {}
 
-        this.bindDependants()
+        this.bindDependents()
         this.bindCheckboxlist()
         this.toggleEmptyTabs()
         this.bindLazyTabs()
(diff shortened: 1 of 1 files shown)


LukeTowers:
Slight problem, dependsOn being dependent on fields outside of the current form scope is permitted behavior and I've used it in production at least twice personally.

LukeTowers:
As long as it doesn't affect that use case (parent form has a switch, child form has a field that depends on the switch) then I'm fine with it (if it's just to say that parentForm has a field that depends on a field in the child form, then that's probably fine to make the change on)

LukeTowers:
This will have to be v1.1.3 however so that the security fixes aren't accompanied by any form of breaking change.

bennothommo:
@LukeTowers haha, I did think this was going to be brought up.

I could tell that it was possible, but I'm wondering if we should be more strict on scope. Generally, when people are adding dependsOn inside of a repeater, they're referencing a field inside the same repeater. There could potentially be collisions if someone is using a field name inside the repeater that's also being used outside of it (ie. `name` field outside for the main model, `name` field inside the repeater for each repeater item, if `name` was used for a dependency, which one would be the correct one to look at?)

We could potentially be smart and try and find the nearest scope and attach to that, or alternatively, we could introduce new syntax to determine if a field inside the same nested form / repeater is used vs. one for the parent form - ie. `name` for the inner repeater field, `^name` for the parent form field.

LukeTowers:
I can't remember what functionality in october already uses the `^` syntax for parent forms (might be the input trigger API?) but that would be fine to me.

PR #5424, opened 2020-12-22
Lists widget relation column's ordering proposal

Rike-cz:
This PR is proposal of Lists widget relation column's ordering.
When relation column is used in Lists widget, and relation is 1:N/M:N, values of selected column are joined by separator, but in undefined order. This PR uses relation's predefined ordering, if set.

I am not sure if I found the best way to achieve my goal, every help or comment is welcome.

### DIFF
--- a/modules/backend/widgets/Lists.php
+++ b/modules/backend/widgets/Lists.php
@@ -499,8 +499,14 @@ class Lists extends WidgetBase
                 $relationObj = $this->model->{$column->relation}();
                 $countQuery = $relationObj->getRelationExistenceQuery($relationObj->getRelated()->newQueryWithoutScopes(), $query);
 
+                $sqlOrder = null;
+                foreach($relationObj->getBaseQuery()->orders ?? [] as $order) {
+                    $sqlOrder[] = $order['column'] . ' ' . $order['direction'];
+                }
+                if (!empty($sqlOrder)) $sqlOrder = ' order by ' . implode(' ', $sqlOrder);
+
                 $joinSql = $this->isColumnRelated($column, true)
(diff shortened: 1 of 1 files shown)


LukeTowers:
I don't really understand the purpose of this PR

Rike-cz:
Well, I will try to explain on use case.
I have hasMany relation and trying to display result of this relation in Lists widget for each row of parent model. For example:
```
Category::$hasMany = [
    Product::class,
    'key' => 'product_id',
    'order' => 'name ASC'
];
```
For Lists widget I have defined column:
```
product_names:
    label: ...
    relation: products
    select: name
```
And the subquery made by Lists widget is: `select group_concat(name separator ', ') from products where categories.id = products.product_id) as product_names` and result looks like:

unordered

But related product names are sorted randomly, I said _order by name ASC_ in relation definition, didn't I?
My little PR proposal trying to solve this problem. It takes order from relation definition (default query) and use it for sorting related product names. Subquery made by Lists widget is `select group_concat(name order by name separator ', ') from products where categories.id = products.product_id) as product_names` and result looks like:

ordered

Finally, I am not sure if I used the best way to achieve this. But meaning of this PR is clear, isn't it?

LukeTowers:
@Rike-cz thanks, that makes a lot more sense. I'm thinking this might not be the correct solution, but I don't really have time to dig deeper at the moment. Perhaps @bennothommo can take a look when he has some free time.

daftspunk:
If the ordering is a concern, you can easily add this yourself in the `select` definition?

```
product_names:
    label: ...
    relation: products
    select: name order by name desc
```

Ping me if I've missed something and we can revisit. Thanks @Rike-cz !

Issue #5428, opened 2020-12-24
Proposal: Allow third-party icons in October components


summercms:
Our Plugins use Font Awesome 6 - alpha right now and I've just spotted an issue with October.

Everything is working fine, except this one code line:

https://github.com/octobercms/october/blob/5591e3fa945ad3c181ccedd9b759e12127b09f3e/modules/cms/formwidgets/components/partials/_component.htm#L3

Currently October is using the container `<div>` to inject the icon.

The issue is that because it's using the `<div>` element October needs to inject `oc-`

This causes the following issue:

```
fad fa-cookie
```

becomes

```
oc-fad fa-cookie
```

This then makes the icon invalid, see screenshot:

![image](https://user-images.githubusercontent.com/57409060/103096188-0650d280-45fb-11eb-9133-2be44c187f81.png)

Notice the square box, where the icon should be placed.

### Proposal fix

Simple fix, we would like to `upgrade` the code to this:

```php
<div class="layout-cell <?= e($component->componentCssClass) ?> <?= $component->isHidden ? 'hidden' : null ?>">
    <div
        class="layout-relative"
        <?php if ($component->inspectorEnabled): ?>data-inspectable<?php endif ?>
        data-inspector-title="<?= $name = e($this->getComponentName($component)) ?>"
        data-inspector-description="<?= $description = e($this->getComponentDescription($component)) ?>"
        data-inspector-config="<?= e($this->getComponentsPropertyConfig($component)) ?>"
        data-inspector-class="<?= get_class($component) ?>">
            <span class="name"><i class="<?= $component->pluginIcon ?>"></i> <?= $name ?></span>
            <span class="description"><?= $description ?></span>
            <span class="alias oc-icon-code"><?= e($component->alias) ?></span>
            <input type="hidden" name="component_properties[]" data-inspector-values value="<?= e($this->getComponentPropertyValues($component)) ?>" />
            <input type="hidden" name="component_names[]" value="<?= e($component->name) ?>" />
            <input type="hidden" name="component_aliases[]" value="<?= e($component->alias) ?>" />
            <a href="#" class="remove">&times;</a>
    </div>
</div>
```

Also upgrading the css code line:

```css
div.control-componentlist div.components div.layout-cell > div span.name {
    padding-left: 38px;
}
```

into

```css
div.control-componentlist div.components div.layout-cell > div span.name {
    padding-right: 20px;
}
```

Code location: modules/cms/assets/css/october.components.css?v471 Line: 200

[...]

PR #5429, opened 2020-12-24
Allow third-party icons in October components

summercms:
Relates to this github issue: https://github.com/octobercms/october/issues/5428

image

I'm using font-awesome 6 icons in my plugins, with this pr. I see no problems in the main navigation bar, sidebar menu, cms page sidebar component list, adding components to pages etc and viewing components list in the builder plugin.

Happy days!

### DIFF
--- a/modules/cms/formwidgets/components/partials/_component.htm
+++ b/modules/cms/formwidgets/components/partials/_component.htm
@@ -6,7 +6,7 @@
         data-inspector-description="<?= $description = e($this->getComponentDescription($component)) ?>"
         data-inspector-config="<?= e($this->getComponentsPropertyConfig($component)) ?>"
         data-inspector-class="<?= get_class($component) ?>">
-            <span class="name"><i class="<?= $component->pluginIcon ?>"></i> <?= $name ?></span>
+            <span class="name"><i class="<?= e(str_replace('oc-', '', $component->pluginIcon)) ?>"></i> <?= $name ?></span>
             <span class="description"><?= $description ?></span>
             <span class="alias oc-icon-code"><?= e($component->alias) ?></span>
             <input type="hidden" name="component_properties[]" data-inspector-values value="<?= e($this->getComponentPropertyValues($component)) ?>" />
--- a/modules/cms/formwidgets/components/partials/_component.htm
+++ b/modules/cms/formwidgets/components/partials/_component.htm
@@ -6,7 +6,7 @@
         data-inspector-description="<?= $description = e($this->getComponentDescription($component)) ?>"
         data-inspector-config="<?= e($this->getComponentsPropertyConfig($component)) ?>"
         data-inspector-class="<?= get_class($component) ?>">
-            <span class="name"><i class="<?= e(preg_replace('/^oc\-/', '', $component->pluginIcon)) ?>"></i> <?= $name ?></span>
+            <span class="name"><i class="<?= e($component->pluginIcon) ?>"></i> <?= $name ?></span>
             <span class="description"><?= $description ?></span>
             <span class="alias oc-icon-code"><?= e($component->alias) ?></span>
             <input type="hidden" name="component_properties[]" data-inspector-values value="<?= e($this->getComponentPropertyValues($component)) ?>" />
--- a/modules/cms/formwidgets/components/partials/_component.htm
+++ b/modules/cms/formwidgets/components/partials/_component.htm
@@ -6,7 +6,7 @@
         data-inspector-description="<?= $description = e($this->getComponentDescription($component)) ?>"
         data-inspector-config="<?= e($this->getComponentsPropertyConfig($component)) ?>"
         data-inspector-class="<?= get_class($component) ?>">
-            <span class="name"><i class="<?= e($component->componentCssClass) ?>"></i> <?= $name ?></span>
+            <span class="name"><i class="<?= $component->pluginIcon ?>"></i> <?= $name ?></span>
             <span class="description"><?= $description ?></span>
             <span class="alias oc-icon-code"><?= e($component->alias) ?></span>
             <input type="hidden" name="component_properties[]" data-inspector-values value="<?= e($this->getComponentPropertyValues($component)) ?>" />
--- a/modules/cms/formwidgets/components/partials/_component.htm
+++ b/modules/cms/formwidgets/components/partials/_component.htm
@@ -6,7 +6,7 @@
         data-inspector-description="<?= $description = e($this->getComponentDescription($component)) ?>"
         data-inspector-config="<?= e($this->getComponentsPropertyConfig($component)) ?>"
         data-inspector-class="<?= get_class($component) ?>">
-            <span class="name"><i class="<?= e(str_replace('oc-', '', $component->pluginIcon)) ?>"></i> <?= $name ?></span>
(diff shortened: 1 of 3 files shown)


LukeTowers:
The existing icon styling is present in lines 52-58: https://github.com/octobercms/october/pull/5429/files#diff-7ff10206102b01f74ffe8fa50bc453d6557d7a08b9bb0e472aa2d91f766fc167R52-R58 so it would need to be tested if it can be removed without affecting the three cases it applies to (`.draggable-component-item,
.component-list .components div.layout-cell,
div.control-componentlist div.components div.layout-cell`) and those styles as they relate to icon size and opacity would need to be replicated on the new icon location.

ayumi-cloud:
https://github.com/octobercms/october/blob/8851a1965149300481f4004bd9f25cf222d36e29/modules/cms/assets/less/october.components.less#L34-L58

summercms:
@LukeTowers 

Ok updated the styling and tested it out.

Explanation:

```css
div.control-componentlist div.components div.layout-cell > div span.name {
    padding: 8px 20px 0 38px;
}
```

Changed the right padding from `15px` to `20px`, the reason was the title was covering the closing icon (cross) so now there is a `5px` gap.

```css
div.control-componentlist div.components div.layout-cell > div span.name i {
    position: absolute;
    font-size: 16px;
    left: 15px;
    top: 10px;
    opacity: 0.7;
}
```

For `<div>` the styling was `top: 7px` and for `<i>` the styling is now `top: 10px`. This was tested and gave the exact same output.

Test results:

image

Above: Before PR.

image

Above: With PR (Note the 1st component has Font Awesome 6 icon).

This pr is done and ready for testing now.

LukeTowers:
If the existing styling no longer applies to anything because of this change then it should be removed as well.

summercms:
@LukeTowers Done.

modules/cms/formwidgets/components/partials/_component.htm
```diff
@@ -1,17 +1,17 @@
...
     <div
-        class="<?= 'oc-'.$component->pluginIcon ?> layout-relative"
+        class="layout-relative"
         <?php if ($component->inspectorEnabled): ?>data-inspectable<?php endif ?>
         data-inspector-title="<?= $name = e($this->getComponentName($component)) ?>"
         data-inspector-description="<?= $description = e($this->getComponentDescription($component)) ?>"
         data-inspector-config="<?= e($this->getComponentsPropertyConfig($component)) ?>"
         data-inspector-class="<?= get_class($component) ?>">
-            <span class="name"><?= $name ?></span>
+            <span class="name"><i class="<?= $component->pluginIcon ?>"></i> <?= $name ?></span>
```
LukeTowers: Doesn't this break it because now pluginIcon would be forced to include the `oc-` prefix?

ayumi-cloud: Good point, I will use `str_replace()`

ayumi-cloud: @LukeTowers Please review and let me know your thoughts: https://github.com/octobercms/october/pull/5429/commits/4652d2902301ef64190cef24461a4eaf1fcbec5e

ayumi-cloud:
Link: https://octobercms.com/docs/ui/icon

Code example:

```html
<i class="icon-camera-retro"></i> icon-camera-retro

<html class="oc-icon-flag-checkered">oc-icon-flag-checkered</html>
```

Adding this here to save me going back and forth.

modules/cms/formwidgets/components/partials/_component.htm
ayumi-cloud: @LukeTowers ok please review this comment: https://github.com/octobercms/october/pull/5429/commits/fac649da1940cc79a6e80fee1f68aa8162221b90

summercms: @LukeTowers Your comment doesn't apply to the cms's setup.

The incoming icon prefix into the plugin is: `icon-flag-checkered` (without the **oc-**). The reason why I opened this issue up in the first place was because October was injecting a hard-coded `oc-` into the string. See code line:

https://github.com/octobercms/october/blob/develop/modules/cms/formwidgets/components/partials/_component.htm#L3

If someone put this `oc-icon-flag-checkered` into their plugin component, then the output from October would be this:

```
oc-oc-icon-flag-checkered
```

Notice the double `oc-oc-`.

So before this issue was open, people could not add `oc-` for the icon plugin component.

So there is no need to remove it, when people haven't been able to add it in the first place.

The sidebar menu in the `cms` section and the **Builder** plugin use `<i>` and not `<div>` to display the icon. This pr makes the plugin's component icon be displayed using `<i>` throughout the cms - something that should've been coded from day one.

ayumi-cloud:
### No Validation rules

Another issue with October CMS is that it has **never** validated the code properly!

A user could enter an icon starting with `oc-` and the cms doesn't give a warning and screws up!

The cms should either give a warning message or strip the `oc-` from the plugin components icon string.

I was just looking at the file here: https://github.com/octobercms/october/blob/develop/modules/cms/formwidgets/Components.php

October does do anything for icons right now.

summercms:
I hate github, it's rubbish to write comments in having to click between `preview` and `write` back and forth. Gonna write my comment in a proper markdown code editor...

summercms:
@LukeTowers Options for you to choose and be aware of - personally I just choose option one and allow this error to take place.

### No Validation rules

Another issue with October CMS is that it has **never** validated the code properly!

A user could add an icon starting with `oc-` and the cms doesn't give a warning and screws up! 

Currently the error would be like this:

### Location 1

image

Above: Is using `<div>` so the output would be this:

```
oc-oc-icon-flag-checkered
```

You will notice the error and would need to remove the `oc-oc-`

#### PR

This PR would fix that issue and output this instead:

```
oc-icon-flag-checkered
```

You probably won't notice the error (due to using `<i>`), but it would be better to remove the `oc-`

### Location 2

image

Above: Is using `<i>` so the output would be this:

```
oc-icon-flag-checkered
```

You probably won't notice the error (due to using `<i>`), but it would be better to remove the `oc-`

### Location 3

image

Above: Is using `<i>` so the output would be this:

```
oc-icon-flag-checkered
```

You probably won't notice the error (due to using `<i>`), but it would be better to remove the `oc-`

### Conclusion

There are three options to choose:

1. Just allow `oc-` with `<i>` element. October was designed to use `oc-` with `<div>` and not `<i>` - but doesn't give any errors when using `oc-` with `<i>`.

2. The cms displays a `warning message` saying to remove the `oc-` when adding to a plugin component. This may annoy users and you may want to just use option 1.

3. The cms strips the `oc-` from the code in a validation check. This won't annoy users, but you need to ask is it necessary as you could just do option 1 and allow it.

The file location:

https://github.com/octobercms/october/blob/develop/modules/cms/formwidgets/Components.php

Doesn't show any Validation for icons.

summercms:
Note: If you choose option one. Then this pr is finished now.

LukeTowers:
It should be fine as it is now, looked into it more and the `oc-` prefix is specifically to allow for the icon to be injected in a `:before` pseudo element without affecting the font-family of the element itself (which the base class of just `icon-` does).

@bennothommo any input on this?

PR #5430, opened 2020-12-24
Add warning message when trying to send an ajax request without using a X-XSRF-TOKEN

summercms:
[edit] I can't tell if this is needed or not? Bit confused right now on how October deals with CSRF.

image

Linking to this conversation: https://github.com/octobercms/october/pull/4701


LukeTowers:
No? There is no point to this as far as I can tell.

Issue #5435, opened 2020-12-28
Proposal: Implement new cross origin security APIs


summercms:
Over the last few months browsers have been adding many new and exciting api's to get a grip on `cross-origin` attacks.

Over the coming weeks to months I would like to slowly upgrade October CMS to start locking down these types of attacks.

### Example of attack

In 2018 the crypto exchange gate.io got hacked. They were using a third party script to measure analytics from the company statcounter.

The hackers didn't touch the gate.io exchange code at all.

Instead they hacked the code hosted on the location `www.statcounter[.]com/counter/counter.js`

When a user went to a certain url on the gate.io exchange website, the code in the `counter.js` would detect that url and perform the attack, under certain criteria. In this case the code was searching for people trading bitcoin and confirmed they have bitcoin in their wallets, before sending the confirmed bitcoin to the hackers address. People who didn't have bitcoin but other cryptocurrencies the hackers code left alone.

### October CMS

To me the backend of the cms should be harden and isolated from the whole world, only allowing the admins access!

This means two things to me.

1. The CMS requesting files and resources from third-parties through cross-origins need to be locked down.
2. Plugin and theme developers need to understand this attack. There are many plugins in the marketplace that allow questionable unchecked third-party cross-origin code to be loaded into the backend of the cms with little to no checks.

### Hardening October

Three main ideas that come to mind are as follows:

1. Start adding these new api's into October and allow users to customize the options in the `config` files. For example:

```php
    /*
    |--------------------------------------------------------------------------
    | Cross-Origin-Opener-Policy
    |--------------------------------------------------------------------------
    |
    | The HTTP Cross-Origin-Opener-Policy (COOP) response header 
    | allows you to ensure a top-level document does not share a 
    | browsing context group with cross-origin documents.
    | 
    | COOP will process-isolate your document and potential attackers 
    | can't access to your global object if they were opening it in a popup, 
    | preventing a set of cross-origin attacks dubbed XS-Leaks.
    |
    | `unsafe-none` - This is the default value. Allows the document to be 
    | added to its opener's browsing context group unless the opener itself 
    | has a COOP of same-origin or same-origin-allow-popups.
    | 
    | `same-origin-allow-popups` - Retains references to newly opened 
    | windows or tabs which either don't set COOP or which opt out of 
    | isolation by setting a COOP of unsafe-none.
    | 
    | `same-origin` - Isolates the browsing context exclusively to same-origin
    | documents. Cross-origin documents are not loaded in the same 
[...]

PR #5436, opened 2020-12-28
Add crossorigin attribute to images being loaded from third-party locations

summercms:
Add crossorigin protection to images be loaded from third-party locations

See github issue: https://github.com/octobercms/october/issues/5435

### DIFF
--- a/modules/backend/layouts/_mainmenu.htm
+++ b/modules/backend/layouts/_mainmenu.htm
@@ -26,7 +26,7 @@
                                 <?php if ($item->iconSvg): ?>
                                     <img
                                         src="<?= Url::asset($item->iconSvg) ?>"
-                                        class="svg-icon" loading="lazy" width="30" height="30" />
+                                        class="svg-icon" loading="lazy" width="30" height="30" crossorigin>
                                 <?php endif ?>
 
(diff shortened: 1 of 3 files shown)


LukeTowers:
What does this attribute achieve?

summercms:
@LukeTowers 

~~The connected issue wasn't sure if you want to put it in the `discussions` section or not by the way?~~

> What does this attribute achieve?

A picture can tell you a thousand words.

image

Above: The new api's turned on - without this pr.

image

Above: The new api's turned on - with this pr.

Note: These images are being loaded `crossorigin` from `gateway.october.com` to the cms backend.

LukeTowers:
When is chrome enabling this API by default?

summercms:
@LukeTowers 

> When is chrome enabling this API by default?

I just checked, the api spec has now been moved into the `html standard` spec (so that's a major move).

Chrome added it to stable version 83 - so it's been around for a while. It's now out of the flag in Canary.

```
Don't merge this pr for now.
```

Not sure if you want to close this pr or put a label on it like `Status: Blocked` ?

I feel this issue needs to be addressed in the future - but can be left alone right now.

summercms:
My reason is because the browser is telling me that the `gateway.october.com` needs to add security headers - because it's cors policy is not giving the complete protection. That's something the admins need to address in the future and I prefer not to mess around with the `gateway.october.com` server for now.

PR #5454, opened 2021-01-13
PHP 8 compatibility

ghost:


LukeTowers:
@tolkamsoft please provide an explanation for each change

tests/unit/backend/models/ExportModelTest.php
```diff
@@ -50,6 +50,11 @@ public function testEncodeArrayValue()
 
     public function testDownload()
     {
+        // MIME detection is improved in PHP8
+        $expectedContentType = PHP_VERSION_ID >= 80000 && extension_loaded('fileinfo')
```
Tolkam-PHP-Components: MIME detection is improved in PHP8 (fileinfo extension), so csv is now detected correctly, not text/plain as before

tests/fixtures/plugins/testvendor/goto/Plugin.php
```diff
@@ -1,4 +1,4 @@
-<?php namespace TestVendor\Goto;
+<?php namespace TestVendor\_Goto;
```
Tolkam-PHP-Components: before php8 "goto" was a reserved word and was not allowed as part of the namespace. Now test checks for validity of plugin namespace according to PSR-4

modules/backend/classes/BackendController.php
```diff
@@ -111,6 +111,14 @@ public static function extend(Closure $callback)
         self::extendableExtendCallback($callback);
     }
 
+    /**
+     * @inheritDoc
+     */
+    public function callAction($method, $parameters)
+    {
+        return parent::callAction($method, array_values($parameters));
```
Tolkam-PHP-Components, line 119: In php8 named parameters were introduced and now it is required to match called method parameter name when setting parameters by array destructing or `call_user_func_array()` etc.

modules/system/classes/CombineAssets.php
```diff
@@ -464,7 +464,7 @@ protected function prepareCombiner(array $assets, $rewritePath = null)
      */
     protected function setHashOnCombinerFilters($hash)
     {
-        $allFilters = call_user_func_array('array_merge', $this->getFilters());
+        $allFilters = array_merge(...array_values($this->getFilters()));
```
Tolkam-PHP-Components, line 467: In php8 named parameters were introduced and now it is required to match called method parameter name when setting parameters by array destructing or call_user_func_array() etc.

tests/fixtures/plugins/testvendor/goto/Plugin.php
bennothommo: @tolkamsoft The parse error here was intentional. It was to test that a [plugin with a parse error](https://github.com/octobercms/october/blob/develop/tests/unit/system/classes/PluginManagerTest.php#L52-L57) is not loaded. If having `goto` in the namespace is no longer a parse error, you would need to introduce a parse error somewhere else in this file.

tests/unit/backend/models/ExportModelTest.php
bennothommo: @tolkamsoft I'd be happy for the expected content type to be `application/csv` or `text/plain`, irrespective of PHP version. The MIME detection can be improved in PHP 7 in other ways, so it's possible this is already happening for people. Would you be willing to refactor this test slightly to accept either content types as acceptable without the checks for PHP8 or `fileinfo`?

Tolkam-PHP-Components: Actually I'm not sure that testing the detected content type is meaningful. AFAIK this depends on server configuration (uses magic.mime file to detect mime type). So I'd prefer not to test for it at all.

bennothommo: @tolkamsoft You're correct, but most of the time, people use the bundled magic database in PHP, so it's reasonable enough to assume that this test will pass in the majority of cases. The test is there so it doesn't come out in some wacky format like `video/mpeg4` or something else very wrong.

tests/fixtures/plugins/testvendor/goto/Plugin.php
Tolkam-PHP-Components: Actually I was referring to test plugin description:

> Test plugin used by unit tests to detect plugins with invalid namespaces

Tolkam-PHP-Components: Fixed now

composer.json
```diff
@@ -37,7 +37,7 @@
         }
     ],
     "require": {
-        "php": ">=7.2",
+        "php": ">=7.3",
```
Tolkam-PHP-Components, line 40: Updated minimum php version to 7.3 to force to use compatible with php8 version of doctrine/dbal (2.12)

Tolkam-PHP-Components:
php 7.x tests are failing because of class 'October\Rain\Support\Facades\Mail' not found. Not sure why?

php 8.x tests failing because of october/rain version. Should I require the forked one in composer.json?

composer.json
bennothommo: @tolkamsoft I'd rather us only require `doctrine/dbal:2.12.x` if using PHP 8, as opposed to upping the minimum PHP version to 7.3. The October 1.1 branch supports PHP 7.2 for the moment.

bennothommo:
@tolkamsoft If you can get the unit tests working locally on PHP8, and can send us a screenshot of them passing in PHPUnit, then I'm happy to merge this into the PHP 8 compatibility branch in lieu of the tests failing here - it's likely failing because of our current testing action setup. You will need to replace `vendor/october/rain` with the changes you've made in your [library PR](https://github.com/octobercms/library/pull/554) when running your tests.

Tolkam-PHP-Components:
@bennothommo Maybe it's better to add php8 test suite for https://github.com/octobercms/library/pull/554 as well? Could you do that, please?

OK, here is the tests ran - https://github.com/tolkamsoft/oc-library/pull/1

composer.json
Tolkam-PHP-Components: @bennothommo nope, this would not work - doctrine/dbal 2.12.0 requires php ^7.3 || ^8

Tolkam-PHP-Components:
@bennothommo any progress on this? Tests ran - tolkamsoft/oc-library#1

composer.json
bennothommo: @tolkamsoft But the only reason we need `2.12.0` is for PHP 8 compatibility, so we can still use the older versions of DBAL for PHP 7.2?

Tolkam-PHP-Components: @bennothommo I'm not sure it is possible since doctrine/dbal prior v2.12 woudn't work with PHP8 and it's minimum PHP version required is 7.3.

PR #5460, opened 2021-01-15
localize all installed themes

mjauvin:
Related to https://github.com/octobercms/october/pull/4960

Opening a discussion regarding a "recent" addition making possible to localize the CURRENTLY ACTIVE.

The problem is that all other installed themes are NOT localized so that all labels using string localization will fail to localize, making theme customization difficult at best before it actually gets activated.

I propose the following change which solves the issue by looping through all installed themes. 
This also uses the theme's code instead of the theme's directory in case the user installs the theme under a different folder.

### DIFF
--- a/modules/cms/ServiceProvider.php
+++ b/modules/cms/ServiceProvider.php
@@ -313,7 +313,8 @@ class ServiceProvider extends ModuleServiceProvider
         foreach (CmsTheme::all() as $theme) {
             $langPath = $theme->getPath() . '/lang';
             if (File::isDirectory($langPath)) {
-                Lang::addNamespace('themes.' . $theme->getId(), $langPath);
+                $config = $theme->getConfig();
+                Lang::addNamespace('themes.' . array_get($config, 'code'), $langPath);
             }
         }
     }
--- a/modules/cms/ServiceProvider.php
+++ b/modules/cms/ServiceProvider.php
@@ -314,7 +314,9 @@ class ServiceProvider extends ModuleServiceProvider
             $langPath = $theme->getPath() . '/lang';
             if (File::isDirectory($langPath)) {
                 $config = $theme->getConfig();
-                Lang::addNamespace('themes.' . array_get($config, 'code'), $langPath);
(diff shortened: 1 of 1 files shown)


LukeTowers:
First problem, the `code` property is optional. Second, why do you need to load every single theme, even the non-active ones? Why do we let non-activated themes be customized?

algoriq:
> First problem, the `code` property is optional. Second, why do you need to load every single theme, even the non-active ones? Why do we let non-activated themes be customized?

In the first case, I think it is worth checking in case the theme code is not specified, then use the directory path.

In the second case, at the moment, users have the ability to edit the customize theme settings of an inactive theme, since the "Customize" button is not disabled for an inactive theme.

mjauvin:
1. the documentation [here](https://octobercms.com/docs/themes/development#localization) mentions the theme-code (i.e. `themes.theme-code::lang.key`) so this should be used first if defined, otherwise we can revert to the theme dirName.
2. Either we disable the customize button or we make all installed themes localizable. I think it can be interesting to be able to customize a theme BEFORE it gets activated in case you are changing the theme on a LIVE website.

algoriq:
> 1. the documentation [here](https://octobercms.com/docs/themes/development#localization) mentions the theme-code (i.e. `themes.theme-code::lang.key`) so this should be used first if defined, otherwise we can revert to the theme dirName.
> 2. Either we disable the customize button or we make all installed themes localizable. I think it can be interesting to be able to customize a theme BEFORE it gets activated in case you are changing the theme on a LIVE website.

I completely agree with you.

mjauvin:
@bennothommo @LukeTowers what's your take on this, should we disable/hide the "customize" buttons for inactive themes or should we allow them to be localized as suggested in this PR?

bennothommo:
@mjauvin IMO, I'd just disable the customize button for inactive themes. No use in customizing it if you cannot see the changes.

mjauvin:
Closing this in favor of another PR to disable the customize button on inactive themes.

PR #5465, opened 2021-01-19
implement artisan october:util purge orphans

mjauvin:


bennothommo:
@mjauvin Orphaned file objects could also be File records with an `attachment_id` and `attachment_type` both set, but pointing to a related record that no longer exists. These would also need to be cleared out by this command.

bennothommo:
I also think the reverse should be cleared - if the File record links to a valid related record but the **file itself** no longer exists in the filesystem (these should perhaps be cleared when a command flag is set, eg. `--missing-files`)

LukeTowers:
@bennothommo could you describe a situation where that would happen and where you'd want the database to be affected? I'm just thinking that you'd want the opposite to occur, the database record to stay behind even when the file is deleted in some cases (i.e. caching of generated files), although to be fair that's a bit of a stretch and could just as easily be achieved with detecting the relationship no longer existing.

I'm just a bit leery of modifying the database from such a terminal command, it feels a bit unexpected and could have negative impacts, especially if people are using validation rules to make sure a given record doesn't exist without a file relationship.

bennothommo:
@LukeTowers I'm mainly thinking that both scenarios are essentially an "orphaned" reference - whether it's the relation that no longer exists, or the file itself. I do know there are scenarios where the file missing is fine - for example, working on a copy of the website for debugging, without necessarily copying over the full storage - this is why I wouldn't necessarily clear them by default, but allow people to option to do so through a flag.

mjauvin:
@LukeTowers @bennothommo any idea why those tests are suddenly failing? It's not even related to this PR.

mjauvin:
I think [this is the commit](https://github.com/octobercms/library/commit/5c312b7ca4cf60b00d5614498dbbf79fc80fe7ab) that broke it.

mjauvin:
@LukeTowers  I verified the above assumption, the following line is breaking the test:

https://github.com/octobercms/library/commit/5c312b7ca4cf60b00d5614498dbbf79fc80fe7ab#diff-b0185ec47eb87315ad4b78f740db185dd938ca525888d249d820c1a7d5be9754R225

mjauvin:
The tests are fixed by https://github.com/octobercms/library/pull/556

modules/system/console/OctoberUtil.php
```diff
@@ -339,7 +340,33 @@ protected function utilPurgeOrphans()
             return;
         }
 
-        // @todo
+        $isLocalStorage = Config::get('cms.storage.uploads.disk', 'local') === 'local';
+        $orphanedFiles = FileModel::whereNull('attachment_id')->orWhereNull('attachment_type')->delete();
+
+        foreach (FileModel::all() as $file) {
```
LukeTowers: Why not just a single `FileModel::whereNotHas('attachment')` instead of loading each file model's attachment?

mjauvin: Because this is a `morphTo` relation and I have to specify types for the query:
```
FileModel::whereDoesntHaveMorph('attachment', $types);
```
Beside, I need to be able to check if the file exists on disk when the `--missing-files` argument is passed.

mjauvin: Ok, nevermind the "types" argument, I can just pass '*' for types and it works for any model.

2nd argument holds, though.

mjauvin: Ok, what about this?

mjauvin:
@LukeTowers @bennothommo what do you think of that last implementation?

modules/system/console/OctoberUtil.php
```diff
@@ -339,7 +340,26 @@ protected function utilPurgeOrphans()
             return;
         }
 
-        // @todo
+        $isLocalStorage = Config::get('cms.storage.uploads.disk', 'local') === 'local';
+
+        $orphanedFiles = FileModel::whereNull('attachment_id')->orWhereNull('attachment_type')->delete();
+        $orphanedFiles += FileModel::whereDoesntHaveMorph('attachment', '*')->delete();
```
LukeTowers: Doesn't this query do the same thing as the previous one as well as evaluating if there's actually related records? Also, calling `->delete()` on the query builder will not delete the underlying files themselves, the delete() method on the model needs to be called instead.

mjauvin: No, the second query doesn't catch all the cases.

I was looping through at some point and you questioned this method @LukeTowers ... !?

bennothommo:
@mjauvin does this work with external storage providers like Amazon S3?

modules/system/console/OctoberUtil.php
LukeTowers: You were looping through every single file on the system and making a separate DB call for each file to attempt to load the relationship. If you had only 1000 records in the system this would load 1000-2000 separate models and make 1001 separate DB queries.

Can you do `FileModel::whereDoesntHaveMorph('attachment', '*')->orWhereNull('attachment_id')->orWhereNull('attachment_type')->get()` and then loop over those results calling delete()?

mjauvin: The problem with this is that I will still need to go through each file model if the "--missing-files" option is provided...

mjauvin:
> @mjauvin does this work with external storage providers like Amazon S3?

Except for the "--missing-files", I would think it does but I didn't test this.

bennothommo:
@mjauvin Looks good to me. We'll probably aim to put this in with OCMS 1.1.3, so might try and test it after 1.1.2 is out.

daftspunk:
Interesting approach

PR #5470, opened 2021-01-23
Fix: Group filter type with modelClass does not work when 500+ options are available

### LINKED ISSUE #5459: List filters adds no items after 500 records
- OctoberCMS Build: 1.1 ### 0634060195cac6c558d46b1d3d906d77e24afd00
- PHP Version: 7.3 / 7.4
- Database Engine: MySQL
- Plugins Installed: Translate/Builder/Pages

### Description:

In the list filters items after 500 are not added. When you search for records and try to add an item after 500 the Apply and Clear button don't show up. A trick with adding one from the first 500 that shows these buttons and adding one after the 500's item only adds the one in the first 500. 

### Steps To Reproduce:
See screenshot. Create a filter with more then 500 records, search for record > 500 and click on + sign. No Apply and Clear button appear.

danielbidala:
This PR solves #5459 

`config_filter.yaml` example:

```
scopes:
    customer:
        label: Client
        type: group
        modelClass: Brightstaff\Brightcrm\Models\Client
        conditions: customer_id in (:filtered)
        nameFrom: name
```

Maybe it's enough to have limit(50) instead of limit(100) in Filter.php?

### DIFF
--- a/modules/backend/widgets/Filter.php
+++ b/modules/backend/widgets/Filter.php
@@ -391,12 +391,6 @@ class Filter extends WidgetBase
         $query = $model->newQuery();
         $query->limit(100);
 
-        // If scope has active filter(s) run additional query and later merge it with base query
-        if ($scope->value) {
-            $modelIds = array_keys($scope->value);
-            $activeOptions = $model::findMany($modelIds);
-        }
-
         /**
          * @event backend.filter.extendQuery
          * Provides an opportunity to extend the query of the list of options
@@ -421,17 +415,21 @@ class Filter extends WidgetBase
         $this->fireSystemEvent('backend.filter.extendQuery', [$query, $scope]);
 
         if (!$searchQuery) {
+            // If scope has active filter(s) run additional query and later merge it with base query
+            if ($scope->value) {
+                $modelIds = array_keys($scope->value);
+                $activeOptions = $model::findMany($modelIds);
+            }
+
             $modelOptions = isset($activeOptions)
                 ? $query->get()->merge($activeOptions)
                 : $query->get();
+                
             return $modelOptions;
         }
 
         $searchFields = [$model->getKeyName(), $this->getScopeNameFrom($scope)];
-        $modelOptions = isset($activeOptions)
-            ? $query->searchWhere($searchQuery, $searchFields)->get()->merge($activeOptions)
-            : $query->searchWhere($searchQuery, $searchFields)->get();
-        return $modelOptions;
+        return $query->searchWhere($searchQuery, $searchFields)->get();
     }
 
     /**
--- a/modules/backend/widgets/Filter.php
+++ b/modules/backend/widgets/Filter.php
@@ -336,7 +336,7 @@ class Filter extends WidgetBase
      * Returns the available options a scope can use, either from the
      * model relation or from a supplied array. Optionally apply a search
      * constraint to the options.
-     * @param  string $scope
+     * @param  \Backend\Classes\FilterScope $scope
      * @param  string $searchQuery
      * @return array
      */
@@ -380,6 +380,8 @@ class Filter extends WidgetBase
 
     /**
      * Looks at the model for defined scope items.
+     * @param \Backend\Classes\FilterScope $scope
+     * @param string $searchQuery
      * @return Collection
      */
     protected function getOptionsFromModel($scope, $searchQuery = null)
@@ -387,12 +389,13 @@ class Filter extends WidgetBase
         $model = $this->scopeModels[$scope->scopeName];
 
         $query = $model->newQuery();
+        $query->limit(100);
 
-        /*
-         * The 'group' scope has trouble supporting more than 500 records at a time
-         * @todo Introduce a more advanced version with robust list support.
-         */
-        $query->limit(500);
+        // If scope has active filter(s) run additional query and leter merge it with base query
+        if ($scope->value) {
+            $modelIds = array_keys($scope->value);
+            $activeOptions = $model::findMany($modelIds);
+        }
 
         /**
          * @event backend.filter.extendQuery
@@ -418,11 +421,17 @@ class Filter extends WidgetBase
         $this->fireSystemEvent('backend.filter.extendQuery', [$query, $scope]);
 
         if (!$searchQuery) {
-            return $query->get();
+            $modelOptions = isset($activeOptions)
+                ? $query->get()->merge($activeOptions)
+                : $query->get();
+            return $modelOptions;
         }
 
         $searchFields = [$model->getKeyName(), $this->getScopeNameFrom($scope)];
-        return $query->searchWhere($searchQuery, $searchFields)->get();
+        $modelOptions = isset($activeOptions)
+            ? $query->searchWhere($searchQuery, $searchFields)->get()->merge($activeOptions)
+            : $query->searchWhere($searchQuery, $searchFields)->get();
+        return $modelOptions;
     }
 
     /**
--- a/modules/backend/widgets/Filter.php
+++ b/modules/backend/widgets/Filter.php
@@ -391,7 +391,7 @@ class Filter extends WidgetBase
         $query = $model->newQuery();
         $query->limit(100);
 
-        // If scope has active filter(s) run additional query and leter merge it with base query
+        // If scope has active filter(s) run additional query and later merge it with base query
         if ($scope->value) {
             $modelIds = array_keys($scope->value);
             $activeOptions = $model::findMany($modelIds);
--- a/modules/backend/widgets/Filter.php
+++ b/modules/backend/widgets/Filter.php
@@ -389,7 +389,9 @@ class Filter extends WidgetBase
         $model = $this->scopeModels[$scope->scopeName];
 
         $query = $model->newQuery();
-        $query->limit(100);
+        
+        // @todo Implement support for lazy loading of options
+        $query->limit(500);
 
         /**
          * @event backend.filter.extendQuery
@@ -415,7 +417,7 @@ class Filter extends WidgetBase
         $this->fireSystemEvent('backend.filter.extendQuery', [$query, $scope]);
 
         if (!$searchQuery) {
-            // If scope has active filter(s) run additional query and later merge it with base query
+            // If scope has active filter(s) run additional query and merge it with base query
             if ($scope->value) {
                 $modelIds = array_keys($scope->value);
                 $activeOptions = $model::findMany($modelIds);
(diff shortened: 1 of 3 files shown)


modules/backend/widgets/Filter.php
```diff
@@ -380,19 +380,16 @@ protected function filterActiveOptions(array $activeKeys, array &$availableOptio
...
         $model = $this->scopeModels[$scope->scopeName];
 
         $query = $model->newQuery();
-
-        /*
-         * The 'group' scope has trouble supporting more than 500 records at a time
-         * @todo Introduce a more advanced version with robust list support.
-         */
-        $query->limit(500);
+        $query->limit(100);
```
bennothommo: @danielbidala Would you be able to explain how dropping the limit to 100 is going to help with the original issue? Wouldn't this just cut off items after 100?

modules/system/assets/ui/js/filter.js
```diff
@@ -302,6 +302,11 @@
 
         if (item)
             toItems.push(item)
+        else
+            toItems.push({
+                'id': itemId,
+                'name': $item.text()
+            })
```
bennothommo: @danielbidala Please wrap conditions with curly brackets in JS:

```suggestion
        if (item) {
            toItems.push(item)
        } else {
            toItems.push({
                'id': itemId,
                'name': $item.text()
            })
        }
```

bennothommo (changes requested):

modules/backend/widgets/Filter.php
danielbidala: @bennothommo When widget loads it gets the first 100 records from the model and populates `scopeAvailable` js array with items on client side and renders `<ul>`.
When you select or deselect an item the `selectItem` method in `filter.js` checks whether the selected item is in the `scopeValues` (active) or `scopeAvailable` array and moves it to the opposite.

**TILL NOW:** When you searched in the group filter another query ran and the `<ul>` was refreshed with search results but the `scopeAvailable` array was not refreshed.

**WITH THIS PR:** If the selected item is not in any of the arrays (`scopeValues` or `scopeAvailable`), `selectItem` method will push it to the right place and you can apply the filter. When page reloads `getOptionsFromModel` in `Filter.php` checks if there are active filters in this scope and if yes it runs additional query to get those records from model and merges the result with the base `$query->limit(100)`. So if you have an active filter which is not in the first 100, it will be in `scopeAvailable` array anyway.

Thanks to this maybe we can limit the query to 50 instead of 100, because the user won't scroll 100 option, he will search instead. What's your opinion?

modules/system/assets/ui/js/filter.js
danielbidala: @bennothommo I added a commit with curly brackets in conditions.

bennothommo:
@danielbidala I don't know how painful it would be to implement, but as an idea, could we perhaps "lazy load" more results when someone scrolls down to the bottom of the options list, to get the next 100 items? I can't discount that some crazy people out there with masses of options are scrolling through, say, an alphabetised list of options and know what they're looking for.

danielbidala:
> I don't know how painful it would be to implement, but as an idea, could we perhaps "lazy load" more results when someone scrolls down to the bottom of the options list, to get the next 100 items?

@bennothommo What if we simply let the developer to define a `limit` property in `config_filter.yaml` and leave the default limit at 500?

bennothommo:
@danielbidala it just feels like a "band-aid" solution doing it that way. I'm sure there's a performance decrease as number of items increases due to the fact that each of the items is essentially a couple of DOM elements. It might also run afoul of the the PHP `max_input_vars` problem that has been cropping up in a couple of spots when people are working with large datasets - especially if people try to increase the number to a large amount like say 1000+.

LukeTowers:
I'm not sure the max input thing is a problem anymore, I thought it had been tweaked to change how the values were sent (in JSON instead of multiple input vars)

bennothommo:
@LukeTowers you're right, however, there's still the DOM issue.

danielbidala:
@bennothommo I think we can trust developers to know what they’re doing in this case. According to what you say, we can set Laravel Query Builder's limit to a fixed value in the whole appliciation, so that accidentally someone does not query too much data at once.

remko-key:
@danielbidala  @bennothommo Is it an idea to make the limit and configurable option with a default of 500?

bennothommo:
@danielbidala

> I think we can trust developers to know what they’re doing in this case.

I'd like to believe that, but when I see people complaining that the Media section is slow because they've uploaded 80,000 unorganised images, it tends to change my opinion a little.

I still would like to see lazy-loading if possible - there's really no reason to try and load a huge amount of items into what amounts to a scrolling and filtering UI element.

danielbidala:
@bennothommo Unfortunatly I don't have time to implement this now, it would require significant changes. Anyway I think this PR could be merged with the original `$query->limit(500)`. This way the group filter will be usable with massive amount of related records, and there won't be any change in performance compared to the current.

@meinmotion We use this PR in production, it works as expected.

modules/backend/widgets/Filter.php
LukeTowers: @danielbidala could you restore the 500 limit and the comment, but adjust it to say `@todo Implement support for lazy loading of options`?

modules/system/assets/ui/js/filter.js
```diff
@@ -295,13 +295,20 @@
...
+        }
 
-        if (item)
+        if (item) {
             toItems.push(item)
+        } else {
+            toItems.push({
+                'id': itemId,
+                'name': $item.text()
+            })
```
LukeTowers: Why was this change required? Also, what is this variable even used for?

danielbidala: @LukeTowers As the searched items added only to the DOM (group filter's `<ul>`), we need to add the item to `scopeValues` or `scopeAvailable` array when it is clicked.

If `item` is selected, `toItems` equals `scopeValues`. If `item` is deselected, `toItems` equals `ScopeAvailable`. 

The `selectItem` method checks the clicked `item` object existance in `scopeValues` and in `scopeAvailable` and push it to the opposite (according it's selected or deselected). If the `item` doesn't exist in none of the arrays, we create the object from the clicked `<li>` and push it anyway.

LukeTowers: I'm not seeing where the `toItems` variable is used though, could you link me to it?

LukeTowers: @danielbidala ^

danielbidala: @LukeTowers I'm not sure I understand your question. The `toItems` array is just a reference to `scopeValues` or `scopeAvailable` array depending on whether the element was selected or deselected. It is used only in `selectItem` method.

LukeTowers: @bennothommo maybe you know what I'm asking? It looks like the toItems variable is populated here but never actually used anywhere else. I don't see a reason for it to exist period.

bennothommo: @LukeTowers yeah, it appears to be unused. I looked through the Git blame and saw it was added about 7 years ago in its original form, but wasn't even used then as far as I can tell.

modules/backend/widgets/Filter.php
danielbidala: @LukeTowers I added a new commit according to your suggestion.

modules/system/assets/ui/js/filter.js
LukeTowers: Let's just get rid of it then. @danielbidala not sure why a change was needed to a variable that doesn't get used, can you explain the logic behind that?

danielbidala: @LukeTowers The variable and the logic was there I've just extended it :) Anyway I added a commit to remove toItems variable.

remko-key:
Hello there,
Any news about this issue? At this point items searched over 500 are not filtered.

alxy:
I think the original problem is that for the scope options simlpy a maximum of 500 records are fetched from teh database and all filtering (reduing the resultset) happens in frontend. What would be much better is an option to define a method in the model that accepts one parameter which contains the current search term to allow the user to filter on teh database level. The method is [already there](https://octobercms.com/docs/backend/lists#filter-scope-options) and called `options`, however, I think it is only evalulated once on initial load. Wouled be nice if it would also trigger on search input by the user... Like that, also the user can decide to how many options to limit the list.

alxy:
Actually I was completely wrong - I've just tested this filtering seems to happen now on the database level. I don't see why the 500 is a limitation then, as on user input the results are filtered to a more sensible amount of records (ususally less than 500)

daftspunk:
Merged in e001fc47422e0dbee7f69a76624837d7825c85b3 🙏 

There are some residual UX issues that remain here, these will be addressed in v2.0 at some point. As mentioned in #5581 this filter type was never intended to handle large datasets, so we're going to limp it through for now

trepmag:
I implemented an options method which limit its number of items and filter its list given the user search key words input. The options list is well updated and the user should find for what he is looking for by refining its keywords.

The issue is that when an item is selected then Apply/Clear buttons doesn't appears.

config_filter.yaml:
```yaml
scopes:
  contact:
    label: Contact
    modelClass: Foo\Bar\Model\Contact
    options: getOptions
    conditions: contact_id in (:filtered)
```

Contact.php:
```php
<php
...
class Contact extends Model
{
    ...
    public static function getOptions()
    {
        // Build query
        $query = self::orderBy('name')
            ->take(50);

        // Lookup for user input search key words if any
        if ($keywords = Input::get('search')) {
            $keywords_a = preg_split('/\s+/', $keywords);
            foreach ($keywords_a as $word) {
                $query->orWhere(function ($query) use ($word) {
                    $query->where('name', 'LIKE', '%' . strtolower($word) . '%');
                    $query->orWhere('first_name', 'LIKE', '%' . strtolower($word) . '%');
                    $query->orWhere('company', 'LIKE', '%' . strtolower($word) . '%');
                });
            }
        }

        return $query->pluck('name', 'id')->toArray();
    }
    ...
}
```

It appears that `modules/system/assets/ui/js/filter.js` is behaving as taking into account only an initial loaded list and not subsequent filtered ones...

daftspunk:
> It appears that modules/system/assets/ui/js/filter.js is behaving as taking into account only an initial loaded list and not subsequent filtered ones...

That's right, the design of this filter pushes most of the logic onto JS. The control is showing its age and will need to be rebuilt to rectify the functional expectation against the design limitations (this is planned). The best we can do is apply workarounds. If you need some core code adjusted, drop us an email and we can review it for you

PR #5476, opened 2021-01-27
Allow to define searchScope for relation manager manage widget

danielbidala:
In your `config_relation.yaml` you can use `searchScope` as follow:

```
family:
    label: Család
    deferredBinding: true
    manage:
        showSearch: true
        searchScope: textSearch
        recordsPerPage: 10
        form: $/brightstaff/brightcrm/models/family/relation_fields.yaml
        list: $/brightstaff/brightcrm/models/family/columns.yaml
    view:
        form: $/brightstaff/brightcrm/models/family/fields.yaml
        toolbarButtons: create|link|unlink
```
Now when you click Link button in relation manager, your scope will be applied when search in the list.

### DIFF
--- a/modules/backend/behaviors/RelationController.php
+++ b/modules/backend/behaviors/RelationController.php
@@ -843,6 +843,9 @@ class RelationController extends ControllerBehavior
             }
 
             $widget = $this->makeWidget('Backend\Widgets\Lists', $config);
+            $widget->setSearchOptions([
+                'scope' => $this->getConfig('manage[searchScope]')
+            ]);
 
             /*
              * Apply defined constraints


modules/backend/behaviors/RelationController.php
```diff
@@ -843,6 +843,9 @@ protected function makeManageWidget()
             }
 
             $widget = $this->makeWidget('Backend\Widgets\Lists', $config);
+            $widget->setSearchOptions([
+                'scope' => $this->getConfig('manage[searchScope]')
+            ]);
```
mjauvin, line 848: @danielbidala What happens if `searchScope` is not defined in `config_relation.yaml` ?

danielbidala: @mjauvin In that case `scope` will be `null` and the default list search will be used based on `getSearchableColumns()` method in `List.php`.

daftspunk:
This has been included in the next feature release v2.2.

Thanks!

PR #5478, opened 2021-01-29
WIP: Add ability to specify custom sorting via map or Model callback

bradietilley:
# Work In Progress.

This PR adds the ability for developers to customise the sort order of any enum-like field, for cases when the numerical value of said enum differs from the ideal sort order of the values displayed.

The best use case example I can think of is the status of a Booking/Order. Lets say a business uses this column to identify the progress of each order, where they can view their progresses in order from start to finish.

For example you may have the following workflow for orders (which is on par with the numerical order of the statuses):
```
1: Confirmed
2: Delivering
3: Completed
4: Cancelled
```

Lets say you then add a 5th enum value later down the track, which represents a status occurring after `Confirmed` and before `Delivering`. Lets call this `Processing`:
```
5: Processing
```

This of course would push all `Processing` orders to the bottom of the list, differing from the workflow.

Insert this PR here.

Now you can define a specific order of the results by defining the order where each enum value should be placed.

columns.yaml:
```yaml
status:
    label: Status
    type: text
    sortable:
        - 1
        - 5
        - 2
        - 3
        - 4
```

This would therefore show `Processing` orders before `Delivering` orders. This is all done through dynamically generated SQL using `CASE WHEN/THEN END` (MySQL, Sqlite, SQL Server compatible)

Better yet, this PR also adds the ability for a custom callback against the relevant model.

columns.yaml:
```yaml
status:
    label: Status
    type: text
    sortable: sortStatusField
```

models\Order.php
```php
...
    public function sortStatusField(\Illuminate\Database\Eloquent\Builder $query, \Backend\Classes\ListColumn $column)
    {
        return OrderStatus::select('id')->orderBy('sequential_order_in_workflow')->get()->pluck('id');
    }
...
```

The model's callback is flexible and can return:

- array (or Arrayable): Similar to the arrays defined in the columns.yaml file
- string: A raw DB query string
- \Illuminate\Database\Query\Expression: A raw DB query expression
- void: Handle all of the ordering external directly in the model

## TODO:

- Ensure non-integer values are escaped in the `WHEN {$field} = '$when'` portion of the SQL `CASE` definition
- Ensure sortDirection is accessible via model callback (for those who are not returning an array or Arrayable object)
- More tests?

### DIFF
--- a/modules/backend/lang/en/lang.php
+++ b/modules/backend/lang/en/lang.php
@@ -227,7 +227,7 @@ return [
         'delete_selected_success' => 'Deleted selected records.',
         'column_switch_true' => 'Yes',
         'column_switch_false' => 'No',
-        'sortable_method_not_exists' => "The model class :model must define a method :method() to either handle or return custom sorting logic.",
+        'sortable_method_not_exists' => 'The model class ":model" must define a method ":method()" to handle custom sorting logic.',
     ],
     'fileupload' => [
         'attachment' => 'Attachment',
--- a/modules/backend/lang/en/lang.php
+++ b/modules/backend/lang/en/lang.php
@@ -227,6 +227,7 @@ return [
         'delete_selected_success' => 'Deleted selected records.',
         'column_switch_true' => 'Yes',
         'column_switch_false' => 'No',
+        'sortable_method_not_exists' => "The model class :model must define a method :method() to either handle or return custom sorting logic.",
     ],
     'fileupload' => [
         'attachment' => 'Attachment',
--- a/modules/backend/widgets/Lists.php
+++ b/modules/backend/widgets/Lists.php
@@ -530,7 +530,8 @@ class Lists extends WidgetBase
 
                 $this->buildQueryCustomSort($query, $customSort);
             } else {
-                if (($column = array_get($this->allColumns, $sortColumn)) && $column->valueFrom) {
+                $column = array_get($this->getColumns(), $sortColumn);
+                if ($column && $column->valueFrom) {
                     $sortColumn = $this->isColumnPivot($column)
                         ? 'pivot_' . $column->valueFrom
                         : $column->valueFrom;
--- a/modules/backend/widgets/Lists.php
+++ b/modules/backend/widgets/Lists.php
@@ -528,7 +528,7 @@ class Lists extends WidgetBase
             if ($this->sortColumnHasCustomSorting()) {
                 $customSort = $this->getColumns()[$this->getSortColumn()]->sortable;
 
-                $this->buildQueryCustomSort($query, $customSort);
+                $this->applyCustomSorting($query, $customSort);
             } else {
                 $column = array_get($this->getColumns(), $sortColumn);
                 if ($column && $column->valueFrom) {
@@ -605,13 +605,13 @@ class Lists extends WidgetBase
     }
 
     /**
-     * Using a custom sort map, build a raw DB query that handles the sorting.
+     * Sort the records by using a custom sortable callback or map
      *
      * @param $query
      * @param array|string $customSort
      * @return void
      */
-    public function buildQueryCustomSort($query, $customSort)
+    public function applyCustomSorting($query, $customSort)
     {
         if (is_string($customSort)) {
             $column = $this->getColumns()[$this->getSortColumn()] ?? [];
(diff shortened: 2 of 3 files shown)


bradietilley:
@bennothommo

modules/backend/lang/en/lang.php
```diff
@@ -227,6 +227,7 @@
         'delete_selected_success' => 'Deleted selected records.',
         'column_switch_true' => 'Yes',
         'column_switch_false' => 'No',
+        'sortable_method_not_exists' => "The model class :model must define a method :method() to either handle or return custom sorting logic.",
```
bennothommo:
```suggestion
        'sortable_method_not_exists' => 'The model class ":model" must define a method ":method()" to handle custom sorting logic.',
```

modules/backend/widgets/Lists.php
```diff
@@ -525,18 +525,24 @@ public function prepareQuery()
...
+            if ($this->sortColumnHasCustomSorting()) {
+                $customSort = $this->getColumns()[$this->getSortColumn()]->sortable;
 
-            // Set the sorting column to $relation_count if useRelationCount enabled
-            if (isset($column->relation) && @$column->config['useRelationCount']) {
-                $sortColumn = $column->relation . '_count';
-            }
+                $this->buildQueryCustomSort($query, $customSort);
+            } else {
+                if (($column = array_get($this->allColumns, $sortColumn)) && $column->valueFrom) {
```
bennothommo: @bradietilley One thing I try to avoid is those variable definitions inside condition calls. I find Xdebug generally has trouble getting those variables until *after* the condition is processed, when sometimes you want to know what it is *before* the condition runs.
```suggestion
                $column = array_get($this->getColumns(), $sortColumn);
                if ($column && $column->valueFrom) {
```

bennothommo: Also, if I'm not mistaken, the `$sortColumn` on this line is no longer defined with your changes in place.

modules/backend/widgets/Lists.php
```diff
@@ -581,6 +587,83 @@ public function prepareQuery()
         return $query;
     }
 
+    /**
+     * Check if the current sort column has a custom map.
+     * We check if the column's sortable property is an array or string.
+     *
+     * array: represents the order of which the values should be returned
+     * string: references a function on the relevant model that defines or handles the custom sort
```
bennothommo:
```suggestion
     * - array: represents the order of which the values should be returned
     * - string: references a function on the relevant model that defines or handles the custom sort
```

modules/backend/widgets/Lists.php
```diff
@@ -581,6 +587,83 @@ public function prepareQuery()
...
+     *
+     * @return boolean
+     */
+    public function sortColumnHasCustomSorting(): bool
```
bennothommo: @bradietilley We're not using return types yet until we do a big refactor of the OCMS code.
```suggestion
    public function sortColumnHasCustomSorting()
```

modules/backend/widgets/Lists.php
```diff
@@ -581,6 +587,83 @@ public function prepareQuery()
...
+        // Get the fieldname that is required for the custom sort
+        $field = $this->getSortColumn();
+
+        // Convert any non-array values to array
+        $customSort = ($customSort instanceof Arrayable) ? $customSort->toArray() : (array) $customSort;
+
+        // Compile a CASE WHEN/THEN END string to handle the sorting logic
+        $bindings = collect();
+        $whens = collect($customSort)
+            ->transform(function ($when, $then) use ($field, $bindings) {
```
bennothommo: @bradietilley Is `$bindings` necessary? Doesn't seem to do anything.

bennothommo (changes requested):

bennothommo:
@LukeTowers I've been chatting with @bradietilley about this, and I think it'll be a good complement to the [`selectable` column type PR](https://github.com/octobercms/october/pull/4377), hence why I suggested he put it up as a PR.

LukeTowers:
Will take a look at it when it's closer to completion, might be some changes to the naming that I'll ask for but if you're vouching for it, it probably makes sense.

modules/backend/widgets/Lists.php
bradietilley: It was used at one point but not anymore...

bradietilley:
@LukeTowers Ready for review when you have a chance :)

I will keep it in Draft/WIP until I'm happy with it

daftspunk:
Thanks for your pull request to October CMS!

Unfortunately, I'm going to delay merging this code for now. To preserve our ability to adequately maintain the platform, we need to be very careful regarding the amount of code we include.

If possible, please consider releasing your code as a library or plugin so that the community can still take advantage of your contributions.

If you feel absolutely certain that this code corrects a bug in the platform, please "@" mention me in a follow-up comment with a further explanation so that GitHub will send me a notification of your response.

---

This particular problem can be solved by joining the `sequential_order_in_workflow` column from the OrderStatus table into the main list, sorting by it, and using `valueFrom` to override its output value as the status name. The outcome is sorted by the status order, displaying the status name

A proof of concept has been demonstrated here: https://github.com/octoberrain/test-plugin/commit/8e5529c9dc7b0a62caa60d3746739817bc4327f8

PR #5480, opened 2021-01-30
Support specifying step, min, and max attributes on number inputs for list filters

zaxbux:
This PR adds support for specifying the `step`, `min`, and `max` attributes on the number inputs for the `number` and `numberrange` list filters. Useful for instances where the developer is aware of the appropriate values that can be used to filter records, thus making the UX less confusing for end users (e.g. specifing a minimum of `1` for a record ID filter, since your record IDs are positive integers). Tested with `number` and `numberrange` filters on integer and decimal column types. Shouldn't break existing feautures, since the changes are mostly cosmetic.

Documentation PR: octobercms/docs#509

### DIFF
--- a/modules/system/assets/ui/js/filter.numbers.js
+++ b/modules/system/assets/ui/js/filter.numbers.js
@@ -242,6 +242,8 @@
             }
 
             numberinput.value = '' !== defaultValue ? defaultValue : '';
+            numberinput.min = scopeData.minInput;
+            numberinput.max = scopeData.maxInput;
         })
     }
 
--- a/modules/system/assets/ui/js/filter.numbers.js
+++ b/modules/system/assets/ui/js/filter.numbers.js
@@ -243,9 +243,15 @@
 
             numberinput.value = '' !== defaultValue ? defaultValue : '';
 
-            if (scopeData.step) numberinput.step = scopeData.step;
-            if (scopeData.minInput) numberinput.min = scopeData.minInput;
-            if (scopeData.maxInput) numberinput.max = scopeData.maxInput;
+            if (scopeData.step) {
+                numberinput.step = scopeData.step
+            }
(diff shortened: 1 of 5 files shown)


modules/system/assets/ui/js/filter.numbers.js
```diff
@@ -242,6 +242,10 @@
             }
 
             numberinput.value = '' !== defaultValue ? defaultValue : '';
+
+            if (scopeData.step) numberinput.step = scopeData.step;
+            if (scopeData.minInput) numberinput.min = scopeData.minInput;
+            if (scopeData.maxInput) numberinput.max = scopeData.maxInput;
```
bennothommo:
```suggestion
            if (scopeData.step) {
                numberinput.step = scopeData.step
            }
            if (scopeData.minInput) {
                numberinput.min = scopeData.minInput
            }
            if (scopeData.maxInput) {
                numberinput.max = scopeData.maxInput
            }
```

bennothommo:
@zaxbux Only thing that you need to add to this is some validation or casting so that these new settings are rendered as numbers when added to the HTML.

LukeTowers:
Why is it `minInput` and `maxInput` instead of just `min` and `max`?

zaxbux:
@LukeTowers
> 
> 
> Why is it `minInput` and `maxInput` instead of just `min` and `max`?

Didn't want the config options to be confused with the `:min` and `:max` condition properties, also the date range filter uses the similarily named settings: `minDate` and `maxDate`.

bennothommo:
@zaxbux It could perhaps be `minValue` / `maxValue` instead.

LukeTowers:
I'm not sure that it would be, there's already precedent of just min & max in the `type: number` form field: https://github.com/octobercms/docs/blob/master/backend-forms.md#number

PR #5481, opened 2021-01-30
Fix: Respect user-defined admin password

mohsin:
During the regular october install using october:install, the user-defined password is overwritten with the random generated string password. Instead, now it uses random password only if the user-defined password defaults to *admin* and if a user did set a custom password, that is used instead.

### DIFF
--- a/modules/backend/database/seeds/DatabaseSeeder.php
+++ b/modules/backend/database/seeds/DatabaseSeeder.php
@@ -14,7 +14,8 @@ class DatabaseSeeder extends Seeder
      */
     public function run()
     {
-        $adminPassword = (SeedSetupAdmin::$password === 'admin') ? Str::random(22) : SeedSetupAdmin::$password;
+        $shouldRandomizePassword = SeedSetupAdmin::$password === 'admin';
+        $adminPassword = $shouldRandomizePassword ? Str::random(22) : SeedSetupAdmin::$password;
 
         Eloquent::unguarded(function () use ($adminPassword) {
             // Generate a random password for the seeded admin account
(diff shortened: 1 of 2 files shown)


modules/backend/database/seeds/DatabaseSeeder.php
```diff
@@ -13,7 +14,7 @@ class DatabaseSeeder extends Seeder
      */
     public function run()
     {
-        $adminPassword = Str::random(22);
+        $adminPassword = (SeedSetupAdmin::$password == 'admin') ? Str::random(22) : SeedSetupAdmin::$password;
```
LukeTowers:
```suggestion
        $adminPassword = (SeedSetupAdmin::$password === 'admin') ? Str::random(22) : SeedSetupAdmin::$password;
```

LukeTowers:
I think the correct fix for this would be to not run this seeder at all when a manual password is specified, otherwise it'll include the manual password at the end saying it was automatically generated.

mohsin:
@LukeTowers If I avoid calling the seeder then the admin account isn't even created at all during the setup, which is pretty bad. The DatabaseSeeder calls SeedSetupAdmin for setting up the admin account and that's essential.

If the issue is merely the message then fixing it would be as simple as changing the return message to not show the "has been automatically generated" message which can be done with a simple if statement in the same run method.

LukeTowers:
@SaifurRahmanMohsin sure, do that and then provide a gif recording of what it looks like please.

mohsin:
@LukeTowers Here's a gif of it in action with a custom admin pass specified:

with-custom-pass

And here's one with an empty password (so OctoberCMS would generate one):
with-empty-pass

modules/system/classes/UpdateManager.php
```diff
@@ -1046,7 +1046,7 @@ protected function addMessage($class, $message)
      */
     protected function printMessages()
     {
-        if (!count($this->messages)) {
+        if (!count($this->messages) || empty(array_filter(array_flatten($this->messages)))) {
```
LukeTowers: Is this meant to filter out empty messages? If so we should filter them when they're added instead of on display

LukeTowers:
Thanks @SaifurRahmanMohsin!

PR #5490, opened 2021-02-03
Missing translations for pt-br

prhost:


modules/system/lang/pt-br/validation.php
```diff
@@ -100,7 +129,11 @@
     |
     */
 
-    'custom' => [],
+    'custom' => [
+        'attribute-name' => [
+            'rule-name' => 'custom-message',
+        ],
+    ],
```
LukeTowers:
```suggestion
    'custom' => [],
```
this is an example, not an actual message to be translated.

prhost: had copied from the original English version, just to make it the same.

PR #5500, opened 2021-02-09
properly set CMS content editor according to documentation

mjauvin:
According to [documentation](https://octobercms.com/docs/cms/content#introduction) , the editor for CMS content files should be a `richeditor` for .htm files and `markdown` for .md files.


daftspunk:
Nice. The functionality has never matched the documentation. This has already been implemented in the next big release coming soon. Thanks for everything Marc

PR #5502, opened 2021-02-11
Change Url to Storage facade

Denoder:
Laravel has the ability to change the storage host url in the filesystem.php: https://laravel.com/docs/6.x/filesystem#file-urls
not sure why `Url::assets` is there, isn't that suppose to lead to an assets folder or was changed to accommodate the cms module? If so what would be the best way to detect if the cms module is disabled that way it can switch to the Storage facade (in any case I just need the benefit of being able to change the url)

### DIFF
--- a/modules/system/models/File.php
+++ b/modules/system/models/File.php
@@ -80,7 +80,7 @@ class File extends FileBase
             $uploadsPath .= '/protected';
         }
 
-        return Url::asset($uploadsPath) . '/';
+        return Storage::url($uploadsPath) . '/';
     }
 
     /**
--- a/config/filesystems.php
+++ b/config/filesystems.php
@@ -46,7 +46,7 @@ return [
         'local' => [
             'driver' => 'local',
             'root'   => storage_path('app'),
-            'url'    => '/storage/app',
+            'url'    => '',
         ],
 
         's3' => [


modules/system/models/File.php
```diff
@@ -80,7 +80,7 @@ public function getPublicPath()
             $uploadsPath .= '/protected';
         }
 
-        return Url::asset($uploadsPath) . '/';
+        return Storage::url($uploadsPath) . '/';
```
LukeTowers, line 83: Isn't the better option to actually do `$this->getDisk()->url($uploadsPath)` without changing the filesystems config? Not sure why you think that change is required but it's definitely not getting merged.

LukeTowers: @Teranode to clarify I meant the filesystems.php config change wasn't going to get merged. Still possibly open to the other change if it solves a specific problem.

Denoder: > @Teranode to clarify I meant the filesystems.php config change wasn't going to get merged. Still possibly open to the other change if it solves a specific problem.

I changed the filesystems.php file because when using the 'url' param with `Storage::url()` and 'storage/app' is defined within it, it appends `storage/aopp` twice, most likely because the `path` param from a storage config is what is being used when defining the base directory  (a safer bet wuld have been to use `env('APP_URL')`.

In any case i've already deleted my fork.

Denoder:
I've remade the fork here @LukeTowers : https://github.com/octobercms/october/pull/5504

PR #5504, opened 2021-02-11
Use Storage facade for file model

Denoder:
Laravel has the ability to change the storage host url in the filesystem.php: https://laravel.com/docs/6.x/filesystem#file-urls
not sure why Url::assets is there, isn't that suppose to lead to an assets folder or was changed to accommodate the cms module? If so what would be the best way to detect if the cms module is disabled that way it can switch to the Storage facade (in any case I just need the benefit of being able to change the url)

filesystem.php has been altered because `/storage/app` gets appended twice when using the `Storage` facade. I'm not sure how you would want to test this.

### DIFF
--- a/modules/system/models/File.php
+++ b/modules/system/models/File.php
@@ -71,9 +71,15 @@ class File extends FileBase
      */
     public function getPublicPath()
     {
-        $uploadsPath = Config::get('cms.storage.uploads.path', '/storage/app/uploads');
+        $uploadsFolder = Config::get('cms.storage.uploads.folder', 'uploads');
 
-        return $this->getDisk()->url($uploadsPath . ($this->isPublic() ? '/public/' : '/protected/'));
+        $path = $this->getDisk()->url('/'. $uploadsFolder . ($this->isPublic() ? '/public/' : '/protected/'));
+
+        if (Config::get('cms.linkPolicy') === 'force') {
+            return Url::asset($path);
+        }
+
+        return $path;
     }
 
     /**
--- a/modules/system/models/File.php
+++ b/modules/system/models/File.php
@@ -71,13 +71,13 @@ class File extends FileBase
      */
     public function getPublicPath()
     {
-        $uploadsFolder = Config::get('cms.storage.uploads.folder', 'uploads');
+        $uploadsFolder = Config::get('cms.storage.uploads.path', '/storage/app/uploads');
 
         if ($this->isPublic()) {
-            return $this->getDisk()->url('/'. $uploadsFolder . '/public/');
+            return $this->getDisk()->url($uploadsFolder . '/public/');
         }
         else {
-            return $this->getDisk()->url('/'. $uploadsFolder . '/protected/');
+            return $this->getDisk()->url($uploadsFolder . '/protected/');
         }
     }
 
--- a/modules/system/models/File.php
+++ b/modules/system/models/File.php
@@ -73,12 +73,7 @@ class File extends FileBase
     {
         $uploadsPath = Config::get('cms.storage.uploads.path', '/storage/app/uploads');
 
-        if ($this->isPublic()) {
-            return $this->getDisk()->url($uploadsPath . '/public/');
-        }
-        else {
-            return $this->getDisk()->url($uploadsPath . '/protected/');
-        }
+        return $this->getDisk()->url($uploadsPath . ($this->isPublic() ? '/public/' : '/protected/'));
     }
 
     /**
--- a/modules/system/models/File.php
+++ b/modules/system/models/File.php
@@ -80,7 +80,7 @@ class File extends FileBase
             $uploadsPath .= '/protected';
         }
 
-        return Url::asset($uploadsPath) . '/';
+        return Storage::url($uploadsPath) . '/';
     }
 
     /**
--- a/modules/system/models/File.php
+++ b/modules/system/models/File.php
@@ -71,16 +71,12 @@ class File extends FileBase
      */
     public function getPublicPath()
     {
-        $uploadsPath = Config::get('cms.storage.uploads.path', '/storage/app/uploads');
-
         if ($this->isPublic()) {
-            $uploadsPath .= '/public';
+            return $this->getDisk()->url('/uploads/public/');
         }
         else {
-            $uploadsPath .= '/protected';
+            return $this->getDisk()->url('/uploads/protected/');
         }
-
-        return Storage::url($uploadsPath) . '/';
     }
 
     /**
--- a/modules/system/models/File.php
+++ b/modules/system/models/File.php
@@ -71,11 +71,13 @@ class File extends FileBase
      */
     public function getPublicPath()
     {
+        $uploadsFolder = Config::get('cms.storage.uploads.folder', 'uploads');
+
         if ($this->isPublic()) {
-            return $this->getDisk()->url('/uploads/public/');
+            return $this->getDisk()->url('/'. $uploadsFolder . '/public/');
         }
         else {
-            return $this->getDisk()->url('/uploads/protected/');
+            return $this->getDisk()->url('/'. $uploadsFolder . '/protected/');
         }
     }
 
--- a/modules/system/models/File.php
+++ b/modules/system/models/File.php
@@ -71,13 +71,13 @@ class File extends FileBase
      */
     public function getPublicPath()
     {
-        $uploadsFolder = Config::get('cms.storage.uploads.path', '/storage/app/uploads');
+        $uploadsPath = Config::get('cms.storage.uploads.path', '/storage/app/uploads');
 
         if ($this->isPublic()) {
-            return $this->getDisk()->url($uploadsFolder . '/public/');
+            return $this->getDisk()->url($uploadsPath . '/public/');
         }
         else {
-            return $this->getDisk()->url($uploadsFolder . '/protected/');
+            return $this->getDisk()->url($uploadsPath . '/protected/');
         }
     }
 
--- a/config/filesystems.php
+++ b/config/filesystems.php
@@ -46,7 +46,7 @@ return [
         'local' => [
             'driver' => 'local',
             'root'   => storage_path('app'),
-            'url'    => env('APP_URL'),
+            'url'    => '/storage/app',
         ],
 
         's3' => [
--- a/config/filesystems.php
+++ b/config/filesystems.php
@@ -46,7 +46,7 @@ return [
         'local' => [
             'driver' => 'local',
             'root'   => storage_path('app'),
-            'url'    => '/storage/app',
+            'url'    => env('APP_URL'),
         ],
 
         's3' => [


modules/system/models/File.php
```diff
@@ -80,7 +80,7 @@ public function getPublicPath()
             $uploadsPath .= '/protected';
         }
 
-        return Url::asset($uploadsPath) . '/';
+        return Storage::url($uploadsPath) . '/';
```
LukeTowers: @Teranode did you try my suggestion of not making any change to the filesystem config, removing the `uploadsPath` config call from this method and just doing `$this->getDisk()->url('/protected');` vs `$this->getDisk()->url('/public')`?

Denoder: that could would but i would also need to append the `uploads` name to it would that work for you @LukeTowers

LukeTowers:
@Teranode can you add a unit test to make sure that this isn't breaking any paths?

Denoder:
@LukeTowers 
gonna be honest, i don't set up unit tests, or rather I haven't set up one in a couple of years, just do things live and if it breaks I fix it.

LukeTowers:
@bennothommo want to take a look at this when you have time (and also optimize the code paths? No need to have two separate calls to getDisk()->url())

mjauvin:
~Tested and it's working!~

LukeTowers:
@mjauvin are you able to resubmit with unit tests?

mjauvin:
I'll see what I can do @LukeTowers

mjauvin:
@LukeTowers Oops, looks like I fucked up when I tested this, it's not honoring the cms.linkPolicy afterall.

LukeTowers:
@mjauvin tbh, I wouldn't expect it to. But it should respect the URL setting in the filesystem config which would allow you to achieve the same results.

mjauvin:
Achieve the same result how @LukeTowers ?

It will only return relative URLs with this PR. (I personally don't mind, I prefer relative URLs anyway), but the goal is to make this configurable with the cms.linkPolicy

LukeTowers:
By setting an absolute URL in the filesystem config

mjauvin:
I don't mean relative to the filesystem, but without a domain (only a path)

LukeTowers:
I know, if someone wants a relative url then they would set a relative url in the filesystem config, and if they want an absolute url they'd set an absolute url in the filesystem config

mjauvin:
That's a little backward @LukeTowers when the setting that should apply this accross the board is already in place, we just need to use it (i.e. cms.linkPolicy) and it's used in most of the other places that generate urls already.

mjauvin:
@LukeTowers this PR breaks the getPublicPath() method, I won't be pursuing this route.

Denoder:
> @LukeTowers this PR breaks the getPublicPath() method, I won't be pursuing this route.

How does it break it?

mjauvin:
It's not using `cms.storage.uploads.path`

Denoder:
> It's not using `cms.storage.uploads.path`

it can

mjauvin:
I'm sorry, a change I made to use a single call to getDisk behaved in an unexpected way... this is actually working fine.

Denoder:
> I'm sorry, a change I made to use a single call to getDisk behaved in an unexpected way... this is actually working fine.

either way it still works out in the end with the change i made (if you can test it out) if there is no url param it uses a relative path, and if a url is specified it uses the url

Denoder:
@LukeTowers 
would something like this be sufficent enough to remove the double method?
```php
    /**
     * Define the public address for the storage path.
     */
    public function getPublicPath()
    {
        $uploadsPath = Config::get('cms.storage.uploads.path');

        return $this->getDisk()->url($uploadsPath . ($this->isPublic() ? '/public/' : '/protected/'));
    }
```

mjauvin:
@Teranode use this, the previous changes don't work:
```
    public function getPublicPath()
    {   
        $uploadsFolder = Config::get('cms.storage.uploads.folder', 'uploads');

        $path = $this->getDisk()->url('/'. $uploadsFolder . ($this->isPublic() ? '/public/' : '/protected/'));

        if (Config::get('cms.linkPolicy') === 'force') {
            return Url::asset($path);
        }

        return $path;
    }
```

Denoder:
> @Teranode use this, the previous changes don't work:
> 
> ```
>     public function getPublicPath()
>     {   
>         $uploadsFolder = Config::get('cms.storage.uploads.folder', 'uploads');
> 
>         $path = $this->getDisk()->url('/'. $uploadsFolder . ($this->isPublic() ? '/public/' : '/protected/'));
> 
>         if (Config::get('cms.linkPolicy') === 'force') {
>             return Url::asset($path);
>         }
> 
>         return $path;
>     }
> ```

i've set it

mjauvin:
@LukeTowers looks like the image resizer unit tests already account for this.

bennothommo:
I'm in two minds about this fix.

The usage of `getPublicPath` is mainly for generating public URLs, so realistically, this should *always* be returning a full URL irrespective of the link policy, because the implication is that should a third-party storage be used (or the filesystem `url` config is defined), this URL may not match the URL of the site itself.

So I think the fact it was routing through `Url::asset` at all times is correct. The only real change required is the `$this->getDisk()->url()` addition.

This will be difficult to unit test because it would require some faking of third-party storage systems, like AWS S3, as well as local paths to ensure that the URL is consistently applied with necessarily requiring real files or connections to the third-party systems. It's doable, but will take time.

Denoder:
> I'm in two minds about this fix.
> 
> The usage of `getPublicPath` is mainly for generating public URLs, so realistically, this should _always_ be returning a full URL irrespective of the link policy, because the implication is that should a third-party storage be used (or the filesystem `url` config is defined), this URL may not match the URL of the site itself.
> 
> So I think the fact it was routing through `Url::asset` at all times is correct. The only real change required is the `$this->getDisk()->url()` addition.
> 
> This will be difficult to unit test because it would require some faking of third-party storage systems, like AWS S3, as well as local paths to ensure that the URL is consistently applied with necessarily requiring real files or connections to the third-party systems. It's doable, but will take time.

The problem with `$this->getDisk()` is that it only applies to the local filesystem by default, there's a default cloud parameter in filesystem.php that uses `Storage::cloud()` to initialize it. The next issue is the fact that `$this->getDisk()` doesn't allow you to specify what disk you want to use so it's only the default one set in filesystem.php that gets to be chosen (this can be ignored since you want to run off a default setting for users, the users would just have to make their own File class and set the methods themselves -- though the problem would arise if they wanna override `getDisk` then they wouldnt be able to because there's do default parameter so theyd have to make a new method).

There needs to be a `$this->getCloud()` equivalent if you want to also account for services like s3. This is what i'm currently using to have the right url and path for cloud (this is a custom different File Model since im not using the System File Model):

```php
    /**
     * Define the public address for the storage path.
     */
    public function getPublicPath()
    {
        $uploadsFolder = Config::get('anime.storage.videos.folder');

        $path = Storage::cloud()->url($uploadsFolder . ($this->isPublic() ? '/public/' : '/protected/'));

        if (Config::get('cms.linkPolicy') === 'force') {
            return Url::asset($path);
        }

        return $path;
    }
```

LukeTowers:
> The problem with $this->getDisk() is that it only applies to the local filesystem by default, there's a default cloud parameter in filesystem.php that uses Storage::cloud() to initialize it. The next issue is the fact that $this->getDisk() doesn't allow you to specify what disk you want to use so it's only the default one set in filesystem.php that gets to be chosen (this can be ignored since you want to run off a default setting for users, the users would just have to make their own File class and set the methods themselves -- though the problem would arise if they wanna override getDisk then they wouldnt be able to because there's do default parameter so theyd have to make a new method).

What are you talking about? The File model uses whatever disk is configured in `cms.storage.uploads.disk`: https://github.com/octobercms/october/blob/develop/modules/system/models/File.php#L106. There is no need for this function to know or care about Storage::cloud().

If anyone has a file model that they don't want to store in the default file model disk configuration, then they just need to extend the filemodel class and implement their own getDisk() method, that is its entire purpose.

bennothommo:
@Teranode as @LukeTowers mentioned, the `url` method in `FilesystemAdapter` (which is essentially what `$this->getDisk()` provides) does work with cloud services: https://github.com/octobercms/library/blob/1fa47b3ab683f1e7a4e9ec3a1d0b484dbe66bde0/src/Filesystem/FilesystemAdapter.php#L21

mjauvin:
@bennothommo I disagree: 

1. A public URL can be just a path, which will resolve relative to the current request.
2. If the filesystem url config is defined and contains a fully qualified URL, this is going to be used irrespective of the cms.linkPolicy, due to the following code in Url::asset:
```
            return $path;
        }   
```

bennothommo:
@mjauvin 

> A public URL can be just a path, which will resolve relative to the current request.

I understand, but that's not what the comments for `getPublicPath` imply.

https://github.com/octobercms/october/blob/develop/modules/system/models/File.php#L70
https://github.com/octobercms/library/blob/develop/src/Database/Attach/File.php#L959
https://github.com/octobercms/library/blob/1fa47b3ab683f1e7a4e9ec3a1d0b484dbe66bde0/src/Filesystem/FilesystemAdapter.php#L14

They all reference a public "address" or URL, which implies a full URL to the location.

mjauvin:
Then maybe the description must be updated (or the method name changed) because, at least for getPublicPath(), we're clearly saying "path", not address or URL.

Denoder:
> cms.storage.uploads.di

ah sorry I was looking at things from the point of the october/library

LukeTowers:
On further examination I'm not comfortable changing the behaviour of the getPublicPath() method. We could potentially expose a new method for using the disk url() method, (although we'd probably want to add a `getDiskPath()` method too)

b8ne:
I've just ran into this problem. Full disclaimer, I don't understand the full inner workings of the Filesystem, but hoping my input may help.  
I'm currently using GCS, which when configured and env variables updated works a treat.  I'm also serving uploaded media/files via CDN.  I've faced a few similar issues sourcing assets from the CDN, such as rendering in templates or showing thumbnails in the media browser - I've bypassed these by using custom form widgets, components or filters.  
To be fair, my work is to solve my use case, so there may be use cases that I haven't thought of.  However, intuitively with any storage or filesystem, I would think that the configuration I use to store a file should be the same configuration used to retrieve it - unless explicitly stated.
Looking at how the `File` models stores, for my use case uses the `copyLocalToStorage()` function in `October\Rain\Database\Attach\File`.  This uses `$this->getDisk()`.
Retrieving the file, I think it should use the exact same, regardless of config, function naming or comments.

Currently I'm using the below, and is working fine.
```
public function getPublicPath()
    {
        $uploadsPath = Config::get('cms.storage.uploads.path', '/storage/app/uploads');

        if ($this->isPublic()) {
            $uploadsPath .= '/public';
        } else {
            $uploadsPath .= '/protected';
        }

        return $this->getDisk()->url($uploadsPath) . '/';
    }
```

daftspunk:
This was an oversight in the original design, I had assumed Url::asset would handle the CDN management

It's been there so long that it might not be a good idea to disrupt existing sites, however, I've fixed it in the next big update coming soon

Thanks for using October CMS

PR #5506, opened 2021-02-13
Permanent Composer 2 compatibility

### LINKED ISSUE #5033: Composer 2.0 compatibility
It seems like composer 2.0 is right around the corner. Downloading `https://getcomposer.org/composer.phar` already gives you a 2.0 build. 

Our deployments just started to fail because of this. It seems like there is an issue with the composer-merge-plugin, as it explicitly requires composer 1.1:
https://github.com/wikimedia/composer-merge-plugin/blob/master/composer.json#L15

I have opened an issue on the plugins' GitHub page:
https://github.com/wikimedia/composer-merge-plugin/issues/184

I guess this needs a bump of the `wikimedia/composer-merge-plugin` dependency in October once a new version has been released.

```
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Root composer.json requires wikimedia/composer-merge-plugin 1.4.1 -> satisfiable by wikimedia/composer-merge-plugin[v1.4.1].
    - wikimedia/composer-merge-plugin v1.4.1 requires composer-plugin-api ^1.0 -> found composer-plugin-api[2.0.0] but it does not match your constraint.
```

VolCh:
https://github.com/wikimedia/composer-merge-plugin/pull/189 has been merged in master and we can use regular "wikimedia/composer-merge-plugin": "~1.5.0" to provide Composer 2.0 compatible  and we can drop repository introduced in https://github.com/octobercms/october/commit/a095c1e54526c64c70c0af85900d64de3924b4bd#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34

Closes https://github.com/octobercms/october/issues/5033

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,7 +37,7 @@
         "october/backend": "1.1.*",
         "october/cms": "1.1.*",
         "laravel/framework": "~6.0",
-        "wikimedia/composer-merge-plugin": "2.0.1"
+        "wikimedia/composer-merge-plugin": "~2.0.1"
     },
     "require-dev": {
         "phpunit/phpunit": "^8.4|^9.3.3",
--- a/composer.json
+++ b/composer.json
@@ -30,12 +30,6 @@
         "docs": "https://octobercms.com/docs/",
         "source": "https://github.com/octobercms/october"
     },
-    "repositories": [
-        {
-            "type":"vcs",
-            "url":"https://github.com/octoberrain/composer-merge-plugin"
-        }
-    ],
     "require": {
         "php": ">=7.2",
         "october/rain": "1.1.*",
@@ -43,7 +37,7 @@
         "october/backend": "1.1.*",
         "october/cms": "1.1.*",
         "laravel/framework": "~6.0",
-        "wikimedia/composer-merge-plugin": "dev-feature/composer-v2 as 1.5.0"
+        "wikimedia/composer-merge-plugin": "2.0.1"
     },
     "require-dev": {
         "phpunit/phpunit": "^8.4|^9.3.3",
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,7 @@
         "source": "https://github.com/octobercms/october"
     },
     "require": {
-        "php": ">=7.2",
+        "php": ">=7.2.9",
         "october/rain": "1.1.*",
         "october/system": "1.1.*",
         "october/backend": "1.1.*",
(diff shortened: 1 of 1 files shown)


VolCh:
@LukeTowers what do you think?

LukeTowers:
@VolCh they haven't released v.1.5.0 yet, and there was some discussion on possibly making it 2.0, in which case we might need to support both

VolCh:
@LukeTowers then let's switch to dev-master from  dev-feature/composer-v2 - it can be deleted at any moment.

LukeTowers:
@VolCh not going to until the official release is out. This file is not easily updated automatically and it's used on all new projects, so it needs to be pretty stable.

VolCh:
@LukeTowers it has been released with v2.0.0 tag even patched with v2.0.1 , so I have used the last one and squashed commits

LukeTowers:
And it works fine with both composer 1.0 and 2.0? Because we could always just accept either option with an `||`

VolCh:
@LukeTowers It works fine with composer 1.10.20 and 2.0.11. The only difference after `composer install` have been executed are in composer.lock, vendor/autoload.php and vendor/composer/* files

composer.json
```diff
@@ -30,20 +30,14 @@
...
-    ],
     "require": {
         "php": ">=7.2",
         "october/rain": "1.1.*",
         "october/system": "1.1.*",
         "october/backend": "1.1.*",
         "october/cms": "1.1.*",
         "laravel/framework": "~6.0",
-        "wikimedia/composer-merge-plugin": "dev-feature/composer-v2 as 1.5.0"
+        "wikimedia/composer-merge-plugin": "2.0.1"
```
LukeTowers:
```suggestion
        "wikimedia/composer-merge-plugin": "~2.0.1"
```

LukeTowers:
blocked by #5357. @VolCh can you remove the platform config from composer.json so we can merge this?

LukeTowers:
Just apply https://github.com/octobercms/october/pull/5357/files#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34L87 and https://github.com/octobercms/october/pull/5357/files#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34R40 to this PR

VolCh:
@LukeTowers done

daftspunk:
Thanks all, nice fix

PR #5516, opened 2021-02-22
Remove possibility of backend username enumeration

cosminbosutar:
## Security fix: Removed backend username enumeration

The current implementation of the reset password functionality allows backend username enumeration because. This is because the platform response for a user that exists is different from the response for a user that does not exist.

### Example:
- Existing user message: *Message sent to your email address with instructions.*
- NOT existing user message: *A user could not be found with a login value of ':login'*

### Proposed solution
1. Display the same **success message** on both cases: 
    - the user exist
    - the user does not exist
2. Silently fail in case the a user with the specified username does not exist or the specified username fails validation.


bennothommo:
@cosminbosutar Security fixes must be disclosed according to our [security policy](https://github.com/octobercms/october/security/policy) to be considered.

cosminbosutar:
> @cosminbosutar Security fixes must be disclosed according to our [security policy](https://github.com/octobercms/october/security/policy) to be considered.

Thank you @bennothommo for your mention. In future, I will follow the instructions you mention.

modules/backend/controllers/Auth.php
```diff
@@ -167,32 +168,28 @@ public function restore_onSubmit()
         ];
 
         $validation = Validator::make(post(), $rules);
-        if ($validation->fails()) {
-            throw new ValidationException($validation);
-        }
```
LukeTowers, line 172: The validation failing should still come into play here shouldn't it? I don't see how that's leaking any information

modules/backend/controllers/Auth.php
```diff
@@ -167,32 +168,28 @@ public function restore_onSubmit()
...
-            'name' => $user->full_name,
-            'link' => $link,
-        ];
-
-        Mail::send('backend::mail.restore', $data, function ($message) use ($user) {
-            $message->to($user->email, $user->full_name)->subject(trans('backend::lang.account.password_reset'));
-        });
-
-        return Backend::redirect('backend/auth/signin');
+        return Redirect::refresh();
```
LukeTowers, line 192: What's your thought process behind refreshing the current reset page vs redirecting to the login page?

LukeTowers:
@bennothommo we can let this one through given that it's mostly a minor UX tweak rather than a major security issue; but yes, generally speaking all security reports should be submitted through the documented process.

modules/backend/controllers/Auth.php
jacobdekeizer: @LukeTowers A different statuscode can also leak information.

Klaasie: Doesn't the ValidationException return a 406 http error code? Really any indication that could give away that an account exists should be avoided.

modules/backend/controllers/Auth.php
rawnetcbosutar: To allow the user to quickly reenter the username in case he/she doesn't type the correct username and the password reset email is not received.

modules/backend/controllers/Auth.php
rawnetcbosutar: @LukeTowers like @jacobdekeizer and @Klaasie mentioned, I consider it's better to not display the information that a username must have a specific length.

cosminbosutar: @LukeTowers like @jacobdekeizer and @Klaasie mentioned, I consider it's better to not display the information that a username must have a specific length.

modules/backend/controllers/Auth.php
cosminbosutar: To allow the user to quickly reenter the username in case he/she doesn't type the correct username and the password reset email is not received.

modules/backend/controllers/Auth.php
LukeTowers: @jacobdekeizer @cosminbosutar @Klaasie the validation exception has nothing to do with whether an account exists or not, it just an early indicator to the user that they screwed up their request. My main problem with it right now is that it's not necessarily accurate given some DB configurations result in a max length of string columns of 191; so on that basis I'm fine with removing the length part, but validating that the input is required period is probably still a good idea.

LukeTowers: @cosminbosutar any thoughts on the above?

cosminbosutar: @LukeTowers I agree with you that validating that the input is received is a good idea.
I'll make an update to the PR in which I'll add the logic that validates that the input exists and the validation messages.

LukeTowers: @cosminbosutar just waiting for your update.

daftspunk:
Thanks @cosminbosutar 🙏  This is a good find. I've implemented a fix in a different way than what is here. It will be available in the next update we've got coming soon. Cheers

PR #5520, opened 2021-02-28
Radio and checkbox didn't show popup "Please select one of these opti…

meesiris:
Radio and checkbox didn't show popup "Please select one of these options." when field definitions required: true. Because it can not focus.

Use display:block;opacity:0 instead of display:none

image

### DIFF
--- a/modules/system/assets/ui/storm.css
+++ b/modules/system/assets/ui/storm.css
@@ -4096,7 +4096,7 @@ html.cssanimations .cursor-loading-indicator.hide {display:none}
 .custom-checkbox input[type=radio],
 .custom-radio input[type=radio],
 .custom-checkbox input[type=checkbox],
-.custom-radio input[type=checkbox] {display:none}
+.custom-radio input[type=checkbox] {display:block;opacity: 0}
 .custom-checkbox label,
 .custom-radio label {display:inline-block;cursor:pointer;position:relative;padding-left:25px;margin-right:15px;margin-left:-20px;font-size:13px;-webkit-user-select:none;-moz-user-select:none;-ms-user-select:none;user-select:none}
 .custom-checkbox label:before,


bennothommo:
@meesiris Does this cause any overlap of corresponding elements (such as the label)?

meesiris:
> @meesiris Does this cause any overlap of corresponding elements (such as the label)?

@bennothommo  Nothing to do with corresponding elements. Cause is checkbox and radio style is display:none. The popup can not focus on that elelments. I change it to display:block but transparent (opacity:0)

daftspunk:
Hi @meesiris, the `required: true` flag is merely a visual indicator that the server will expect this value to be entered, it doesn't activate the client-side validation

Check out the `Validation` trait in the documentation for more details: https://octobercms.com/docs/database/traits#validation

If you require client-side validation, you could implement it using a partial form field type:
https://octobercms.com/docs/backend/forms#field-partial

Hope this helps. Thanks for using October CMS! 🙏

meesiris:
Hi @daftspunk required: true effect to input, checkbox, radio attribute required. 

`<input type="text" name="fiestname" id="Form-field-PersonalInformation-lastname" value="" placeholder="" class="form-control" autocomplete="off" maxlength="255" required>`

When present, it specifies that an input field must be filled out before submitting the form. It works fine on text field. But didn't show popup on Radio or Checkbox.

Please try it from following link.
https://www.w3schools.com/tags/att_input_required.asp
https://www.w3schools.com/tags/tryit.asp?filename=tryhtml5_input_required

daftspunk:
@meesiris

daftspunk:
This would make sense for larger forms, however, be sure to include the support via the `attributes` property.

Eg:

```
    party_mode:
        label: Party mode
        type: checkbox
        attributes:
            required: 1
```

This will trigger client-side behaviors, although it is not the preferred approach, we can still support it. Thanks!

PR #5522, opened 2021-03-01
Fix: Set default value for fields

meesiris:
/**
     * Checks if default values should be taken from data.
     * This should be done when model exists or when explicitly configured
     */
    protected function shouldFetchDefaultValues()
    {
        $enableDefaults = object_get($this->config, 'enableDefaults');
        if ($enableDefaults === false) {
            return false;
        }
        return !$this->model->exists || $enableDefaults;
    }

Function should be done when model exists but function return FALSE when model exists. I remove "!" from line 1132.

Test with following configuration. works fine.

```
fields:
    field_text:
        label: field_text
        type: text
        default: default text
        span: left
    field_number:
        label: field_number
        type: number
        default: 123
        span: left
    field_radio:
        label: field_radio
        type: radio
        default: 2
        options:
            1: 1)
            2: 2)
            3: 3)
            4: 4)
            span: left
    field_dropdown:
        label: field_dropdown
        type: dropdown
        default: 3
        options:
            1: 1)
            2: 2)
            3: 3)
            4: 4)
        span: left
```

image

### DIFF
--- a/modules/backend/widgets/Form.php
+++ b/modules/backend/widgets/Form.php
@@ -1129,7 +1129,7 @@ class Form extends WidgetBase
         if ($enableDefaults === false) {
             return false;
         }
-        return !$this->model->exists || $enableDefaults;
+        return $this->model->exists || $enableDefaults;
     }
 
     /**


bennothommo:
@meesiris The function itself should not be changed, but the comment should be.

The following line:

>  * This should be done when model exists or when explicitly configured

should be:

>  * This should be done when the model does not exist or when explicitly configured.

The default values should be retrieved from the data if the model doesn't exist, because if the model does exist, then that becomes the primary source of truth for the form.

meesiris:
@bennothommo If function not change ,field configuration below will not work.

```
fields:
    field_text:
        label: field_text
        type: text
        default: default text
        span: left
    field_number:
        label: field_number
        type: number
        default: 123
        span: left
    field_radio:
        label: field_radio
        type: radio
        default: 2
        options:
            1: 1)
            2: 2)
            3: 3)
            4: 4)
            span: left
    field_dropdown:
        label: field_dropdown
        type: dropdown
        default: 3
        options:
            1: 1)
            2: 2)
            3: 3)
            4: 4)
        span: left
```

daftspunk:
That's right, default field values are not applied when the model exists. Consider using 

- `filterFields` - https://octobercms.com/docs/backend/forms#filter-form-fields
- A model accessor  - https://octobercms.com/docs/database/mutators#accessors-and-mutators

PR #5528, opened 2021-03-04
Dynamically configure per page options

thomas4Bitcraft:
It is now possible to change the perPageOptions of a list dynamically according to the config_list.yaml file. This allows the user to be more flexible.


daftspunk:
Yes! Works a treat, thanks! I will add to the docs for the next release

PR #5536, opened 2021-03-17
Trigger dispose-control on relation popup hide

sixgweb:
ea4ca1802bdbdb46ed20c127ae1c1a4e68a52a1c fixes #5112 but the FormWidget.prototype.dispose function is not triggered during 'hide.oc.popup'.

https://github.com/octobercms/october/blob/ea4ca1802bdbdb46ed20c127ae1c1a4e68a52a1c/modules/backend/widgets/form/assets/js/october.form.js#L41

This commit modifies october.relation.js bindToPopups() and triggers the formwidget 'dispose-control' during 'hide.oc.popup' event


daftspunk:
A better solution was proposed, which is why I assume this was closed. Thanks @sixgweb!

PR #5537, opened 2021-03-17
call controlUtils.disposeControls in popup.js

sixgweb:
#5112 adds the necessary cleanup for the formwidget but 'dispose-control' listener is not triggered when a popup is closed
https://github.com/octobercms/october/blob/ea4ca1802bdbdb46ed20c127ae1c1a4e68a52a1c/modules/backend/widgets/form/assets/js/october.form.js#L41

This commit calls $.oc.foundation.controlUtils.disposeControls on the popup container, similar to popover.js and tab.js
https://github.com/octobercms/october/blob/ea4ca1802bdbdb46ed20c127ae1c1a4e68a52a1c/modules/system/assets/ui/js/popover.js#L39
https://github.com/octobercms/october/blob/ea4ca1802bdbdb46ed20c127ae1c1a4e68a52a1c/modules/system/assets/ui/js/tab.js#L220

### DIFF
--- a/modules/system/assets/ui/js/popup.js
+++ b/modules/system/assets/ui/js/popup.js
@@ -145,6 +145,7 @@
 
         this.$modal.on('hidden.bs.modal', function(){
             self.triggerEvent('hidden.oc.popup')
+            $.oc.foundation.controlUtils.disposeControls(self.$container.get(0))
             self.$container.remove()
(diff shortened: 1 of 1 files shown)


daftspunk:
Good stuff!

msimkunas:
@sixgweb Thanks for finalizing this!

PR #5549, opened 2021-04-02
Fix setting ignoreTimezone for date filters 

KonstantinObuhov:
Set ignoreTimezone to true if at least one of the filter has a data-ignore-timezone attribute


daftspunk:
LGTM, Thanks @KonstantinObuhov !

PR #5552, opened 2021-04-04
Add event to override recordUrl in Lists widget

alxy:
The use case should be clear from the example in code.

Basically, in my controller I want to be able to decide if I want to preview or update a record based on some custom logic:

```php
public function listOverrideRecordUrl($url, $model)
    {
        if($model->user_id !== BackendAuth::getUser()->id) {
            return 'acme/test/services/preview/' . $model->id;
        }
    }
```


daftspunk:
An improved version of this has been included in v2.0

Thanks @alxy !

PR #5553, opened 2021-04-06
Fix pagelinks plugin in richeditor

alxy:
This is a re-opening of PR #3086
All credits to the original author @tomsseisums

I have tested the fix and it is still valid.


daftspunk:
Thanks @alxy and @tomsseisums !

PR #5598, opened 2021-05-16
Add resized to White listed folders

igor-tv:
Resizer adds images to `storage/app/resized` directory, but this directory is not whitelisted


daftspunk:
Hi @igor-tv!

The `resized` path is included in the 1.1 project scaffold: https://github.com/octobercms/october/blob/1.1/.htaccess#L40

It looks like you may be using the 2.0 project scaffold with the 1.1 modules

If you plan on using 1.1, try installing with the following command to get the best .htaccess file:

```
composer create-project october/october:1.1.* myoctober
```

In 2.0 the `resized` path is called `resources`. You could also just rename this to `resized` instead.

PR #5629, opened 2021-11-13
Fix column preview

alxy:
The interface for league/csv changed a bit for v9, see [documentation here](https://csv.thephpleague.com/9.0/reader/).

As the `october/rain` library v1.1 [depends on `league/csv` v9.1](https://github.com/octobercms/library/blob/8f06437ab772b48dd7ea7079c293f436438f867d/composer.json#L39), this change is required to make the column preview work again.


daftspunk:
Thanks @alxy!

PR #5631, opened 2021-11-27
Added default avatar thumb support

ZhiweiWu0425:
Added default avatar thumb support because at china we didn't use gravatar.com and it load very slow. And which is also very helpful for LAN projects


daftspunk:
Thanks for this, @ZhiweiWu0425. However, it has been submitted to the wrong repo. Please email hello@octobercms.com for getting support with platform change proposals.

PR #5661, opened 2022-10-05
Update fr.json

Ladylain:
"Limace" is not a good translation for "Slug" , i think it would be better to let "Slug" but we can also translate it by "URL rewriting" or "Permalink" or even "Short URL". Maybe you are the best to choose. 

best,


daftspunk:
"Slug" may be too technical. "URL Code" is something we can consider as an alternative.

PR #5663, opened 2022-10-09
Update de.json

DavidBrogli:
The correct translation of ID in German is also "ID".


alxy:
This is actually hilarious 😂

daftspunk:
Thanks!

PR #5665, opened 2022-10-24
Fix Indonesia Translations

syehan:
Hi @daftspunk..  recently i try to redefine indonesia translations for set the precisely meaning on October CMS modules. hope this enhancement would helpful. :)


daftspunk:
Thanks @syehan, looks great!

PR #5668, opened 2022-11-12
fix ua and ru auto-translation errors

layerok:
In this PR I fixed auto-translation errors for ua and ru languages for the tailor module.
The word "slug" was auto-translated literally. The Ukrainian and Russian language don't have their own word for slug. 
We use the English word and write it the way it sounds.


layerok:
I'm not quite right. We have an abbreviation "ЧПУ" which stands for "Human Readable URL". But I think the word "слаг" would work too, considering that earlier the slug was literally auto-translated as an earth slug

daftspunk:
Thanks!

PR #5708, opened 2023-03-24
Update fr.json

najbo:
Based on en.json ; added missing keys, related to new version 3.3 enhancements (and some other general missing keys from past versions)


daftspunk:
Thanks!

PR #5736, opened 2023-07-06
Better French translations

OlinB:
I went around the french localisation files to add missing translations, and to correct a few existing ones


daftspunk:
Thank you, @OlinB !

PR #5755, opened 2023-09-29
use current PHP rather than cli default

timrspratt:
When multiple versions of PHP are installed and the non-default PHP version is used to run `composer create-project`, this `post-create-project-cmd` script uses the cli default PHP version rather than the initiated PHP version.


daftspunk:
Thanks!

PR #5878, opened 2025-06-19
Fix Twig 3.21 deprecation warning in GetAttrNode

anik1ng:
Screenshot 2025-06-19 at 1 54 45 PM


anik1ng:
`strlen(): Passing null to parameter #1 ($string) of type string is deprecated in [/var/www/html/modules/backend/behaviors/FormController.php](javascript:) on line 479`

PR #5881, opened 2025-06-20
Fix strlen() deprecation warnings in FormController

anik1ng:
`strlen(): Passing null to parameter #1 ($string) of type string is deprecated in [/var/www/html/modules/backend/behaviors/FormController.php](javascript:) on line 479`


daftspunk:
🙏

PR #5883, opened 2025-06-20
Fix strtolower() null deprecation in RelationController

anik1ng:
`strtolower(): Passing null to parameter #1 ($string) of type string is deprecated in [/var/www/html/modules/backend/behaviors/RelationController.php](javascript:) on line 527`


daftspunk:
Nice refactor! Thanks

PR #5884, opened 2025-06-20
Fix str_contains() null deprecation

anik1ng:
`str_contains(): Passing null to parameter #1 ($haystack) of type string is deprecated in [/var/www/html/modules/backend/widgets/lists/partials/_column_partial.php](javascript:) on line 11`


daftspunk:
Thanks!

PR #5894, opened 2025-06-30
Fix trim() null deprecation in HasAjaxRequests.php

anik1ng:
`trim(): Passing null to parameter #1 ($string) of type string is deprecated in [/var/www/.../modules/cms/classes/controller/HasAjaxRequests.php](javascript:) on line 87`


daftspunk:
Thanks!

PR #5895, opened 2025-06-30
Fix strlen() null deprecation in TrafficLogger.php

anik1ng:
`strlen(): Passing null to parameter #1 ($string) of type string is deprecated in [/var/www/.../modules/dashboard/classes/TrafficLogger.php](javascript:) on line 54`


daftspunk:
Thanks!

PR #5897, opened 2025-07-01
str_replace, null deprecated in Lists.php

pvullioud:


anik1ng:
@daftspunk found a bug in the original code — https://github.com/octobercms/october/blob/8a1c7546af2d496239117b3e689e87a7caa440b9/modules/backend/widgets/Lists.php#L1547C9-L1556C58

Here PR — https://github.com/octobercms/october/pull/5899

PR #5899, opened 2025-07-01
Fix lost changes in $result due to wrong variable in str_replace

anik1ng:
@daftspunk found a bug in the original code — https://github.com/octobercms/october/blob/8a1c7546af2d496239117b3e689e87a7caa440b9/modules/backend/widgets/Lists.php#L1547C9-L1556C58

```php
...

$result = $value;
$result = str_replace(' <', '<', $result);

// Add natural spacing between HTML nodes
$result = str_replace("><", '> <', $value); // should use $result instead of $value

// Strip HTML
$result = $original = trim(Html::strip($result));

...
```


daftspunk:
Thanks @anik1ng !

PR #5906, opened 2025-07-04
Add missing french translations

OlinB:
I've added missing french translations mainly for the dashboard, but also for the missing "Update" label (shows up in the update modal for relations).

Can I also modify the translations in `modules/system/assets/js/lang/lang.fr.js` ? or are they somehow generated ?


daftspunk:
Thanks!

PR #5907, opened 2025-07-04
Add a configuration to enable a non-strict role hierarchy.

### LINKED ISSUE #5876: Configurable non strict role hierarchy
For some of our clients, it is necessary for admins to be able to create and manage other admins with the same level of permissions.

Is there a possibility to add a configuration option (e.g. `backend.loose_role_comparaison`) allowing this on a site per site basis ?
As far as I can tell, the only required changes youlw be to the `getRankedRoleOptions` and
`applyRankPermissionsToQuery` methods in the `Users` controller.

OlinB:
Here is my proposition to add a config option to make roles have a non-strict hierarchy (as it was in OCv1 and 2).

In some cases, our customers need to be able to manage each other's accounts without resorting to a shared admin account.

closes #5876

### DIFF
--- a/config/backend.php
+++ b/config/backend.php
@@ -158,6 +158,20 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Strict Role hierarchy
+    |--------------------------------------------------------------------------
+    |
+    | Defines if the backend should enforce a strict role hierarchy.
+    |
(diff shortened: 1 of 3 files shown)


daftspunk:
Thanks for this, it is an interesting concept to allow horizontal management. Using config is the way to go!

PR #5923, opened 2025-08-07
Fix partial name parameter conflict

mihailbishkek:
## Description
Fixes a conflict when using `name` as a parameter in partial tags.

## Problem
When using `{% partial 'test' name='value' %}`, the `name` parameter conflicts with the internal storage of the partial name, causing the partial to fail to load.

## Solution
- Changed internal partial name storage from `'name'` to `'_partial_name'`
- Updated both `PartialTokenParser` and `PartialNode` classes
- Maintains full backward compatibility

## Testing
- [x] Tested with `{% partial 'test' name='123' %}` - works correctly
- [x] Tested with other parameters - works correctly  
- [x] Tested existing functionality - no regressions

## Example

Before (broken):

```twig
{% partial 'sidebar' name='John' %}
```

After (working):

```twig
{% partial 'sidebar' name='John' %}
```

### DIFF
--- a/modules/cms/twig/PartialNode.php
+++ b/modules/cms/twig/PartialNode.php
@@ -58,7 +58,7 @@ class PartialNode extends TwigNode
         if ($options['isAjax']) {
             $compiler
                 ->write("yield '<div data-ajax-partial=\"'.")
-                ->subcompile($this->getNode('name'))
+                ->subcompile($this->getNode('_partial_name'))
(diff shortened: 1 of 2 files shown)


FlusherDock1:
@daftspunk Can confirm that this PR fixes the headache with partial param renaming after upgrade from 3.x to 4.x.

daftspunk:
Thanks!

PR #5936, opened 2025-08-22
Extend .htaccess index.php rule to include website root

johrclaessens:
This PR adds 2 new rules to the .htaccess file to ensure that not only /page/index.php is redirected to /page/, but also the website root: /index.php will redirect to /


daftspunk:
Thanks!

Issue #5926, opened 2025-08-10
Add trigger condition: not_empty


Eeesbksbk:
Please add condition "not_empty" in file october.js

1)
this.conditionValid = "checked" == this.triggerCondition || "unchecked" == this.triggerCondition || "value" == this.triggerCondition;
to
this.conditionValid = "checked" == this.triggerCondition || "unchecked" == this.triggerCondition ||  "not_empty" == this.triggerCondition || "value" == this.triggerCondition;

2)
onConditionChanged() {
                    if ("checked" == this.triggerCondition) this.updateTarget(!!t(this.triggerSelector + ":checked", this.triggerParent).length);
                    else if ("unchecked" == this.triggerCondition) this.updateTarget(!t(this.triggerSelector + ":checked", this.triggerParent).length);

to

onConditionChanged() {
                    if ("checked" == this.triggerCondition) this.updateTarget(!!t(this.triggerSelector + ":checked", this.triggerParent).length);
                    else if ("unchecked" == this.triggerCondition) this.updateTarget(!t(this.triggerSelector + ":checked", this.triggerParent).length);
                   else if ("not_empty" == this.triggerCondition) this.updateTarget(t(this.triggerSelector, this.triggerParent).val() != '');

Thanks...

daftspunk:
This could be `value` without a value specified. Value means not empty, and without a value, it could mean any value (not empty).

Could you provide a usage example of your `not_empty`?

Eeesbksbk:
you mean value, not value[] ?

executioner90:
Hi! 👋

I’d like to work on implementing the `not_empty` trigger condition. 

To clarify its usage, here are some examples across different field types:

1. **Text Field**
```yaml
triggers:
  - field: title
    condition: not_empty
    action: notify_admin
```
Fires when title has any text.

2. **Textarea Field**

```yaml
triggers:
  - field: description
    condition: not_empty
    action: send_email
```
Fires when description has content.

3. **Number Field**

```yaml
triggers:
  - field: quantity
    condition: not_empty
    action: update_inventory
```
Fires if a number is entered (zero counts as a value if valid).

4. **Checkbox Field**

```yaml
triggers:
  - field: terms_accepted
    condition: not_empty
    action: proceed_to_checkout
```
Fires if the checkbox is checked.

So basically:
  -  not_empty is value-agnostic; no exact value is needed.
  -  It should work across different input types.

[...]

PR #5937, opened 2025-08-26
[ISSUE #5926] Trigger action on not empty value when no value passed

executioner90:
Implementing solution for issue #5926


daftspunk:
Sorry @executioner90 

~I think based on #5944 we need to use `not_empty` explicitly. Could you introduce `not_empty` instead of using `value`.~

Edit: Hold the phone. The solution could be `value[*]`. Keep an eye on #5944 for the solution.

Best regards

daftspunk:
Fixed by #5944

PR #5944, opened 2025-09-05
Input trigger with compound empty condition

OlinB:
This PR solves a bug when using trigger conditions and trying to combine an empty condition with a value condition.

All of these worked: 
```yaml
trigger:
    action: hide
    field: my_field
    condition: value[]
    condition: value[myvalue]
    condition: value[myvalue][othervalue]
```

But these did not:
```yaml
trigger:
    action: hide
    field: my_field
    condition: value[][myvalue]
    condition: value[myvalue][]
    condition: value[myvalue][][othervalue]
```

Slightly modifying the regex allows it to match the empty brackets and add an empty string to the list of matches


daftspunk:
Before we merge this, can you check the following works:

`value[]` -- empty value
`value[*]` - any value (not empty)
`value[foobar]` -- value is "foobar"
`value[foobar*]` - value beginning with foobar
`value[*foobar]` - value ending with foobar
`value[*foobar*]` - value containing foobar

It would help solve #5937 and #5926

OlinB:
The wildcards work, but I had to add a special case to handle the `value[*]` case, since it matched with empty values as well.

if I want to add input trigger examples using this syntax to the test-plugin, where should I add them ?

daftspunk:
Looks good!

Potentially add it to the Post update model (plugins/october/test/models/post/fields.yaml) in the Tags tab.

OlinB:
While trying to add an example to the test plugin in tags, I realized the value[] syntax works for empty text inputs, but not for inputs that output an empty array as a value. 
In the case the condition triggers on an empty array, I add an empty string as a value before.

This shouldn't break existing logic, but there may be a case I'm unaware of.

daftspunk:
Thanks!

PR #5948, opened 2025-09-16
Harden plugin discovery to avoid boot-time fatal on unreadable/broken directories

sotiriskast:
- Summary
    - This PR hardens plugin discovery by replacing the RecursiveDirectoryIterator scan with a safe two-level DirectoryIterator loop in PluginManager::getVendorAndPluginNames(). It gracefully skips unreadable or broken directory entries and only registers plugins that contain a Plugin.php file

- Motivation
    - In some environments, RecursiveDirectoryIterator->getChildren() can throw UnexpectedValueException when it encounters a missing/unreadable directory or a broken symlink (e.g., during deployments or partial syncs). This results in a boot-time fatal. The proposed approach avoids recursion exceptions and keeps plugin enumeration resilient.

- What changed
    - Use DirectoryIterator to list vendor and plugin directories.
    - Add is_dir/is_readable checks and skip dot entries.
    - Confirm Plugin.php exists before registering a plugin path.
    - No changes to how plugin namespaces are built or cached.

- Behavior and compatibility
    - Valid plugins continue to be discovered exactly as before.
    - Unreadable/missing/broken directories are skipped instead of causing a fatal.
    - No configuration changes required.

- Tests / Verification
    - With standard plugins under plugins// /Plugin.php, plugins load as usual.
    - Create a broken symlink or unreadable directory under plugins/ and verify the application boots without errors and skips the invalid entry.
    - Clear caches to ensure manifests regenerate:
        - php artisan cache:clear
        - php artisan config:clear

- Performance
    - Equivalent in practice; the scan is shallow (two levels) and avoids exception overhead.

- Risks
    - Minimal; the change narrows discovery to directories that contain Plugin.php, which is already the de facto requirement for a valid plugin.

### DIFF
--- a/modules/system/classes/PluginManager.php
+++ b/modules/system/classes/PluginManager.php
@@ -502,25 +502,54 @@ class PluginManager
         $plugins = [];
 
         $dirPath = plugins_path();
-        if (!is_dir($dirPath)) {
+        if (!is_dir($dirPath) || !is_readable($dirPath)) {
             return $plugins;
         }
 
-        $it = new RecursiveIteratorIterator(
-            new RecursiveDirectoryIterator($dirPath, RecursiveDirectoryIterator::FOLLOW_SYMLINKS)
-        );
-        $it->setMaxDepth(2);
-        $it->rewind();
+        // Iterate vendors directly to avoid recursion exceptions on unreadable children
+        try {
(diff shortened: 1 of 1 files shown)


daftspunk:
Thanks!

PR #5955, opened 2025-10-11
Tailor: translate content group names in Entry Type dropdown

burzuy:
Wrap name with trans() in EntryBlueprint::getEntryTypeOptions() so the Entry Type selector shows translated labels.
Ref: https://talk.octobercms.com/t/how-do-i-set-a-translation-for-a-content-group/2685/5


daftspunk:
Thanks!

PR #5956, opened 2025-10-11
Backend Form: hide locale indicator for `ruler` field (align with `section`/`hint`)

burzuy:
**Problem:** In a translatable context (e.g., Tailor with `multisite: sync`), the `ruler` field shows a locale badge next to its generic label, even though it’s a structural (non-data) field. `section` and `hint` don’t show the badge because they render their own headers in their partials.

**Solution:** Make `ruler` behave like `section`/`hint`:

* Don’t render the generic label for `ruler` (so the locale badge doesn’t appear).
* If `label`/`comment` are set, render them inside `field_ruler.htm` (local to the field, no locale badge).

**Changes:**

1. `modules/backend/widgets/Form.php` — add `ruler` to the exclusions in `showFieldLabels()` (generic label not rendered).
2. `modules/backend/widgets/form/partials/field_ruler.htm` — render:

   * `<label for="…"
     class="form-label">…</label>` when `label` is provided;
   * `<hr>`;
   * `<p class="form-text">…</p>` when `comment` is provided (respects `commentHtml`).

**Why:** `ruler` is structural; showing a translation badge is misleading. This aligns UI with existing `section`/`hint` behavior. No data/schema/save logic is affected.

**Reproduction (before fix):**

1. Blueprint with `multisite: sync`.
2. Field:

   ```yaml
   _r:
     type: ruler
     label: TEST
   ```
3. Locale badge appears next to the label.

**Expected (after fix):** No locale badge for `ruler`; `label`/`comment` render correctly inside the field’s partial.

**Backward compatibility:** No changes for regular input fields; UI-only adjustment for the structural `ruler` field.

### DIFF
--- a/modules/backend/widgets/Form.php
+++ b/modules/backend/widgets/Form.php
@@ -1065,7 +1065,7 @@ class Form extends WidgetBase implements FormElement
      */
     protected function showFieldLabels(FormField $field): bool
     {
-        if (in_array($field->type, ['checkbox', 'switch', 'section', 'hint', 'rule'])) {
+        if (in_array($field->type, ['checkbox', 'switch', 'section', 'hint', 'ruler'])) {
             return false;
         }
 
--- a/modules/backend/widgets/Form.php
+++ b/modules/backend/widgets/Form.php
@@ -1065,7 +1065,7 @@ class Form extends WidgetBase implements FormElement
      */
     protected function showFieldLabels(FormField $field): bool
     {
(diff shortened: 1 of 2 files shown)


daftspunk:
Thanks for this. The field name is **ruler** instead of **rule**

burzuy:
Thank you for the note. I’ve updated it to ruler in showFieldLabels() and adjusted the markup in field_ruler.htm (container class field-horizontalrule). Changes have been pushed.

daftspunk:
Thanks!

Issue #5924, opened 2025-08-07
Invalid aggregate function: none


sqwk:
Using `ReportMetric::AGGREGATE_NONE` as per https://docs.octobercms.com/4.x/extend/dashboards/data-sources.html#registering-metrics produces the error `Invalid aggregate function: none`

Some background: I am trying to display durations (time between a start_time and an end_time) on a widget using the new DataSource interface. Since aliases cannot be summed the following does not work:

```
        $this->registerMetric(new ReportMetric(
            self::METRIC_ENTRY_DURATION,
            'entry_duration',
            'Recorded Entry Duration',
            ReportMetric::AGGREGATE_SUM
        ));

        (...)

        $reportQueryBuilder->onConfigureQuery(
            function(Builder $query, ReportDimension $dimension, array $metrics) {
                $query->addSelect([
                    Db::raw('TIMESTAMPDIFF(SECOND, start_time, end_time) / 3600 AS entry_duration')
                ]);
            }
        );
```
as that will add `SUM(entry_duration)` to the query.

Instead I figured one could skip the aggregate function and do the sum as part of the query manually:

```
        $this->registerMetric(new ReportMetric(
            self::METRIC_ENTRY_DURATION,
            'id',
            'Recorded Entry Duration',
            ReportMetric::AGGREGATE_NONE
        ));

        (...)

        $reportQueryBuilder->onConfigureQuery(
            function(Builder $query, ReportDimension $dimension, array $metrics) {
                $query->addSelect([
                    Db::raw('SUM(TIMESTAMPDIFF(SECOND, start_time, end_time) / 3600) AS oc_metric_entry_duration')
                ]);
            }
        );
```

however, that produces the error above.

sqwk:
```php
class TimeDataSource extends ReportDataSourceBase
{

    const METRIC_ENTRY_DURATION = 'entry_duration';

    public function __construct()
    {
        $this->registerDimension(new ReportDimension(
            ReportDimension::CODE_DATE,
            'CAST(end_time as DATE)',
            'Date'
        ));

        $this->registerMetric(new ReportMetric(
            self::METRIC_ENTRY_DURATION,
            'SUM(TIMESTAMPDIFF(SECOND, start_time, end_time) / 3600)',
            'Recorded Entry Duration (hours)',
            ReportMetric::AGGREGATE_NONE
        ));
    }

    protected function fetchData(ReportFetchData $data): ReportFetchDataResult
    {
[...]

PR #5966, opened 2025-10-29
Adds ReportMetric::AGGREGATE_NONE, Fix ReportMetric::AGGREGATE_COUNT_DISTINCT_NOT_NULL

sqwk:
Adds `ReportMetric::AGGREGATE_NONE` and fixes `ReportMetric::AGGREGATE_COUNT_DISTINCT_NOT_NULL.`

See #5924 .

Theoretically I would also like to reduce `validateDbObjectName()` to just the `trim` to allow SQL methods like `CAST` or `TIMESTAMPDIFF`. To what extend is removing or opening up the regex a risk?


daftspunk:
Thanks for this!

> To what extend is removing or opening up the regex a risk? 

This is likely just internal validation mechanics. It should be safe to do.

PR #5974, opened 2025-11-11
Update parameter types to support both int and string in Repeater to support ULIDs and UUIDs

RickAcb:
**Summary of the Changes:**
- Updated parameter types in the Repeater component to support both `int` and `string` values.
- This allows Repeater field keys to handle not only auto-increment integer IDs (the previous default) but also string-based identifiers such as ULIDs and UUIDs.

**Solution Implemented:**
- Modified the codebase where the Repeater expects field keys, updating type hints and value handling logic so IDs can be integers or strings.

**Why:**
- Modern applications increasingly use string-based unique identifiers (ULIDs, UUIDs) for scalability and uniqueness, especially in distributed systems.
- Supporting both identifier types enables developers to integrate the Repeater widget with systems or databases that use non-integer keys.

**Backward Compatibility:**
- The changes are designed to be backward compatible: existing Repeater fields that use integer IDs will continue to work without modification.
- No breaking change for sites relying on integer-based keys.

### DIFF
--- a/modules/backend/formwidgets/Repeater.php
+++ b/modules/backend/formwidgets/Repeater.php
@@ -310,12 +310,12 @@ class Repeater extends FormWidgetBase
 
     /**
      * makeItemFormWidget creates a form widget based on a field index and optional group code
-     * @param int $index
+     * @param int|string $index
      * @param string $groupCode
-     * @param int $fromIndex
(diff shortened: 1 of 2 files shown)


daftspunk:
Thanks!

Issue #5977, opened 2025-11-19
Report Widgets: title property default value is not used


PhilippeBlanko:
### Problem

In October CMS v4, when adding a **Report Widget** through the admin UI, the `default` value defined for the title property in `defineProperties()` is ignored. The Title field in the configuration popup always shows **"My Custom Widget"**.

### Example

```
public function defineProperties()
{
    return [
        'title' => [
            'title'   => 'Widget Title',
            'default' => 'Hello Administrator',
            'type'    => 'string',
        ],
    ];
}
```

### Likely cause

In `modules/dashboard/vuecomponents/dashboard/assets/js/widget-static.js`:

```
makeDefaultConfigAndData: function () {
    Vue.set(this.widget.configuration, 'title', 'My Custom Widget');
},
```

The label is hardcoded, preventing October from using the widget’s actual default value.

### Expected behavior

Use the `default` value defined in the widget, falling back only if none is provided.

### In picture

<img width="650" height="245" alt="Image" src="https://github.com/user-attachments/assets/997133ed-f937-4995-ab8d-5d05d4928809" />

PR #5979, opened 2025-11-20
Set default title for static dashboard widgets from loaded properties

PhilippeBlanko:
This PR improves the handling of the `default` title for static dashboard widgets.

Previously, the widget would always initialize with a hardcoded title ("**My Custom Widget**"), ignoring any default value provided in the widget's properties. This PR updates the `makeDefaultConfigAndData` method to:

- Check if `loadedValue.properties` contains a `title` property with a `default` value.
- Use the `default` from properties if available.
- Fall back to "**My Custom Widget**" only if no default is provided.

This ensures that any default titles defined in widget data are respected, improving user experience and consistency with other widgets.

Related issue: [Report Widgets: title property default value is not used #5977](https://github.com/octobercms/october/issues/5977)


daftspunk:
Nice! Thanks

PR #5980, opened 2025-11-20
Fix default backend branding configuration options

PhilippeBlanko:
This PR updates the default backend branding configuration for login background types.

Previously, the configuration did not include all supported options. This update ensures that the following option is correctly listed:

- login_background_type: color, wallpaper, gradient, ai_images

No functionality changes are introduced; this is purely a configuration correction to align the default settings with the supported options.


daftspunk:
Thanks!

Issue #5987, opened 2025-11-26
Dashboards with dashes in code are inaccessible


PhilippeBlanko:
### Description:

It is currently impossible to access a dashboard whose code contains a dash (`-`).

### Example:

If a dashboard is created with the code `abc-123`, the following URL does not work as expected:

```
http://localhost:8080/administration/dashboard/index/abc-123
```

### Cause:

In the `BackendController` class, at line [108](https://github.com/octobercms/october/blob/4.x/modules/backend/classes/BackendController.php#L108), the action is transformed using the `parseAction` method at line [201](https://github.com/octobercms/october/blob/4.x/modules/backend/classes/BackendController.php#L201) to make it compatible with PHP.

This automatically converts the URL:

```
http://localhost:8080/administration/dashboard/index/abc-123
```
to

```
http://localhost:8080/administration/dashboard/index/abc123
```

### Impact:

The dashboard `Index` controller receives `abc123` as the code instead of `abc-123`, preventing proper access to the originally created dashboard.

daftspunk:
This should probably do a check `if ($controllerClass instanceof WildcardController)` then skip transformation.

daftspunk:
A fix for this has been included in v4.0.19

PR #5988, opened 2025-11-26
Allow dashboards with dashes in code to be accessed

PhilippeBlanko:
Currently, dashboards with codes containing dashes (`abc-123`) are inaccessible because `BackendController::run` transforms the action using `parseAction`, removing dashes (`abc123`).

This PR updates the `run` method to:

- Keep the original action for `WildcardController`.
- Apply `parseAction` only for regular controllers.

This ensures dashboards with dashed codes are correctly loaded without affecting other controllers.

Related issue: [octobercms/october#5987](https://github.com/octobercms/october/issues/5987)


daftspunk:
Thanks!

PR #5989, opened 2025-11-26
Add the missing French translations to the modules

PhilippeBlanko:
Add missing French translations in the modules since version 4.0.18


daftspunk:
Thanks!

PR #5998, opened 2026-01-06
Fix role based peer management check before User is saved

Pindagus:
An administrator without super user access can't give a user the same role as itself while user peer management is enabled. This change fixes that.


daftspunk:
Thanks!

PR #6001, opened 2026-01-17
Remove misleading @return self from NoteMaker setNotesOutput phpdoc

mariavilaro:
Removes a misleading `@return self` phpdoc from `NoteMaker::setNotesOutput()`.

When extending `System\Classes\UpdateManager`, the annotation resolves to the base
class and breaks fluent chaining of subclass methods in static analysis tools.
Runtime behavior is unchanged.


daftspunk:
Thanks!

PR #6010, opened 2026-02-06
Add new buttons to mediafinder

Ladylain:
For multi mode Mediafinder form widget, adds new buttons :
- Select / Deselect all
- Copy selected
- Paste

image

image

image

- - - 
I was not sure about Lang so i just created EN and FR version on files, let me know if you want that i change other files ... :)

UPDATE : i think i will add a Storage Event to be able to update buttons based on changes made on local storage key.

### DIFF
--- a/modules/media/formwidgets/mediafinder/assets/js/mediafinder.js
+++ b/modules/media/formwidgets/mediafinder/assets/js/mediafinder.js
@@ -18,6 +18,7 @@ oc.registerControl('mediafinder', class extends oc.ControlBase {
         this.$el = $(this.element);
         this.instanceNumber = this.constructor.instanceCounter;
 
+        
         if (this.config.isMulti === undefined) {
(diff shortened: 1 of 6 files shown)


daftspunk:
Hey @Ladylain 

This will copy and paste references between `mediafinder` fields, useful inside `repeater` fields, is that correct?

Ladylain:
> Hey @Ladylain
> 
> This will copy and paste references between `mediafinder` fields, useful inside `repeater` fields, is that correct?

Yes, exactly, I wasn't thinking about the 'repeater' when I did it, but that's another good possible use.

daftspunk:
Ok great! You are copying between different database records?

Ladylain:
@daftspunk yes, my workmate needed to copy paste images between galleries cause we have a lot of images.

daftspunk:
Thanks @Ladylain !

PR #6029, opened 2026-03-29
Fixing some german strings in modules.

SamBrishes:
Hello,

I used the German admin dashboard for the first time and noticed that, alongside the manual translations, some automated tooling (e.g. Google Translate, DeepL, or similar) seems to have been used in parts.

This led to a few unintentionally amusing results — for example:

- "Dashboard" → "Armaturenbrett" (that's literally "instrument panel", as in cars or aircraft)
- "LESS" → "WENIGER" (technically correct in isolation, but not when referring to the CSS preprocessor)
- "Enclosure Character" → "Textqualifizierer" (technically accurate too, but rather uncommon, like only bureaucrats love to say it that way)
- "Traffic Statistics" → "Verkehrsstatistik" ("Verkehr" is typically associated with road traffic rather than website usage)

I've revised and corrected as many of these cases as I could find, including a number of smaller inconsistencies and typos.

One thing I wasn’t sure about is the preferred tone for the German UI. At the moment, both formal ("Sie") and informal ("Du") forms are used, which feels kinda odd.

Sam.


daftspunk:
Thanks Sam!

PR #6033, opened 2026-04-05
Jackyhuynh0407 fix vuecomponents uploader

jackyhuynh0407:
I found that, "Upload From Computer" in richedtor isn't working right! The upload progress bar didn't update, and didn't show error or success after upload complete.


daftspunk:
Thanks for finding this @jackyhuynh0407!

This is a Vue 3 upgrade problem. Fixed in ebd356aef13aeef29ba9fbfaad7f2806bac4e06f

Best regards

jackyhuynh0407:
Thanks for fixing it with better solution.

PR #6037, opened 2026-04-17
Do not match ignored columns

vojtasvoboda:
When using the [Import feature](https://docs.octobercms.com/4.x/extend/importexport/importexport-controller.html), a user can ignore a file column by clicking the ignore icon on that row (which calls $.oc.importBehavior.ignoreFileColumn(this)). However, if the user then clicks the Auto-match columns button, the ignored columns get matched anyway and are passed to the import model.

Fix: Skip columns marked as ignored during auto-matching so they are not passed to the import model.

This issue also applies to OctoberCMS 3.6 and 3.7. I haven't tested the older ones.

### DIFF
--- a/modules/backend/behaviors/importexportcontroller/assets/js/october.import.js
+++ b/modules/backend/behaviors/importexportcontroller/assets/js/october.import.js
@@ -119,6 +119,10 @@
                 dbColumnName = $dbItem.data('column-name'),
                 $dbItemMatchInput = $('[data-column-match-input]', $dbItem);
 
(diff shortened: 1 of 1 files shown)


daftspunk:
Thanks!

vojtasvoboda:
Any chance to get this fix into 3.7 or even 3.6? :-) Thanks

daftspunk:
Sure, I merged it in to 3.x (3.7) so it will be included in the next patch for that

PR #6049, opened 2026-05-29
fix infinite scroll in firefox

VladStarynets:
There's a problem in firefox, scroll become infinite.

https://github.com/user-attachments/assets/1da19317-8421-4209-8b26-a9b5127c48fb


daftspunk:
Thanks for this!

PR #6050, opened 2026-05-31
security: add allowed_classes to unserialize() in CMS and Media cache to prevent PHP Object Injection

XananasX7:
## Summary

All `unserialize()` call sites in the CMS and Media modules operate on cache-backend data. If the cache backend (Redis, Memcached, or file cache) can be poisoned by an attacker — e.g., via a compromised or shared cache instance — the lack of `allowed_classes` restriction enables **PHP Object Injection**, allowing an attacker to achieve arbitrary code execution via gadget chains.

## Changes

| File | Stored type | Fix |
|------|-------------|-----|
| `modules/media/classes/MediaLibrary.php` | `MediaLibraryItem` objects | `allowed_classes => [MediaLibraryItem::class]` |
| `modules/cms/classes/Router.php` (×2) | Plain string/array URL maps | `allowed_classes => false` |
| `modules/cms/classes/CodeParser.php` | Plain arrays of strings | `allowed_classes => false` |
| `modules/cms/classes/CmsCompoundObject.php` | Plain property arrays | `allowed_classes => false` |
| `modules/cms/classes/SnippetManager.php` | Plain snippet info arrays | `allowed_classes => false` |
| `modules/cms/traits/UrlMaker.php` | Plain scalar arrays | `allowed_classes => false` |
| `modules/tailor/behaviors/PreviewController.php` | Plain scalar arrays | `allowed_classes => false` |

## Security Impact

Without `allowed_classes`, PHP will instantiate any class present in the serialized data during deserialization. This is the root cause of PHP Object Injection attacks. Adding this restriction means any unexpected class in cache data is returned as `__PHP_Incomplete_Class` instead of a live object — neutralizing gadget chain exploitation.

## Testing

All changes are functionally transparent for legitimate cache data: the stored types match the `allowed_classes` restrictions exactly, so existing behavior is unchanged for valid data.


daftspunk:
Thanks!

PR #6053, opened 2026-06-09
Added missing localization in the Internal Traffic Statistics section.

syarmolyuk-coder:
Some UI strings in the view file were displayed without using the OctoberCMS localization mechanism.


daftspunk:
Thanks!

PR #6055, opened 2026-06-14
Add a devcontainer with codespace support

scottbedard:
This PR adds a devcontainer that is compatible with GitHub Codespaces.

Setting up October is now a no-install and one-click process 🎉


modules/backend/controllers/Auth.php
```diff
@@ -353,6 +353,9 @@ protected function handleSubmitSetup()
         $user = UserModel::createDefaultAdmin(post());
         BackendAuth::login($user);
 
+        // Fresh setup should not resume a stale guest URL (e.g. https://localhost/admin).
+        session()->forget('url.intended');
+
```
scottbedard, line 358: Using October inside a codespace revealed an issue where stale session values could break a redirect path if `APP_URL` and the browser's URL were of different origins. The simplest fix is to just clear the potentially stale value on setup.

[More info here &rarr;](https://gist.github.com/scottbedard/85e03024eda85374732292fc01c462eb)

daftspunk:
🚀 🚀 🚀

PR #6061, opened 2026-06-22
[DOCKER] Use detect link policy for local devcontainers

scottbedard:
This enables local devcontainers, and remains compatible with GitHub Codespaces.

Most IDEs support devcontainers nowadays, so local dev is now just a matter of installing Docker Desktop, OrbStack, Podman, etc... then open your repo and go! 🚀 

Here is a demo with OrbStack + Cursor. Codespaces are nice, but working locally is obviously a much smoother experience 🙂

[october-devcontainer.webm](https://github.com/user-attachments/assets/9e35328c-6a2f-410c-a93a-a6905865241c)


daftspunk:
Thanks!

PR #6075, opened 2026-07-16
i18n: Spanish (es) translation

mariavilaro:
Add new Spanish translations for the Editor and Tailor module. Update, fix and cleanup existing lang fiels across CMS, Dashboard, System, and Tailor modules.


daftspunk:
Thank you @mariavilaro !

PR #6076, opened 2026-07-16
i18n: Catalan (ca) translation

mariavilaro:
Add new Catalan translations for the Editor and Tailor modules. Update, fix and cleanup existing  lang files across CMS, Dashboard and System modules.


daftspunk:
Thanks!

PR #6081, opened 2026-07-24
Add system.updater.beforeUpdate event

scottbedard:
Create a `system.updater.beforeUpdate` event that plugins can use to intercept update calls from the backend. This is necessary when working with ephemeral or read-only file systems [like the one described here](https://www.reddit.com/r/octobercms/comments/1v1p5bd/october_cloud_preview/).

Once intercepted, we can use `php artisan october:update` for more fine-grained control of the update.

```php
Event::listen('system.updater.beforeUpdate', function (string $context) {
    Flash::success('Update started!');

    // trigger custom update workflow

    return Redirect::refresh();
});
```

Similarly, a plugin can now disable backend updates entirely by throwing an exception.

```php
Event::listen('system.updater.beforeUpdate', function (string $context) {
    throw new Exception("Backend updates are disabled, use 'php artisan october:update'");
});
```

### DIFF
--- a/modules/system/classes/UpdateManager.php
+++ b/modules/system/classes/UpdateManager.php
@@ -84,6 +84,29 @@ class UpdateManager
     }
 
     /**
+     * fireBeforeUpdate triggers the before update event.
+     */
+    public function fireBeforeUpdate(): mixed
+    {
+        /**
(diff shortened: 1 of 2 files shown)


daftspunk:
Thanks!

PR #6083, opened 2026-07-27
feat(mail): add live HTML preview tab to mail templates, layouts

mahmoud-linkness:
Hello,
we always in team had this problem when modifying layout. we can't preview the changes immediately and instead have to send an email to the administrator, which is inefficient.

## Before:
image

## After:
image


skyfallduck1:
bump!

daftspunk:
Nice! Thanks!!

PR #6102, opened 2026-09-02
Remove the duplicate exists key from the Finnish validation strings

darkdi:
`modules/system/lang/fi/validation.php` sets `exists` twice in the same array, once in the aligned alphabetical block and again eleven lines below it, so the first one has never been used.

I dropped the later one because it sits out of alphabetical order in a block with different indentation, which is where a second translation pass looks to have been pasted in, and because its text is character for character the same as the `in` message right under it. Laravel does ship the same English string for both keys, so that is not wrong on its own, it just means the earlier line reads as the deliberate `exists` wording.

This does change which Finnish sentence users see. If the team prefers the other wording, the fix is to move that text onto the surviving line rather than to keep both keys.


daftspunk:
Thanks!

PR #6106, opened 2026-09-04
Fix MediaFinder paste button not updating across fields on the same page

### LINKED ISSUE #6103: MediaFinder: Paste button remains disabled when copying media between fields on the same page
Hi! I found something that would be great to improve.

Copying media from one `MediaFinder` field to another `MediaFinder` field located on the same page does not work correctly.

After copying media from the first field, the **Paste** button in another `MediaFinder` field remains disabled.

Refreshing the entire page makes the **Paste** button available and the copied media can then be pasted successfully.

It would be very handy to do this without refreshing the whole page.

**CMS version:** 4.4.1

wakqasahmed:
Closes #6103

Copying items from one MediaFinder field and pasting into another MediaFinder field on the same page doesn't work until the page is reloaded. The paste button in the second field stays disabled right after the copy, even though the clipboard now has items it could paste.

Each MediaFinder instance keeps its own toolbar button state and only recalculates it on its own DOM events, so when instance A writes to the shared `localStorage` clipboard, instance B never re-runs `updateButtonsState()` to notice. This patch dispatches a document-level `oc.mediafinder.clipboardUpdated` event after a copy and has every multi/copy-paste-enabled instance listen for it and refresh its own button state, mirroring the `oc.Events` pub/sub pattern already used by other form widgets like colorpicker and paletteeditor.

Tested with two MediaFinder fields on the same form: copying in one now immediately enables the paste button (with the correct new-item count) in the other, without a reload.

### DIFF
--- a/modules/media/formwidgets/mediafinder/assets/js/mediafinder.js
+++ b/modules/media/formwidgets/mediafinder/assets/js/mediafinder.js
@@ -63,6 +63,10 @@ oc.registerControl('mediafinder', class extends oc.ControlBase {
             this.listen('click', '.toolbar-select-all', this.onSelectAllClick);
             this.listen('click', '.toolbar-copy-selected', this.onCopySelectedClick);
             this.listen('click', '.toolbar-paste-items', this.onPasteItemsClick);
+
+            // Other MediaFinder instances on the same page need to refresh their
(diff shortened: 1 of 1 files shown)


daftspunk:
Thanks!

PR #6111, opened 2026-09-09
Fix scope filter option labels being HTML-escaped (#5935)

### LINKED ISSUE #5935: Scopes filter view html tags as text
Image

`php
public function getStatusScopesOptions()
{
	foreach(Status::all() as $v) $res[$v['code']] = "<span class='select-status status-indicator' style='background:".$v['color']."'></span> ".$v['name'];
	return $res;
}
`

Image

wakqasahmed:
Fixes #5935.

Group filter scope option labels are meant to carry raw HTML — the reported use case is a status dropdown returning `<span class='select-status'></span> Active` per option, and that's a pretty common pattern for icon/color-coded scope options. The PHP side already treats it that way: `Group::optionsToAjax()` sends the label through `__()` with no `e()` call, so nothing gets escaped before it reaches the client.

The AJAX response gets rendered into the group filter dropdown via a Mustache template, though, and that template used `{{ name }}`, which HTML-escapes by Mustache's own spec. So the label came out as literal escaped markup instead of rendering. Switched both occurrences (available and active option lists) to `{{{ name }}}`, Mustache's unescaped tag, matching what the PHP side already assumes.

Added a test on `FilterWidgetTest` asserting the `onGetGroupOptions` AJAX handler preserves an HTML label unescaped end-to-end (the actual escaping happens client-side in the Mustache template, which isn't something PHPUnit can exercise directly, but this locks in the PHP-side contract the fix depends on). Ran the full `FilterWidgetTest` suite — 9 tests, 19 assertions, all green.

### DIFF
--- a/modules/backend/filterwidgets/group/partials/_group_form.php
+++ b/modules/backend/filterwidgets/group/partials/_group_form.php
@@ -46,7 +46,7 @@
         <div class="filter-items">
             <ul>
                 {{#available}}
-                    <li data-item-id="{{ id }}"><a href="javascript:;">{{ name }}</a></li>
+                    <li data-item-id="{{ id }}"><a href="javascript:;">{{{ name }}}</a></li>
                 {{/available}}
                 {{#loading}}
                     <li class="loading"><span></span></li>
@@ -56,7 +56,7 @@
         <div class="filter-active-items">
             <ul>
                 {{#active}}
-                    <li data-item-id="{{ id }}"><a href="javascript:;">{{ name }}</a></li>
+                    <li data-item-id="{{ id }}"><a href="javascript:;">{{{ name }}}</a></li>
                 {{/active}}
             </ul>
         </div>
--- a/modules/backend/tests/widgets/FilterWidgetTest.php
+++ b/modules/backend/tests/widgets/FilterWidgetTest.php
@@ -54,6 +54,53 @@ class FilterWidgetTest extends PluginTestCase
         $this->assertSame($rawLabel, $result['options']['available'][0]['name']);
     }
 
+    /**
+     * A group scope backed by a related model (no `options`/`optionsMethod`
+     * configured) pulls its label from a plain database column, which is not
+     * developer-authored, so it must stay HTML-escaped even though the
+     * template renders it with the unescaped `{{{ name }}}` Mustache tag —
+     * matching the RecordFinder precedent (`_record_single.php`,
+     * `_record_multi.php`, both call `e()` on their model-derived name).
+     * https://github.com/octobercms/october/issues/5935
+     */
+    public function testGroupScopeOptionLabelFromModelRelationIsEscaped()
+    {
+        $unsafeLogin = '<script>alert(1)</script>';
+
+        $user = new User;
+        $user->fill([
+            'first_name' => 'Test',
+            'last_name' => 'User',
+            'login' => $unsafeLogin,
+            'email' => 'grouptest@test.com',
+            'password' => 'password',
+            'password_confirmation' => 'password',
+        ]);
+        $user->save();
+
+        $filter = $this->makeFilterWidget([
+            'model' => new User,
+            'arrayName' => 'array',
+            'scopes' => [
+                'status' => [
+                    'type' => 'group',
+                    'label' => 'Status',
+                    'modelClass' => User::class,
+                    'nameFrom' => 'login',
+                ],
+            ],
+        ]);
+        $filter->render();
+
+        $widget = $filter->getFilterWidgets()['status'];
+        $result = $widget->onGetGroupOptions();
+
+        $names = array_column($result['options']['available'], 'name');
+
+        $this->assertContains(e($unsafeLogin), $names);
+        $this->assertNotContains($unsafeLogin, $names);
+    }
+
     public function testFilterWidgetsAreRegistered()
     {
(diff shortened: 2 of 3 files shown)


wakqasahmed (commented):
The core fix direction is right, but I think it's unescaping more than it should. See inline comments.

modules/backend/filterwidgets/group/partials/_group_form.php
```diff
@@ -46,7 +46,7 @@ class="filter-search-input form-control popup-allow-focus"
         <div class="filter-items">
             <ul>
                 {{#available}}
-                    <li data-item-id="{{ id }}"><a href="javascript:;">{{ name }}</a></li>
+                    <li data-item-id="{{ id }}"><a href="javascript:;">{{{ name }}}</a></li>
```
wakqasahmed, line 49: This unescapes `name` for every group scope, but `name` doesn't always come from developer-authored config. Look at `Group::getAvailableOptions()` — when a scope has neither `options` nor `optionsMethod` set, it falls through to `getOptionsFromModel()` and does `$available[$option->getKey()] = $option->{$nameColumn}`, i.e. a plain column value off a related model (whatever `nameFrom` points at). That's live DB data, not something the plugin developer typed into a config array.

Compare that to `RecordFinder`, which does the same "pick a name column off a related model" thing and explicitly `e()`-escapes it (`_record_single.php`, `_record_multi.php`). This PR makes the Group filter inconsistent with that: any category/tag/whatever-relation name that ends up containing `<`/`>` (typed by another backend user, imported, synced from an external source) now renders as raw HTML in this popup instead of literal text.

The issue's own repro (`getStatusScopesOptions()`) is a case where the developer is deliberately building an HTML string in code they wrote, which is a different trust level than "whatever text sits in a related row's name column." I'd scope this so the model-relation path still gets escaped, and only the explicit `options`/`optionsMethod` path (where a developer opts into returning label markup) skips escaping.

modules/backend/filterwidgets/group/partials/_group_form.php
```diff
@@ -56,7 +56,7 @@ class="filter-search-input form-control popup-allow-focus"
         <div class="filter-active-items">
             <ul>
                 {{#active}}
-                    <li data-item-id="{{ id }}"><a href="javascript:;">{{ name }}</a></li>
+                    <li data-item-id="{{ id }}"><a href="javascript:;">{{{ name }}}</a></li>
```
wakqasahmed, line 59: Same concern as the `available` list above — applies here too since `active` is built from the same `getAvailableOptions()` output.

modules/backend/tests/widgets/FilterWidgetTest.php
```diff
@@ -18,9 +18,42 @@ public function setUp(): void
...
+                    ],
+                ],
+            ],
+        ]);
+        $filter->render();
+
+        $widget = $filter->getFilterWidgets()['status'];
+        $result = $widget->onGetGroupOptions();
+
+        $this->assertSame($rawLabel, $result['options']['available'][0]['name']);
```
wakqasahmed, line 54: This test only exercises the static `options` array path, which is the safe/developer-authored case. It doesn't cover `getOptionsFromModel()` (no `options`/`optionsMethod`, scope backed by a model relation), which is the path I'm flagging in the template comments as now emitting unescaped HTML. Given the fix touches both branches equally, a second case using a related model with a `nameFrom` column containing `<`/`>` would actually pin down whether that data should still be escaped or not — right now nothing in this suite would catch a regression there.

modules/backend/filterwidgets/group/partials/_group_form.php
wakqasahmed: Confirmed the concern — `getAvailableOptions()` does have two branches with different trust levels, and the model-relation branch (`getOptionsFromModel()`, plain column off a related model via `nameFrom`) shouldn't skip escaping just because the developer-config branch needs to.

Fixed in 8c76e0d: the model-relation branch now does `$available[$option->getKey()] = e($option->{$nameColumn});` before it reaches the template, matching the `e()` calls in RecordFinder's `_record_single.php`/`_record_multi.php` for the same pattern. The `options`/`optionsMethod` branch is untouched and still passes labels through raw, since that's developer-authored config. Template stays `{{{ name }}}` for both — the escaping decision now happens server-side at the point where the source is known, rather than in the template.

Also added `testGroupScopeOptionLabelFromModelRelationIsEscaped`, which creates a real `Backend\Models\User` row with `<script>alert(1)</script>` as the `login` column, wires a group scope with `modelClass`/`nameFrom` pointing at it (no `options`/`optionsMethod`), and asserts the AJAX response contains the escaped string and not the raw one. Ran both this test and the existing `testGroupScopeOptionLabelPreservesRawHtml` together (php 8.4, phpunit 9.6) — all 10 tests in `FilterWidgetTest.php` pass.

modules/backend/filterwidgets/group/partials/_group_form.php
wakqasahmed: Same fix covers this — see my reply on the `available` list comment above (8c76e0d).

daftspunk:
Thanks for this, great catch on the escaping split. 

Differentiating developer-authored config from model-sourced data is exactly the right trust boundary, and following the RecordFinder precedent keeps it consistent. The two tests covering both directions are appreciated.

PR #6116, opened 2026-09-11
Cache theme data between requests

scottbedard:
Adds a `CMS_THEME_DATA_CACHE` environment variable that caches theme data between requests.

Currently it's only cached in memory, so that data is lost on the next request and causes a database hit. The in memory data is still preferred, so when a cache hit occurs it places the value there, as if it had been fetched normally.

There are no user facing behavior changes.

### DIFF
--- a/config/cms.php
+++ b/config/cms.php
@@ -163,7 +163,7 @@ return [
     |
     */
 
-    'enable_data_cache' => env('CMS_DATA_CACHE', false),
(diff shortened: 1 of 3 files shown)


daftspunk:
Thanks, I'll tweak these since it doesn't really need config -- SettingModel follow this pattern already and it is fine to stay on par with this.

PR #6117, opened 2026-09-13
Fix Backspace clearing every tag in a customTags TagList

### LINKED ISSUE #6108: TagList with `customTags` and no options: empty dropdown after the first tag, and Backspace removes every tag on a create form
## Environment

- October CMS 4.3.8 (`october/system` v4.3.8), backend form
- Field definition, no `options` at all:

```yaml
blocked_keywords:
    type: taglist
    mode: string
    separator: comma
    customTags: true
```

## Bug 1: on an update form, an empty dropdown bar opens although there is nothing to choose from

**Steps**

1. Add the field above to any model form (a settings model is enough).
2. Save the form with the field empty, reopen it, click into the field: no dropdown, as expected.
3. Type `foo`, press Enter, save. Reopen the record on the update form.
4. Click into the field.

**Expected:** no dropdown. There are no predefined options and the only existing tag is already selected, so there is nothing to choose from.shows up on every update form once the record has at least one saved tag.

**Cause**

`modules/backend/formwidgets/taglist/partials/_taglist.php` adds the `select-no-dropdown` class only when the option list is empty:

```php
class="form-control custom-select <?= !count($fieldOptions) ? 'select-no-dropdown' : '' ?> select-hide-selected"
```

With `customTags` the widget sets `useKey = false`, and `getKeylessOptions()` (`modules/backend/formwidgets/TagList.php:206`) appends every already selected value to `$fieldOptions`, so the list stops being empty as soon as one tag has been saved. Without the class, `custom-select-control.js:152` no longer applies `selectOnClose` / `minimumInputLength`, so select2 opens its dropdown, and `select-hide-selected` hides every row because every option is also selected. The result is a dropdown with nothing in it.

Passing `options: []` or `options: null` makes no difference, because the option list is rebuilt from the selected values anyway.

**Suggested fix**

Decide the class from the options the widget actually received, not from the list after the selected values were merged in, for example by computing it before `getKeylessOptions()` is applied, or by counting only options that are not already selected.

## Bug 2: on a create form, Backspace removes every tag at once

**Steps**

1. Same field, open the **create** form of the model (no record saved yet).
2. Type `foo`, press Enter; type `bar`, press Enter; type `baz`, press Enter.
3. With the search input empty, press Backspace once.

**Expected:** only `baz` is removed, as with any select2 tag input.

**Actual:** `foo`, `bar` and `baz` are all removed.

On the **update** form of a record that already has tags the same steps behave correctly: one Backspace removes only the last tag. So the difference is between a form that started with no selected values and one that started with some.

I have not traced this one down. It may be select2 with `tags: true` plus the `tokenSeparators` built in `custom-select-control.js:147`, or the October layer on top of it. Happy to test a patch.

wakqasahmed:
Fixes #6108 (the Backspace-clears-everything part).

\`buildSelectOptions()\` turns on Select2's \`allowClear\` whenever the field has a placeholder, and \`buildTokenSeparatorOptions()\` (used for TagList's \`customTags\` mode) never turned it back off. Select2 treats \`allowClear\` as "this whole field is clearable" — with a multi-value tags select, that makes a single Backspace clear every tag at once instead of just removing the last one, since Select2 doesn't distinguish "clear one token" from "clear the field" when \`allowClear\` is on.

Set \`allowClear = false\` in \`buildTokenSeparatorOptions()\` so token-separator tag fields keep normal per-tag Backspace behavior. Regular (non-tag) selects with a placeholder are untouched — \`buildSelectOptions()\` still sets \`allowClear\` there, this only overrides it for the tags path.

Verified with a small jsdom + Select2 harness that loads the actual control file and dispatches a real Backspace keydown: before the fix, backspacing once with 3 tags present clears all of them; after, it removes only the last one. Also confirmed a regular placeholder select's own clear button still works, unaffected by this change.


daftspunk:
Thanks!

PR #6120, opened 2026-09-14
Default formSubmitted to false in SubmissionComponent (#6118)

### LINKED ISSUE #6118: Demo theme: blog/comment-form.htm throws "Variable formSubmitted does not exist" with strict Twig variables
# Demo theme: `blog/comment-form.htm` throws "Variable formSubmitted does not exist" with strict Twig variables

## Environment
- October CMS 4.4 (also affects 3.x, file introduced with the v3.0 Tailor blog demo)
- Bundled demo theme (`themes/demo`)
-`CMS_STRICT_VARIABLES=true` (strict Twig variables enabled in `.env`)
- PHP 8.x, fresh install, `php artisan tailor:migrate` executed

## Steps to reproduce
1. Fresh October CMS 4.4 install, activate the bundled demo theme
2. Set `CMS_STRICT_VARIABLES=true` in `.env`
3. Run `php artisan tailor:migrate`
4. Open any blog post page (`/blog/:slug`) that renders the comment form

## Actual result
```
Twig\Error\RuntimeError: Variable "formSubmitted" does not exist
in "partials/blog/comment-form.htm"
```

## Expected result
The comment form renders. With `CMS_STRICT_VARIABLES=false` (the default) the page works fine — which is why this easily goes unnoticed.

## Root cause
`themes/demo/partials/blog/comment-form.htm` line 6 checks:

```twig
{% if formSubmitted %}
```

`formSubmitted` is only injected by the Submission component during the `onFormSubmit` AJAX handler. On a regular page render (GET) the variable is undefined. With strict Twig variables enabled, this becomes a runtime error and takes down the whole page — for a visitor this is a 500 on every blog post.

## History
- Introduced with the Tailor blog demo in v3.0: commit 61b8be697c56 (2022-05-04)
- Still unguarded after the 4.4 refactor of the very same line: commit da619ceff8c3 (2026-09-05)

## Suggested fix
Minimal theme-side fix (one line):

```twig
{% if formSubmitted is defined and formSubmitted %}
```

Alternatively — and more robust for all theme developers — the Submission component could always define `formSubmitted` (defaulting to `false`) on the initial render, so templates don't need the `is defined` guard at all. Both approaches could be combined.

I searched the issue tracker and did not find an existing report for this.

## Workaround
Theme authors can apply the `is defined` guard locally in `comment-form.htm`.

wakqasahmed:
Fixes #6118.

`themes/demo/partials/blog/comment-form.htm` and `survey-form.htm` both check `{% if formSubmitted %}`, but `formSubmitted` is only ever set inside `onFormSubmit()` — never on a plain GET render. With `CMS_STRICT_VARIABLES` on, that undefined read throws and takes down the page, which is basically every blog post and (via the footer) most pages on the demo theme.

Added an `onRun()` to `SubmissionComponent` that defaults `formSubmitted` to `false`. Used a null-coalescing assignment rather than a flat one because `onRun()` runs after the handler on the non-JS postback path (`execPostbackHandler` falls through to `execPageCycle`), so a flat assignment would stomp the `true` that `onFormSubmit()` just set. The genuine-AJAX path never reaches `onRun()` at all — `execAjaxHandlers()` returns before `execPageCycle()` runs — so that path was never at risk either way.

Added two tests to the existing `SubmissionComponentTest`: one confirming `onRun()` alone gives `false`, one confirming a real submit followed by `onRun()` keeps `true`.


daftspunk:
We fixed this in the behavior instead

Thank you for your contribution

PR #6122, opened 2026-09-15
Fix file column click being swallowed by row record-navigation link

### LINKED ISSUE #6121: File list column does not open files when used in a relation widget
Hi! I have a relation widget with the following configuration (`config_relation.yaml`):

```
reports: 
    label: Reports
    view:
        list: 
            columns:
                files:
                    label: Files
                    type: file
                    limit: 5
                    clickable: true
        toolbarButtons: add|remove|create
        showSearch: true
    manage:
        list: $/test/reports/models/reports/columns.yaml
        form: $/test/reports/models/reports/fields.yaml
        showSearch: true
```

Clicking the file icon should open the file. Instead, it opens the relation record edit form.

wakqasahmed:
Fixes #6121.

A `type: file` column with `clickable: true` never opens the file when clicked — the click always navigates to the record instead, as reported (e.g. inside a relation widget's list config).

Two things compound this. `_list_body_row.php` wraps a clickable column's whole cell in a record-navigation `<a href="...">` when the list has a `recordUrl`/`recordOnClick` configured. Separately, `_column_file.php` rendered each file as a bare, non-linked `<span>` whenever the column was `clickable`, and only rendered a real `<a href={fileUrl}>` when it wasn't. So with `clickable: true`, every file item sat inside the outer record-navigation anchor with no link of its own — HTML can't nest two `<a>` tags anyway, so the outer one always won the click regardless of where on the cell you clicked.

A file column's whole point is that each item is its own link target, so `clickable` doesn't really make sense as a row-navigation toggle for it the way it does for a plain text column. Excluded file columns from the row-level wrap entirely (regardless of their own `clickable` setting), and made `_column_file.php` always render a real, individually clickable link for every file item.

Added a test: a `Lists` widget with a `recordUrl` configured (so the row-wrap actually triggers) and a `file` column with `clickable: true`, asserting the rendered file item is a real link to the file and that it isn't the record's own URL.


daftspunk:
Thanks for trying!

PR #6125, opened 2026-09-18
Load demo code highlighting on demand

samuelpatro:
Pages without code samples currently download the demo theme’s CodeMirror JavaScript and CSS. Load these assets only when code samples exist, including samples inserted by AJAX. Wait for the stylesheet before initializing editors, deduplicate requests, and keep plain code and Expand usable when downloads fail; a later render retries failed downloads.

Also lazy-load the three below-the-fold homepage illustrations and add their intrinsic dimensions. The hero image stays eager. The generated editor bundle retains CodeMirror 5.65.5 and uses the existing global jQuery.

Validation:
- Browser fixture passes for initial code and AJAX-inserted code, repeated renders, replaced fragments, PHP/Twig text, read-only editors and Expand.
- Cold-browser blocked CSS and blocked JavaScript checks preserve readable, expandable code; unblocking and rendering again succeeds.
- JavaScript syntax checks and git diff whitespace checks pass.
- Independent standards and spec reviews completed; identified failure-path issues were fixed and rechecked.

Draft for testing: full CMS integration smoke testing remains pending. Reproduction and CMS checks are documented in themes/demo/tests/README.md. This change avoids optional assets on pages without samples; no LCP or page-load timing improvement is claimed. Gallery changes are outside this PR.


daftspunk:
This is an improvement we want to make to Larajax. Essentially convert codeblocks to a hot control then lazy load it when the `data-control` element is first seen on the page. It means passing a closure to the registration

```js
oc.registerControl('codeblock', () => import('./codeblocks.js'));
```

That script would also import the codeblocks library rather than include it on every page. Then the whole bundle only imports when this is first seen:

```html
<div data-control="codeblock"></div>
```

Issue #681, opened 2014-10-11
Add support for Import/Export in Lists


daftspunk:
A complimentary behavior or widget to the ListController behavior / List widget, or even straight up baking it in to the ListController/List widget itself (not sure which yet)

#### Exporting
The list should allow exporting of the current data in the view in various formats (CSV, PDF, etc). The export format should be easily compatible with the import format.

#### Importing
Importing feature should have a dedicated page where a CSV can be uploaded, the header row is analysed and a "match maker" screen is shown, where the CSV fields can be matched to the List columns.

The match maker screen will automatically marry fields that have the same name, to compliment the export process.

Eg:

```
Step 1:
[  Upload import file ] ......
```

then
```
Step 2
Match the data fields:

File           |         Matched columns    |       List columns
------------------------------------------------------------------
Name           ->        [ Name  ]          |       URL
Website        ->        [       ]          |       Content
Description    ->        [       ]          |       Country
                                            |       State

[   Import this data   ]
```

The user can drag from the *Available columns* to match with the *File* header columns.

daftspunk:
Added in Build 284

sushmak2216:
Integration with list behavior is not working
I have added the following lines into list configuration files:
export:
    useList: true

PR #6126, opened 2026-09-18
Cache remaining cms queries

scottbedard:
This PR is dependent on https://github.com/octobercms/library/pull/681

A frontend request still queries theme language lines and for the database template index. This pull request caches both so those queries do not run on every request and a sleeping database can stay asleep.

Language lines stay cached until a save, tombstone, or purge. The template index is cached in `DbDatasource`, this pull request only clears it after a theme purge, which deletes rows with a raw query that skips the normal write path.


scottbedard:
@daftspunk I think this is ready for your review. The only question I have is if we want to set a TTL using a config / environment variable. I think it's safe without one, but I'll defer to your judgement.

modules/cms/classes/ThemeManager.php
```diff
@@ -469,6 +520,8 @@ public function importDatabaseTemplates(string $dirName, ?string $srcDirName = n
     public function purgeDatabaseTemplates(string $dirName)
     {
         Db::table('cms_theme_templates')->where('source', $dirName)->delete();
+
+        DbDatasource::clearCache($dirName, 'cms_theme_templates');
```
scottbedard, line 524: This depends on [library #681](https://github.com/octobercms/library/pull/681/changes#diff-98157453abeb8dbdc62be9b05c39b39cc13f63407ef16cafb0428b501563aa66R347-R358)

daftspunk:
Thanks!

